PR #91, opened 2015-01-29
Modifies default logging behavior for IssueCommand

ehankland:
Modifies logging so that command results with non-zero return codes will  be printed by default. This gives immediate feedback when something goes
wrong rather than forcing the user to go through the log file.


voellm:
LGTM

PR #135, opened 2015-02-25
Provides a flag to execute UnixBench with the > 16 cores patch

meteorfox:
UnixBench by default is designed to only execute up to 16 jobs
concurrently. If your system has more than 16 CPUs available, UnixBench
will not execute enough jobs to fully utilize all the CPUs in the machine,
which will result in a lower score that what the system is capable of.
A patch was submitted to the UnixBench issues that addresses this issue.
This commit allows the user to optionally use this patch by setting the
following flag:

  --unixbench_all_cores

Official link to issue:
https://code.google.com/p/byte-unixbench/issues/detail?id=4

### DIFF
--- a/perfkitbenchmarker/benchmarks/unixbench_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/unixbench_benchmark.py
@@ -22,6 +22,7 @@ some memory bandwidth, and disk.
 
 import logging
 
+from perfkitbenchmarker import data
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import regex_util
 from perfkitbenchmarker import sample
@@ -34,7 +35,11 @@ BENCHMARK_INFO = {'name': 'unixbench',
                   'scratch_disk': True,
                   'num_machines': 1}
 
+flags.DEFINE_boolean('unixbench_all_cores', default=False,
+                     help='Applies patch to UnixBench to fix limitation of '
+                     '16 cores limit.')
 
+UNIXBENCH_PATCH_FILE = 'unixbench-16core-limitation.patch'
 SYSTEM_SCORE_REGEX = r'\nSystem Benchmarks Index Score\s+([-+]?[0-9]*\.?[0-9]+)'
 RESULT_REGEX = (
     r'\n([A-Z][\w\-\(\) ]+)\s+([-+]?[0-9]*\.?[0-9]+) (\w+)\s+\('
@@ -50,6 +55,17 @@ def GetInfo():
   return BENCHMARK_INFO
 
 
+def CheckPrerequisites():
+  """Verifies that the required resources for UnixBench are present.
+
+  Raises:
+    perfkitbenchmarker.data.ResourceNotFound: On missing resource.
+  """
+
+  if FLAGS.unixbench_all_cores:
+    data.ResourcePath(UNIXBENCH_PATCH_FILE)
+
+
 def Prepare(benchmark_spec):
   """Install Unixbench on the target vm.
 
@@ -61,6 +77,13 @@ def Prepare(benchmark_spec):
   vm = vms[0]
   logging.info('Unixbench prepare on %s', vm)
   vm.Install('unixbench')
+
+  if FLAGS.unixbench_all_cores:
+    vm.PushDataFile(UNIXBENCH_PATCH_FILE)
+    vm.RemoteCommand('cp %s %s' %
+                     (UNIXBENCH_PATCH_FILE, unixbench.UNIXBENCH_DIR))
+    vm.RemoteCommand('cd %s && patch ./Run %s' %
+                     (unixbench.UNIXBENCH_DIR, UNIXBENCH_PATCH_FILE))
 
 
 def ParseResults(results):
--- a/perfkitbenchmarker/benchmarks/unixbench_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/unixbench_benchmark.py
@@ -36,8 +36,9 @@ BENCHMARK_INFO = {'name': 'unixbench',
                   'num_machines': 1}
 
 flags.DEFINE_boolean('unixbench_all_cores', default=False,
-                     help='Applies patch to UnixBench to fix limitation of '
-                     '16 cores limit.')
+                     help='Setting this flag changes the default behavior of '
+                     'Unix bench. It will now scale to the number of CPUs on '
+                     'the machine vs the limit of 16 CPUs today.')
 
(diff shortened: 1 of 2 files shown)


meteorfox:
DO NOT MERGE YET!

This is WIP, I just want to get early feedback.

voellm:
No worries... well just look at the code.

Connor why does the bot keep pinging?

cmccoy:
ok to test.

cmccoy:
Sorry for the noise from the bot - I'll look at turning it down.

voellm:
Looks pretty good so far.  The only comment is to make the flag more descriptive like "Setting this flag changes the default behavior of Unix bench.  It will now scale to the number of processors on the machine."

Is the 2D and 3D line in the patch file adding extra tests?

meteorfox:
Ok, I will add that to the flag description. 

It doesn't add extra tests, I think the GitHub diff viewer is kind of confusing. If you look at the file directly here you'll see it doesn't add them, they are already there.

https://github.com/meteorfox/PerfKitBenchmarker/blob/unixbench-16-cores-patch/perfkitbenchmarker/data/unixbench-16core-limitation.patch

I'm having an issue with the regex for the score for some reason. I'm debugging it right now, will commit patch soon.

This is the error I'm getting:

<code>
NoMatchError: No match for pattern "\n([A-Z][\w\-\(\) ]+)\s+([-+]?[0-9]*\.?[0-9]+) (\w+)\s+\(([-+]?[0-9]*\.?[0-9]+) (\w+), (\d+) samples\)" in "Benchmark Run: Wed Feb 25 2015 23:16:02 - 23:16:02
32 CPUs in system; running 32 parallel copies of tests

" 
</code>

Here's the full UnixBench output where is failing on:
https://gist.github.com/meteorfox/558408e6138fac5d90ed

meteorfox:
I think because the output was longer due to number of processors in the system, it didn't flush all of it or something, the output stops abruptly at the ""Benchmark Run: Wed Feb 25 2015 23:16:02 - 23:16:02 32 CPUs in system; running 32 parallel copies of tests"" line and there's nothing else.

voellm:
It will be good to add some tests for the output parser :)

On Thu, Feb 26, 2015 at 1:31 AM, Carlos Torres <notifications@github.com>
wrote:

> I think because the output was longer due to number of processors in the
> system, it didn't flush all of it or something, the output stops abruptly
> at the ""Benchmark Run: Wed Feb 25 2015 23:16:02 - 23:16:02 32 CPUs in
> system; running 32 parallel copies of tests"" line and there's nothing else.
>
> --
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/135#issuecomment-76097035>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

meteorfox:
Ok, I figured out the problem. It wasn't the size of the output with the default UnixBench you cannot simply run more than 16 "copies", which is exactly the problem this pull-request request is trying to address. It's easy to replicate, either run UnixBench as is with no-patch in a box with more than 16 CPUs, or execute it with this flag:

$ ./Run -c 17 

-c  is the number of copies, which it will not execute beyond 16. If -c is <= 16, it will work as expected. Executing it this way will skip directly to an incomplete output just like above, and produce an exception on PerfKit.

meteorfox:
Can someone confirm and reproduce the steps I mentioned on the previous comment?

* Setup UnixBench (default, with no-patch)
* Execute with ./Run -c N  where N > 16
  * When no flags are specified, UnixBench does a run with -c 1, and then another one with -c M, where M is the number of CPUs in the system.
* Expected outcome is that it runs N parallel copies and produces the full results report
* Actual outcome, only the following message is printed and benchmarks do not run, hence reports no data.
  * "M CPUs in system; running N parallel copies of tests"

meteorfox:
Regarding the issue about not getting data from the parallel copies test with vanilla UnixBench on a box with more than 16 CPUs, it's actually a limitation on UnixBench and not on PerfKit. This bug is already in master, and it is not introduced by this pull-request.

This pull-request enables a user to optionally overcome this limitation, by adding the --unixbench_all_cores=true flag to the pkb command. It doesn't address the problem of PerfKit throwing the NoMatchError exception when running vanilla UnixBench on >16 CPUs systems, I believe another issue and pull-request should be opened to address that problem. 

We will need to decide how will PerfKit deal when executing unixbench on a >16 CPUs system and without this flag.

As of the last commit, I consider this change to be Ready for Review.

voellm:
Connor,

Can you please help Carlos with this commit.

  Thanks,
     Tony

On Fri, Feb 27, 2015 at 3:55 PM, Carlos Torres <notifications@github.com>
wrote:

> Regarding the issue about not getting data from the parallel copies test
> with vanilla UnixBench on a box with more than 16 CPUs, it's actually a
> limitation on UnixBench and not on PerfKit. This bug is already in master,
> and it is not introduced by this pull-request.
>
> This pull-request enables a user to optionally overcome this limitation,
> by adding the --unixbench_all_cores=true flag to the pkb command. It
> doesn't address the problem of PerfKit throwing the NoMatchError exception
> when running vanilla UnixBench on >16 CPUs systems, I believe another issue
> and pull-request should be opened to address that problem.
>
> We will need to decide how will PerfKit deal when executing unixbench on a
> >16 CPUs system and without this flag.
>
> As of the last commit, I consider this change to be Ready for Review.
>
> --
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/135#issuecomment-76495031>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

cmccoy:
Sure, I'm looking now.

cmccoy:
> Can someone confirm and reproduce the steps I mentioned on the previous comment?
> 
> * Setup UnixBench (default, with no-patch)
> * Execute with ./Run -c N where N > 16
> * When no flags are specified, UnixBench does a run with -c 1, and then another one with -c M, where M is the number of CPUs in the system.
>
>     * Expected outcome is that it runs N parallel copies and produces the full results report
>
> * Actual outcome, only the following message is printed and benchmarks do not run, hence reports no data.
>
>     * "M CPUs in system; running N parallel copies of tests"

I get the same behavior.  I opened #138 to track cleaning up the behavior when running on a machine with >16 cores and without `--unixbench_all_cores`.

I'm running UnixBench with the patch to verify, otherwise looks great.

meteorfox:
@cmccoy Thanks for reviewing this commit.

cmccoy:
Thanks, @meteorfox.

PR #154, opened 2015-03-06
Installs EPEL through downloaded RPM if 'epel-release' is not available

ehankland:
Fixes benchmarks that require EPEL on images where 'epel-release' is not available (as is the case for RHEL6/7 on GCE)

### DIFF
--- a/perfkitbenchmarker/package_managers.py
+++ b/perfkitbenchmarker/package_managers.py
@@ -26,6 +26,8 @@ file name minus .py). The framework will take care of all cleanup
 for you.
 """
 
+import re
+
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import packages
@@ -34,6 +36,11 @@ from perfkitbenchmarker import vm_util
 
 RHEL = 'rhel'
 DEBIAN = 'debian'
+
+EPEL6_RPM = ('http://dl.fedoraproject.org/pub/epel/'
+             '6/x86_64/epel-release-6-8.noarch.rpm')
+EPEL7_RPM = ('http://dl.fedoraproject.org/pub/epel/'
+             '7/x86_64/e/epel-release-7-5.noarch.rpm')
 
 flags.DEFINE_enum('os_type', DEBIAN,
                   [DEBIAN, RHEL],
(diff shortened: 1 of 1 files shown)


ehankland:
test this please.

perfkitbenchmarker/package_managers.py
```diff
@@ -109,7 +116,18 @@ def Startup(self):
 
   def InstallEpelRepo(self):
     """Installs the Extra Packages for Enterprise Linux repository."""
-    self.InstallPackages('epel-release')
+    try:
+      self.InstallPackages('epel-release')
+    except errors.VmUtil.SshConnectionError as e:
+      stdout, _ = self.RemoteCommand('cat /etc/redhat-release')
```
cmccoy: What's the format for `/etc/redhat-release`? Any way to make this regex more specific?

perfkitbenchmarker/package_managers.py
```diff
@@ -109,7 +116,18 @@ def Startup(self):
...
+      major_version = int(re.search('[0-9]', stdout).group(0))
+      if major_version == 6:
+        self.RemoteCommand('sudo rpm -ivh --force %s' % EPEL6_RPM)
```
cmccoy: Nit: this would be cleaner if you set a local `rpm` based on `major_version`, then had a single `sudo rpm install` command.

perfkitbenchmarker/package_managers.py
```diff
@@ -109,7 +116,18 @@ def Startup(self):
...
+      elif major_version == 7:
+        self.RemoteCommand('sudo rpm -ivh --force %s' % EPEL7_RPM)
+      else:
+        raise e
+
```
cmccoy: Nit: extra newline here.

cmccoy:
A few minor comments, otherwise LGTM.

perfkitbenchmarker/package_managers.py
ehankland: specified "release" before the major version number

perfkitbenchmarker/package_managers.py
ehankland: Done

perfkitbenchmarker/package_managers.py
ehankland: Done

ehankland:
test this please.

Issue #159, opened 2015-03-10
Error with df command in fio benchmark


akrzos:
Running against a static vm that is using device mapper, df can indent the output it provides which leads to an error attempting to run the fio benchmark:

```
# df -k
Filesystem           1K-blocks    Used Available Use% Mounted on
/dev/mapper/vg_dhcp23108-lv_root
                      51475068 2121292  46732336   5% /
tmpfs                  1960452       0   1960452   0% /dev/shm
/dev/vda1               487652   33396    428656   8% /boot
/dev/mapper/vg_dhcp23108-lv_scratch0
                      46967160   53116  44521572   1% /scratch0
```

This output breaks [vm.GetDeviceSizeFromPath(vm.GetScratchDir())] (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/5de105ba9a9edd536603f831ac1d9d175f386c02/perfkitbenchmarker/benchmarks/fio_benchmark.py#L65) in the fio benchmark.

This might only affect the version of df on this vm which is 8.4.

cmccoy:
Closed by #160.

Issue #127, opened 2015-02-23
allow scratch space to be an aribtirary directory


cburroughs:
It would be useful to be able to run PerfKitBenchmarker on VMs/containers/physical-servers where there isn't a dedicated mount point for the scratch space, or where creating a mountpoint per service is not supported.

Currenlty trying to run fio with `"scratch_disk_mountpoints": "/tmp/google-pkb"` results in```
Traceback (most recent call last):
  File "./pkb.py", line 21, in <module>
    sys.exit(Main())
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 358, in Main
    return RunBenchmarks()
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 324, in RunBenchmarks
    RunBenchmark(benchmark, collector)
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 229, in RunBenchmark
    benchmark.Prepare(benchmark_specification)
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/benchmarks/fio_benchmark.py", line 65, in Prepare
    disk_size_kb = vm.GetDeviceSizeFromPath(vm.GetScratchDir())
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/virtual_machine.py", line 405, in GetDeviceSizeFromPath
    return int(stdout)
ValueError: invalid literal for int() with base 10: '210411904 16419404 '
```

While omitting results in the only marginally more helpful error message:
```
Traceback (most recent call last):
  File "./pkb.py", line 21, in <module>
    sys.exit(Main())
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 358, in Main
    return RunBenchmarks()
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 324, in RunBenchmarks
    RunBenchmark(benchmark, collector)
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 229, in RunBenchmark
    benchmark.Prepare(benchmark_specification)
  File "/home/csb/prog/github/PerfKitBenchmarker/perfkitbenchmarker/benchmarks/fio_benchmark.py", line 65, in Prepare
[...]

PR #160, opened 2015-03-10
Pass '-P' to df.

cmccoy:
Which enforces the output format as described at
http://pubs.opengroup.org/onlinepubs/007904975/utilities/df.html

Fix for #159. Tested on GCE against Ubuntu {14.04,12.04,14.10}, CentOS{6,7}, RHEL7.

### DIFF
--- a/perfkitbenchmarker/virtual_machine.py
+++ b/perfkitbenchmarker/virtual_machine.py
@@ -395,12 +395,12 @@ class BaseVirtualMachine(resource.BaseResource):
 
     Args:
       path: The function will return the amount of space on the file system
-            that conatins this file name.
+            that contains this file name.
 
     Returns:
       The size in 1K blocks of the file system containing the file.
     """
-    df_command = 'df -k | grep %s | awk \'{ print $2 }\'' % (path)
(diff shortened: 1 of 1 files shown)


cmccoy:
@akrzos - any chance you'd be willing to test this against your static VM?

akrzos:
@cmccoy I tested it.  The -P options fixes the issue with the version of df for this static vm.  However there are other issues with static_virtual_machine.py  My scratch directory was always appearing to be / rather than what I had assigned it to be /scratch0.  This is a problem with [GetScratchDir](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/df-fix/perfkitbenchmarker/static_virtual_machine.py#L91)  scratch_dir which is a string is indexed to a one character string.

Additionally, if you were to configure a directory that is not /scratch0 as the scratch mountpoint the fio.job file will fail however that is a different issue.  Do you need me to open another issue for this or is /scratch0 a constraint to PerfKitBenchmarker?

voellm:
LGTM for the -P

cmccoy:
Merging this. I believe the scratch disk issue is being tracked in #127.

cmccoy:
Opened #161 to track the fio issue.

PR #164, opened 2015-03-11
Oldisim integration into Perfkitbenchmarker

huinghuang:
The following changes are made:
- add oldisim in README.md
- add oldisim_dependencies.py in /packages
- add oldisim_benchmark.py in /benchmarks
- add oldisim tarball in /data (need to be removed & use git clone after oldisim repo goes public)


cmccoy:
add to whitelist.

voellm:
Huing - Can we move this to pull the source from GitHub vs checking in a tarball?  You can label builds on GitHub to be sure you pull exactly what you want.

huinghuang:
I planned to replace the tarball with git clone after oldisim goes public.  Currently the oldisim git clone in a vm will fail since it tries to access a private project. Any suggested solution for this?

Issue #155, opened 2015-03-07
Please record timestamp when benchmark begins


timothyeharris:
Right now, we get the timestamp of when a benchmark finishes along with end-to-end runtime.  The end-to-end runtime is inclusive of setting up the environment and booting the machine, etc.

It would be nice if we got a timestamp when the whole process starts, a timestamp when the benchmark itself starts, a timestamp when the benchmark ends, and a timestamp when the whole process ends.  IMO, the "end-to-end runtime" field could be deprecated.  I'd rather do the math myself :-)

This will enable more apples-to-apples comparisons between instance types.

akrzos:
:+1: 
When running against a static vm, I still receive an end-to-end runtime metric too.  It would definitely be useful to include timestamps / runtimes throughout all stages of PerfKitBenchmarker (Prepare, run, cleanup, total)

voellm:
Skyler want to take this on?

skschneider:
Sure, sounds pretty straightforward and useful.

skschneider:
More detailed timing information has been added in #171. Timing information for each benchmark is reported as a sample, similar to End to End Runtime. You can enable the most information with --timing_measurements=runtimes,timestamps

PR #171, opened 2015-03-17
Add per-phase timestamp Samples

skschneider:
Add four new samples per phase for each of the Prepare, Run, and
Cleanup phases:
- The start and stop time of the corresponding benchmark module's
  function for that phase.
- The start and stop time for all work associated with that phase,
  including any work required by the BenchmarkSpec object.


skschneider:
test this please

perfkitbenchmarker/pkb.py
```diff
@@ -195,35 +258,46 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
         logging.exception('Prerequisite check failed for %s', benchmark_name)
         raise
 
-    start_time = time.time()
+    # Initialize timed interval objects.
+    end_to_end_interval = timed_interval.TimedInterval('End to End')
+    prepare_phase_interval = timed_interval.TimedInterval('Prepare Phase')
```
cmccoy: Recommend converting "Prepare Phase" to "Resource Provisioning", exclusive of "Benchmark Prepare" time.

perfkitbenchmarker/pkb.py
```diff
@@ -195,35 +258,46 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
...
+    bm_prepare_interval = timed_interval.TimedInterval('Benchmark Prepare')
+    run_phase_interval = timed_interval.TimedInterval('Run Phase')
```
cmccoy: What's the difference between "Run Phase" and "Benchmark Run"?

perfkitbenchmarker/pkb.py
```diff
@@ -195,35 +258,46 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
...
+    bm_run_interval = timed_interval.TimedInterval('Benchmark Run')
+    cleanup_phase_interval = timed_interval.TimedInterval('Cleanup Phase')
+    bm_cleanup_interval = timed_interval.TimedInterval('Benchmark Cleanup')
```
cmccoy: As above, recommend separating these into "Resource Teardown" and "Benchmark Cleanup".

perfkitbenchmarker/timed_interval.py
```diff
@@ -0,0 +1,67 @@
...
+
+  def __init__(self, name):
+    """Create a named interval.
+
+    Args:
+      name: A string used to prefix the generated samples' names.
+    """
+    self._name = name
+    self._start_time = None
+    self._stop_time = None
```
cmccoy: Recommend making these three attributes public.

perfkitbenchmarker/timed_interval.py
```diff
@@ -0,0 +1,67 @@
...
+    Returns:
+      A list of Samples. If Measure has not been called, the list is empty.
+      Otherwise, the list contains at least two Samples. The first Sample
+      contains the timestamp before the enclosed interval, and the second
+      contains the timestamp after the enclosed interval. If include_runtime is
+      True, a third sample is included that contains the elapsed time between
+      the start and stop timestamps.
+    """
+    samples = []
+    if self._stop_time is not None:
```
cmccoy: Consider verifying that neither or both start_time and stop_time are `None`

perfkitbenchmarker/timed_interval.py
```diff
@@ -0,0 +1,67 @@
...
+      samples.append(sample.Sample(
+          self._name + ' Start Timestamp', self._start_time, 'seconds'))
+      samples.append(sample.Sample(
+          self._name + ' Stop Timestamp', self._stop_time, 'seconds'))
```
cmccoy: Run time seems more useful to me than the absolute timestamps.

skschneider: Original issue (#155) favored timestamps in place of run time, presumably because timestamps give more information and can be used to calculate run time. I selectively enabled run time only to maintain the existing end to end runtime metric. Would you prefer having both timestamps and run times included in the results?

cmccoy: I'd lean toward run times only by default (1 sample versus 2)/phase, or omitting these samples entirely unless enabled by a flag.  As it stands now, there's a lot of noise for users just interested in the benchmark result. For example, running Coremark produces one performance sample and 15 timing samples:

```
-------------------------PerfKitBenchmarker Results Summary-------------------------
COREMARK:
        Coremark Score 43544.524276
        Prepare Phase Start Timestamp 1426612739.59 seconds
        Prepare Phase Stop Timestamp 1426612858.99 seconds
        Benchmark Prepare Start Timestamp 1426612834.24 seconds
        Benchmark Prepare Stop Timestamp 1426612858.99 seconds
        Run Phase Start Timestamp 1426612858.99 seconds
        Run Phase Stop Timestamp 1426612899.91 seconds
        Benchmark Run Start Timestamp 1426612858.99 seconds
        Benchmark Run Stop Timestamp 1426612899.91 seconds
        Cleanup Phase Start Timestamp 1426612899.91 seconds
        Cleanup Phase Stop Timestamp 1426612980.97 seconds
        Benchmark Cleanup Start Timestamp 1426612899.91 seconds
        Benchmark Cleanup Stop Timestamp 1426612902.35 seconds
        End to End Start Timestamp 1426612739.59 seconds
        End to End Stop Timestamp 1426612980.97 seconds
        End to End Runtime 241.380285978 seconds
```

perfkitbenchmarker/pkb.py
skschneider: For each of the phases, I added two sets of start/stop timestamps: one set for the entire phase, and one set for just the benchmark module's associated Prepare/Run/Cleanup function. "Benchmark Run" is timing just the call to the benchmark module's Run function. "Run Phase" includes the BEFORE_RUN_HOOKS and the collector.AddSamples call, which would also be called if running just pkb's run phase.

perfkitbenchmarker/timed_interval.py
skschneider: The flag idea sounds good. I think it's reasonable that if someone wants specific phase timing information, the printed summary is not the best way to extract that information, so getting detailed results would be fine if a detailed timing flag is enabled.

perfkitbenchmarker/pkb.py
skschneider: Changed in 48d9e23.

perfkitbenchmarker/pkb.py
skschneider: Removed overlapping phase measurements in 48d9e23.

perfkitbenchmarker/pkb.py
skschneider: Changed in 48d9e23.

perfkitbenchmarker/timed_interval.py
skschneider: Changed in 48d9e23.

perfkitbenchmarker/timed_interval.py
skschneider: Added in 48d9e23.

skschneider:
test this please

perfkitbenchmarker/pkb.py
```diff
@@ -114,6 +113,12 @@
                     'characters in length.')
 flags.DEFINE_string('owner', getpass.getuser(), 'Owner name. '
                     'Used to tag created resources and performance records.')
+flags.DEFINE_boolean('enable_detailed_timing', False, 'Includes detailed '
```
cmccoy: Maybe consider an enum here for different verbosities:
* `none`: No timing samples emitted.
* `end_to_end`: Only the end-to-end runtime (current behavior, default).
* `phase_runtime`: Runtime by phase
* `phase_timestamps`: Runtime by phase + timestamps.

perfkitbenchmarker/timed_interval.py
```diff
@@ -0,0 +1,87 @@
...
+  def name(self):
+    return self._name
+
+  @property
+  def start_time(self):
+    return self._start_time
+
+  @property
+  def stop_time(self):
+    return self._stop_time
```
cmccoy: Nit: this is a bit verbose. Generally we would just make these public attributes rather than properties.

skschneider:
test this please

perfkitbenchmarker/pkb.py
skschneider: Converted it to a list in 30b489d. Runtimes and timestamps can be specified individually.

perfkitbenchmarker/timed_interval.py
skschneider: Fixed in 30b489d.

perfkitbenchmarker/pkb.py
```diff
@@ -195,35 +257,52 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
...
+          DoRunPhase(
+              benchmark, benchmark_name, spec, collector, bm_run_interval)
+
+        if FLAGS.run_stage in [STAGE_ALL, STAGE_CLEANUP]:
+          DoCleanupPhase(
+              benchmark, benchmark_name, spec, bm_cleanup_interval,
+              resource_teardown_interval)
+
+      # Add samples for any timed interval that was measured.
+      include_end_to_end = timing_util.TimingMeasurementsFlag.end_to_end_runtime
```
cmccoy: It looks like members of TimingMeasurementsFlag should be capitalized.

cmccoy: NM, I missed the Initialize method.

skschneider:
test this please

skschneider:
test this please

perfkitbenchmarker/pkb.py
```diff
@@ -195,35 +257,55 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
...
 
-    start_time = time.time()
+    # Initialize timed interval objects.
+    TimedInterval = timing_util.TimedInterval
+    end_to_end_interval = TimedInterval('End to End')
+    resource_provisioning_interval = TimedInterval('Resource Provisioning')
+    bm_prepare_interval = TimedInterval('Benchmark Prepare')
+    bm_run_interval = TimedInterval('Benchmark Run')
+    bm_cleanup_interval = TimedInterval('Benchmark Cleanup')
+    resource_teardown_interval = TimedInterval('Resource Teardown')
```
cmccoy: This is a big diff to enable timing. 

I wonder if this would be cleaner with an `IntervalTimer`, with a `Measure(self, interval_name)` method which could be invoked multiple times. End to end might need special treatment, but the rest can be handled identically, and you'd only need to pass a single extra arg to `DoPrepare`, etc.

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+
+from contextlib import contextmanager
+import time
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import flags_validators
+from perfkitbenchmarker import sample
+
+
+# Name of the measurements configuration flag as a user would type it.
```
cmccoy: Nit: superfluous comment.

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+MEASUREMENTS_FLAG_NAME = 'timing_measurements'
+# Valid options that can be included in the flag's list value.
+MEASUREMENTS_NONE = 'NONE'
```
cmccoy: Please make the values lower case. We haven't been 100% consistent on this, but that's the preferred format.
Please document what each flag value means.

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+MEASUREMENTS_ALL = (
+    MEASUREMENTS_NONE, MEASUREMENTS_END_TO_END_RUNTIME, MEASUREMENTS_RUNTIMES,
+    MEASUREMENTS_TIMESTAMPS)
+
+
+def ValidateMeasurementsFlag(options_list):
+  """Verifies correct usage of the measurements configuration flag.
+
+  The user of the flag must provide at least one option. All provided options
+  must be valid. The NONE option cannot be combined with other options.
```
cmccoy: Could no value be used in place of NONE?

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+
+  Returns:
+    True if the list of options provided as the value for the flag meets all
+    the documented requirements.
+
+  Raises:
+    flags_validators.Error: If the list of options provided as the value for
+      the flag does not meet the documented requirements.
+  """
+  if len(options_list) is 0:
```
cmccoy: Prefer `not options_list`. Also, `==` should be preferred in pretty much all equality comparisons other than `is None` / `is not None`

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+
+  Args:
+    option: A string containing a valid option for the timing measurements
+      configuration flag.
+
+  Returns:
+    A Boolean that is True if the specified option is present in the global
+    measurements configuration flag value, or False if it is not present.
+  """
+  return option in getattr(flags.FLAGS, MEASUREMENTS_FLAG_NAME)
```
cmccoy: Prefer accessing the flag value directly to `getattr`.
Recommend replacing this method with `IsEndToEndRuntimeEnabled`, `IsPhaseRuntimeEnabled`, etc or similar.

perfkitbenchmarker/timing_util.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+flags.DEFINE_list(
+    MEASUREMENTS_FLAG_NAME, MEASUREMENTS_END_TO_END_RUNTIME,
+    'Comma-separated list of values from <%s> that selects which timing '
+    'measurements to enable.' % '|'.join(MEASUREMENTS_ALL))
```
cmccoy: Please briefly describe what the different options mean here.

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class ValidateMeasurementsFlagTestCase(unittest.TestCase):
+  """Tests exercising ValidateMeasurementsFlag."""
+
+  def testEmptyList(self):
+    """Passing an empty list is not allowed."""
+    e = None
+    try:
+      timing_util.ValidateMeasurementsFlag([])
```
cmccoy: Use `assertRaises` or `assertRaisesRegexp`.

cmccoy: (https://docs.python.org/2/library/unittest.html?highlight=unittest#unittest.TestCase.assertRaises)

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+    except flags_validators.Error as e:
+      pass
+    self.assertIsNotNone(e)
+    self.assertEqual(str(e), 'option --timing_measurements requires argument')
+
+  def testInvalidValue(self):
+    """Passing an unrecognized value is not allowed."""
+    e = None
+    try:
+      timing_util.ValidateMeasurementsFlag(['TEST'])
```
cmccoy: Use `assertRaises` or `assertRaisesRegexp`.

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+    except flags_validators.Error as e:
+      pass
+    self.assertIsNotNone(e)
+    self.assertEqual(str(e), 'TEST: Invalid value for --timing_measurements')
+
+  def testNoneWithAnother(self):
+    """Passing NONE with another value is not allowed."""
+    e = None
+    try:
+      timing_util.ValidateMeasurementsFlag(['NONE', 'RUNTIMES'])
```
cmccoy: Use `assertRaises` or `assertRaisesRegexp`.

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+    self.assertIsNotNone(e)
+    self.assertEqual(
+        str(e), 'NONE: Cannot combine with other --timing_measurements options')
+
+  def testValid(self):
+    """Test various valid combinations."""
+    validate = timing_util.ValidateMeasurementsFlag
+    e = None
+    try:
+      self.assertEqual(validate(['NONE']), True)
```
cmccoy: Nit: use assertTrue here.

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+      self.assertEqual(validate(['END_TO_END_RUNTIME']), True)
+      self.assertEqual(validate(['RUNTIMES']), True)
+      self.assertEqual(validate(['TIMESTAMPS']), True)
+      self.assertEqual(validate(['END_TO_END_RUNTIME', 'RUNTIMES']), True)
+      self.assertEqual(validate(['END_TO_END_RUNTIME', 'TIMESTAMPS']), True)
+      self.assertEqual(validate(['RUNTIMES', 'TIMESTAMPS']), True)
+      self.assertEqual(
+          validate(['END_TO_END_RUNTIME', 'RUNTIMES', 'TIMESTAMPS']),
+          True)
+    except flags_validators.Error as e:
```
cmccoy: Let this fail rather than catching the error.

tests/timing_util_test.py
```diff
@@ -0,0 +1,171 @@
...
+    self.assertIsNone(interval.stop_time)
+    with interval.Measure():
+      pass
+    self.assertIsNotNone(interval.start_time)
+    self.assertIsNotNone(interval.stop_time)
+    self.assertTrue(interval.start_time <= interval.stop_time)
+    samples = interval.GenerateSamples(
+        include_runtime=True, include_timestamps=False)
+    self.assertEqual(len(samples), 1)
+    self.assertEqual(samples[0].metric, 'Test Interval Runtime')
```
cmccoy: Consider `assertEqual(samples[0], <expected value>)` rather than testing all properties individually.

perfkitbenchmarker/pkb.py
skschneider: Good idea, will do.

perfkitbenchmarker/timing_util.py
skschneider: To clarify, you mean the string values, not the variable names?

perfkitbenchmarker/timing_util.py
skschneider: If none were the default value, this would be fine, because the flag could be omitted, but to match previous behavior, the default is END_TO_END_RUNTIME. As a result, the user would need to specify --timing_measurements='' or similar to mark an empty value, and that would be more confusing than --timing-measurements=none.

perfkitbenchmarker/timing_util.py
cmccoy: Exactly.

tests/timing_util_test.py
skschneider: I think assertIs(value, True) would be most appropriate here. assertTrue and assertEqual both have behaviors where non-True values pass (ex: ["non-empty list"] for assertTrue, or 1 for assertEqual).

perfkitbenchmarker/timing_util.py
cmccoy: Sorry, I meant both: passing either `--timing-measurements=none` or `--timing-measurements=""` would prevent any runtime measurements.

skschneider: I originally thought the empty string was ambiguous and could lead to unintended behavior. But on second thought, --timing_measurements without a value is disallowed by gflags, so a user would need to specifically provide a list that is empty to get the no-value behavior from this validator, and if the user says "I want all the timing measurements in this list" and then gives an empty list, it makes sense that none are collected. I guess I will change this.

skschneider:
test this please

perfkitbenchmarker/pkb.py
skschneider: Changed in d856f33.

perfkitbenchmarker/timing_util.py
skschneider: Removed in d856f33.

perfkitbenchmarker/timing_util.py
skschneider: Changed in d856f33.

perfkitbenchmarker/timing_util.py
skschneider: Changed to allow in d856f33.

perfkitbenchmarker/timing_util.py
skschneider: Removed in d856f33 as part of allowing empty lists to be treated like the none option.

perfkitbenchmarker/timing_util.py
skschneider: Changed in d856f33.

perfkitbenchmarker/timing_util.py
skschneider: Added descriptions in MEASUREMENTS_ALL dict in d856f33.

tests/timing_util_test.py
skschneider: Changed in d856f33.

tests/timing_util_test.py
skschneider: Changed in d856f33.

tests/timing_util_test.py
skschneider: Changed in d856f33.

tests/timing_util_test.py
skschneider: Changed in d856f33.

tests/timing_util_test.py
skschneider: Changed in d856f33.

cmccoy:
This is looking good. A few small comments.

skschneider:
test this please

cmccoy:
Two small things, otherwise LGTM.

skschneider:
test this please

skschneider:
Test failures seem to have occurred in setup (git checkout). Trying rerun.

test this please

skschneider:
test this please

PR #202, opened 2015-04-02
New nosql metric

voellm:
This pull requests adds new metrics for the Redis and Aerospike benchmarks.  What those benchmarks publish today is a latency vs throughput series.  That is very useful.  What this adds is a decision point of publishing the max throughput using 1ms.

1ms was chosen because this is how many users rate and tune the system.

This does not modify the benchmark or any of the existing published metrics.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# v0.11.1
+
+* Fixed CHANGES.md
+
 # v0.11.0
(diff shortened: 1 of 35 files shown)


voellm:
Looks like a bad push... hold on reviewing.

voellm:
test this please

PR #203, opened 2015-04-02
New nosql metric

voellm:
This pull requests adds new metrics for the Redis and Aerospike benchmarks. What those benchmarks publish today is a latency vs throughput series. That is very useful. What this adds is a decision point of publishing the max throughput using 1ms.

1ms was chosen because this is how many users rate and tune the system.

This does not modify the benchmark or any of the existing published metrics.

### DIFF
--- a/perfkitbenchmarker/benchmarks/aerospike_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/aerospike_benchmark.py
@@ -191,10 +191,11 @@ def Run(benchmark_spec):
     samples.append(sample.Sample('Average Latency', latency, 'ms', metadata))
     if latency < 1.0:
       max_throughput_for_completion_latency_under_1ms = tps
-      
-  samples.append(sample.Sample('max_throughput_for_completion_latency_under_1ms',
-                                max_throughput_for_completion_latency_under_1ms ,
-                                'req/s'))
+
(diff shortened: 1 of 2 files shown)


voellm:
PTAL

perfkitbenchmarker/benchmarks/aerospike_benchmark.py
```diff
@@ -188,6 +189,13 @@ def ParseOutput(output):
     if FLAGS.aerospike_storage_type == DISK:
       metadata['Disk Type'] = 'Local' if FLAGS.use_local_disk else 'Remote'
     samples.append(sample.Sample('Average Latency', latency, 'ms', metadata))
+    if latency < 1.0:
+      max_throughput_for_completion_latency_under_1ms = tps
```
cmccoy: Is `tps` guaranteed to be monotonically increasing (that is: could `max_throughput_for_completion_latency_under_1m` ever decrease here)?

voellm: That is a very good point.  Indeed it may not be.  Let me fix this.

voellm:
PTAL - Did a full run of Aerospike and Redis.

cmccoy:
LGTM.

PR #223, opened 2015-04-24
Add the option to skip installing packages

ehankland:
Add the option to skip installing packages (for static vms only)

### DIFF
--- a/perfkitbenchmarker/package_managers.py
+++ b/perfkitbenchmarker/package_managers.py
@@ -161,6 +161,8 @@ class YumMixin(BasePackageMixin):
 
   def Install(self, package_name):
     """Installs a PerfKit package on the VM."""
+    if self.is_static and not self.install_packages:
+      return
     if package_name not in self._installed_packages:
       package = packages.PACKAGES[package_name]
       package.YumInstall(self)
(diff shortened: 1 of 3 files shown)


ehankland:
test this please.

perfkitbenchmarker/package_managers.py
```diff
@@ -161,6 +161,8 @@ def InstallPackageGroup(self, package_group):
 
   def Install(self, package_name):
     """Installs a PerfKit package on the VM."""
+    if self.is_static and not self.install_packages:
```
voellm: Is these a reason to restrict this to only static VMs?  I can see this being useful for even custom images.

perfkitbenchmarker/package_managers.py
```diff
@@ -237,6 +239,8 @@ def InstallPackages(self, packages):
...
```
voellm: Is these a reason to restrict this to only static VMs? I can see this being useful for even custom images.

voellm:
Added some comments in the code.  Just ping me to look again once you decide how to go with the feature.

ehankland:
Added a flag override - please take another look

voellm:
LGTM

PR #226, opened 2015-04-28
Fix a DivideByZeroError in side-by-side

cmccoy:
@voellm: can you take a look? Tested with 0 in one of the sample pairs.


voellm:
LGTM

PR #227, opened 2015-04-28
Adds a "force_delete" flag

ehankland:
Adds an override so that resources can be deleted even when things go wrong by specifying
--force_delete=True --run_stage=cleanup --benchmarks=*your_benchmarks_here* 
--run_uri=*your_uri_here*


ehankland:
test this please.

voellm:
Eric and I talked about this and rather than have a new flag we are going to make cleanup more robust and try to cleanup anything that might have been created.

PR #231, opened 2015-04-29
Add ability to inject environment variables to IssueCommand

meteorfox:
Adds an optional parameter to the functions Issue*Command() that  allows a user to set append/overwrite/unset environment variables  before the command is issued. Included tests demonstrate usage.

Background:
For some providers like Rackspace, and OpenStack, the command line utilities provided by these need certain environments variables to be set to pass configuration data such as credentials, endpoints, regions, etc. Before this patch, PerfKitBenchmarker ran with the same environment variables which the process was started with, and did not provide a way to extend, overwrite, or unset environment variables.

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -324,7 +324,7 @@ def Retry(poll_interval=POLL_INTERVAL, max_retries=MAX_RETRIES,
   return Wrap
 
 
-def IssueCommand(cmd, force_info_log=False, suppress_warning=False):
+def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
   """Tries running the provided command once.
 
   Args:
@@ -337,13 +337,24 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False):
         not be logged at the info level in the event of a non-zero
         return code. When force_info_log is True, the output is logged
         regardless of suppress_warning's value.
+    env: A dict of key/value strings, such as is given to the subprocess.Popen()
+        constructor, that contains environment variables to be injected.
 
   Returns:
     A tuple of stdout, stderr, and retcode from running the provided command.
   """
+
+  if env is not None:
+    cmd_env = env
+  else:
+    cmd_env = os.environ.copy()
+
+  logging.debug('Environment variables: %s' % cmd_env)
+
   full_cmd = ' '.join(cmd)
   logging.info('Running: %s', full_cmd)
-  process = subprocess.Popen(cmd,
+
+  process = subprocess.Popen(cmd, env=cmd_env,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
   stdout, stderr = process.communicate()
@@ -361,23 +372,33 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False):
   return stdout, stderr, process.returncode
 
 
-def IssueBackgroundCommand(cmd, stdout_path, stderr_path):
+def IssueBackgroundCommand(cmd, stdout_path, stderr_path, env=None):
   """Run the provided command once in the background.
 
   Args:
     cmd: Command to be run, as expected by subprocess.Popen.
     stdout_path: Redirect stdout here. Overwritten.
     stderr_path: Redirect stderr here. Overwritten.
+    env: A dict of key/value strings, such as is given to the subprocess.Popen()
+        constructor, that contains environment variables to be injected.
   """
+  if env is not None:
+    cmd_env = env
+  else:
+    cmd_env = os.environ.copy()
+
+  logging.debug('Environment variables: %s' % cmd_env)
+
   full_cmd = ' '.join(cmd)
   logging.info('Spawning: %s', full_cmd)
   outfile = open(stdout_path, 'w')
   errfile = open(stderr_path, 'w')
-  subprocess.Popen(cmd, stdout=outfile, stderr=errfile, close_fds=True)
+  subprocess.Popen(cmd, env=cmd_env,
+                   stdout=outfile, stderr=errfile, close_fds=True)
 
 
 @Retry()
-def IssueRetryableCommand(cmd):
+def IssueRetryableCommand(cmd, env=None):
   """Tries running the provided command until it succeeds or times out.
 
   Args:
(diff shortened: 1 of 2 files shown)


tests/vm_util_test.py
```diff
@@ -17,10 +17,80 @@
...
+  def testOverwriteEnvArgument(self):
+    with mock.patch.dict('os.environ', {'perfkit': 'old'}, clear=True):
+      with mock.patch('perfkitbenchmarker.vm_util.subprocess') as subprocess:
+        process_mock = mock.Mock()
+        attributes = {'communicate.return_value': ('output', 'error',),
+                      'returncode': 0}
+        process_mock.configure_mock(**attributes)
+        subprocess.Popen.return_value = process_mock
+
+        env = dict({'perfkit': 'old'}, **{'perfkit': 'new'})
```
cmccoy: Can this just be `{'perfkit': 'new'}`?

tests/vm_util_test.py
```diff
@@ -17,10 +17,80 @@
...
+  def testAppendEnvArgument(self):
+    with mock.patch.dict('os.environ', {'perfkit': 'old'}, clear=True):
+      with mock.patch('perfkitbenchmarker.vm_util.subprocess') as subprocess:
+        process_mock = mock.Mock()
+        attributes = {'communicate.return_value': ('output', 'error',),
+                      'returncode': 0}
+        process_mock.configure_mock(**attributes)
+        subprocess.Popen.return_value = process_mock
+
+        env = dict({'perfkit': 'old'}, **{'newkey': 'newvalue'})
```
cmccoy: Could you simplify this to `{'perfkit': 'old', 'newkey': 'newvalue'}`?

cmccoy:
Thanks for this! Is the expectation that the client will merge `os.environ` with desired environment variable updates before calling IssueCommand?

meteorfox:
Yes, originally I was going to let the user  specify a dict for environment, and then just do the merge with `os.environ` automatically, but that makes it harder to fully unset an environment variable, I didn't want to have empty environment variables. 

I decided to go with the same approach as subprocess.Popen(), which is a full env replace, not merge.

tests/vm_util_test.py
meteorfox: Done

tests/vm_util_test.py
meteorfox: Done

perfkitbenchmarker/vm_util.py
```diff
@@ -362,23 +373,33 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False):
...
     cmd: Command to be run, as expected by subprocess.Popen.
     stdout_path: Redirect stdout here. Overwritten.
     stderr_path: Redirect stderr here. Overwritten.
+    env: A dict of key/value strings, such as is given to the subprocess.Popen()
+        constructor, that contains environment variables to be injected.
   """
+  if env is not None:
+    cmd_env = env
+  else:
+    cmd_env = os.environ.copy()
```
cmccoy: I think passing `env` directly should be fine - `None` is the default value; as best I can tell the Popen constructor selects `os.execvp` or [`os.execvpe`](https://docs.python.org/2/library/os.html#os.execvpe) based on whether `env` is None.

cmccoy:
> I decided to go with the same approach as subprocess.Popen(), which is a full env replace, not merge.
Sounds good - thanks for explaining.

tests/vm_util_test.py
```diff
@@ -17,10 +17,56 @@
 import unittest
 
 import mock
+import os
```
meteorfox: I tried coming up with some mock/patch to capture the execvp/execvpe function calls, but ran into a ton of problems. Took a platform-specific approach, since it executes /usr/bin/env and looks at the stdout.

What are you thoughts? I feel like I should just drop these tests, since all I'm doing is passing a parameter through the function to pass to Popen(), and not adding any logic around it.

tests/vm_util_test.py
```diff
@@ -17,10 +17,56 @@
...
 
 from perfkitbenchmarker import vm_util
 
 
+if os.name is 'posix':
```
meteorfox: I tried coming up with some mock/patch to capture the `execvp/execvpe` function calls, but ran into a ton of problems. Took a platform-specific approach, since it executes `/usr/bin/env` and looks at the stdout.

What are you thoughts? I feel like I should just drop these tests, since all I'm doing is passing a parameter through the function to pass to `Popen()`, and not adding any logic around it.

cmccoy: Sorry to be slow responding to this - I was traveling last week and missed this update.

I agree that this is overkill since there's no new logic.  You could mock `subprocess.Popen` itself (and verify that it was constructed properly). Otherwise removing the test is fine.

meteorfox: I think I'm just going to drop the tests, even if I test for making sure subprocess.Popen gets the right argument, I wouldn't be testing any new logic, it will just be a "variable assignment".

cmccoy: Sounds good.

On Tue, May 5, 2015 at 8:48 AM, Carlos Torres <notifications@github.com>
wrote:

> In tests/vm_util_test.py
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/231#discussion_r29682144>
> :
>
> >
> >  from perfkitbenchmarker import vm_util
> >
> >
> > +if os.name is 'posix':
>
> I think I'm just going to drop the tests, even if I test for making sure
> subprocess.Popen gets the right argument, I wouldn't be testing any new
> logic, it will just be a "variable assignment".
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/231/files#r29682144>
> .
>



-- 
Connor McCoy | Software Engineer | connormccoy@google.com | 206-876-1657

meteorfox:
Ready for review/merge

cmccoy:
test this please.

cmccoy:
Thanks, @meteorfox!

meteorfox:
You are welcome. Thanks for merging!

voellm:
One thing we need to start working on is a developers guide on the API's we
expose for adding new benchmarks, packages, etc.

I'll open an issue to track it so cool and helpful features can be exposed
more easily.

On Tue, May 5, 2015 at 1:16 PM, Carlos Torres <notifications@github.com>
wrote:

> You are welcome. Thanks for merging!
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/231#issuecomment-99208585>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

PR #236, opened 2015-04-30
Add iops and against device flag.

yuyantingzero:


yuyantingzero:
@stfeng , PTAL

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -33,6 +34,8 @@
 flags.DEFINE_string('fio_jobfile', 'fio.job', 'job file that fio will use')
 flags.DEFINE_integer('memory_multiple', 10,
                      'size of fio scratch file compared to main memory size.')
+flags.DEFINE_boolean('against_device', False,
+                     'Test direct against scratch disk.')
```
stfeng: add a explanation here that this will modify the job file (or ignore the directory and filename part of the job file).

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -74,6 +77,25 @@ def Prepare(benchmark_spec):
     #    invalid. Once the benchmark results are returned from Run() an
     #    invalid (or is that rather a 'valid' flag should be added.
     exit(1)
+  if FLAGS.against_device:
+    logging.info('Fill scratch disk on %s', vm)
```
stfeng: print the device path here in the logging.info.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -92,14 +114,21 @@ def Run(benchmark_spec):
   vms = benchmark_spec.vms
   vm = vms[0]
   logging.info('FIO running on %s', vm)
-  fio_command = '%s --output-format=json --directory=%s %s' % (
-      fio.FIO_PATH, vm.GetScratchDir(), flags.FLAGS.fio_jobfile)
+  if FLAGS.against_device:
+    file_path = vm_util.PrependTempDir(FLAGS.fio_jobfile)
+    fio_command = '--filename=%s %s' % (vm.scratch_disks[0].GetDevicePath(),
```
stfeng: just to double check: the --filename option specified here will apply to all jobs in the job file, right?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -74,6 +77,25 @@ def Prepare(benchmark_spec):
     #    invalid. Once the benchmark results are returned from Run() an
     #    invalid (or is that rather a 'valid' flag should be added.
     exit(1)
+  if FLAGS.against_device:
+    logging.info('Fill scratch disk on %s', vm)
+    command = (
+        ('sudo %s --filename=%s --ioengine=libaio '
+         '--name=fill-device --blocksize=512k --iodepth=64 '
+         '--rw=write --fill_device=1 --direct=1') %
```
stfeng: will --fill_device actually work for raw devices like /dev/sdb? I thought we tried it and it didn't work? 

Alternatively, can we just specify --size=device_size here and let fio write exactly that many bytes to the device? This will also allow people who want to save time and fill just part of the device (and use only that part for read testing).

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: Yes, if specified through command line, it is treated as a global parameter, which applied to all jobs (same as --directory parameter in the orignal test).

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: --fill_device only works with sequential_write from documentation (http://linux.die.net/man/1/fio).

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -33,6 +34,8 @@
 flags.DEFINE_string('fio_jobfile', 'fio.job', 'job file that fio will use')
 flags.DEFINE_integer('memory_multiple', 10,
                      'size of fio scratch file compared to main memory size.')
+flags.DEFINE_boolean('against_device', False,
```
stfeng: Thinking more about this, we will likely need to add another flag here:

flags.DEFINE_integer(device_fill_size, -1, "the amount of bytes to write to the device when against_device is set to true. Default -1 means fill the entire device.")

This way, for some providers that do not require pre-warm of the volume, the user can use this parameter to only fill part of the device, e.g., 10GB, and then use that part for subsequent read testing (by specifying --size=10GB in their read job).

perfkitbenchmarker/packages/fio.py
```diff
@@ -153,4 +153,23 @@ def ParseResults(job_file, fio_json_result):
             sample.Sample('%s:latency' % metric_name,
                           job[mode]['lat']['mean'],
```
stfeng: the parsing here is not entirely correct: we should get 'clat' rather than 'lat'

and the unit is not always usec, some times it's msec, fio changes the unit dynamically based on the outcome.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: Sounds good to me

perfkitbenchmarker/packages/fio.py
yuyantingzero: In the parsing it transferred to json format, I believe has consistent unit (always use usec, even the number is large), I will double check

stfeng:
Minor comments mostly, thanks for great work!

yuyantingzero:
Tested locally on fio and block_storage_workload.

PR #238, opened 2015-05-01
Fix the command that creates a storage account for Azure. (was broken du...

stfeng:
...e to a recent Azure CLI syntax change that was not backward compatible)


stfeng:
@yuyantingzero 

A small but important fix. Azure's recent change in CLI syntax broke PKB's azure object storage service test.

yuyantingzero:
LGTM

PR #255, opened 2015-05-07
Lets you not specify --run_uri under certain conditions

ehankland:
Check the /tmp/perfkitbenchmarker directory for the last modified file in order to find the last run. This lets us skip specifying --run_uri when using --run_stage as long as it is the same as the last pkb run.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -294,7 +294,7 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks):
       logging.exception('Error during benchmark %s', benchmark_name)
       # If the particular benchmark requests us to always call cleanup, do it
       # here.
-      if spec.always_call_cleanup:
+      if spec and spec.always_call_cleanup:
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/pkb.py
```diff
@@ -320,9 +321,17 @@ def RunBenchmarks(publish=True):
 
   if FLAGS.run_uri is None:
     if FLAGS.run_stage not in [STAGE_ALL, STAGE_PREPARE]:
-      logging.error(
-          'Cannot run "%s" with unspecified run_uri.', FLAGS.run_stage)
-      return 1
+      # Attempt to get the last modified run directory.
+      run_uri = vm_util.GetLastRunUri()
+      if run_uri:
+        FLAGS.run_uri = run_uri
```
voellm: should you make sure the user has not already set this?  Seems like you might be overriding the user here.

ehankland: This code only gets executed if FLAGS.run_uri is None (see line 322), which means they didn't set it.

voellm:
LGTM

PR #264, opened 2015-05-11
Use the boto API differently when downloading a one byte object to av…

stfeng:
…oid the unnecessary HEAD request (and issue GET request only).


stfeng:
@yuyantingzero PTAL. Thanks!

yuyantingzero:
LGTM

PR #265, opened 2015-05-15
Move VM name to BaseVirtualMachine.

cmccoy:
* Adds _instance_counter and associated lock to BaseVirtualMachine.
* Removes (non-thread-safe) implementations from GCE VM, Azure VM (see
  #118)

PR #266, opened 2015-05-15
Write an SSH config file to the run temp directory

cmccoy:
All VMs have names as of #265.

This change makes it simpler to SSH into VMs by name, to aid debugging.
Two aliases are provided for each VM: the VM name, and vm<index>, where
<index> is a unique number for each VM. The config file sets the same
SSH options used on the command line.

Example usage:

    ssh -F /tmp/perfkitbenchmarker/run_ea2f81a3/ssh_config perfkit-ea2f81a3-0

or (equivalent):

    ssh -F /tmp/perfkitbenchmarker/run_ea2f81a3/ssh_config vm0

### DIFF
--- a/perfkitbenchmarker/data/ssh_config.j2
+++ b/perfkitbenchmarker/data/ssh_config.j2
@@ -1,0 +1,33 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


cmccoy:
test this please.

perfkitbenchmarker/data/ssh_config.j2
```diff
@@ -0,0 +1,20 @@
+Host *
```
yuyantingzero: Add licensing header.

yuyantingzero:
LGTM

cmccoy:
License added. Thanks, @yuyantingzero.

PR #272, opened 2015-05-19
benchmark_sets: add all_set, sort sets in help summary.

klausw:
Add all_set containing all implemented benchmarks, and sort the
list of known sets when printing the usage summary.

### DIFF
--- a/perfkitbenchmarker/benchmark_sets.py
+++ b/perfkitbenchmarker/benchmark_sets.py
@@ -22,6 +22,7 @@ FLAGS = flags.FLAGS
 MESSAGE = 'message'
 BENCHMARK_LIST = 'benchmark_list'
 STANDARD_SET = 'standard_set'
+ALL_SET = 'all_set'
 
 BENCHMARK_SETS = {
     STANDARD_SET: {
@@ -33,6 +34,10 @@ BENCHMARK_SETS = {
             'hadoop_terasort', 'hpcc', 'iperf', 'mesh_network', 'mongodb',
             'netperf', 'ping', 'redis', 'speccpu2006',
             'block_storage_workload', 'sysbench_oltp', 'unixbench']
+    },
+    ALL_SET: {
+        MESSAGE: ('This set contains all implemented benchmarks.'),
+        BENCHMARK_LIST: []  # Special cased below.
     },
     'arm_set': {
         MESSAGE: 'ARM benchmark set.',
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/benchmark_sets.py
```diff
@@ -125,9 +130,18 @@ def GetBenchmarksFromFlags():
...
+  # build the dictionary
+  valid_benchmarks = {}
+  for benchmark_module in benchmarks.BENCHMARKS:
+    valid_benchmarks[benchmark_module.GetInfo()['name']] = benchmark_module
+
   benchmark_names = set()
   for benchmark in FLAGS.benchmarks:
-    if benchmark in BENCHMARK_SETS:
+    if benchmark == ALL_SET:
+      benchmark_names.update(valid_benchmarks)
```
voellm: Do you really want to do this for each benchmark?  Seems like this should be outside the "for" loop.

perfkitbenchmarker/benchmark_sets.py
```diff
@@ -125,9 +130,18 @@ def GetBenchmarksFromFlags():
...
+  # TODO(voellm): I really think the benchmarks package init should
+  # build the dictionary
+  valid_benchmarks = {}
+  for benchmark_module in benchmarks.BENCHMARKS:
+    valid_benchmarks[benchmark_module.GetInfo()['name']] = benchmark_module
+
   benchmark_names = set()
   for benchmark in FLAGS.benchmarks:
-    if benchmark in BENCHMARK_SETS:
+    if benchmark == ALL_SET:
```
voellm: There is one bug here... looks like the user would need to pass "all_set" vs "all".  The _set is assumed.

perfkitbenchmarker/benchmark_sets.py
```diff
@@ -125,9 +130,18 @@ def GetBenchmarksFromFlags():
...
+      benchmark_names.update(valid_benchmarks)
+    elif benchmark in BENCHMARK_SETS:
```
voellm: The code below handles the expansion.  Dont think we need this.

klausw: This is a bit tricky. It would have been nicer to have the ALL_SET entry in BENCHMARK_SETS contain the full BENCHMARK_LIST, but those values aren't available yet at the time BENCHMARK_SETS is assigned, it's only calculated a few lines above in the code block I moved. Could patch the value in at runtime, but it seems dubious to write to a module constant.

The loop iterates over the individual benchmark names provided by the user in the --benchmarks flag value list, and if it encounters the ALL_SET name it replaces that with the list of all known benchmarks. It's not equivalent to the recursive expansion below since the ALL_SET doesn't recursively expand to other sets.

Does that seem ok, or would you prefer me to change it to populate BENCHMARK_SETS[ALL_SET][BENCHMARK_LIST] via patching it in initialization?

perfkitbenchmarker/benchmark_sets.py
klausw: Not sure what you mean by "_set is assumed" - as far as I can tell it's just part of the name and I didn't see any special logic that would automatically add or remove a _set suffix. I named this set 'all_set' for consistency with the other sets that also have '_set' suffixes, assuming that names without '_set' suffix are individual benchmarks and not sets. Should I change the name from 'all_set' to just 'all'?

voellm:
Is this ready for review?  Its not marked which is why I am asking.

klausw:
Marked as ready for review. Overall question is if this is useful - if not, I'm also ok with withdrawing the patch.

voellm:
I still wonder if we should explicitly list out all benchmarks manually.  That said in the future "all" but being auto-generated can select just LINUX or just WINDOWS depending on what is selected.

voellm:
We talked about this offline and decided we should improve the filtering before checking in.  Right now it is possible some benchmarks that should not be attempted, will be, causing everything to fail.  For example if a user does not have the Coremark or SPECCPU tarballs in the data directory everything will fail.

The plan is to remove benchmarks that dont have the necessary requirements.

klausw:
Closing this pull request for now since it's unclear if this is what we want, can revisit the issue once the overall flags/filtering refactor that @voellm mentioned is going on.

PR #273, opened 2015-05-19
GCE: do not restart on host failure.

cmccoy:
Setting `--maintenance-policy TERMINATE` *without* `--no-restart-on-failure`
will cause the VM to be rebooted on host failure.


cmccoy:
test this please.

yuyantingzero:
LGTM

PR #276, opened 2015-05-20
Cleans up cleanup

ehankland:
running with --run_stage=cleanup should now work to teardown resources in the vast majority of cases


ehankland:
test this please.

perfkitbenchmarker/gcp/gce_network.py
```diff
@@ -30,6 +33,56 @@
...
+
+  def __init__(self, name, project, port):
+    super(GceFirewallRule, self).__init__()
+    self.name = name
+    self.project = project
+    self.port = port
+
+  def __eq__(self, other):
+    """Defines equality to make comparison easy."""
+    return self.__dict__ == other.__dict__
```
cmccoy: Would you consider two firewall rules to be non-equal if the only differences are in `created` or `deleted`?

perfkitbenchmarker/gcp/gce_network.py
```diff
@@ -66,29 +119,16 @@ def AllowPort(self, vm, port):
     with self._lock:
       firewall_name = ('perfkit-firewall-%s-%d' %
                        (FLAGS.run_uri, port))
-      if firewall_name in self.firewall_names:
+      firewall_rule = GceFirewallRule(firewall_name, self.project, port)
+      if firewall_rule in self.firewall_rules:
```
cmccoy: I believe this will always return `False` see: https://docs.python.org/2/reference/datamodel.html#object.__hash__

```
User-defined classes have __cmp__() and __hash__() methods by default; with them, all objects compare unequal (except with themselves) and x.__hash__() returns a result derived from id(x).
```

perfkitbenchmarker/gcp/gce_network.py
ehankland: oops thanks - fixed

perfkitbenchmarker/gcp/gce_network.py
ehankland: oops - it used to be a list, and I wasn't thinking when changed it to a set - changed back

ehankland:
PTAL

perfkitbenchmarker/gcp/gce_network.py
```diff
@@ -30,6 +33,58 @@
...
+    """Returns True if the Firewall Rule exists."""
+    describe_cmd = [FLAGS.gcloud_path,
+                    'compute',
+                    'firewall-rules',
+                    'describe',
+                    self.name]
+    describe_cmd.extend(util.GetDefaultGcloudFlags(self))
+    stdout, _, _ = vm_util.IssueCommand(describe_cmd, suppress_warning=True)
+    try:
+      json.loads(stdout)
```
cmccoy: Maybe check the return code rather than relying on JSON parsing failure to determine whether the rule exists?

cmccoy:
Discussed offline: would you mind adding a comment to `BaseResource._Delete` indicating that subclass implementations should be idempotent?
One other small comment, otherwise LGTM.

PR #279, opened 2015-05-21
Switch MongoDB to use common YCSB code.

cmccoy:
This PR adds a new benchmark: `mongodb_ycsb`, using the code common to `perfkitbenchmarker/packages/ycsb.py`. Differences from the current YCSB benchmark:
* Uses YCSB Executor, to report additional details.
* Stores data on scratch disks.
* Allows a user-configurable number of YCSB client vms, operations, initial database sizes.
* Allows a user-specifiable write concern, defaulting to `safe`.

This better matches other YCSB-based benchmarks, but is still slightly less
stringent than others:

* HBase flushes to 3 replicas by default before acking a write.
* Cassandra is configured in QUORUM write mode.

MongoDB write concern `JOURNALED` would better match these configurations,
but writes are committed in batches, so latency will be quite high.

Also deprecates the `mongodb` benchmark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ In its current release these are the benchmarks that are executed:
   - `iperf`: BSD license(http://iperf.sourceforge.net/)
   - `memtier_benchmark`: GPL v2 (https://github.com/RedisLabs/memtier_benchmark)
   - `mesh_network`: HP license (http://www.calculate-linux.org/packages/licenses/netperf)
-  - `mongodb`: GNU AGPL v3.0 (http://www.mongodb.org/about/licensing/)
+  - `mongodb`: **Deprecated**. GNU AGPL v3.0 (http://www.mongodb.org/about/licensing/)
+  - `mongodb_ycsb`: GNU AGPL v3.0 (http://www.mongodb.org/about/licensing/)
   - `netperf`: HP license (http://www.calculate-linux.org/packages/licenses/netperf)
   - `oldisim`: Apache v2.
   - `object_storage_service`: Apache v2.
--- a/perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
@@ -1,0 +1,112 @@
+# Copyright 2014 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Run YCSB against MongoDB.
+
+YCSB is a load generator for many 'cloud' databases. MongoDB is a NoSQL
+database.
+
+MongoDB homepage: http://www.mongodb.org/
+YCSB homepage: https://github.com/brianfrankcooper/YCSB/wiki
+"""
+
+import os
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker.packages import ycsb
+
+FLAGS = flags.FLAGS
+
+
+BENCHMARK_INFO = {'name': 'mongodb_ycsb',
+                  'description': 'Run YCSB against MongoDB.',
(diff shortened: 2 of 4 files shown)


voellm:
Connor until the community agrees to remove MongoDB in favor of MongoDB_YCSB I think we should keep both benchmarks.

For the google_set can may the change with this checkin.

Make sense?

cmccoy:
The old mongodb benchmark used YCSB, but didn't use very reasonable parameters, specifically:
* The database was initially empty.
* 1000 read/write operations were applied to the initially empty database, so most read operations just fail.
* What little data were written were stored on the machine boot disk.

I don't think the resulting numbers are very informative from a performance standpoint.

I agree that we should get feedback from the community before merging this, but having two YCSB-based benchmarks in PKB seems confusing to me.

voellm:
Please open an issue to deprecate the old version.

On Wed, Jun 3, 2015 at 11:47 AM, Connor McCoy <notifications@github.com>
wrote:

> The old mongodb benchmark used YCSB, but didn't use very reasonable
> parameters, specifically:
>
>    - The database was initially empty.
>    - 1000 read/write operations were applied to the initially empty
>    database, so most read operations just fail.
>    - What little data were written were stored on the machine boot disk.
>
> I don't think the resulting numbers are very informative from a
> performance standpoint.
>
> I agree that we should get feedback from the community before merging
> this, but having two YCSB-based benchmarks in PKB seems confusing to me.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/279#issuecomment-108571526>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

README.md
```diff
@@ -37,7 +37,7 @@ In its current release these are the benchmarks that are executed:
   - `iperf`: BSD license(http://iperf.sourceforge.net/)
   - `memtier_benchmark`: GPL v2 (https://github.com/RedisLabs/memtier_benchmark)
   - `mesh_network`: HP license (http://www.calculate-linux.org/packages/licenses/netperf)
-  - `mongodb`: GNU AGPL v3.0 (http://www.mongodb.org/about/licensing/)
+  - `mongodb_ycsb`: GNU AGPL v3.0 (http://www.mongodb.org/about/licensing/)
```
voellm: should list both mongodb and mongodb_ycsb

cmccoy:
@voellm: updated to leave `mongodb` unchanged, opened #294 to track deprecating it. This PR now just adds `mongodb_ycsb`. PTAL.

perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
```diff
@@ -0,0 +1,116 @@
...
+
+FLAGS = flags.FLAGS
+
+
+BENCHMARK_INFO = {'name': 'mongodb_ycsb',
+                  'description': 'Run YCSB against MongoDB.',
+                  'scratch_disk': True,
+                  'num_machines': 2}
+
+RESULT_REGEX = r'\[(\w+)\], (\w+)\(([\w/]+)\), ([-+]?[0-9]*\.?[0-9]+)'
```
voellm: Its helpful to give an example of what you expect to parse correctly.  From this I can tell if it is correct or not.

perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
```diff
@@ -0,0 +1,116 @@
...
+  Returns:
+    A list of samples in the form of 3 or 4 tuples. The tuples contain
+        the sample metric (string), value (float), and unit (string).
+        If a 4th element is included, it is a dictionary of sample
+        metadata.
+  """
+  vms = benchmark_spec.vms
+  vm = vms[1]
+
+  executor = ycsb.YCSBExecutor('mongodb')
```
voellm: Does the executor log what is running + parameters?  If not a log message here would be useful.

voellm:
LGTM

perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
cmccoy: Errant copy from [mongodb_benchmark.py](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/benchmarks/mongodb_benchmark.py#L39`). This isn't used, removed.

perfkitbenchmarker/benchmarks/mongodb_ycsb_benchmark.py
cmccoy: Yes, it logs all commands run.

Issue #180, opened 2015-03-18
Collect and publish data on machine load during benchmark runs


cmccoy:
e.g., with `dstat`.

cc: @meteorfox

meteorfox:
Other possible tools to collect system utilization data with time range addressing capabilities:

[sysstat package (sar)](http://sebastien.godard.pagesperso-orange.fr/)
[Performance Co-Pilot (pcp)](http://www.pcp.io/)

With time range addressing you can zoom in to particular time intervals, or time spans instead of relying on when the tool was started and stopped.

hungvelo:
It's good and useful to collect system utilization data during benchmark runs but be careful as this may effect performance results for some benchmarks that are latency sensitive. What's about the host itself? Maybe not but is it possible to find out the host system configuration (OS version, CPU, memory, I/O storage, Network, etc.)? :-)?

cmccoy:
Host configuration is being tracked in #183.

cmccoy:
I added dstat (because it's simple to run and parse) in #282. Does that seem reasonable as a first step? PCP and sar sound great too, but I'm not as familiar. #282 includes the unix epoch for each sample, so filtering to time windows of interest is possible.

cmccoy:
Comment by @meteorfox on #282, which sounds like a great idea to me:

>A nother suggestion. It'd be great if we could collect error metrics too (I'm not entirely sure if dstat already does). I'm looking for metrics like dropped packets, TCP re-transmissions and timeouts, storage IO errors, etc. 
This follows the USE method methodology, which monitors utilization, saturation, and error metrics for each main component in a system, and serves as checklist to quickly identify bottlenecks.

> Here's more info: http://www.brendangregg.com/USEmethod/use-linux.html

This won't be addressed by #282, as `dstat` doesn't seem to collect error metrics in the list above.

cmccoy:
The collectd support added in #398 includes TCP errors, but not everything from the USE method @meteorfox mentioned above.

I'm going to close this - if there are specific additional metrics we should be collecting, let's open separate issues for them.

PR #282, opened 2015-05-23
Add a --dstat flag

cmccoy:
When set, runs 'dstat' during the benchmark run, copies the results to
the run temp dir once the benchmark run completes.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -206,8 +206,10 @@ def DoRunPhase(benchmark, name, spec, collector, timer):
   logging.info('Running benchmark %s', name)
   for before_run_hook in BEFORE_RUN_HOOKS:
     before_run_hook(benchmark=benchmark, benchmark_spec=spec)
-  with timer.Measure('Benchmark Run'):
-    samples = benchmark.Run(spec)
+
+  with vm_util.RunDStatIfConfigured(spec.vms, suffix='-{0}-dstat'.format(name)):
+    with timer.Measure('Benchmark Run'):
+      samples = benchmark.Run(spec)
   collector.AddSamples(samples, name, spec)
 
 
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -14,8 +14,10 @@
 
 """Set of utility functions for working with virtual machines."""
 
+import contextlib
 import logging
 import os
+import posixpath
 import random
 import re
 import socket
@@ -56,6 +58,13 @@ flags.DEFINE_integer('default_timeout', TIMEOUT, 'The default timeout for '
 flags.DEFINE_integer('burn_cpu_seconds', 0,
                      'Amount of time in seconds to burn cpu on vm.')
 flags.DEFINE_integer('burn_cpu_threads', 1, 'Number of threads to burn cpu.')
+flags.DEFINE_boolean('dstat', False,
+                     'Run dstat (http://dag.wiee.rs/home-made/dstat/) '
+                     'on each VM to collect system performance metrics during '
+                     'each benchmark run.')
+flags.DEFINE_integer('dstat_interval', None,
+                     'dstat sample collection frequency, in seconds. Only '
+                     'applicable when --dstat is specified.')
 
 
 class IpAddressSubset(object):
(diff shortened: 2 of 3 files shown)


cmccoy:
test this please.

perfkitbenchmarker/vm_util.py
```diff
@@ -501,3 +504,51 @@ def GenerateSSHConfig(vms):
...
+
+  lock = threading.Lock()
+  pids = {}
+  file_names = {}
+
+  def StartDStat(vm):
+    vm.Install('dstat')
+    dstat_file = os.path.join(VM_TMP_DIR, '{0}-dstat.csv'.format(vm.name))
+    cmd = ('nohup dstat --epoch --all --mem --vm --unix -o {0} '
+           '> /dev/null 2>&1 & echo $!').format(dstat_file)
```
voellm: here is a DSTAT command I've used a lot with customers.  It may be a superset of what you have;

'/usr/bin/nohup /usr/bin/dstat -T -C total,%s'
        ' -D total,sda,sdb,sdc'
        ' -clrdngyi -pms --fs --ipc --tcp'
        ' --udp --raw --socket --unix --vm --rpc'
        ' --noheaders --output %s > /dev/null &'
        ' echo $!' %

perfkitbenchmarker/vm_util.py
```diff
@@ -501,3 +504,51 @@ def GenerateSSHConfig(vms):
...
+    yield
+    return
+
+  lock = threading.Lock()
+  pids = {}
+  file_names = {}
+
+  def StartDStat(vm):
+    vm.Install('dstat')
+    dstat_file = os.path.join(VM_TMP_DIR, '{0}-dstat.csv'.format(vm.name))
```
voellm: I think the filename should be a parameter ... or at least part of the filename.  They way you have this right now the user has no control over the name.  If they start and stop a couple of times it looks like they will only get the latest.

I like passing the name because it will allow us to add processors over DSTAT files more easily.

voellm:
I like DSTAT because it has easy to read and understand text output.  An alternative is collectd.

perfkitbenchmarker/pkb.py
```diff
@@ -206,7 +206,7 @@ def DoRunPhase(benchmark, name, spec, collector, timer):
   logging.info('Running benchmark %s', name)
   for before_run_hook in BEFORE_RUN_HOOKS:
     before_run_hook(benchmark=benchmark, benchmark_spec=spec)
-  with timer.Measure('Benchmark Run'):
+  with vm_util.RunDStatIfConfigured(spec.vms), timer.Measure('Benchmark Run'):
```
voellm: This will only work for LINUX and not Windows.  For Windows we should look into xperf.

perfkitbenchmarker/vm_util.py
cmccoy: > I think the filename should be a parameter ... or at least part of the filename. The way you have this right now the user has no control over the name. If they start and stop a couple of times it looks like they will only get the latest.

True, but only if they re-use the same run URI. I can add an additional URI to the file name, or a user-specified suffix, but since each VM will create a local file, a single file name provided by the user won't work.

voellm: Are you getting one dstat file per VM?  Per Benchmark?

On Tue, May 26, 2015 at 9:57 AM, Connor McCoy <notifications@github.com>
wrote:

> In perfkitbenchmarker/vm_util.py
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/282#discussion_r31055176>
> :
>
> > +    vms: List of virtual machines.
> > +
> > +  Yields:
> > +    None
> > +  """
> > +  if not FLAGS.dstat:
> > +    yield
> > +    return
> > +
> > +  lock = threading.Lock()
> > +  pids = {}
> > +  file_names = {}
> > +
> > +  def StartDStat(vm):
> > +    vm.Install('dstat')
> > +    dstat_file = os.path.join(VM_TMP_DIR, '{0}-dstat.csv'.format(vm.name))
>
>  I think the filename should be a parameter ... or at least part of the
> filename. The way you have this right now the user has no control over the
> name. If they start and stop a couple of times it looks like they will only
> get the latest.
>
> True, but only if they re-use the same run URI. I can add an additional
> URI to the file name, or a user-specified suffix, but since each VM will
> create a local file, a single file name provided by the user won't work.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/282/files#r31055176>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

ehankland: I'd prefer posixpath over os.path since posixpath.join() will work on windows controllers targeting linux guests while os.path.join() won't.

cmccoy: @ehankland: done, thanks.
@voellm: one dstat file per VM. You're right, they would be overridden if multiple benchmarks are specified to `--benchmarks`. I can add a uuid, or we could tag each file with the benchmark run. Any preference?

perfkitbenchmarker/vm_util.py
```diff
@@ -56,6 +57,8 @@
 flags.DEFINE_integer('burn_cpu_seconds', 0,
                      'Amount of time in seconds to burn cpu on vm.')
 flags.DEFINE_integer('burn_cpu_threads', 1, 'Number of threads to burn cpu.')
+flags.DEFINE_boolean('dstat', False, 'Run dstat on the VM during the benchmark '
```
meteorfox: Can we give more details in the help message? For example, a short description of what dstat actually does. Honestly, I've never heard of dstat before until this project. :)

Maybe something like:

'Collects system resource metrics, for each VM during the execution of each benchmark'

perfkitbenchmarker/vm_util.py
meteorfox: What's the poll period of the metrics? 1 sec? 5 sec? Could it be configurable?

perfkitbenchmarker/vm_util.py
cmccoy: Good point - thanks!

How about: 

> Run dstat (http://dag.wiee.rs/home-made/dstat/) on each VM to collect system performance metrics during benchmark runs.

perfkitbenchmarker/vm_util.py
cmccoy: 1s by default. I'll add a flag for the collection interval.

meteorfox: Another suggestion. It'd be great if we could collect error metrics too (I'm not entirely sure if dstat already does). I'm looking for metrics like dropped packets, TCP re-transmissions and timeouts, storage IO errors, etc. 
This follows the USE method methodology, which monitors utilization, saturation, and error metrics for each main component in a system, and serves as checklist to quickly identify bottlenecks.

Here's more info: http://www.brendangregg.com/USEmethod/use-linux.html

perfkitbenchmarker/vm_util.py
meteorfox: Yes, much better.

perfkitbenchmarker/vm_util.py
cmccoy: Added benchmark name to file in 3c55a7e5c752847354492b84973f166ae083303d.

perfkitbenchmarker/vm_util.py
cmccoy: Switched to this, with a few small changes.

cmccoy: This is a great idea, but it [doesn't appear](http://dag.wiee.rs/home-made/dstat/dstat.1.html) that dstat can quantify any of the above. I've copied your comment to #180 - maybe we could discuss the pieces not addressed by this PR there?

cmccoy:
@voellm: expanded the metrics collected, added the benchmark name to the file names. PTAL.

cmccoy:
test this please.

voellm:
LGTM

cmccoy:
test this please.

cmccoy:
test this please.

PR #283, opened 2015-05-27
Add cloud_bigtable_ycsb benchmark.

cmccoy:


voellm:
You should add the benchmark to the google_set.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
```diff
@@ -0,0 +1,231 @@
...
+    'https://storage.googleapis.com/cloud-bigtable/jars/'
+    'bigtable-hbase/bigtable-hbase-0.1.5.jar',
+    'URL for the Bigtable-HBase client JAR.')
+
+
+BENCHMARK_INFO = {'name': 'cloud_bigtable_ycsb',
+                  'description': 'Run YCSB against Cloud Bigtable. '
+                  'configure the number of client VMs via --num_vms.',
+                  'scratch_disk': False,
+                  'num_machines': None}
```
voellm: Should this be 1?

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
```diff
@@ -0,0 +1,231 @@
...
+
+  for scope in REQUIRED_SCOPES:
+    if scope not in FLAGS.gcloud_scopes:
+      raise ValueError('Scope {0} required.'.format(scope))
+
+  if not FLAGS.google_bigtable_cluster_name:
+    raise ValueError('Missing --google_bigtable_cluster_name')
+  if not FLAGS.google_bigtable_zone_name:
+    raise ValueError('Missing --google_bigtable_zone_name')
+  if not FLAGS.project:
```
voellm: Is this really a requirement?  Today the project default is configure by the gcloud tool.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
```diff
@@ -0,0 +1,231 @@
...
+
+  hbase.CheckPrerequisites()
+  ycsb.CheckPrerequisites()
+
+  for scope in REQUIRED_SCOPES:
+    if scope not in FLAGS.gcloud_scopes:
+      raise ValueError('Scope {0} required.'.format(scope))
+
+  if not FLAGS.google_bigtable_cluster_name:
+    raise ValueError('Missing --google_bigtable_cluster_name')
```
voellm: I'm ok checking these here but gcloud should support these defaults.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
```diff
@@ -0,0 +1,231 @@
...
+  Raises:
+    perfkitbenchmarker.data.ResourceNotFound: On missing resource.
+  """
+  for resource in HBASE_CONF_FILES:
+    data.ResourcePath(resource)
+
+  hbase.CheckPrerequisites()
+  ycsb.CheckPrerequisites()
+
+  for scope in REQUIRED_SCOPES:
```
voellm: If you know certain scope are required why not just add them?  Seems like you are having the user do busy work by adding the scopes

voellm:
Can't say I understand all the details but overall it looks good.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
cmccoy: Wehn `num_machines` is `None`, it takes the value of `--num_vms`, defaulting to 1. This way, more client VMs can be used if desired.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
cmccoy: We need this to populate the [`hbase-site.xml`](https://github.com/cmccoy/PerfKitBenchmarker/blob/cloud-bigtable-ycsb/perfkitbenchmarker/data/cloudbigtable/hbase-site.xml.j2). I could also get the current default project from `gcloud config list`.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
cmccoy: Yes, but I don't believe there's a way (presently) for the VM spec to modify the specification for VMs used prior to the VMs being created.  Modifying the flag value here would change a global for all benchmarks. Thoughts?

cmccoy:
> You should add the benchmark to the google_set.

I think we should wait until the enhancement @klausw proposed to skip benchmarks without prerequisites. This test will fail if the user doesn't have a Bigtable cluster already set up.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
voellm: Good point... probably dont way this for all benchmarks and introduce a side-effect.

Guess this is the best that can be done.  We should address it as part of the flags refactor

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
voellm: If you can obtain it I think that is better.  So if project is set use it otherwise get it from running gcloud.

cmccoy: Done in 4aaf62668d50a9438cdd036cf9de3c584eceac2a.

perfkitbenchmarker/benchmarks/cloud_bigtable_ycsb_benchmark.py
cmccoy: I don't believe these are stored in the gcloud config. Added a TODO.

cmccoy:
Tested manually.

Issue #288, opened 2015-06-03
Add timeout for vm_util.IssueCommand


ehankland:
We should add a timeout parameter to IssueCommand which will kill any subprocesses that don't complete within the time limit. This will prevent PKB from hanging if any of the tools we use (gcloud, azure, aws, etc) don't exit.

PR #289, opened 2015-06-03
Add an IssueCommand timeout

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/288

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -380,11 +380,12 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
 
-  def _KillProcess(process):
-    logging.warning('IssueCommand timed out. Killing Process.')
+  def _KillProcess():
+    logging.error('IssueCommand timed out after %d seconds. '
+                  'Killing command "%s".', timeout, full_cmd)
     process.kill()
 
-  timer = threading.Timer(timeout, _KillProcess, (process,))
+  timer = threading.Timer(timeout, _KillProcess)
   timer.start()
 
   stdout, stderr = process.communicate()
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -47,6 +47,9 @@ TEMP_DIR = os.path.join(tempfile.gettempdir(), 'perfkitbenchmarker')
 # to use this directory as a base for other module level constants.
 VM_TMP_DIR = '/tmp/pkb'
 
+# Default timeout for issuing a command.
+DEFAULT_TIMEOUT = 300
+
 # Defaults for retrying commands.
 POLL_INTERVAL = 30
 TIMEOUT = 1200
@@ -344,7 +347,8 @@ def Retry(poll_interval=POLL_INTERVAL, max_retries=MAX_RETRIES,
   return Wrap
 
 
-def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
+def IssueCommand(cmd, force_info_log=False, suppress_warning=False,
+                 env=None, timeout=DEFAULT_TIMEOUT):
   """Tries running the provided command once.
 
   Args:
@@ -359,6 +363,9 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
         regardless of suppress_warning's value.
     env: A dict of key/value strings, such as is given to the subprocess.Popen()
         constructor, that contains environment variables to be injected.
+    timeout: Timeout for the command in seconds. If the command has not finished
+        before the timeout is reached, it will be killed. Set timeout to -1 to
+        let the command run indefinitely.
 
   Returns:
     A tuple of stdout, stderr, and retcode from running the provided command.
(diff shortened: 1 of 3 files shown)


ehankland:
test this please.

perfkitbenchmarker/vm_util.py
```diff
@@ -47,6 +47,9 @@
 # to use this directory as a base for other module level constants.
 VM_TMP_DIR = '/tmp/pkb'
 
+# Default timeout for issuing a command.
+DEFAULT_TIMEOUT = 300
```
cmccoy: How about keeping the existing behavior by default (no timeout)?

perfkitbenchmarker/vm_util.py
```diff
@@ -359,6 +363,9 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
         regardless of suppress_warning's value.
     env: A dict of key/value strings, such as is given to the subprocess.Popen()
         constructor, that contains environment variables to be injected.
+    timeout: Timeout for the command in seconds. If the command has not finished
+        before the timeout is reached, it will be killed. Set timeout to -1 to
+        let the command run indefinitely.
```
cmccoy: Prefer `None` or for handling the no timeout case.

perfkitbenchmarker/vm_util.py
```diff
@@ -369,11 +376,21 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
...
 
   shell_value = RunningOnWindows()
+
   process = subprocess.Popen(cmd, env=env, shell=shell_value,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
+
+  timer = None
+  if timeout != -1:
+    timer = threading.Timer(timeout, lambda p: p.kill(), (process,))
```
cmccoy: Could you log that the process is being killed?

perfkitbenchmarker/vm_util.py
ehankland: Done

perfkitbenchmarker/vm_util.py
ehankland: Done

perfkitbenchmarker/vm_util.py
ehankland: I think the default should be to timeout. The only cases I can think of that we don't want to time out after waiting for 5 minutes would be long running ssh/scp commands. In every other case, if the process is still running, something has almost certainly gone wrong.

ehankland:
test this please.

ehankland:
PTAL

perfkitbenchmarker/vm_util.py
```diff
@@ -372,8 +379,18 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
   process = subprocess.Popen(cmd, env=env, shell=shell_value,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
+
+  def _KillProcess(process):
+    logging.warning('IssueCommand timed out. Killing Process.')
```
cmccoy: Nit: this can take no arguments (`process` is in the enclosing scope).
I think it would also be good to:
* Make this `error` level, rather than `warning`.
* Include the value of `timeout` and `cmd`.

perfkitbenchmarker/vm_util.py
cmccoy: Ah, I didn't see that you kept the default in `RemoteCommand`. SGTM.

perfkitbenchmarker/vm_util.py
ehankland: Good points - done.

ehankland:
PTAL

cmccoy:
Looks good.
* Would you mind adding some tests?
* Could you document the behavior when the process is terminated? Is there an exception?

ehankland:
Added a test and updated the description of the timeout parameter

ehankland:
PTAL

tests/vm_util_test.py
```diff
@@ -67,5 +67,16 @@ def testReachable_Unreachable(self):
         False, vm_util.IpAddressSubset.REACHABLE, False)
 
 
+class IssueCommandTestCase(unittest.TestCase):
+
+  def testTimeoutNotReached(self):
```
cmccoy: Could you add a test with no timeout?

perfkitbenchmarker/vm_util.py
```diff
@@ -372,8 +382,19 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False, env=None):
   process = subprocess.Popen(cmd, env=env, shell=shell_value,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
+
+  def _KillProcess():
+    logging.error('IssueCommand timed out after %d seconds. '
+                  'Killing command "%s".', timeout, full_cmd)
+    process.kill()
+
+  timer = threading.Timer(timeout, _KillProcess)
```
cmccoy: Will this behave reasonably if `timeout` is None?

tests/vm_util_test.py
ehankland: Done

perfkitbenchmarker/vm_util.py
ehankland: yes

tests/vm_util_test.py
cmccoy: Also, will this work on Windows? If not, maybe add a `unittest.skipIf` annotation?

cmccoy:
LGTM.

tests/vm_util_test.py
ehankland: sleep works fine on windows

cmccoy: Thanks, I'm ignorant. :)

PR #290, opened 2015-06-04
Add initial support for DigitalOcean as a cloud provider.

klausw:
Add initial support for DigitalOcean as a cloud provider.

This uses the `curl` tool to access the DigitalOcean v2 API. You must set
 up authentication before use, see README.md for instructions.

Since DigitalOcean does not provide add-on disks, droplets always use local
SSD storage as mounted on a single root partition, and AFAIK this cannot be
repartitioned. Benchmarks that request a data directory via vm.GetScratchDir()
will use a subdirectory of the root partition.

Benchmarks that need multiple scratch disks or raw devices are **NOT**
supported and will fail to run in such configurations. This includes
*copy_throughput* in disk-to-disk mode, or *aerospike* when using
non-memory storage modes.

Usage example:

```
./pkb.py --cloud=DigitalOcean --machine_type=16gb --benchmarks=iperf
```

Working benchmarks on a 16gb VM (droplet):

* aerospike, block_storage_workload, bonnie++, cassandra_stress,
  cluster_boot, coremark, fio, hpcc, iperf, mesh_network, mongodb, netperf,
  oldisim, ping, redis, scimark2, silo, sysbench_oltp, unixbench

Test in progress:

* speccpu2006

Not yet tested:

* hadoop_terasort, hbase_ycsb

Unsupported benchmarks:

* _copy_throughput:_ this needs multiple scratch disks.
* _object_storage_service:_ DigitalOcean doesn't offer an object store.
  It should work to run the benchmark using an external object store
  provider, I haven't tested that.


klausw:
Please ignore, wrong base branch.

PR #271, opened 2015-05-19
Add SciMark2 benchmark

klausw:
This addresses Issue #23.

See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/207 for the original pull request contributed by @zlim, I've made some minor modifications as discussed in the pull request.

PR #292, opened 2015-06-04
pkb: add scimark2 + its license to README.md's benchmarks

klausw:
Also fix some formatting inconsistencies for surrounding benchmarks in the list.

https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/271 added scimark2 support but
didn't update the README.


voellm:
Nice catch... thanks for fixing it.

PR #293, opened 2015-06-04
Ensure the _KillProcess task is always cancelled.

cmccoy:
Errors raised in [`subprocess.Popen.communicate`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/651c7558321db1d5dc6b641453a376467cc1b2a1/perfkitbenchmarker/vm_util.py#L394) (e.g. `KeyboardInterrupt`) cause the process to hang without exiting, since the `Timer` thread does not exit.

When a timeout is set, this will exit after the timer elapses. Without a timeout, the process hangs indefinitely.

Newly-added test hangs prior to making this change.


ehankland:
oops - thanks for catching this

PR #304, opened 2015-06-11
Include coremark and speccpu2006 tarballs in gitignore.

klausw:
These files require separate download for licensing reasons,
and they must not be republished. Adding them to the ignore list
to help avoid accidentally uploading them into the repository.


voellm:
Was already marked LGTM... looks fine to me.

PR #307, opened 2015-06-16
Specify psutil version in test-requirements.txt. Fix use in tests.

skschneider:


klausw:
Looks good, didn't know that the API was so version dependent. Sorry about the breakage.

PR #309, opened 2015-06-16
Proxy support

kivio:
Added possibility to configure proxy on VM when setup must stay behind the proxy server.

### DIFF
--- a/perfkitbenchmarker/package_managers.py
+++ b/perfkitbenchmarker/package_managers.py
@@ -146,6 +146,8 @@ class YumMixin(BasePackageMixin):
 
   def Startup(self):
     """Eliminates the need to have a tty to run sudo commands."""
+    if FLAGS.http_proxy or FLAGS.https_proxy:
+        self.SetupProxy()
     self.RemoteCommand('echo \'Defaults:%s !requiretty\' | '
                        'sudo tee /etc/sudoers.d/pkb' % self.user_name,
                        login_shell=True)
@@ -230,6 +232,24 @@ class YumMixin(BasePackageMixin):
     """
     package = packages.PACKAGES[package_name]
     return package.YumGetServiceName(self)
+
+  def SetupProxy(self):
+    """
+        This function setup proxy if your cloud is currently installed behind proxy server
+    """
+    super(YumMixin, self).SetupProxy()
+    yum_proxy_file = "/etc/yum.conf"
+
+    commands = []
+
+    add_proxy_conf = "sudo touch %s" % yum_proxy_file
+    commands.append(add_proxy_conf)
+
+    if FLAGS.http_proxy:
+        add_http_proxy = "echo -e 'proxy= \"%s\";' | sudo tee -a %s" % (FLAGS.http_proxy, yum_proxy_file)
+        commands.append(add_http_proxy)
+
+    self.RemoteCommand(";".join(commands))
 
 
 class AptMixin(BasePackageMixin):
--- a/perfkitbenchmarker/package_managers.py
+++ b/perfkitbenchmarker/package_managers.py
@@ -1,0 +1,343 @@
+# Copyright 2014 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module containing mixin classes for package management.
+
+These classes allow installation on both Debian and RHEL based linuxes.
+They also handle some intial setup (especially on RHEL based linuxes
+since by default sudo commands without a tty don't work) and
+can restore the VM to the state it was in before packages were
+installed.
+
+To install a package on a VM, just call vm.Install(package_name).
+The package name is just the name of the package module (i.e. the
+file name minus .py). The framework will take care of all cleanup
+for you.
+"""
+
+import logging
+import re
+
+from perfkitbenchmarker import errors
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import packages
+from perfkitbenchmarker import vm_util
+
+
+RHEL = 'rhel'
+DEBIAN = 'debian'
+
+EPEL6_RPM = ('http://dl.fedoraproject.org/pub/epel/'
+             '6/x86_64/epel-release-6-8.noarch.rpm')
+EPEL7_RPM = ('http://dl.fedoraproject.org/pub/epel/'
+             '7/x86_64/e/epel-release-7-5.noarch.rpm')
+
+UPDATE_RETRIES = 5
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_enum('os_type', DEBIAN,
+                  [DEBIAN, RHEL],
+                  'The version of linux that the VM\'s os is based on. '
+                  'This will determine the package manager used, among '
+                  'other things.')
+flags.DEFINE_bool('install_packages', True,
+                  'Override for determining whether packages should be '
+                  'installed. If this is false, no packages will be installed '
+                  'on any VMs. This option should probably only ever be used '
+                  'if you have already created an image with all relevant '
+                  'packages installed.')
+
+
+class BasePackageMixin(object):
+
+  def __init__(self):
+    self._installed_packages = set()
+    super(BasePackageMixin, self).__init__()
+
+  def Startup(self):
+    """Performs any necessary setup on the VM specific to the OS."""
+    pass
+
+  def SnapshotPackages(self):
+    """Grabs a snapshot of the currently installed packages."""
+    pass
+
+  def RestorePackages(self):
+    """Restores the currently installed packages to those snapshotted."""
+    pass
+
+  def PackageCleanup(self):
+    """Cleans up all installed packages.
+
+    Deletes the temp directory, restores packages, and uninstalls all
+    PerfKit packages.
+    """
+    for package_name in self._installed_packages:
+      self.Uninstall(package_name)
+    self.RestorePackages()
+    self.RemoteCommand('rm -rf %s' % vm_util.VM_TMP_DIR)
+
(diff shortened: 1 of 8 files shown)


perfkitbenchmarker/package_managers.py
```diff
@@ -115,11 +115,39 @@ def GetServiceName(self, package_name):
...
+        commands.append(add_ftp_proxy)
+
+    self.RemoteCommand(";".join(commands))
+
 
 class YumMixin(BasePackageMixin):
 
   def Startup(self):
     """Eliminates the need to have a tty to run sudo commands."""
+    if FLAGS.http_proxy or FLAGS.https_proxy:
```
klausw: or FLAGS.ftp_proxy

perfkitbenchmarker/package_managers.py
```diff
@@ -115,11 +115,39 @@ def GetServiceName(self, package_name):
...
+    """
+    env_file = "/etc/environment"
+
+    commands = []
+
+    add_proxy_conf = "sudo touch %s" % env_file
+    commands.append(add_proxy_conf)
+
+    if FLAGS.http_proxy:
+        add_http_proxy = "echo -e 'http_proxy=%s' | sudo tee -a %s" % (FLAGS.http_proxy, env_file)
```
klausw: Why `-e`? I don't think there's really a need to interpret escape sequences, and doing so may have odd side effects, i.e. spurious newlines if the proxy value contains `\n`.

perfkitbenchmarker/package_managers.py
```diff
@@ -115,11 +115,39 @@ def GetServiceName(self, package_name):
...
 
+  def SetupProxy(self):
+    """
+        This function setup proxy if your cloud is currently installed behind proxy server
+    """
+    env_file = "/etc/environment"
+
+    commands = []
+
+    add_proxy_conf = "sudo touch %s" % env_file
```
klausw: This shouldn't be needed since `tee` also creates its destination if it doesn't exist yet. Currently, this also has the side effect of always touching the file even if it isn't modifying it.

perfkitbenchmarker/package_managers.py
```diff
@@ -115,11 +115,39 @@ def GetServiceName(self, package_name):
...
+
+    if FLAGS.https_proxy:
+        add_https_proxy = "echo -e 'https_proxy=%s' | sudo tee -a %s" % (FLAGS.https_proxy, env_file)
+        commands.append(add_https_proxy)
+
+    if FLAGS.ftp_proxy:
+        add_ftp_proxy = "echo -e 'ftp_proxy=%s' | sudo tee -a %s" % (FLAGS.ftp_proxy, env_file)
+        commands.append(add_ftp_proxy)
+
+    self.RemoteCommand(";".join(commands))
```
klausw: Add a `if commands:` here, should only execute if there's something to do. (Assuming the `touch` is removed.)

perfkitbenchmarker/package_managers.py
```diff
@@ -205,11 +233,31 @@ def GetServiceName(self, package_name):
...
+  def SetupProxy(self):
+    """
+        This function setup proxy if your cloud is currently installed behind proxy server
+    """
+    super(YumMixin, self).SetupProxy()
+    yum_proxy_file = "/etc/yum.conf"
+
+    commands = []
+
+    add_proxy_conf = "sudo touch %s" % yum_proxy_file
```
klausw: See previous comment, touch shouldn't be needed. Yum is http only?

perfkitbenchmarker/package_managers.py
```diff
@@ -205,11 +233,31 @@ def GetServiceName(self, package_name):
...
+    commands.append(add_proxy_conf)
+
+    if FLAGS.http_proxy:
+        add_http_proxy = "echo -e 'proxy= \"%s\";' | sudo tee -a %s" % (FLAGS.http_proxy, yum_proxy_file)
```
klausw: remove `-e`?

Nitpick: whitespace on only one side of `= ` looks strange.

perfkitbenchmarker/package_managers.py
```diff
@@ -205,11 +233,31 @@ def GetServiceName(self, package_name):
...
+        commands.append(add_http_proxy)
+
+    self.RemoteCommand(";".join(commands))
```
klausw: If there's only going to be one command, may as well execute it directly without a command array. But ok either way.

perfkitbenchmarker/package_managers.py
```diff
@@ -205,11 +233,31 @@ def GetServiceName(self, package_name):
...
+
 
 class AptMixin(BasePackageMixin):
 
   def Startup(self):
     """Runs apt-get update so InstallPackages shouldn't need to."""
+    if FLAGS.http_proxy or FLAGS.https_proxy:
```
klausw: or FLAGS.ftp_proxy ?

klausw: On second thought not sure if it's needed, but would be good to document.

perfkitbenchmarker/pkb.py
```diff
@@ -147,6 +147,14 @@
                      'all disks together.',
                      lower_bound=1)
 
+# support for proxy
+flags.DEFINE_string('http_proxy', '',
+                    'Specify a proxy for HTTP in the form [user:passwd@]proxy.server:port.')
```
klausw: There's some ambiguity about where the proxies get used. Maybe clarify that these are for use within the cloud deployed OS? Presumably if people need proxies on the machine where they are running pkb.py, those would be configured separately.

perfkitbenchmarker/packages/ycsb.py
```diff
@@ -130,6 +144,26 @@ def _Install(vm):
   vm.RemoteCommand(('mkdir -p {0} && curl -L {1} | '
                     'tar -C {0} --strip-components=1 -xzf -').format(
                         YCSB_BUILD_DIR, YCSB_TAR_URL))
+
+  settings_file = ".m2/settings.xml"
+  proxy_enable = False
```
klausw: You've got quite a lot of separate state in variables here. Alternatively, how about something like this:

```python
proxy_nodes = []
if FLAGS.http_proxy:
  proxy_nodes.append(CreateProxyElement('http', FLAGS.http_proxy))
...
if proxy_nodes:
  root = Element('settings')
  proxies = SubElement(root, 'proxies')
  proxies.extend(proxy_nodes)
  ...
```

For better isolation, would be nice to move this into a CreateSettingsFile method.

perfkitbenchmarker/packages/ycsb.py
```diff
@@ -97,6 +99,18 @@
                      'each workload / client count combination. Set to 0 for '
                      'unlimited time.')
 
+def create_proxy_element(proxy_type, proxy, parent):
```
klausw: Style nitpicks - use CamelCaseNames for functions, and prefer single quotes over double quotes for strings.

See comments on _Install below, I think it would be a bit cleaner to return an element instead of passing in and modifying the parent.

perfkitbenchmarker/package_managers.py
kivio: yes, yum just need http proxy

klausw:
FYI, review comments are intended as suggestions - we don't want the review process to turn into a barrier to contributions. Feel free to push back on requests and/or ask us to take care of issues.

kivio:
OK, i think now it looks good

perfkitbenchmarker/pkb.py
kivio: Proxy is used on Guest VM created by Perfkit. If your cloud is deployed behind the proxy, this feature is required. Currently this feature is needed just for OpenStack but in next time if someone add support for VMWare private clouds or Azure private setups. This feature can be useful for all.

kivio: I think this flags must be defined in pkb.py or in package_manager

klausw: I had meant adding an explanation to the flags so that people looking at the help text know about it, or a note in the README. But it's probably reasonably clear from context.

klausw:
LGTM, thanks for making the changes. @cmccoy , any comments from your side?

klausw:
test this please

kivio:
OK, i was added documentation now.

In subject of test, i have that environment in my work then proxy works good, do you think about unit tests?

cmccoy:
LGTM. There are going to be some merge conflicts due to #321. Let us know if we can help.

klausw:
Working on the merge now - seeing some lint errors from flake8, I'll take care of them.

klausw:
I've merged this manually to the dev branch, but this didn't work right since it erroneously applied changes to the deleted
package_managers.py file instead of the new linux_virtual_machine.py file. Working on  https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/337 , please wait for that before continuing. I'll update this thread once that's done.

klausw:
The dev branch is now up to date. I had combined your patches into the single change d0ebc169 with no modifications, then followed up with the two fixup changes - in case you run into merge conflicts, you could try cherry picking those separately on top of your original changes.

Also, can you please verify that things work right for you? I had slightly changed the logic as part of the move to take better advantage of inheritance, it should be equivalent but I could only do limited testing.
(I get the expected connection errors from *apt* when setting a nonexistent --http_proxy.)

For reference:

```
commit b8e2771b0d66a2d8ed2845601e9f03f097380e93
Author: Klaus Weidner <klausw@google.com>
Date:   Tue Jun 23 11:39:57 2015 -0700

    Proxy config fixup for bad merge.
    
    The previous proxy config patch erroneously applied changes to the deleted
    file package_managers.py, reviving it in the process. Delete it again, and
    apply its changes to linux_virtual_machine.py.

commit 6c061d2475312c88742e5c605ce6e94525c34fae
Author: Klaus Weidner <klausw@google.com>
Date:   Tue Jun 23 10:45:47 2015 -0700

    Lint fixes for proxy configuration.
    
    Minor doc edits, fix some over-long lines, and use the ElementTree prefix
    instead of directly importing functions into the toplevel namespace.

commit d0ebc1699b04b22dd7fab0bbe73808e5485ab8c3
Author: Marcin Karkocha <kivio@kivio.pl>
Date:   Tue Jun 23 10:41:59 2015 -0700

    Add support for using proxies in the cloud environment.
    
    This introduces new flags --http_proxy, --https_proxy and --ftp_proxy.
    The flag values use the same syntax as the corresponding environment variables,
    for example `--http_proxy=proxy.example.com:8080` .

    See README.md for additional information.
```

PR #310, opened 2015-06-16
Execute long running commands via a remote agent.

cmccoy:
This makes `LongRunningRemoteCommand` a little more robust via two scripts run on the guest VM:
* `remote_command.py`: executed via `nohup`, runs a command, holding a lock on a status file until the command exits.
* `wait_for_command.py`: waits for the long running command to finish by waiting on the lock held by `remote_command.py`, then returns the long-running command's `stdout` and `stderr`, exiting with the command's exit status.  This can be safely retried on SSH failure.

We gain a bit of resilience on SSH flakiness, will catch non-zero exits from the long running command, and avoid parsing the command name to find the PID.

Ran HPCC against CentOS 6 and Ubuntu 14.04 to test.

### DIFF
--- a/perfkitbenchmarker/scripts/execute_command.py
+++ b/perfkitbenchmarker/scripts/execute_command.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# -*- coding: utf-8 -*-
+"""Runs a command, saving stdout, stderr, and the return code in files.
+
+Simplifies executing long-running commands on a remote host.
+The status file (as specified by --status) is exclusively locked until the
+child process running the user-specified command exits.
+This command will fail if the status file cannot be successfully locked.
+
+To await completion, "wait_for_command.py" acquires a shared lock on the
+status file, which blocks until the process completes.
+
+*Runs on the guest VM. Supports Python 2.6, 2.7, and 3.x.*
+"""
+import fcntl
+import logging
+import optparse
+import sys
+import subprocess
+
+
+def main():
+  parser = optparse.OptionParser()
+  parser.add_option('-o', '--stdout', dest='stdout', metavar='FILE',
+                    help="""Write stdout to FILE. Required.""")
+  parser.add_option('-e', '--stderr', dest='stderr', metavar='FILE',
+                    help="""Write stderr to FILE. Required.""")
+  parser.add_option('-p', '--pid', dest='pid', help="""Write PID to FILE.""",
+                    metavar='FILE')
+  parser.add_option('-s', '--status', dest='status', help="""Write process exit
+                    status to FILE. An exclusive lock will be placed on FILE
+                    until this process exits. Required.""", metavar='FILE')
+  parser.add_option('-c', '--command', dest='command', help="""Shell command to
+                    execute. Required.""")
+  options, args = parser.parse_args()
+  if args:
+    sys.stderr.write('Unexpected arguments: {0}\n'.format(args))
+    return 1
+
+  missing = []
+  for option in ('stdout', 'stderr', 'status', 'command'):
+    if getattr(options, option) is None:
+      missing.append(option)
+
+  if missing:
+    parser.print_usage()
+    msg = 'Missing required flag(s): {0}\n'.format(
+        ', '.join('--' + i for i in missing))
+    sys.stderr.write(msg)
+    return 1
+
+  with open(options.status, 'w+') as status:
+    with open(options.stdout, 'w') as stdout:
+      with open(options.stderr, 'w') as stderr:
+        logging.info('Acquiring lock on %s', options.status)
+        # Non-blocking exclusive lock acquisition; will raise an IOError if
+        # acquisition fails, which is desirable here.
+        fcntl.lockf(status, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        # Initialize the status to -1; will be filled with the exit status
+        # on subprocess completion.
+        status.write('-1')
+        status.flush()
+        status.seek(0)
+
+        p = subprocess.Popen(options.command, stdout=stdout, stderr=stderr,
+                             shell=True)
+        logging.info('Started pid %d: %s', p.pid, options.command)
+
+        if options.pid:
+          with open(options.pid, 'w') as pid:
+            pid.write(str(p.pid))
+
+        logging.info('Waiting on PID %s', p.pid)
+        return_code = p.wait()
+        logging.info('Return code: %s', return_code)
+        status.truncate()
+        status.write(str(return_code))
+
+        # File lock will be released when the status file is closed.
+        return return_code
+
+if __name__ == '__main__':
+  logging.basicConfig(level=logging.INFO)
+  sys.exit(main())
--- a/perfkitbenchmarker/scripts/wait_for_command.py
+++ b/perfkitbenchmarker/scripts/wait_for_command.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python2
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# -*- coding: utf-8 -*-
+"""Waits for a command started by execute_command.py to complete.
+
+Blocks until a command wrapped by "execute_command.py" completes, then mimics
+the wrapped command, copying the wrapped command's stdout/stderr to this
+process' stdout/stderr, and exiting with the wrapped command's status.
+
+*Runs on the guest VM. Supports Python 2.6, 2.7, and 3.x.*
+"""
+
+import fcntl
+import optparse
+import os
+import shutil
+import sys
+import threading
+
+
+def main():
+  p = optparse.OptionParser()
+  p.add_option('-o', '--stdout', dest='stdout',
+               help="""Read stdout from FILE.""", metavar='FILE')
+  p.add_option('-e', '--stderr', dest='stderr',
+               help="""Read stderr from FILE.""", metavar='FILE')
+  p.add_option('-s', '--status', dest='status', metavar='FILE',
+               help='Get process exit status from FILE. '
+               'Will block until a shared lock is acquired on FILE.')
+  p.add_option('-d', '--delete', dest='delete', action='store_true',
+               help='Delete stdout, stderr, and status files when finished.')
+  options, args = p.parse_args()
+  if args:
+    sys.stderr.write('Unexpected arguments: {0}\n'.format(args))
+    return 1
+
+  missing = []
+  for option in ('stdout', 'stderr', 'status'):
+    if getattr(options, option) is None:
+      missing.append(option)
+
+  if missing:
+    p.print_usage()
+    msg = 'Missing required flag(s): {0}\n'.format(
+        ', '.join('--' + i for i in missing))
+    sys.stderr.write(msg)
+    return 1
+
+  with open(options.stdout, 'r') as stdout:
+    with open(options.stderr, 'r') as stderr:
+      with open(options.status, 'r') as status:
+        fcntl.lockf(status, fcntl.LOCK_SH)
+        return_code = int(status.read())
+        status.close()
+
+        stderr_copier = threading.Thread(target=shutil.copyfileobj,
+                                         args=[stderr, sys.stderr],
+                                         name='stderr-copier')
+        stderr_copier.daemon = True
+        stderr_copier.start()
+        try:
+          shutil.copyfileobj(stdout, sys.stdout)
+        finally:
+          stderr_copier.join()
+
+  if options.delete:
+    for f in [options.stdout, options.stderr, options.status]:
+      os.unlink(f)
+
+  return return_code
+
+if __name__ == '__main__':
+  sys.exit(main())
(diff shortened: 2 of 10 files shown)


perfkitbenchmarker/scripts/execute_command.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+import fcntl
+import logging
+import optparse
+import sys
+import subprocess
+
+
+def main():
+  p = optparse.OptionParser()
```
stfeng: consider using full name to "optionParser"

perfkitbenchmarker/scripts/execute_command.py
```diff
@@ -0,0 +1,94 @@
...
+  p.add_option('-o', '--stdout', dest='stdout', metavar='FILE',
+               help="""Write stdout to FILE. Required.""")
+  p.add_option('-e', '--stderr', dest='stderr', metavar='FILE',
+               help="""Write stderr to FILE.  Required.""")
+  p.add_option('-p', '--pid', dest='pid', help="""Write PID to FILE.""",
+               metavar='FILE')
+  p.add_option('-s', '--status', dest='status', help="""Write process exit
+               status to FILE. FILE with have an exclusive lock during run.
```
stfeng: s/with/will

stfeng: quickly describe lock life cycle, e.g, when is the lock acquired, released (list all the conditions, especially those error conditions: process crashed/hang, etc)

Either in this comment here, or above.

perfkitbenchmarker/scripts/execute_command.py
```diff
@@ -0,0 +1,94 @@
...
+        ', '.join('--' + i for i in missing))
+    print >> sys.stderr, msg
+    return 1
+
+  with open(options.stdout, 'w') as stdout:
+    with open(options.stderr, 'w') as stderr:
+      with open(options.status, 'w+') as status:
+        # Initialize the status to -1; will be filled with the exit status
+        # on subprocess completion.
+        status.write('-1')
```
stfeng: Is it possible to use some more descriptive text than -1 for status? Does it have to be an integer code?

perfkitbenchmarker/scripts/execute_command.py
```diff
@@ -0,0 +1,94 @@
...
+                             shell=True)
+        logging.info('Started pid %d: %s', p.pid, options.command)
+
+        if options.pid:
+          with open(options.pid, 'w') as pid:
+            pid.write(str(p.pid))
+
+        logging.info('Acquiring lock on %s', options.status)
+
+        fcntl.lockf(status, fcntl.LOCK_EX)
```
stfeng: does this lock acquring function allow a timeout of some sort?

perfkitbenchmarker/scripts/execute_command.py
```diff
@@ -0,0 +1,94 @@
...
+        logging.info('Waiting on PID %s', p.pid)
+        return_code = p.wait()
```
stfeng: why do we wait for pid while holding the exclusive lock to the file? Can the wait happen before the lock acquisition?

perfkitbenchmarker/scripts/wait_for_command.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python2
```
stfeng: [ignore this]

A note to myself: continue review from this file here...

perfkitbenchmarker/scripts/execute_command.py
cmccoy: renamed to `parser` - we don't use camelCase.

perfkitbenchmarker/scripts/execute_command.py
cmccoy: It does have to be an int: it's eventually the exit status code of `options.command`.

perfkitbenchmarker/scripts/execute_command.py
cmccoy: Changed to nonblocking acquisition; the command will now fail if the lock cannot be acquired.
PKB generates UUIDs for each command run with this script, so the file should not exist prior to the command running. If the file cannot be locked, it's an error.

perfkitbenchmarker/scripts/execute_command.py
cmccoy: Acquiring the lock while the subprocess runs is intentional here.

The status file is locked while `p` is running. Once `p` completes, the return code from `p` is written to the status file.

Other processes interested in the results of `p` can wait on the locked status file to block until the process completes (see `wait_for_command.py`). Does that make sense?

stfeng: ok, let me try to state my understanding: the lock is to ensure that the status code is properly populated into the status file before the waiting process can consume it by opening and reading the code from status file, right?

If the waiting process has other mechnisms of knowing the status, we wouldn't have needed the lock here because other process can simply call "wait pid" to wait for the process p to terminate. (instead of waiting on the lock).

cmccoy: > the lock is to ensure that the status code is properly populated into the status file before the waiting process can consume it by opening and reading the code from status file, right?

Right: status code, plus process stdout and stderr output to files.

> If the waiting process has other mechnisms of knowing the status, we wouldn't have needed the lock here because other process can simply call "wait pid" to wait for the process p to terminate. (instead of waiting on the lock).

I don't think this would work when `pid` is not a child of the waiting process, which is the case here.

stfeng: sgtm. thanks!

perfkitbenchmarker/scripts/wait_for_command.py
```diff
@@ -0,0 +1,83 @@
...
+    print >> sys.stderr, msg
+    return 1
+
+  with open(options.stdout, 'r') as stdout:
+    with open(options.stderr, 'r') as stderr:
+      with open(options.status, 'r') as status:
+        fcntl.lockf(status, fcntl.LOCK_SH)
+        return_code = int(status.read())
+        status.close()
+
```
stfeng: Would be useful at here to log how large the stderr file to be copied is, so people would have an expectation on how soon it would finish.

perfkitbenchmarker/scripts/wait_for_command.py
```diff
@@ -0,0 +1,83 @@
...
+        stderr_copier = threading.Thread(target=shutil.copyfileobj,
+                                         args=[stderr, sys.stderr],
+                                         name='stderr-copier')
+        stderr_copier.daemon = True
+        stderr_copier.start()
+        try:
+          shutil.copyfileobj(stdout, sys.stdout)
```
stfeng: does this copyfileobj() function display progress information? Would be nice to have some progress information made visible to the user.

perfkitbenchmarker/virtual_machine.py
```diff
@@ -148,6 +163,17 @@ def __repr__(self):
   def __str__(self):
     return self.ip_address
 
+  def __getstate__(self):
+    """Get state, excluding _has_remote_command_script_lock."""
```
stfeng: Can you add comment here explaining why the return result needs to have the _has_remote_command_script_lock removed from this function before returning to the caller? Can the caller just ignore this field then?

cmccoy: `__getstate__` and `__setstate__` are used for pickling (Python serialization) objects. Locks can't be serialized, so these two methods:
* Drop the lock from the state to be serialized.
* Create a new lock when the object is deserialized.

Added some comments.

perfkitbenchmarker/scripts/wait_for_command.py
cmccoy: It doesn't. Progress would be nice, but there's nowhere to write it :).  `wait_for_command.py` is mimicing the command executed via `execute_command.py`: it writes that process' stdout to its own stdout, stderr to stderr. Writing progress to stderr would confuse a consumer expecting a certain output format from the command.

perfkitbenchmarker/scripts/execute_command.py
cmccoy: Added a description to the text here and to the top-level docstring.

perfkitbenchmarker/virtual_machine.py
```diff
@@ -138,6 +150,9 @@ def __init__(self, vm_spec):
     self._total_memory_kb = None
     self._num_cpus = None
 
+    self._has_remote_command_script_lock = threading.Lock()
```
stfeng: threading.lock() returns a lock object, not a boolean. Please consider renaming this to _remote_command_script_lock (basically, drop the "has" prefix which implies a boolean).

perfkitbenchmarker/virtual_machine.py
```diff
@@ -285,36 +314,65 @@ def RemoteCopy(self, file_path, remote_path='', copy_to=True):
                     (retcode, full_cmd, stdout, stderr))
       raise errors.VmUtil.SshConnectionError(error_text)
 
+  def _PushLongRunningCommandScripts(self):
```
stfeng: Can this function be called to push script for any purpose (in addition to pushing it only for long running commands)? I don't see anything specific here in this function wrt long running commands (but I could be wrong here), please consider renaming this function and make it generic so anyone can use it to push a script (safely) to a VM.

perfkitbenchmarker/virtual_machine.py
```diff
@@ -285,36 +314,65 @@ def RemoteCopy(self, file_path, remote_path='', copy_to=True):
...
+    with self._has_remote_command_script_lock:
+      if not self._has_remote_command_script:
+        for f in (EXECUTE_COMMAND, WAIT_FOR_COMMAND):
+          self.PushDataFile(f, os.path.join(vm_util.VM_TMP_DIR,
+                                            os.path.basename(f)))
+
   def LongRunningRemoteCommand(self, command):
     """Runs a long running command on the VM in a robust way.
 
+    Executes a long running command via a pair of scripts on the VM:
```
stfeng: [I know it was not clear in the old code before but I think this would be helpful to have:]

Consider adding a clarifying comment to indicate at which threshold the caller should consider a command as a "long running" command and invoke this function. e.g., 10 minutes, 30 minutes, or 1 hour?

Also, at a design level, AFAIK, this function should just work for all commands independent of execution duration (but it does come with a small overhead of async waiting, etc), I'm wondering why it can't be the default "run remote command function". Not blocking this CL but something to consider...

perfkitbenchmarker/virtual_machine.py
```diff
@@ -285,36 +314,65 @@ def RemoteCopy(self, file_path, remote_path='', copy_to=True):
...
+    file_base = os.path.join(vm_util.VM_TMP_DIR, 'cmd%s' % uid)
+    stdout_file = file_base + '.stdout'
+    stderr_file = file_base + '.stderr'
+    status_file = file_base + '.status'
 
-    return stdout, stderr
+    if not isinstance(command, basestring):
+      command = ' '.join(command)
+
+    start_command = ['nohup', 'python', execute_path,
```
stfeng: I noticed "python" in this line, if this is a generic OS command, why do we need to involve python here?

stfeng:
I've done the first pass, mostly LGTM, just some questions need clarifications but almost ready to go.

Thanks for making this happen Connor!

perfkitbenchmarker/virtual_machine.py
cmccoy: I don't think I follow the generic OS command bit.

Using `python` here so we don't rely on the script on the host to have the execute bit. `chmod`ing after upload would do the same thing.

perfkitbenchmarker/virtual_machine.py
cmccoy: `_has_remote_command_script_lock` protects the boolean value `_has_remote_command_script`. Naming here was to be consistent with `_instance_counter_lock` / `_instance_counter`.  I'm worried `_remote_command_script_lock` will imply that a single remote command should be run at a time, which is not the case. 

Renamed to `_remote_command_script_upload_lock`.

perfkitbenchmarker/virtual_machine.py
cmccoy: `PushDataFile` and `PushFile` are generic functions for pushing files to the remote VM.
This function defers most of the work `PushDataFile`, under a lock so this work is done only once.  I don't think it adds a lot of generic value.

perfkitbenchmarker/virtual_machine.py
cmccoy: Good point. Maybe a name change would make this clearer. AsyncRemoteCommand?

You're right that it should work for all commands independent of execution duration. The overhead is:

1. A second SSH connection.
1. stdout/stderr being buffered to disk rather than streamed to the client.
1. More data in the logs (as implemented).

I think the first and third are good reasons to leave most of the code as-is, but agree that we should move more code to this method.

cmccoy: I removed "long running" from the docs here, and described the differences from RemoteCommand in more detail.

cmccoy:
Addressed comments. PTAL.

perfkitbenchmarker/virtual_machine.py
stfeng: By "Generic OS command", I mean those command that does not depend on a python interpreter...

The question here is: what if the host does not have python or the script to be executed cannot run properly wrapped in the python environment?

perfkitbenchmarker/virtual_machine.py
```diff
@@ -285,8 +314,30 @@ def RemoteCopy(self, file_path, remote_path='', copy_to=True):
...
+    If the scripts have already been placed on the VM, this is a noop.
+    """
+    with self._has_remote_command_script_lock:
+      if not self._has_remote_command_script:
+        for f in (EXECUTE_COMMAND, WAIT_FOR_COMMAND):
+          self.PushDataFile(f, os.path.join(vm_util.VM_TMP_DIR,
+                                            os.path.basename(f)))
+        self._has_remote_command_script = True
+
+  def AsyncRemoteCommand(self, command):
```
stfeng: I think "RobustRemoteCommand" might be a better name than AsyncRemoteCommand. Having Async*** starting a method name gives me an immediate impression that this API will return immediately to the caller when called, and starts doing its work asynchronously, and it would be the *caller*'s responsibility to poll or wait for the completion of this work. But that's not the case here, this method is still a synchronous method, it will start two concurrent commands, and wait for the commands to finish before it returns control to the caller.

The intention of this method is to provide a better, and more robust way of executing a remote command and deliver the results back to the caller safely. It's really just a better implementation of the original RemoteCommand() method (with some extra overhead). We can also name it as RemoteCommandV2() or something like that. (if you don't like the "RobustRemoteCommand()" suggestion. :-)

perfkitbenchmarker/virtual_machine.py
cmccoy: Ah, I understand.
No, this absolutely requires a python interpreter. There's no difference between invoking via `python` and making the script executable and invoking it directly - in the second case the shabang line specifies that the script be executed by `/usr/bin/env python`, which will be the same `python` as the first case.

I was careful to make sure that:
* The script has no dependencies other than those that ship with python.
* The script works on Python 2.6 (required for CentOS 6), 2.7, and 3.x.

perfkitbenchmarker/virtual_machine.py
cmccoy: Good point. Renamed.

cmccoy:
@stfeng: PTAL.

stfeng:
LGTM. Ship it!

PR #313, opened 2015-06-17
Fix aws on windows

ehankland:
The AWS CLI does not correctly set either $LastExitCode or $? when it
fails (at least with version 1.7.28 which was what I was running with on Windows). 
This leads to us thinking a command succeeded when it didn't.
By adding the option to retry the command in IssueRetryableCommand when
there is output on stderr, we can work around this.

### DIFF
--- a/perfkitbenchmarker/aws/aws_disk.py
+++ b/perfkitbenchmarker/aws/aws_disk.py
@@ -88,7 +88,8 @@ class AwsDisk(disk.BaseDisk):
         'describe-volumes',
         '--region=%s' % self.region,
         '--filter=Name=volume-id,Values=%s' % self.id]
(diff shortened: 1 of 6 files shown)


ehankland:
test this please.

ehankland:
test this please.

cmccoy:
If this is only applicable to `aws`, maybe add a function to `perfkitbenchmarker.aws.util` rather than add more options to IssueRetryableCommand?

ehankland:
Added an aws.util.IssueRetryableCommand function and removed modifications to vm_util.IssueRetryableCommand

PTAL

ehankland:
test this please.

cmccoy:
LGTM.

Issue #205, opened 2015-04-03
Proposal: add event hooks


cmccoy:
## TL;DR
Some bugs (e.g. #146), and some additional desired functionality (e.g., annotating samples with host load information) could be fairly easily addressed by adding event hooks to PKB.

I suggest the following:
* Add a new module: `perfkitbenchmarker.events`.
* Define a class `EventDispatcher` within that module with methods:
   * `Subscribe(event_name, function)`
     Adds function to subscriber list for `event_name`.
   * `Unsubscribe(event_name, function)`
     Removes `function` from subscriber list to `event_name`.
   * `FireEvent(event_name, *args, **kwargs)`
     Logs event, calls all listeners of `event_name` with `*args`, `**kwargs`.
* Create a singleton `EventDispatcher` for the application as a whole.
  Listeners will be registered there, and events fired.

## Issues to think about:

* Threading.  Multiple benchmarks may be running simultaneously; multiple threads may be running for a single benchmark.
* How will we document the interfaces for event handlers (expected arguments)?
* Can events have side effects?
   * Internally: definitely yes (e.g., populating metadata for a metadata provider).
   * Otherwise: likely no.

## Useful events

This is a (non-comprehensive) list of events what could enhance the product.

#### Arguments to listener:

* Benchmark
* Flag name
* Flag value

#### Triggered when:

A benchmark uses a command-line flag.

#### Listeners:

* `FlagMetadataProvider`: Adds distinct used flags and their values to sample metadata.

#### Arguments to listener:

* Stage (benchmark, prepare, run, cleanup)
* benchmark_spec.

#### Triggered when

A run phase (prepare, run, cleanup) starts/ends.

#### Listeners:

   * A function/class that creates Sample objects with end-to-end time for each phase?

### `MachineReachable`

*Low priority.*

#### Arguments to listener:

* Virtual machine

#### Triggered when

The first `hostname` command succeeds on a VM.

#### Listeners:

* Cluster boot benchmark - gives time to boot before apt-get update runs.

Comments welcome!

ivansmf:
Can we design this to also handle streamed results to BigQuery? How is an event different from a sample besides the listener end point?

voellm:
Ivan ... Perf your request you could register and even handler to send events to BQ.  

The intent here however is not on performance timing but as an extension point.

voellm:
[...]

PR #315, opened 2015-06-17
Add event support

cmccoy:
* Add a dependency on [`blinker`](https://pythonhosted.org/blinker/) to provide signal support.
* Add `perfkitbenchmarker.events` module, which defines a set of events
  for subscription.
* Switch `dstat` collector to use `{before,after}_phase` events to
  start/stop dstat.

See #205.

### DIFF
--- a/perfkitbenchmarker/events.py
+++ b/perfkitbenchmarker/events.py
@@ -16,6 +16,9 @@
 
 All events are passed keyword arguments, and possibly a sender. See event
 definitions below.
+
+Event handlers are run synchronously in an unspecified order; any exceptions
+raised will be propagated.
 """
 
 from blinker import Namespace
--- a/perfkitbenchmarker/events.py
+++ b/perfkitbenchmarker/events.py
@@ -45,3 +45,11 @@ successful.
 
 Sender: the phase. Currently only RUN_PHASE.
 Payload: benchmark_spec.""")
+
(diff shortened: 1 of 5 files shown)


cmccoy:
@klausw: do you have time to look this over?

perfkitbenchmarker/events.py
```diff
@@ -0,0 +1,55 @@
...
+Signal sent immediately after a phase runs, regardless of whether it was
+successful.
+
+Sender: the phase. Currently only RUN_PHASE.
+Payload: benchmark_spec.""")
+
+sample_created = _events.signal('sample-created', doc="""
+Called with sample object and benchmark spec.
+
+Signal sent immediately after a sample is created by a publisher.
```
klausw: Clarify if the sample is mutable or not? Is it ok to add metadata and expect it'll be stored?

klausw:
LGTM overall, the event system looks like a nice abstraction.

What's the semantics for failing hooks? For example, if a subscriber throws an exception in a before-phase hook, does that get propagated or ignored?

cmccoy:
Errors in subscribers are propagated. I'll note this in `events.py`. Does that seem sufficient?

klausw:
LGTM

perfkitbenchmarker/events.py
cmccoy: It is mutable; now noted.

PR #316, opened 2015-06-17
Add resource creation/deletion times

ehankland:
Instead of recording the creation time in the VM class record it in the resource class.
Also record some other time info about resources that we can use later if we decide to publish resource creation/deletion times.

### DIFF
--- a/perfkitbenchmarker/virtual_machine.py
+++ b/perfkitbenchmarker/virtual_machine.py
@@ -514,10 +514,10 @@ class BaseVirtualMachine(resource.BaseResource):
     """Gets the time it took to boot this VM.
 
     Returns:
-      Boot time (in seconds), or None if the boot is incomplete.
+      Boot time (in seconds).
     """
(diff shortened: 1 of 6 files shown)


perfkitbenchmarker/virtual_machine.py
```diff
@@ -520,10 +516,10 @@ def TimeToBoot(self):
     Returns:
       Boot time (in seconds), or None if the boot is incomplete.
     """
-    if not self.bootable_time or not self.create_time:
+    if not self.bootable_time or not self.create_start_time:
       return None
```
yuyantingzero: Consider use exceptions here instead of using None as a special return value.

yuyantingzero:
A small comment, LGTM

perfkitbenchmarker/virtual_machine.py
ehankland: Done.

PR #317, opened 2015-06-18
netperf: Support confidence intervals, increase test length, report stddev

cmccoy:
* Increases default netperf test length to 60 seconds (from 10).
* Adds support to run each netperf test for a user-specified amount of time (`--netperf_test_length`).
* Adds support for multiple test iterations, which will also generate confidence estimates (`--netperf_max_iter`).
* Adds standard deviation for RR tests to the output.

### DIFF
--- a/perfkitbenchmarker/benchmarks/netperf_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/netperf_benchmark.py
@@ -35,7 +35,7 @@ flags.DEFINE_integer('netperf_max_iter', None,
                      'Maximum number of iterations to run during '
                      'confidence interval estimation. If unset, '
                      'a single iteration will be run.',
-                     lower_bound=1, upper_bound=30)
+                     lower_bound=3, upper_bound=30)
 
 flags.DEFINE_integer('netperf_test_length', 60,
                      'netperf test length, in seconds',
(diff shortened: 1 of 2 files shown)


cmccoy:
test this please.

perfkitbenchmarker/benchmarks/netperf_benchmark.py
```diff
@@ -31,11 +31,22 @@
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import netperf
 
+flags.DEFINE_integer('netperf_max_iter', None,
+                     'Maximum number of iterations to run during '
+                     'confidence interval estimation. If unset, '
+                     'a single iteration will be run.',
```
skschneider: Any reason why the default isn't just 1 w/ no None behavior?

cmccoy: I want to distinguish between performing confidence interval estimation and not, which was the reason for using None here.
netperf will silently ceil the max iters to 3 (docs [here](http://www.netperf.org/svn/netperf2/tags/netperf-2.6.0/doc/netperf.html), search for `-i <sizespec>`), so I don't want to blindly pass the value of `--netperf_max_iter`.  But the current lower bound doesn't make sense; I'll update it to a default of `None`, `lower_bound` of 3.

skschneider:
LGTM

PR #319, opened 2015-06-18
Add --quiet flag to Cloud Bigtable cluster list

cmccoy:
Prevents prompting when user doesn't have the component installed.


skschneider:
LGTM

PR #320, opened 2015-06-18
Improve error message

ehankland:
Improve error message when a machine doesn't have enough local disks to run a benchmark.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -174,7 +174,7 @@ class BenchmarkSpec(object):
         if (FLAGS.scratch_disk_type == disk.LOCAL and
             benchmark_info['scratch_disk'] and
             not FLAGS['num_striped_disks'].present):
-          num_striped_disks = (vm.max_local_disks /
+          num_striped_disks = (vm.max_local_disks //
                                benchmark_info['scratch_disk'])
           if num_striped_disks == 0:
             raise errors.Error(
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/benchmark_spec.py
```diff
@@ -175,6 +176,13 @@ def __init__(self, benchmark_info):
             not FLAGS['num_striped_disks'].present):
           num_striped_disks = (vm.max_local_disks /
```
meteorfox: Also, we should probably use explicit integer floor division `//`, just to be safe when PerfKit moves to Python3

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -175,6 +176,13 @@ def __init__(self, benchmark_info):
...
                                benchmark_info['scratch_disk'])
+          if num_striped_disks == 0:
+            raise errors.Error(
+                'Not enough local disks to run benchmark "%s". It requires at '
+                'least %d local disk(s). The specified machine type has %d '
+                'local disk(s).' % (benchmark_info['name'],
+                                    int(benchmark_info['scratch_disk']),
+                                    vm.max_local_disks))
```
meteorfox: Yes, this is much better. Thanks

perfkitbenchmarker/benchmark_spec.py
ehankland: Done.

PR #322, opened 2015-06-18
Fix 'no attribute pkb_thread_log_context' error for sub-thread logs

klausw:
The logging filter assumes that log_util.SetThreadLogContext was called
in the current thread, but this is not the case for ad-hoc threads
created by threading.Thread() or similar. This broke timeout handling
since it tried to log a message before killing subprocesses.

Includes a test to ensure logging works for plain threads.


ehankland:
LGTM

PR #323, opened 2015-06-18
Work around annoying API difference across psutil versions.

klausw:
We ask for v3.0.0 in test-requirements.txt, but the system
may have an older version installed which would get used
instead of that.

### DIFF
--- a/tests/vm_util_test.py
+++ b/tests/vm_util_test.py
@@ -122,6 +122,15 @@ class WaitUntilSleepTimer(threading.Thread):
 
 class IssueCommandTestCase(unittest.TestCase):
 
+  def setUp(self):
+    # psutil changed its API in v2.0, and we ask for a current one
+    # in test-requirements.txt. Unfortunately we may still get an old
+    # version if there's a system package already installed.
+    # Catch this and show a diagnostic to avoid obscure errors.
+    if psutil.version_info < (2, 0):
+      raise Exception('Wrong psutil version, found %s. Use virtualenv?' % (
+          psutil.__version__))
+
   def testTimeoutNotReached(self):
     _, _, retcode = vm_util.IssueCommand(['sleep', '0s'])
     self.assertEqual(retcode, 0)
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -16,4 +16,4 @@ nose>=1.3
 nose-progressive>=1.5
 flake8>=2.1.0
 futures>=2.1.0
-psutil==3.0.0
+psutil>=2.0.0
--- a/tests/requirements_test.py
+++ b/tests/requirements_test.py
@@ -1,0 +1,51 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Verifies that test requirements are satisfied."""
+
+import os
+import pkg_resources
+import psutil
+import sys
+import unittest
+
+class CheckResourcesTestCase(unittest.TestCase):
+  def testRequirements(self):
+    testreq_file = os.path.join(
+        os.path.dirname(__file__), os.pardir, 'test-requirements.txt')
+    expected_modules = set(['psutil'])
+    checked_modules = set()
+    with open(testreq_file) as requirements:
+      for line in requirements:
+        try:
+          req = pkg_resources.Requirement.parse(line)
+          # There's no obvious mapping from project name as used in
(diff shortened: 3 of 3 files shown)


klausw:
This is very ugly, but I couldn't find an easy way to use the pip-installed version of the package instead of the system-installed `python-psutil` on my system. Open for other suggestions.

cmccoy:
Use a [virtualenv](https://virtualenv.pypa.io/en/latest/)?

cmccoy:
I think we should target a single version of `psutil`, which doesn't strictly need to be the latest.

klausw:
PTAL, I've replaced the workaround hack with a simple version check to verify that we're using
a sane version. Anything >= v2.0 should be ok according to https://github.com/giampaolo/psutil/blob/master/HISTORY.rst , unfortunately my system's python-psutil package is v1.2.1.

tests/vm_util_test.py
```diff
@@ -122,6 +122,15 @@ def run(self):
 
 class IssueCommandTestCase(unittest.TestCase):
 
+  def setUp(self):
+    # psutil changed its API in v2.0, and we ask for a current one
+    # in test-requirements.txt. Unfortunately we may still get an old
+    # version if there's a system package already installed.
+    # Catch this and show a diagnostic to avoid obscure errors.
+    if psutil.version_info < (2, 0):
+      raise Exception('Wrong psutil version, found %s. Use virtualenv?' % (
```
cmccoy: Maybe include the required version here too and/or refer to `test-requirements.txt`?

cmccoy:
Confirmed that this behaves reasonably with psutil 1.2.1, 2.0.0, 3.0.1.

klausw:
I spent entirely too much time yesterday experimenting with package versioning. TL;DR: it's messy, I think best we can do is catch misconfigurations and point people to virtualenv or similar. Below is a summary for posterity.

The *pkg_resources* module is supposed to be able to help with version management, but it doesn't quite behave as expected.

Putting required versions into `__main__.__requires__[]` should work, but doesn't really help for nosetests since (AFAIK) there's no easy way to get this set in the main module. `__requires__` in other modules are ignored, and the `require()` function doesn't help:

```
$ python
>>> import pkg_resources
>>> pkg_resources.require('psutil>=2.0.0')
[psutil 3.0.0 (/usr/local/lib/python2.7/dist-packages)]
>>> import psutil
>>> print psutil.__version__
1.2.1
>>> import sys
>>> print sys.modules['psutil']
<module 'psutil' from '/usr/lib/python2.7/dist-packages/psutil/__init__.pyc'>
```

Core issue seems to be that the system psutil is effectively invisible to pkg_resources due to missing metadata, so it doesn't detect a version conflict.

I had tried to use pkg_resources to parse requirements.txt and check if they are met, but this doesn't really work right when the module it locates isn't necessarily the one that an "import" would load, so I guess we're stuck with checking version numbers.

I can update the current sanity check to parse the requirements file, though I'm not sure if it's worth the extra complexity.

klausw:
FWIW, `__requires__` in `__main__` also doesn't work for this case:

```
>>> __requires__=['psutil>=2.0.0']
>>> import pkg_resources
>>> import psutil
>>> psutil.__version__
'1.2.1'
```

It's supposed to throw an error like this:
```
>>> __requires__=['psutil>=4.0.0']
>>> import pkg_resources
Traceback (most recent call last): ...
pkg_resources.DistributionNotFound: psutil>=4.0.0
```

skschneider:
Order might matter. Try this:
```
>>> __requires__=['psutil>=2.0.0']
>>> import psutil
>>> import pkg_resources
```

klausw:
@skschneider, the pkg_resources import is supposed to rearrange sys.path based on the `__requires__` setting as a side effect to ensure that a following import loads the expected package. I tested that variant also to confirm it won't work.

@cmccoy, I added a separate commit with a new `requirements_test.py` that uses info from the resources file. Personally I think it's overcomplicated since there's a bunch of special cases, and at this point it effectively only tests psutil anyway. What do you think? I'm leaning towards keeping the hardcoded version number in the sanity check unless there's a way to do this more cleanly.

cmccoy:
@klausw: Up to you, but I'm not sure we should invest so much to ensure that the user has set up their environment correctly to run unit tests.
I'd rather invest in automating test environment creation and runs (I'm quite fond of [`tox`](https://tox.readthedocs.org/en/latest/) for that). That way, *every* user gets the same isolated predictable test environment.

klausw:
*tox* sounds like the right way to go - IMHO the current approach of asking the user to use pip for installing prerequisites seems dubious if there's no guarantee that the downloaded packages will actually end up being used.

OK to keep the simple warning for wrong psutil version? I'm also ok with abandoning this pull request entirely, if needed we can add a note to a FAQ to explain the issue for reference. Not sure how common it will be for people to run into this.

cmccoy:
> OK to keep the simple warning for wrong psutil version?

Sounds good to me. I'll open a PR for tox support.

cmccoy:
tox support added in #330. Since pre-submits and CI runs are now running via tox, maybe we could close this?

klausw:
Closing. The tox support handles this much better than standard *pip*. While some kind of diagnostic for misconfigurations might be nice, probably not worth the extra complexity if it's just for the test setup.

We may want to revisit this if we run into versioning issues with non-test pkb, though in that case we may want a tox-like wrapper to ensure the environment is sane.

PR #325, opened 2015-06-18
Fix lint error in log_util_test.py

klausw:
Sorry about that, need to see why flake8 doesn't seem to be running
in my presubmits.


ehankland:
LGTM

PR #328, opened 2015-06-19
Fix a bad SSH options in hadoop_terasort.

cmccoy:
The version of `ssh` on CentOS 6 is case sensitive. Not an issue on Ubuntu / CentOS 7.


ehankland:
LGTM

PR #333, opened 2015-06-23
Adding support to ContainerizedVM

gablg1:
This PR adds the ContainerizedDebian class, which allows anyone to run benchmarks within a Ubuntu docker container. Every call to RemoteCommand runs within the Docker container via docker exec, whereas any call to RemoteHostCommand is guaranteed to run the command in the VM, outside the container.

This was only tested in GCE so far.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -15,7 +15,6 @@
 """Container for all data required for a benchmark to run."""
 
 import logging
-import os
 import pickle
 
 from perfkitbenchmarker import disk
@@ -97,8 +96,10 @@ flags.DEFINE_enum(
     'os_type is "rhel". In general if two OS\'s use the same package manager, '
     'and are otherwise very similar, the same os_type should work on both of '
     'them.')
-flags.DEFINE_string('scratch_dir', '/',
-                    'Directory in the VM where scratch disks will be mounted.')
+flags.DEFINE_string('scratch_dir', '/scratch',
+                    'Base name for all scratch disk directories in the VM.'
+                    'Upon creation, these directories will have numbers'
+                    'appended to them (for example /scratch0, /scratch1, etc).')
 
 
 class BenchmarkSpec(object):
@@ -165,7 +166,7 @@ class BenchmarkSpec(object):
         else:
           num_striped_disks = FLAGS.num_striped_disks
         for i in range(benchmark_info['scratch_disk']):
-          mount_point = os.path.join(FLAGS.scratch_dir, 'scratch%d' % i)
+          mount_point = '%s%d' % (FLAGS.scratch_dir, i)
           disk_spec = disk.BaseDiskSpec(
               self.scratch_disk_size, self.scratch_disk_type,
               mount_point, self.scratch_disk_iops,
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -294,6 +294,7 @@ class BenchmarkSpec(object):
       firewall.AllowPort(vm, port)
     vm.AddMetadata(benchmark=self.benchmark_name)
     vm.WaitForBootCompletion()
+    vm.OnStartup()
     if FLAGS.scratch_disk_type == disk.LOCAL:
       vm.SetupLocalDisks()
     for disk_spec in vm.disk_specs:
@@ -301,7 +302,7 @@ class BenchmarkSpec(object):
 
     # This must come after Scratch Disk creation to support the
     # Containerized VM case
-    vm.OnStartup()
+    vm.PrepareVMEnvironment()
 
   def DeleteVm(self, vm):
     """Deletes a single vm and scratch disk if required.
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -33,7 +33,7 @@ import perfkitbenchmarker.deployment.shared.ini_constants as ini_constants
 from perfkitbenchmarker.digitalocean import digitalocean_network
 from perfkitbenchmarker.digitalocean import digitalocean_virtual_machine
 from perfkitbenchmarker.gcp import gce_network
-from perfkitbenchmarker.gcp import gce_virtual_machine
+from perfkitbenchmarker.gcp import gce_virtual_machine as gce_vm
 
 GCP = 'GCP'
 AZURE = 'Azure'
@@ -53,10 +53,10 @@ FIREWALL = 'firewall'
 CLASSES = {
     GCP: {
         VIRTUAL_MACHINE: {
-            DEBIAN: gce_virtual_machine.DebianBasedGceVirtualMachine,
-            RHEL: gce_virtual_machine.RhelBasedGceVirtualMachine,
-            UBUNTU_CONTAINER: gce_virtual_machine.ContainerizedGceVirtualMachine,
-            WINDOWS: gce_virtual_machine.WindowsGceVirtualMachine
+            DEBIAN: gce_vm.DebianBasedGceVirtualMachine,
+            RHEL: gce_vm.RhelBasedGceVirtualMachine,
+            UBUNTU_CONTAINER: gce_vm.ContainerizedGceVirtualMachine,
+            WINDOWS: gce_vm.WindowsGceVirtualMachine
         },
         FIREWALL: gce_network.GceFirewall
     },
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -15,6 +15,7 @@
 """Container for all data required for a benchmark to run."""
 
 import logging
+import os
 import pickle
 
 from perfkitbenchmarker import disk
@@ -42,6 +43,7 @@ DIGITALOCEAN = 'DigitalOcean'
 DEBIAN = 'debian'
 RHEL = 'rhel'
 WINDOWS = 'windows'
+UBUNTU_CONTAINER = 'ubuntu_container'
 IMAGE = 'image'
 WINDOWS_IMAGE = 'windows_image'
 MACHINE_TYPE = 'machine_type'
@@ -54,6 +56,7 @@ CLASSES = {
         VIRTUAL_MACHINE: {
             DEBIAN: gce_virtual_machine.DebianBasedGceVirtualMachine,
             RHEL: gce_virtual_machine.RhelBasedGceVirtualMachine,
+            UBUNTU_CONTAINER: gce_virtual_machine.ContainerizedGceVirtualMachine,
             WINDOWS: gce_virtual_machine.WindowsGceVirtualMachine
         },
         FIREWALL: gce_network.GceFirewall
@@ -88,12 +91,14 @@ FLAGS = flags.FLAGS
 flags.DEFINE_enum('cloud', GCP, [GCP, AZURE, AWS, DIGITALOCEAN],
                   'Name of the cloud to use.')
 flags.DEFINE_enum(
-    'os_type', DEBIAN, [DEBIAN, RHEL, WINDOWS],
+    'os_type', DEBIAN, [DEBIAN, RHEL, UBUNTU_CONTAINER, WINDOWS],
     'The VM\'s OS type. Ubuntu\'s os_type is "debian" because it is largely '
     'built on Debian and uses the same package manager. Likewise, CentOS\'s '
     'os_type is "rhel". In general if two OS\'s use the same package manager, '
     'and are otherwise very similar, the same os_type should work on both of '
     'them.')
+flags.DEFINE_string('scratch_dir', '/',
+                    'Directory in the VM where scratch disks will be mounted.')
 
 
 class BenchmarkSpec(object):
@@ -160,9 +165,10 @@ class BenchmarkSpec(object):
         else:
           num_striped_disks = FLAGS.num_striped_disks
         for i in range(benchmark_info['scratch_disk']):
+          mount_point = os.path.join(FLAGS.scratch_dir, 'scratch%d' % i)
           disk_spec = disk.BaseDiskSpec(
               self.scratch_disk_size, self.scratch_disk_type,
-              '/scratch%d' % i, self.scratch_disk_iops,
+              mount_point, self.scratch_disk_iops,
               num_striped_disks)
           vm.disk_specs.append(disk_spec)
 
@@ -288,11 +294,14 @@ class BenchmarkSpec(object):
       firewall.AllowPort(vm, port)
     vm.AddMetadata(benchmark=self.benchmark_name)
     vm.WaitForBootCompletion()
-    vm.OnStartup()
     if FLAGS.scratch_disk_type == disk.LOCAL:
       vm.SetupLocalDisks()
     for disk_spec in vm.disk_specs:
       vm.CreateScratchDisk(disk_spec)
+
+    # This must come after Scratch Disk creation to support the
+    # Containerized VM case
+    vm.OnStartup()
 
   def DeleteVm(self, vm):
     """Deletes a single vm and scratch disk if required.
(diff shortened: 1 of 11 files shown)


dyema:
@ehankland

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -309,12 +314,15 @@ def PrepareVm(self, vm, firewall):
...
     vm.WaitForBootCompletion()
-    vm.OnStartup()
     if FLAGS.scratch_disk_type == disk.LOCAL:
       vm.SetupLocalDisks()
     for disk_spec in vm.disk_specs:
       vm.CreateScratchDisk(disk_spec)
 
+    # This must come after Scratch Disk creation to support the
+    # Containerized VM case
+    vm.OnStartup()
```
ehankland: I think this will break RHEL for benchmarks that use disks - when they boot, you can't use sudo over ssh without using the '-t' option - we do some setup in the startup that gets around this. The Windows support I'm adding also expects OnStartup to be called immediately after boot before configuring the disks.

ehankland:
test this please.

ehankland:
("test this please" triggers our integration/unit tests)

ehankland:
Looks like our linter failed - can you please run "flake8" locally and fix the errors it gives you?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -660,3 +678,131 @@ def GetServiceName(self, package_name):
...
+
+      # Default container directory is /
+      if container_path == '':
+        container_path = '/%s' % file_name
+
+      # Need to ensure container_path is not a directory
+      if container_path[-1] == '/':
+        container_path = os.path.join(container_path, file_name)
+
+      command = "cat %s | sudo docker exec -i %s bash -c 'cat > %s'" \
```
ehankland: Is this the recommended way to do the copy? The solution here seems cleaner to me:
http://stackoverflow.com/questions/22907231/copying-files-from-host-to-docker-container

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -99,10 +99,11 @@ def Prepare(benchmark_spec):
                  '1 loader node, 3 data nodes')
     vm_dict[LOADER_NODE] = [vm_dict['default'][-1]]
     vm_dict[DATA_NODE] = vm_dict['default'][:3]
+    mount_point = os.path.join(vm_util.GetTempDir(), 'cassandra_data')
```
ehankland: GetTempDir gets you the temporary directory on the controller - not on the VM

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -110,13 +113,14 @@
...
 flags.DEFINE_enum(
-    'os_type', DEBIAN, [DEBIAN, RHEL],
+    'os_type', DEBIAN, [DEBIAN, RHEL, UBUNTU_CONTAINER],
     'The VM\'s OS type. Ubuntu\'s os_type is "debian" because it is largely '
     'built on Debian and uses the same package manager. Likewise, CentOS\'s '
     'os_type is "rhel". In general if two OS\'s use the same package manager, '
     'and are otherwise very similar, the same os_type should work on both of '
     'them.')
-
+flags.DEFINE_string('scratch_dir', '/',
```
ehankland: Does the default work when running with docker or do you set this to something else? If not, it should be documented and you may also want to set it to whatever value does work if the user hasn't supplied a value.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -257,8 +261,19 @@ def RemoteCommand(self, command,
                     should_log=False, retries=SSH_RETRIES,
                     ignore_failure=False, login_shell=False,
                     suppress_warning=False):
+    return self.RemoteHostCommand(command, should_log, retries,
+                                  ignore_failure, login_shell,
+                                  suppress_warning)
+
+  def RemoteHostCommand(self, command,
```
ehankland: Since RemoteHostCommand, RemoteHostCopy, and MoveHostFile won't be used outside of the containerized vm class, I'd prefer them only defined inside that class - this also means overriding a few other methods that use those methods

gablg1: The issue here is that RemoteHostCommand, RemoteHostCopy, and MoveHostFile are used outside of the containerized vm class. Namely in methods like FormatDisk, WaitForBootCompletion, and other stuff within BaseLinuxMixin

ehankland: Thats what I was suggesting you override in the containerized class

perfkitbenchmarker/benchmark_spec.py
gablg1: The default works when running with docker. This was added because '/' is read only in the specific image I'm using, so for this case I need scratch_dir=/usr/local but that's a particularity of my use case.

ehankland: Sounds good - thanks for explaining

perfkitbenchmarker/linux_virtual_machine.py
gablg1: This would lead to a lot of repeated code though... For example I'd copy paste the entirety of BaseLinuxMixin.RemoteCopy into ContainerizedDebianMixin.RemoteHostCopy
Same thing with the methods listed above.

Do you really think this is the best approach? If so, I can go ahead and make the changes.

perfkitbenchmarker/benchmark_spec.py
gablg1: In that case, is moving the package-manager specific commands (i.e. AptUpdate) to a different method (maybe PrepareVMEnvironment?) that gets called after Scratch disk creation a viable idea? docker run has to be called after mounting the volumes, but it has to be called before any apt-get commands

perfkitbenchmarker/linux_virtual_machine.py
ehankland: ContainerizedDebianMixin.RemoteHostCopy can just call super(ContainerizedDebianMixin, self).RemoteCopy(...)   - same for remotehostcommand and movehostfile - there wouldn't be too much duplicate code. And I do feel that adding Host methods to the linuxmixin complicates things unnecessarily

gablg1:
test this please.

gablg1:
@ehankland

ehankland:
test this please.

perfkitbenchmarker/gcp/gce_virtual_machine.py
```diff
@@ -215,11 +215,16 @@ def AddMetadata(self, **kwargs):
     vm_util.IssueCommand(cmd)
 
 
+class ContainerizedGceVirtualMachine(GceVirtualMachine,
+                                     linux_vm.ContainerizedDebianMixin):
+  pass
```
ehankland: Please add a default image here

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+                       '--workdir=%s '
+                       '-v %s:%s ' % (CONTAINER_WORK_DIR,
+                                      vm_util.VM_TMP_DIR,
+                                      CONTAINER_MOUNT_DIR))
+    for sd in self.scratch_disks:
+      init_docker_cmd += '-v %s:%s ' % (sd.mount_point, sd.mount_point)
+    init_docker_cmd += '%s sleep infinity ' % UBUNTU_DOCKER_IMAGE
+
+    resp, _ = self.RemoteHostCommand(init_docker_cmd)
+    self.docker_id = resp[:-1]
```
ehankland: Are you removing a new line? Consider using rstrip() instead.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
 
     if commands:
       self.RemoteCommand(";".join(commands))
+
+
+class ContainerizedDebianMixin(DebianMixin):
+  """Class representing a Virtual Machine that runs Remote
```
ehankland: For docstrings try to use a single line (<80 chars) summary  - you can include more info after the summary if needed. See https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Comments#Comments

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+    # Escapes bash sequences
+    command = command.replace('\'', '\'\\\'\'')
+
+    logging.info("Docker running: %s" % command)
+    command = "sudo docker exec %s bash -c '%s'" % (self.docker_id, command)
+    return self.RemoteHostCommand(command, should_log, retries,
+                                  ignore_failure, login_shell, suppress_warning)
+
+  def ContainerCopy(self, file_name, container_path='', copy_to=True):
+    """Copies a file from container_path to the host's vm_util.VM_TMP_DIR
```
ehankland: Same comment about docstring summary length

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+      self.RemoteHostCopy(file_path, tmp_path, copy_to)
+      self.ContainerCopy(file_name, remote_path, copy_to)
+    else:
+      file_name = os.path.basename(remote_path)
+      tmp_path = os.path.join(vm_util.VM_TMP_DIR, file_name)
+      self.ContainerCopy(file_name, remote_path, copy_to)
+      self.RemoteHostCopy(file_path, tmp_path, copy_to)
+
+  def MoveFile(self, target, source_path, remote_path=''):
+    """Copies a file from one Container in a source VM to
```
ehankland: same comment about length

perfkitbenchmarker/linux_virtual_machine.py
gablg1: I just did this because that's what WaitForBootCompletion and CheckJavaVersion do. Should I change them to rstrip as well?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+  def InitDocker(self):
+    """Initializes the docker container daemon."""
+    init_docker_cmd = ('sudo docker run -d '
+                       '--net=host '
+                       '--workdir=%s '
+                       '-v %s:%s ' % (CONTAINER_WORK_DIR,
+                                      vm_util.VM_TMP_DIR,
+                                      CONTAINER_MOUNT_DIR))
+    for sd in self.scratch_disks:
+      init_docker_cmd += '-v %s:%s ' % (sd.mount_point, sd.mount_point)
```
ehankland: Try to avoid loops where you accumulate strings using += - instead create a list and join them. See https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Strings#Strings

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+      ignore_failure: Ignore any failure if set to true.
+      login_shell: Run command in a login shell.
+      suppress_warning: Suppress the result logging from IssueCommand when the
+          return code is non-zero.
+
+    Returns:
+      A tuple of stdout and stderr from running the command.
+    """
+    # Escapes bash sequences
+    command = command.replace('\'', '\'\\\'\'')
```
ehankland: Using " for the outside quotes instead of ' would make this more readable - see https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Strings#Strings

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,127 @@ def SetupProxy(self):
...
+    and vice versa.
+    Args:
+      file_name: Name of the file in the host's vm_util.VM_TMP_DIR.
+      container_path: Optional path of where to copy file on container.
+      copy_to: True to copy to container, False to copy from container.
+    Raises:
+      RemoteExceptionError: If the source container_path is blank.
+    """
+    if copy_to:
+      if container_path == '':
```
ehankland: why not just make container_path default to CONTAINER_WORK_DIR instead of ''?

ehankland: ahh nevermind - I didn't look below

perfkitbenchmarker/linux_virtual_machine.py
ehankland: Yeah - they should use rstrip as well if they are getting rid of newlines at the end

perfkitbenchmarker/virtual_machine.py
```diff
@@ -259,10 +259,18 @@ def WaitForBootCompletion(self):
...
     pass
 
+  def PrepareVMEnvironment(self):
+    """Performs any necessary setup on the VM specific to the OS.
+
+    This will be called once after setting up scratch disks.
+    """
+    pass
+
+
```
ehankland: nit: remove 1 of the 2 blank lines

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +731,137 @@ def SetupProxy(self):
...
+        container_path = CONTAINER_WORK_DIR
+
+      # Everything in vm_util.VM_TMP_DIR is directly accessible
+      # both in the host and in the container
+      source_path = os.path.join(CONTAINER_MOUNT_DIR, file_name)
```
ehankland: For paths inside a linux VM/container use posixpath rather than os.path - otherwise it will fail in windows

gablg1: I believe we have a bigger problem here. Many benchmarks use os.path, so ContainerizedVM running on a Windows machine would break for those benchmarks for the same reason, right?

ehankland: It's possible that a few slipped through, but by and large benchmarks do not use os.path *unless* they are referring to a file on the controller.

gablg1:
Fixed the issues from yesterday @ehankland

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -708,3 +732,137 @@ def SetupProxy(self):
...
+  def RemoteCopy(self, file_path, remote_path='', copy_to=True):
+    """Copies a file to or from the container in the remote VM.
+
+    Args:
+      file_path: Local path to file.
+      remote_path: Optional path of where to copy file inside the container.
+      copy_to: True to copy to VM, False to copy from VM.
+    """
+    if copy_to:
+      file_name = posixpath.basename(file_path)
```
ehankland: I think this has to be:
    if copy_to:
      base_name = os.path.basename(file_path)
      tmp_path = posixpath.join(vm_util.VM_TMP_DIR, base_name)
      self.RemoteHostCopy(file_path, tmp_path, copy_to)
      self.ContainerCopy(base_name, remote_path, copy_to)
    else:
....

gablg1: Good catch

gablg1: Fixed now

ehankland:
Thanks - seems to be working on a windows controller. One last question - are there any benchmarks that this doesn't work on? Or does it work for everything?

gablg1:
Awesome. I tested everything that doesn't require a .boto file (we weren't
very interested in that so I didn't spend time configuring that). It works
on all benchmarks but bonnie++ because it can't run as root by default. The
workaround for Bonnie++ is to add -u root but my adviser said this was
likely not a long term solution so I left it out of the PR.

ehankland:
Can you document that you can't run bonnie++ with that VM type? Other than that, LGTM.

PR #338, opened 2015-06-23
Windows on GCE

ehankland:


ehankland:
test this please.

perfkitbenchmarker/benchmark_sets.py
```diff
@@ -147,12 +149,10 @@ def GetBenchmarksFromFlags():
...
-  # Create a dictionary of valid benchmark names and modules
-  # TODO(voellm): I really think the benchmarks package init should
-  # build the dictionary
-  valid_benchmarks = {}
-  for benchmark_module in benchmarks.BENCHMARKS:
-    valid_benchmarks[benchmark_module.GetInfo()['name']] = benchmark_module
+  if FLAGS.os_type == benchmark_spec.WINDOWS:
+    valid_benchmarks = windows_benchmarks.VALID_BENCHMARKS
+  else:
+    valid_benchmarks = benchmarks.VALID_BENCHMARKS
```
ivansmf: Line 163 below - can you expand the error message to say "benchmark %s not valid on %platform"?

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -247,7 +256,11 @@ def CreateVirtualMachine(self, opt_zone=None):
     if vm:
       return vm
 
-    vm_class = CLASSES[self.cloud][VIRTUAL_MACHINE][FLAGS.os_type]
+    vm_classes = CLASSES[self.cloud][VIRTUAL_MACHINE]
+    if FLAGS.os_type not in vm_classes:
+      raise errors.Error('The cloud "%s" does not support VMs of type "%s".' %
```
ivansmf: I'd phrase it "VMs of type %type are currently not supported on provider %cloud"

perfkitbenchmarker/gcp/gce_virtual_machine.py
```diff
@@ -101,6 +102,8 @@ def _Create(self):
                     'sshKeys=%s' % tf.name,
                     '--metadata',
                     'owner=%s' % FLAGS.owner]
+      for key, value in self.boot_metadata.iteritems():
+        create_cmd.append('%s=%s' % (key, value))
       ssd_interface_option = NVME if NVME in self.image else SCSI
```
ivansmf: I don't think nvme is supported on Windows.

perfkitbenchmarker/windows_benchmarks/cluster_boot_benchmark.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2014 Google Inc. All rights reserved.
```
ivansmf: This seems to be a copy of the linux version. This is a somewhat trivial benchmark, but we should figure out a way not to duplicate code.

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+
+    Args:
+      local_path: Local path to file.
+      remote_path: Optional path of where to copy file on remote host.
+      copy_to: True to copy to vm, False to copy from vm.
+
+    Raises:
+      RemoteCommandError: If there was a problem copying the file.
+    """
+    if ':' in remote_path:
```
ivansmf: Please consider 
drive, path_and_file = os.path.splitdrive(path)

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    set_error_pref = '$ErrorActionPreference="Stop"'
+
+    password = self.password.replace("'", "''")
+    create_cred = (
+        '$pw = convertto-securestring -AsPlainText -Force \'%s\';'
+        '$cred = new-object -typename System.Management.Automation'
+        '.PSCredential -argumentlist %s,$pw' % (password, self.user_name))
+
+    psdrive_name = self.name
+    root = '\\\\%s\\c$' % self.ip_address
```
ivansmf: os.getenv("SystemDrive") is probably safer.

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+  def WaitForBootCompletion(self):
+    """Waits until VM is has booted."""
+    resp, _ = self.RemoteCommand('hostname', suppress_warning=True)
+    if self.bootable_time is None:
+      self.bootable_time = time.time()
+    if self.hostname is None:
+      self.hostname = resp[:-1]
+
+  def OnStartup(self):
+    stdout, _ = self.RemoteCommand('echo $env:TEMP')
```
ivansmf: Surprisingly enough there is a library for temp file. https://docs.python.org/2/library/tempfile.html

Also, if you're getting the local of the temp folder you can use os.getenv("TEMP")

ehankland: This is the *remote* temp directory - we use the tempfile library for the local temp stuff

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    self.temp_dir = os.path.join(stdout.strip(), 'pkb')
+    self.RemoteCommand('mkdir %s' % self.temp_dir)
+    self.DisableGuestFirewall()
```
ivansmf: This arguably belongs on the benchmark, no?

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    if hasattr(package, 'Uninstall'):
+      package.Uninstall()
+
+  def PackageCleanup(self):
+    """Cleans up all installed packages.
+
+    Deletes the temp directory and uninstalls all PerfKit packages.
+    """
+    for package_name in self._installed_packages:
+      self.Uninstall(package_name)
```
ivansmf: What happens if the package was installed to begin with?

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    self.RemoteCommand('rm -recurse -force %s' % self.temp_dir)
+    self.DisableGuestFirewall()
+
```
ivansmf: Why to disable it here?

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+        'Get-WmiObject -class Win32_PhysicalMemory | '
+        'select -exp Capacity')
+    return int(stdout) / 1024
+
+  def _TestReachable(self, ip):
+    """Returns True if the VM can reach the ip address and False otherwise."""
+    try:
+      self.RemoteCommand('ping -n 1 %s' % ip)
+    except errors.VirtualMachine.RemoteCommandError:
+      return False
```
ivansmf: This conflates two errors, no?

perfkitbenchmarker/gcp/gce_virtual_machine.py
ehankland: Right - but the ssd_interface_option will only be "nvme" if the string "nvme" is present in the image name - which it isn't for windows. So this works as it's supposed to

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Meant to enable - already had this fixed - just hadn't pushed because I wanted to wait till you finished reviewing

perfkitbenchmarker/windows_virtual_machine.py
ehankland: If the first VM couldn't be reached we wouldn't have reached this point in the benchmark. Unless something really wrong happens (like the VM goes down), the only reason we'd fail here is if the internal ip isn't reachable. If a VM goes down, I'd argue we have larger problems than thinking we can't reach the internal ip address.

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+
+  def UnzipFile(self, zip_file, dest):
+    """Unzips the file with the given path."""
+    command = ('Add-Type -A System.IO.Compression.FileSystem; '
+               '[IO.Compression.ZipFile]::ExtractToDirectory(\'{zip_file}\', '
+               '\'{dest}\')').format(zip_file=zip_file, dest=dest)
+    self.RemoteCommand(command)
+
+  def DisableGuestFirewall(self):
+    """Disables the guest firewall."""
```
ivansmf: If you want to go powershell all the way there is always Get-NetFirewallProfile | Set-NetFirewallProfile –Enabled False

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    command = 'netsh advfirewall set allprofiles state off'
+    self.RemoteCommand(command)
+
+  def EnableGuestFirewall(self):
+    """Enables the guest firewall."""
```
ivansmf: Same here: Get-NetFirewallProfile | Set-NetFirewallProfile –Enabled False

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    command = 'netsh advfirewall set allprofiles state on'
+    self.RemoteCommand(command)
+
+  def _RunDiskpartScript(self, script):
+    """Runs the supplied Diskpart script on the VM."""
```
ivansmf: The powershell version of this is documented onhttps://technet.microsoft.com/en-us/library/hh848705(v=wps.630).aspx

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+    with vm_util.NamedTemporaryFile(prefix='diskpart') as tf:
+      tf.write(script)
+      tf.close()
+      self.RemoteCopy(tf.name, self.temp_dir)
+      script_path = os.path.join(self.temp_dir, os.path.basename(tf.name))
+      self.RemoteCommand('diskpart /s {script_path}'.format(
+          script_path=script_path))
+
+  def _CreateScratchDiskFromDisks(self, disk_spec, disks):
+    """Helper method to prepare data disks.
```
ivansmf: There is a version of this without the need for disk numbers here: http://www.karafilis.net/windows-2012-storage-virtualization/
He queries for disks that can pool, which excludes the OS disk.

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Looks like that won't be compatible with windows 2008 r2

perfkitbenchmarker/windows_virtual_machine.py
ehankland: That may mix local disks and remote disks

perfkitbenchmarker/windows_virtual_machine.py
ehankland: I think this will break compatibility with 2008 r2 - https://technet.microsoft.com/en-us/library/jj554896(v=wps.630).aspx says 2012 r2 and windows 8.1

perfkitbenchmarker/windows_virtual_machine.py
ehankland: same as above

perfkitbenchmarker/windows_virtual_machine.py
ehankland: The only case where this could happen is with a static vm. With static vms, we don't try to detect if the package is already present on the system - we just have two modes - either we installed it (in which case we will uninstall it here), or we didn't install any packages (in which case we don't uninstall it here). There is a config option to set this on a per VM basis and a flag override.

perfkitbenchmarker/benchmark_sets.py
ehankland: Done

perfkitbenchmarker/benchmark_spec.py
ehankland: Done

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Good point - done

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Done - now we default to the drive that was part of the path passed to copy file - if that didn't have a drive, then we use the system drive environment variable that we got during startup.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -157,7 +162,10 @@ def __init__(self, benchmark_info):
       self.project = FLAGS.project
       defaults = DEFAULTS[self.cloud]
       self.zones = FLAGS.zones or [defaults[ZONE]]
-      self.image = FLAGS.image or defaults[IMAGE]
+      if FLAGS.os_type == WINDOWS:
+        self.image = FLAGS.image or defaults.get(WINDOWS_IMAGE)
```
klausw: I think this would be a lot cleaner if each cloud provider had a {os_type: default_image} map. Currently, would setting --os_type=rhel still install Ubuntu if you don't separately override --image?

perfkitbenchmarker/disk.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, disk_spec):
     self.mount_point = disk_spec.mount_point
     self.iops = disk_spec.iops
 
+    # Windows related attributes.
```
klausw: This isn't necessarily Windows specific. I think keeping count of the per-instance disks would benefit some Linux implementations also, such as the rather ugly /dev/vdN name allocation in azure_disk.py. If the base class can track this generically, would be nice to document this.

perfkitbenchmarker/gcp/gce_virtual_machine.py
```diff
@@ -47,8 +48,8 @@
 FLAGS = flags.FLAGS
 
 SET_INTERRUPTS_SH = 'set-interrupts.sh'
-BOOT_DISK_SIZE_GB = 10
-BOOT_DISK_TYPE = disk.STANDARD
+BOOT_DISK_SIZE_GB = 50
+BOOT_DISK_TYPE = disk.REMOTE_SSD
```
klausw: Is it ok to just change the defaults as part of this CL? If it could affect benchmark performance, seems preferable to do this separately.

ehankland: I can change it back to standard rather than ssd if you'd like - performance will be somewhat poor on Windows VMs in many cases if I do that. It does need a 50 GB boot disk because of the size of the image though.

perfkitbenchmarker/pkb.py
```diff
@@ -179,13 +180,13 @@ def ValidateBenchmarkInfo(benchmark_info):
...
-  valid_benchmark_names = frozenset(benchmark.GetInfo()['name']
-                                    for benchmark in benchmarks.BENCHMARKS)
+  valid_windows_benchmark_names = frozenset(windows_benchmarks.VALID_BENCHMARKS)
+  valid_benchmark_names = frozenset(benchmarks.VALID_BENCHMARKS)
   valid_benchmark_sets = frozenset(benchmark_sets.BENCHMARK_SETS)
   specified_benchmark_names = frozenset(FLAGS.benchmarks)
 
-  return sorted((specified_benchmark_names - valid_benchmark_names) -
-                valid_benchmark_sets)
+  return sorted(((specified_benchmark_names - valid_benchmark_names) -
```
klausw: This logic is getting increasingly convoluted. Is it really worth the effort to check this separately? Looks like benchmark_sets.GetBenchmarkFromFlags already throws "Invalid benchmark" errors for unrecognized ones, and the benefit of being able to report multiple invalid ones in one go seems marginal.

perfkitbenchmarker/windows_packages/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+Packages should, at a minimum, define an install function (Install(vm)).
+If the package manually places files in locations other than the VM's temp
+directory, then it also needs to define an uninstall function (Uninstall(vm)).
+"""
+
+from perfkitbenchmarker import import_util
+
+
+def _LoadPackages():
+  return dict([(module.__name__.split('.')[-1], module) for module in
```
klausw: I think this deserves a comment. Or consider rewriting as a plain loop to make the side effects and return values more explicit.

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+FLAGS = flags.FLAGS
+
+STARTUP_SCRIPT = ('powershell -Command "Enable-PSRemoting -force; '
+                  'Set-Item wsman:\\localhost\\client\\trustedhosts * -Force; '
+                  'Restart-Service WinRM; '
+                  'netsh advfirewall firewall add rule name=\'Port 5985\' '
+                  'dir=in action=allow protocol=TCP localport=5985"')
+DEFAULT_WINRM_PORT = 5985
```
klausw: This number looks suspiciously similar to the separately configured constants in STARTUP_SCRIPT. Consider using .format() to interpolate them there so they stay consistent?

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+      raise errors.VirtualMachine.RemoteCommandError(error_text)
+
+  @vm_util.Retry(log_errors=False, poll_interval=1)
+  def WaitForBootCompletion(self):
+    """Waits until VM is has booted."""
+    resp, _ = self.RemoteCommand('hostname', suppress_warning=True)
+    if self.bootable_time is None:
+      self.bootable_time = time.time()
+    if self.hostname is None:
+      self.hostname = resp[:-1]
```
klausw: Consider resp.rstrip(... whatever it's removing)?

perfkitbenchmarker/windows_virtual_machine.py
```diff
@@ -0,0 +1,306 @@
...
+  def Uninstall(self, package_name):
+    """Uninstalls a Perfkit package on the VM."""
+    package = windows_packages.PACKAGES[package_name]
+    if hasattr(package, 'Uninstall'):
+      package.Uninstall()
+
+  def PackageCleanup(self):
+    """Cleans up all installed packages.
+
+    Deletes the temp directory and uninstalls all PerfKit packages.
```
klausw: This sounds somewhat scary, especially when potentially combined with is_static for a pre-existing machine. Clarify that this is a perfkit specific temp dir (assuming that's what it is)?

klausw:
Overall I think it looks ok, though I don't know enough about Windows scripting to make meaningful comments on that. As discussed offline, at some point it would be nice to replace various `if FLAGS.os_type == WINDOWS` conditionals with controller subclasses or similar, but that can happen as a separate refactor.

perfkitbenchmarker/benchmark_spec.py
ehankland: Fixed in another PR

perfkitbenchmarker/pkb.py
ehankland: Agreed - removed.

perfkitbenchmarker/gcp/gce_virtual_machine.py
ehankland: Now only windows has the different disk type/size

perfkitbenchmarker/windows_packages/__init__.py
ehankland: Commented

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Done

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Done

perfkitbenchmarker/windows_virtual_machine.py
ehankland: Done

perfkitbenchmarker/disk.py
ehankland: Added more documentation in case we want to use it later.

ehankland:
PTAL

perfkitbenchmarker/windows_virtual_machine.py
klausw: Hm, next problem. WindowsGceVirtualMachine uses windows_virtual_machine.STARTUP_SCRIPT which will always contain the hardcoded DEFAULT_WINRM_PORT. If a subclass were to override self.winrm_port, they'd get out of sync again.

Consider either moving the startup script generation to a method, so that it can use the current self attribute value, or removing the DEFAULT_ prefix if it's unlikely that people will want to change it.

klausw:
Looks good as far as overall integration is concerned, can't really comment on the Windows specific side including @ivansmf 's comments.

Please flatten the changes before merging, should avoid having "respond to review comments" commits in the main branch history.

PR #342, opened 2015-06-25
Move defaults from benchmark_spec to vm classes

ehankland:


perfkitbenchmarker/publisher.py
```diff
@@ -132,7 +132,7 @@ def AddMetadata(self, metadata, benchmark_spec):
     metadata = metadata.copy()
     metadata['perfkitbenchmarker_version'] = version.VERSION
     metadata['cloud'] = benchmark_spec.cloud
-    metadata['zones'] = ','.join(benchmark_spec.zones)
+    metadata['zones'] = ','.join(set([vm.zone for vm in benchmark_spec.vms]))
```
klausw: I was confused by the set() wrapper until I realized it's for making them unique. May deserve a comment.

klausw:
Looks good, but the change description should have a bit more detail since the network changes are nontrivial.

perfkitbenchmarker/publisher.py
ehankland: Done.

ehankland:
Added additional description to the commit message

PR #355, opened 2015-06-30
Minor fixes.

klausw:
I've kept the commits separate, let me know if I should also do separate pull requests for each.


ehankland:
LGTM

PR #357, opened 2015-07-01
Fix publisher

ehankland:


perfkitbenchmarker/publisher.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2014 Google Inc. All rights reserved.
+# Copyright 2015 Google Inc. All rights reserved.
```
klausw: AFAIK there's no need to update the copyright date in existing headers. My earlier change just changed the templates, so that people get the current year when they copy them into new files (as prompted by the presubmit check).

klausw:
LGTM - I'll look into what's going wrong with the boilerplate check and handle that separately.

Issue #354, opened 2015-06-29
Aerospike service-threads and transaction-queues should match vm.num_cpus


cmccoy:
These parameters are recommended to match the number of cores per http://www.aerospike.com/docs/reference/configuration/, currently they are fixed at 4.

voellm:
+1 to fix this.  We should follow the [default] recommendations.  Any idea
why Aerospike does not default to this behavior?

On Mon, Jun 29, 2015 at 3:21 PM, Connor McCoy <notifications@github.com>
wrote:

> These parameters are recommended to match the number of cores per
> http://www.aerospike.com/docs/reference/configuration/, currently they
> are fixed at 4.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/354>.
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382
<https://www.google.com/voice/b/0?pli=1#phones>
Blog: http://perfguy.blogspot.com/
Benchmark like a pro... PerfKitBenchmarker
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker>

cmccoy:
I'm not sure, but it's hardly the only product where recommended values are in the docs rather than automatically set.  Opened #360 to address this.

cmccoy:
Resolved by #360.

PR #360, opened 2015-07-01
Aerospike: Use vm.num_cpus where recommended.

cmccoy:
Use vm.num_cpus for Aerospike configurations service-threads, transaction-queues.
This is the recommended value per the documentation.

See #354.

Also modifies `vm_util.NamedTemporaryFile` to optionally delete the created
file (rather than always deleting it), so that rendered templates can be saved
to the run temp dir for debugging.  I can separate that out if preferred.

### DIFF
--- a/perfkitbenchmarker/benchmarks/aerospike_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/aerospike_benchmark.py
@@ -108,10 +108,12 @@ def _PrepareServer(server):
     else:
       devices = [scratch_disk.GetDevicePath()
                  for scratch_disk in server.scratch_disks]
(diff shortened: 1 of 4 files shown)


cmccoy:
test this please.

ehankland:
LGTM

voellm:
It looks like we are missing one thing with this change and that is the ability to override the num_cpus to change the default behavior back to use just 4 CPUs.

This means there will be some surprises in results and we should message it clearly in the coming release.  It was wrong in the first place so I dont think we should add yet another flag to preserve the behavior that was there.

We follow the defaults and when we find an error we should fix it.

PR #361, opened 2015-07-01
Added Iptables configuration

gablg1:
This allows PKB to ensure that UDP and TCP internal connections on a GCP project are always allowed. Some distros have a DROP all iptables policy by default, so we need this for those.

### DIFF
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -801,7 +801,7 @@ class ContainerizedDebianMixin(DebianMixin):
   def RemoteCommand(self, command,
                     should_log=False, retries=SSH_RETRIES,
                     ignore_failure=False, login_shell=False,
-                    suppress_warning=False):
+                    suppress_warning=False, timeout=None):
     """Runs a command inside the container.
 
     Args:
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -157,6 +157,18 @@ class BaseLinuxMixin(virtual_machine.BaseOsMixin):
                     '--delete']
     return self.RemoteCommand(' '.join(wait_command), should_log=False)
 
+  def SetupHostFirewall(self):
+    """Sets up IP table configurations on the VM."""
+    if hasattr(self, 'allowed_ip_range') and self.allowed_ip_range:
+      self.RemoteHostCommand('sudo iptables -A --source %s '
+                             '-p tcp -j ACCEPT' % self.allowed_ip_range)
+      self.RemoteHostCommand('sudo iptables -A INPUT --source %s '
+                             '-p udp -j ACCEPT' % self.allowed_ip_range)
+      self.RemoteHostCommand('sudo iptables -A OUTPUT --destination %s '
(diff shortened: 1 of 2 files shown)


voellm:
Gabriel - This pull request is changing the default behavior of the image which will change default performance.  If you would like this you will need to put it under a flag that is off by default.  We have worked hard to capture a non-modified view of what a customer sees.

gablg1:
Ok. Just added it as a flag! Let me know if there is anything else.

voellm:
Do you still need this pull request or can we close it?  Thanks

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -70,6 +70,10 @@
 # by EXECUTE_COMMAND.
 WAIT_FOR_COMMAND = 'wait_for_command.py'
 
+flags.DEFINE_bool('setup_host_firewall', False,
```
voellm: Can you called this setup_guest_firewall.  Host generally means the server running the virtual machine so I'd like to reduce confusion.

gablg1: We still need this since our image disables all ports by default.

gablg1: What do you think about setup_remote_firewall ?

voellm: Thank works for me.  Its more clear than host.

gablg1:
Just rebased it and also added timeout=None to ContainerizedMixin's RemoteCommand as well (since it was added to the LinuxMixin RemoteCommand).

voellm:
LGTM

@ehankland  - Look ok to you?

gablg1:
@ehankland any updates?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -157,6 +161,18 @@ def RobustRemoteCommand(self, command):
                     '--delete']
     return self.RemoteCommand(' '.join(wait_command), should_log=False)
 
+  def SetupRemoteFirewall(self):
+    """Sets up IP table configurations on the VM."""
+    if hasattr(self, 'allowed_ip_range') and self.allowed_ip_range:
```
ehankland: Can we make this more generic so it will work on other providers? Is there any problem with just allowing all traffic (and not using the allowed ip range)? In most cases there should be a firewall outside of the guest anyway...

gablg1: This way is more secure of course. And other cloud providers could easily do that just by passing allowed_ip_range = everything in their virtual_machine classes

ehankland: This doesn't really supply any additional security unless the cloud provider doesn't have a firewall - and many images dont have iptables on by default anyway. And I'm not sure how likely it is that someone will go to the trouble of modifying the other cloud providers to set allowed_ip_range - there's no documentation that this is what is needed, so unless they happened to see this PR, they will have to figure that out for themselves.

gablg1: Ok. Just made it more generic. Should work for other cloud providers as well!

gablg1:
Ok. Just made it more generic. Should work for other cloud providers as well!

ehankland:
Thanks - LGTM

PR #362, opened 2015-07-06
Removing singleton pattern for Openstack Novaclient.

mateusz-blaszkowski:
Novaclient lib is not designed to be used by multiple threads.
In particular, it may cause race conditions when connecting to different
Openstack services (like Nova and Cinder).


klausw:
Looks good.

Now that you don't have the singleton pattern anymore, would it be possible to make NovaClient a plain subclass of novaclient.Client instead of the `__getattribute__` hack? This may require renaming attributes such as `url` private to avoid conflicts - better to do this as a separate change if applicable.

PR #363, opened 2015-07-06
Refactor hooks, add opt-in pre-push hook.

klausw:
See hooks/README.md for implementation notes.

Motivtion: the old hooks are hard to run manually, and code reuse among
hooks is difficult. Goal of this refactor is to cleanly split the steps
involved:

- determine set of changed files
- run checks on specific files
- report issues based on check results

The old logic was based on running checks on individual files, printing
"1" on STDOUT for bad files. Now, check-* scripts take a list of files
as args and print filenames of bad files on STDOUT. A toplevel "check"
script drives these individual checks and contains common hook logic
such as problem reporting.  The toplevel hooks themselves use the
"check" script to take appropriate actions.


hooks/lib/check-boilerplate.sh
```diff
@@ -0,0 +1,70 @@
...
+  if [[ "${EXT}" == "py" && -x "${FILE}" ]]; then
+    # remove shabang and blank line from top of executable python files.
+    HEADER=$(cat "${FILE}" | tail --lines=+3 | head "-${LINES}")
+  else
+    HEADER=$(head "-${LINES}" "${FILE}")
+  fi
+
+  # Treat errors from diff other than the usual "files differ"
+  # code 1 as fatal problems.
+  DIFFER=$(echo "${HEADER}" | diff - "${REF_FILE}" || [[ $? -le 1 ]])
```
klausw: FYI, the content of this file is mostly just copied from the previous `boilerplate.sh` file, replacing "echo 0" with "continue". I added the diff exit code check here since the original wasn't using strict `-euo pipefail` error checking.

hooks/check
```diff
@@ -0,0 +1,75 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+HOOKS_DIR="$(dirname "$(test -L "$0" && echo "$(dirname $0)/$(readlink "$0")" || echo "$0")")"
+
+status=0
```
cmccoy: Could you be consistent about lower- or upper-case variable names?

hooks/prepare-commit-msg
```diff
@@ -14,78 +14,21 @@
...
-    (
-      echo
-      echo "# *** ERROR: *** Unit test failure."
-      echo "# Your commit will be aborted unless you fix these."
-      echo "#   COMMIT_BLOCKED_ON_UNIT_TESTS"
-      echo
-    ) >> $1
-  fi
+if [[ "${#modified_files[@]}" -ne 0 ]]; then
+  PREPARE_COMMIT_HOOK="true" \
```
klausw: FYI, this environment var is a bit of a hack, but I couldn't think of an easier way to do context-dependent reports. I thought about always printing the extra lines and grepping them away if not needed, but that seems even messier.

hooks/pre-push
```diff
@@ -0,0 +1,58 @@
...
+# check opt-in.
+
+set -euo pipefail
+
+HOOKS_DIR="$(dirname "$(test -L "$0" && echo "$(dirname $0)/$(readlink "$0")" || echo "$0")")"
+remote="$1"
+url="$2"
+z40=0000000000000000000000000000000000000000
+
+if [ -z "$PKB_ORIGIN" ]; then
```
cmccoy: I think we should be consistent about `[[ ]]` versus `[ ]` conditional checks.

cmccoy:
Regarding the pre-push hook: how about `PKB_BASE` instead of `PKB_ORIGIN`? `origin` is usually the name of a remote, but the hook requires a revision.

hooks/check
klausw: Done. Now using uppercase for top-of-file constants or vars exported to the environment, otherwise lowercase.

klausw:
Changed to use PKB_BASE.

klausw:
PTAL - I've changed the error handling for the `prepare-commit-msg` and `check` scripts so that they can use a nonzero exit code iff there are unexpected errors that weren't covered by the text report.

hooks/check
```diff
@@ -0,0 +1,82 @@
...
+# offscreen by diagnostics.
+
+if [[ "${#files_need_boilerplate[@]}" -ne 0 ]]; then
+  echo
+  echo "*** ERROR: *** Some files are missing the required boilerplate"
+  echo "header from hooks/boilerplate.txt:"
+  for file in "${files_need_boilerplate[@]}"; do
+    echo "  ${file}"
+  done
+  echo "See hooks/boilerplate.*.txt for required headers."
```
cmccoy: Should be: `hooks/lib/boilerplate.*.txt`.

cmccoy:
One more small thing, otherwise LGTM. Thanks for fixing this up!

hooks/check
klausw: Fixed.

PR #365, opened 2015-07-06
Fix file headers that don't match expected boilerplate.

klausw:
Running the new `hooks/check-everything` script complains about these
pre-existing files. Fix them to get clean output by default.


klausw:
For background, see https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/363 which refactored the hook scripts. I checked file history for silo_benchmark.py to get its copyright year.

cmccoy:
LGTM.

PR #367, opened 2015-07-07
Fix bug in aerospike

ehankland:
FLAGS.use_local_disk no longer exists. Scratch disk type is already published by default


klausw:
Looks good except that docstrings in the file still refer to the flag. Ok to merge once those are fixed.

klausw:
Thanks! LGTM.

PR #368, opened 2015-07-07
Rackspace: change default machine type to single-core.

klausw:
Other cloud providers have been using single-core machines as the
default machine type, so this keeps things more consistent if run
without explicit machine type. Users can (and should) explicitly
set `--machine_type=...` if they want to test other combinations.


klausw:
@meteorfox - is this ok with you?

meteorfox:
@klausw Sounds good. :+1: 

The reason I chose `general1-4` over `general1-1` was mainly for capacity so most benchmarks have plenty of head room. Benchmarks like cassandra_stress, run very constrained on 1 CPU and 1 GB RAM (CPU is completely saturated, RAM at 70% util for the cassandra process).

ehankland:
LGTM

PR #375, opened 2015-07-08
Add a timeout to iperf parallel benchmark runs to handle iperf hangs

voellm:
This commit adds the ability to set RemoteCommand timeouts to fix an issue where iperf can hang indefinitely.  For iperf we know the length of each benchmark and so here a timeout really makes sense.


klausw:
LGTM

PR #376, opened 2015-07-08
Remove tools/pre-commit/, use hooks/check-everything instead.

klausw:
The old tools/pre-commit/run-precommit.sh duplicates logic in
the hooks directory and wasn't working since it got out of sync,
delete it.


voellm:
LGTM

PR #377, opened 2015-07-08
Expose Azure disk options via two new flags.

ehankland:
The "azure_host_caching" flag determines what kind of host caching (if
any) that will be used on Azure disks. The "azure_storage_type" flag
determines what kind of Azure Storage Account gets created. This allows
users to make use of remote SSD scratch disks.

### DIFF
--- a/perfkitbenchmarker/azure/azure_disk.py
+++ b/perfkitbenchmarker/azure/azure_disk.py
@@ -26,12 +26,25 @@ import logging
 import threading
 
 from perfkitbenchmarker import disk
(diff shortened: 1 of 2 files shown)


stfeng:
Should we also check that the machine type is DS series? (Or at least document it?)

stfeng:
because DS-series is required when using the premium storage on Azure.

ehankland:
I can document it - right now VM creation fails if people don't use DS series

stfeng:
LGTM

Issue #327, opened 2015-06-19
Compose VMs from OS/cloud-specific class and package manager


cmccoy:
Rather than using multiple inheritance as we do now. See #321.

PR #378, opened 2015-07-09
Containerized refactor to add Rhel support

gablg1:
Slight refactor that abstracts ContainerizedMixin away from a specific Linux type. Without this, containers only work on Debian systems (Debian image running Ubuntu container). With this, containers work on both Rhel and Debian systems (as well as any linux flavor).

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -48,6 +48,7 @@ RACKSPACE = 'Rackspace'
 DEBIAN = 'debian'
 RHEL = 'rhel'
 WINDOWS = 'windows'
+CENTOS_CONTAINER = 'centos_container'
 UBUNTU_CONTAINER = 'ubuntu_container'
 IMAGE = 'image'
 WINDOWS_IMAGE = 'windows_image'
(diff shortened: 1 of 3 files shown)


voellm:
Is this ready for review?  If so can you add the label "Ready for review".  Thanks!

voellm:
I plan to roll the next release today.  Is this ready to integrate?

gablg1:
I can fix the conflicts to integrate this, but @ehankland and I are not sure the triple inheritance is a good design decision.

voellm:
Looks like this CL has been not been updated since Sept because of the CLA.

Is there something we can help with?  Should be close the pull request?

Thanks.

ehankland:
I think this was blocked because of the comment above, not because of the CLA. If we fix https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/327 , then fixing this should be much easier.

voellm:
@gablg1 - We would like to complete the pull request but the CLA is not signed and the code needs updating.  For now I'm going to close this because it has been open since July.

We really do appreciate you taking the time to put this together.

gablg1:
Hi Tony, that's fine! The refactor was too bulky with multiple inheritance
anyway

PR #382, opened 2015-07-13
netperf/iperf: add firewall rules for external IP tests only.

cmccoy:
Speeds up runs for internal IP tests.


voellm:
LGTM but the real thing we should do is move every network benchmark to create a GCE network and then configure that.

PR #386, opened 2015-07-16
Fix network deletion when using run stages

ehankland:


stfeng:
LGTM. As long as you can verify this works, check-in! 

Thanks for quickly fixing this!

PR #391, opened 2015-07-21
Adding a timeout to 'apt-get update'

voellm:
Adding a timeout to 'apt-get update' which has known to get stuck indefinitely.


cmccoy:
LGTM.

PR #392, opened 2015-07-22
Iperf package support for RHEL 6

Vukasin92:
Added support for installing iperf on RHEL 6. Previously, it was assumed that the version of RHEL being used is 7.

### DIFF
--- a/perfkitbenchmarker/packages/iperf.py
+++ b/perfkitbenchmarker/packages/iperf.py
@@ -15,9 +15,13 @@
 
 """Module containing iperf installation and cleanup functions."""
 
+import re
+
 from perfkitbenchmarker import errors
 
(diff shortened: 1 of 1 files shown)


voellm:
LGTM

ehankland:
Could you add what error you were hitting with this installation? It should install iperf from the EPEL repository when running on RHEL 6. Testing it myself (before this change), it seems to work correctly.

Vukasin92:
When I run it I get this error:
error: Failed dependencies:
	libc.so.6(GLIBC_2.14)(64bit) is needed by iperf-2.0.4-1.el7.rf.x86_64
	libc.so.6(GLIBC_2.15)(64bit) is needed by iperf-2.0.4-1.el7.rf.x86_64
It happens when it doesn't find iperf in the EPEL repository and tries to install the one for el7 (which requirements are not matched in the version 6).
My linux version is actually Scientific Linux 6.4, but its RHEL base is version 6.

PR #396, opened 2015-07-23
DigitalOcean: fix package installs failing with chfn PAM errors.

klausw:
DigitalOcean sets up the root account with a temporary
password that's set as expired, requiring it to be changed
immediately. This breaks dpkg postinst scripts, for example
running adduser will produce errors:

  # chfn -f 'RabbitMQ messaging server' rabbitmq
  You are required to change your password immediately (root enforced)
  chfn: PAM: Authentication token is no longer valid; new one required

To avoid this, just disable the root password (we don't need it),
and remove the forced expiration.


ehankland:
LGTM

PR #397, opened 2015-07-24
speccpu2006: avoid reporting incomplete results.

klausw:
The benchmark will report an aggregate score even if some of
the component tests failed with a "NR" status, this can be
very misleading. Currently, this happens for machines with
less than 2GB RAM per core.

By default, partial failures are now treated as errors, preventing
reporting of results.

Optionally, you can set --runspec_keep_partial_results to save the
available partial results. In this case, the aggregate score will
be renamed with a "_partial" suffix, for example:
  SPECfp(R)_base2006_partial

### DIFF
--- a/perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
@@ -55,6 +55,13 @@ flags.DEFINE_boolean('runspec_enable_32bit', default=False,
                      'instance types where 64-bit execution may be problematic '
                      ' - i.e. < 1.5-2GB/core)')
 
+flags.DEFINE_boolean('runspec_keep_partial_results', False,
+                     'speccpu will report an aggregate score even if some of '
+                     'the component tests failed with a "NR" status. If this '
+                     'flag is set to true, save the available results and '
+                     'mark metadata with partial=true. If unset, partial '
+                     'failures are treated as errors.')
+
 BENCHMARK_INFO = {'name': 'speccpu2006',
                   'description': 'Run Spec CPU2006',
                   'scratch_disk': True,
@@ -107,7 +114,8 @@ def Prepare(benchmark_spec):
                                              SPECCPU2006_TAR))
 
 
-def ExtractScore(stdout, vm):
+def ExtractScore(stdout, vm,
+                 keep_partial_results=FLAGS.runspec_keep_partial_results):
   """Exact the Spec (int|fp) score from stdout.
 
   Args:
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
```diff
@@ -186,16 +193,26 @@ def ExtractScore(stdout, vm):
...
-  results.append(sample.Sample(spec_name, spec_score, '', metadata))
 
+  missing_results = False
   for benchmark in result_section:
-    # ignore failed runs
-    if re.search('NR', benchmark):
+    # Skip over failed runs, but count them since they make the overall
+    # result invalid.
+    if 'NR' in benchmark:
+      missing_results = True
```
cmccoy: Maybe track the benchmarks which are missing results?

perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
```diff
@@ -186,16 +193,26 @@ def ExtractScore(stdout, vm):
...
       continue
     # name, ref_time, time, score, misc
     name, _, _, score, _ = benchmark.split()
     results.append(sample.Sample(str(name), float(score), '', metadata))
 
+  if missing_results:
+    if FLAGS.runspec_keep_partial_results:
+      spec_name += '_partial'
```
klausw: On second thought, this won't work right if the resulting data uses SQL such as `metric contains "SPECint(R)_rate_base2006"`, I saw an example of that today. A prefix or suffix modified name would still match this.

Looks like it'll need to insert the "partial" in the middle somewhere.

klausw: After further discussion, consensus was that dashboards that use 'contains' should be fixed instead of doing more complex name mangling here.

perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
klausw: Adding warning level logs. Running a test to verify.

perfkitbenchmarker/benchmarks/speccpu2006_benchmark.py
voellm: Please make this a label vs changing the name of the metric.

klausw: Done, see the (shiny new) unit test for sample output.

klausw:
Please review - I've updated the PR and added a unit test for speccpu results parsing.

ehankland:
LGTM

PR #408, opened 2015-07-30
Move Vm SSH authentication away from copy_throughput benchmark

gablg1:
For Private Keys, we must call vm.PushHostFile instead of vm.PushFile. AuthenticateVM in linux_virtual_machine already has that, so it's better to keep that functionality there.


ehankland:
LGTM

PR #412, opened 2015-08-05
On exception: Only cleanup if run stage is all/cleanup.

cmccoy:
The current behavior cleans up whenever `spec.always_call_cleanup` is `True`,
even if `run_stage` is `run`.  This can be annoying when developing benchmarks.


ehankland:
LGTM

cmccoy:
Cherry-picked to master.

PR #415, opened 2015-08-07
Preemptible gcp vms

noahl:


tests/gce_virtual_machine_test.py
```diff
@@ -0,0 +1,47 @@
...
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import virtual_machine
+from perfkitbenchmarker import vm_util
+from perfkitbenchmarker.gcp import gce_virtual_machine
+
+
+class GCEPreemptibleVMFlagTestCase(unittest.TestCase):
+  def testPreemptibleVMFlag(self):
+    with mock.patch(vm_util.__name__ + '.IssueCommand') as issueCommand, \
+            mock.patch('__builtin__.open'), \
```
cmccoy: Why do you need to patch `open` and `NamedTemporaryFile` here?

tests/gce_virtual_machine_test.py
```diff
@@ -0,0 +1,47 @@
...
+            mock.patch(vm_util.__name__ + '.NamedTemporaryFile'):
+      try:
+        old_preemptible_flag = flags.FLAGS.gce_preemptible_vms
```
cmccoy: Prefer patching `gce_virtual_machine.FLAGS`.

tests/gce_virtual_machine_test.py
```diff
@@ -0,0 +1,47 @@
...
+from perfkitbenchmarker import pkb  # noqa
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import virtual_machine
+from perfkitbenchmarker import vm_util
+from perfkitbenchmarker.gcp import gce_virtual_machine
+
+
+class GCEPreemptibleVMFlagTestCase(unittest.TestCase):
+  def testPreemptibleVMFlag(self):
+    with mock.patch(vm_util.__name__ + '.IssueCommand') as issueCommand, \
```
cmccoy: Use `_`-delimited local variable names.

voellm:
Noah - Whats the status of this change?  Have you addressed Connors feedback?

tests/gce_virtual_machine_test.py
noahl: Done.

tests/gce_virtual_machine_test.py
noahl: Because GceVirtualMachine._Create() uses those methods.

cmccoy: Will creating a temporary file in your test cause issues? If not, might be more realistic to leave them.

tests/gce_virtual_machine_test.py
noahl: Done. I just pushed an update which I think addresses all of your comments.

tests/gce_virtual_machine_test.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for perfkitbenchmarker.gcp.gce_virtual_machine"""
+
+import unittest
+import mock
+
+from perfkitbenchmarker import pkb  # noqa
```
cmccoy: Nit: Add a comment indicating why this module was imported.

cmccoy:
Is there any indication on the resulting samples that preemptible VMs were used? Otherwise, LGTM.

tests/gce_virtual_machine_test.py
noahl: Done.

noahl:
I don't think there will be any indication in the samples that the VMs were preemptible. I see two ways to accomplish that. The first option is to record data under fake machine types, like n1-standard-8-preemptible or something like that. I think I could make that change locally to gce_virtual_machine.py. The other option is to make it a flag in the metadata. In the second case, I think I would need to add a new interface for a VM to provide metadata to the publisher. It wouldn't be a huge change, but it would be bigger than the first option.

I don't have strong feelings about this. I think it depends on whether we think preemptible images are sufficiently different than regular ones that they should be considered their own type of machine, and I don't know the answer to that. What do you think?

noahl:
I just uploaded a new commit that annotates samples from GCP VMs with a metadata label 'preemptible'. Possible values are True and False.

perfkitbenchmarker/gcp/gce_virtual_machine.py
```diff
@@ -82,6 +85,13 @@ def __init__(self, vm_spec):
     self.max_local_disks = FLAGS.gce_num_local_ssds
     self.boot_metadata = {}
 
+    if FLAGS.gce_preemptible_vms:
```
cmccoy: self.preemptible = FLAGS.gce_preemptible_vms?

perfkitbenchmarker/gcp/gce_virtual_machine.py
```diff
@@ -226,6 +238,12 @@ def AddMetadata(self, **kwargs):
     cmd.extend(util.GetDefaultGcloudFlags(self))
     vm_util.IssueCommand(cmd)
 
+  def AnnotateSample(self, unused_sender, benchmark_spec, sample):
+    if self.preemptible:
```
cmccoy: No need to switch here.
`sample['metadata']['preemptible'] = self.preemptible`

cmccoy:
Two nits, otherwise LGTM.

perfkitbenchmarker/gcp/gce_virtual_machine.py
noahl: Done.

perfkitbenchmarker/gcp/gce_virtual_machine.py
noahl: Done.

PR #417, opened 2015-08-07
Umount disk when running fio against raw device.

yuyantingzero:


yuyantingzero:
@stfeng, can you take a look?

ehankland:
LGTM

PR #421, opened 2015-08-13
Generate Fio Job Files Automatically

noahl:
Make the fio benchmark generate job files automatically based on
user-specified parameters. Can still be overridden with a
user-supplied job file.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -32,6 +32,8 @@ from perfkitbenchmarker.packages import fio
 LOCAL_JOB_FILE_NAME = 'fio.job'  # used with vm_util.PrependTempDir()
 REMOTE_JOB_FILE_PATH = posixpath.join(vm_util.VM_TMP_DIR, 'fio.job')
 DEFAULT_TEMP_FILE_NAME = 'fio-temp-file'
+MAX_FILE_SIZE_GB = 100
+DISK_USABLE_SPACE_FRACTION = 0.9
 
 
 FLAGS = flags.FLAGS
@@ -40,9 +42,7 @@ flags.DEFINE_string('fio_jobfile', None,
                     'Job file that fio will use. Giving a custom job file '
                     'overrides the other fio options.')
 flags.DEFINE_boolean('against_device', False,
-                     'Test direct against scratch disk. If set True, will '
-                     'create a modified job file in temp directory, '
-                     'which ignores directory and filename parameter.')
+                     'If true, test directly against the scratch disk.')
 flags.DEFINE_string('device_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'This flag is only valid when against_device=True. '
@@ -53,6 +53,10 @@ flags.DEFINE_string('device_fill_size', '100%',
 flags.DEFINE_string('io_depths', '1',
                     'IO queue depths to run on. Can specify a single number, '
                     'like --io_depths=1, or a range, like --io_depths=1-4')
+
+
+FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
+    'against_device', 'device_fill_size', 'io_depth'}
 
 
 IODEPTHS_REGEXP = re.compile(r'(\d+)(-(\d+))?')
@@ -123,44 +127,60 @@ def GetIODepths(io_depths):
     return range(int(match.group(1)), int(match.group(3)) + 1)
 
 
-def GenerateJobFileString(mount_point):
+def GenerateJobFileString(mount_point, against_device,
+                          device_fill_size, io_depths, device_size):
   """Write a fio job file.
 
   Args:
     mount_point: the mount point of the disk we're testing against.
+    against_device: bool. True if we're using a raw disk.
+    device_fill_size: string. The amount of the disk to pre-fill.
+    io_depths: string. The contents of the io_depths flag.
+    device_size: int. The size of the device, in gigabytes.
 
   Returns:
     The contents of a fio job file, as a string.
   """
 
-  if FLAGS.against_device:
+  if against_device:
     filename = mount_point
-    size = FLAGS.device_fill_size
+    size = device_fill_size
   else:
     filename = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
-    size = '100G'
+    size = str(min(MAX_FILE_SIZE_GB,
+                   int(DISK_USABLE_SPACE_FRACTION * device_size))) + 'G'
 
-  return jinja2.Template(JOB_FILE_TEMPLATE).render(
+  return str(jinja2.Template(JOB_FILE_TEMPLATE,
+                             undefined=jinja2.StrictUndefined).render(
       filename=filename,
       size=size,
-      iodepths=GetIODepths(FLAGS.io_depths))
+      iodepths=GetIODepths(io_depths)))
 
 
-def JobFileString(vm):
+def JobFileString(fio_jobfile, mount_point, against_device,
+                  device_fill_size, io_depths, device_size):
   """Get the contents of our job file.
 
   Args:
+    fio_jobfile: string or False. The path to the user's jobfile, if provided.
+    mount_point: the mount point of the disk we're testing against.
+    against_device: bool. True if we're using a raw disk.
+    device_fill_size: string. The amount of the disk to pre-fill.
+    io_depths: string. The contents of the io_depths flag.
+    device_size: int. The size of the device, in gigabytes.
+
     vm: the virtual_machine.BaseVirtualMachine that we will run on.
 
   Returns:
     A string containing the user's job file.
   """
 
-  if FLAGS.fio_jobfile:
-    with open(FLAGS.fio_jobfile, 'r') as jobfile:
+  if fio_jobfile:
+    with open(fio_jobfile, 'r') as jobfile:
       return jobfile.read()
   else:
-    return GenerateJobFileString(vm.scratch_disks[0].mount_point)
+    return GenerateJobFileString(mount_point, against_device,
+                                 device_fill_size, io_depths, device_size)
 
 
 def GetInfo():
@@ -181,33 +201,44 @@ def Prepare(benchmark_spec):
 
   """
 
+
   if FLAGS.fio_jobfile:
-    ignored_flags = []
-    if FLAGS.against_device:
-      ignored_flags.append('--against_device')
-    if FLAGS.device_fill_size != '100%':
-      ignored_flags.append('--device_fill_size')
-    if FLAGS.io_depths != '1':
-      ignored_flags.append('--io_depths')
+    ignored_flags = {flag_name for flag_name in FLAGS_IGNORED_FOR_CUSTOM_JOBFILE
+                     if FLAGS[flag_name].present}
 
     if ignored_flags:
       logging.warning('Fio job file specified. Ignoring options "%s"',
                       ', '.join(ignored_flags))
-  if FLAGS.device_fill_size != '100%' and not FLAGS.against_device:
+  if FLAGS['device_fill_size'].present and not FLAGS['against_device'].present:
     logging.warning('--device_fill_size has no effect without --against_device')
 
   vm = benchmark_spec.vms[0]
   logging.info('FIO prepare on %s', vm)
   vm.Install('fio')
 
+  disk = vm.scratch_disks[0]
   if FLAGS.against_device and not FLAGS.fio_jobfile:
-    mount_point = vm.scratch_disks[0].mount_point
+    mount_point = disk.mount_point
     logging.info('Umount scratch disk on %s at %s', vm, mount_point)
     vm.RemoteCommand('sudo umount %s' % mount_point)
 
+    device_path = disk.GetDevicePath()
+    logging.info('Fill scratch disk on %s at %s', vm, device_path)
+    command = (
+        ('sudo %s --filename=%s --ioengine=libaio '
+         '--name=fill-device --blocksize=512k --iodepth=64 '
+         '--rw=write --direct=1 --size=%s') %
+        (fio.FIO_PATH, device_path, FLAGS.device_fill_size))
+    vm.RemoteCommand(command)
+
   job_file_path = vm_util.PrependTempDir(LOCAL_JOB_FILE_NAME)
   with open(job_file_path, 'w') as job_file:
-    job_file.write(JobFileString(vm))
+    job_file.write(JobFileString(FLAGS.fio_jobfile,
+                                 disk.mount_point,
+                                 FLAGS.against_device,
+                                 FLAGS.device_fill_size,
+                                 FLAGS.io_depths,
+                                 disk.disk_size))
     logging.info('Wrote fio job file at %s', job_file_path)
 
   vm.PushFile(job_file_path, REMOTE_JOB_FILE_PATH)
@@ -237,7 +268,14 @@ def Run(benchmark_spec):
   logging.info('FIO Results:')
   stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
 
-  return fio.ParseResults(JobFileString(vm), json.loads(stdout))
+  disk = vm.scratch_disks[0]
+  return fio.ParseResults(JobFileString(FLAGS.fio_jobfile,
+                                        disk.mount_point,
+                                        FLAGS.against_device,
+                                        FLAGS.device_fill_size,
+                                        FLAGS.io_depths,
+                                        disk.disk_size),
+                          json.loads(stdout))
 
 
 def Cleanup(benchmark_spec):
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -81,7 +81,8 @@ flags.DEFINE_string('device_fill_size', '100%',
                     'The valid value can either be an integer, which '
                     'represents the number of bytes to fill or a '
                     'percentage, which represents the percentage '
-                    'of the device.')
+                    'of the device. Default to filling 100% of a raw device or '
+                    '90% of a filesystem.')
 flags.DEFINE_string('io_depths', '1',
                     'IO queue depths to run on. Can specify a single number, '
                     'like --io_depths=1, or a range, like --io_depths=1-4')
@@ -183,7 +184,7 @@ def GetIODepths(io_depths):
 
 def GenerateJobFileString(disk, against_device,
                           scenarios, io_depths, working_set_size):
-  """Write a fio job file.
+  """Make a string with our fio job file.
 
   Args:
     disk: the disk.BaseDisk object we're benchmarking with.
@@ -221,7 +222,10 @@ def GenerateJobFileString(disk, against_device,
 
 def JobFileString(fio_jobfile, disk, against_device,
                   scenario_strings, io_depths, working_set_size):
-  """Get the contents of our job file.
+  """Get the contents of the fio job file we're working with.
+
+  This will either read the user's job file, if given, or generate a
+  new one.
 
   Args:
     fio_jobfile: string or None. The path to the user's jobfile, if provided.
@@ -231,10 +235,9 @@ def JobFileString(fio_jobfile, disk, against_device,
       generate.
     io_depths: iterable. The IO queue depths to test.
     working_set_size: int or None. If int, the size of the working set in GB.
-    vm: the virtual_machine.BaseVirtualMachine that we will run on.
 
   Returns:
-    A string containing the user's job file.
+    A string containing a fio job file.
   """
 
   # The default behavior is to run with the standard fio job file.
@@ -336,10 +339,7 @@ def Run(benchmark_spec):
         required to run the benchmark.
 
   Returns:
-    A list of samples in the form of 3 or 4 tuples. The tuples contain
-        the sample metric (string), value (float), and unit (string).
-        If a 4th element is included, it is a dictionary of sample
-        metadata.
+    A list of sample.Sample objects.
   """
   vm = benchmark_spec.vms[0]
   logging.info('FIO running on %s', vm)
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -20,14 +20,18 @@ Quick howto: http://www.bluestop.org/fio/HOWTO.txt
 
 import json
 import logging
-import os
+import posixpath
+import re
+
+import jinja2
 
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import fio
 
 LOCAL_JOB_FILE_NAME = 'fio.job'  # used with vm_util.PrependTempDir()
-REMOTE_JOB_FILE_PATH = '~/fio.job'
+REMOTE_JOB_FILE_PATH = posixpath.join(vm_util.VM_TMP_DIR, 'fio.job')
+DEFAULT_TEMP_FILE_NAME = 'fio-temp-file'
 
 
 FLAGS = flags.FLAGS
@@ -51,13 +55,28 @@ flags.DEFINE_string('io_depths', '1',
                     'like --io_depths=1, or a range, like --io_depths=1-4')
 
 
+IODEPTHS_REGEXP = re.compile(r'(\d+)(-(\d+))?')
+
+
+def IODepthsValidator(string):
+  match = IODEPTHS_REGEXP.match(string)
+  return match and match.end() == len(string) and int(match.group(1)) > 0
+
+
+flags.RegisterValidator('io_depths',
+                        IODepthsValidator,
+                        message='--io_depths must be an integer '
+                                'or a range of integers, all > 0')
+
+
 BENCHMARK_INFO = {'name': 'fio',
                   'description': 'Runs fio in sequential, random, read '
                                  'and write modes.',
                   'scratch_disk': True,
                   'num_machines': 1}
 
-GLOBALS_TEMPLATE = """
+
+JOB_FILE_TEMPLATE = """
 [global]
 ioengine=libaio
 invalidate=1
@@ -65,20 +84,21 @@ blocksize=4k
 direct=1
 runtime=10m
 time_based
-filename={filename}
+filename={{filename}}
 do_verify=0
 verify_fatal=0
 randrepeat=0
-size={size}
+size={{size}}
 
-"""
-
-SINGLE_JOB_TEMPLATE = """
-[{rwkind}-io-depth-{iodepth}]
+{% for rwkind in ('randread', 'randwrite') %}
+{% for iodepth in iodepths %}
+[{{rwkind}}-io-depth-{{iodepth}}]
 stonewall
-rw={rwkind}
-iodepth={iodepth}
+rw={{rwkind}}
+iodepth={{iodepth}}
 
+{% endfor %}
+{% endfor %}
 """
 
 
@@ -95,15 +115,12 @@ def GetIODepths(io_depths):
     ValueError if FLAGS.io_depths doesn't follow a format it recognizes.
   """
 
-  try:
-    return [int(io_depths)]
-  except ValueError:
-    bounds = io_depths.split('-', 1)
+  match = IODEPTHS_REGEXP.match(io_depths)
 
-    if len(bounds) != 2:
-      raise ValueError
-
-    return range(int(bounds[0]), int(bounds[1]) + 1)
+  if match.group(2) is None:
+    return [int(match.group(1))]
+  else:
+    return range(int(match.group(1)), int(match.group(3)) + 1)
 
 
 def WriteJobFile(mount_point):
@@ -120,15 +137,13 @@ def WriteJobFile(mount_point):
     filename = mount_point
     size = FLAGS.device_fill_size
   else:
-    filename = os.path.join(mount_point, 'fio-temp-file')
+    filename = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
     size = '100G'
-  return (GLOBALS_TEMPLATE.format(filename=filename, size=size) +
-          '\n'.join((SINGLE_JOB_TEMPLATE.format(rwkind='randread',
-                                                iodepth=str(i))
-                     for i in GetIODepths(FLAGS.io_depths))) +
-          '\n'.join((SINGLE_JOB_TEMPLATE.format(rwkind='randwrite',
-                                                iodepth=str(i))
-                     for i in GetIODepths(FLAGS.io_depths))))
+
+  return jinja2.Template(JOB_FILE_TEMPLATE).render(
+      filename=filename,
+      size=size,
+      iodepths=GetIODepths(FLAGS.io_depths))
 
 
 def JobFileString(vm):
@@ -142,7 +157,8 @@ def JobFileString(vm):
   """
 
   if FLAGS.fio_jobfile:
-    return open(FLAGS.fio_jobfile, 'r').read()
+    with open(FLAGS.fio_jobfile, 'r') as jobfile:
+      return jobfile.read()
   else:
     return WriteJobFile(vm.scratch_disks[0].mount_point)
 
@@ -177,7 +193,7 @@ def Prepare(benchmark_spec):
     if ignored_flags:
       logging.warning('Fio job file specified. Ignoring options "%s"',
                       ', '.join(ignored_flags))
-  if FLAGS.device_fill_size and not FLAGS.against_device:
+  if FLAGS.device_fill_size != '100%' and not FLAGS.against_device:
     logging.warning('--device_fill_size has no effect without --against_device')
 
   vm = benchmark_spec.vms[0]
@@ -237,4 +253,4 @@ def Cleanup(benchmark_spec):
   if not FLAGS.against_device and not FLAGS.fio_jobfile:
     # If the user supplies their own job file, then they have to clean
     # up after themselves, because we don't know their temp file name.
-    vm.RemoveFile(vm.GetScratchDir() + 'fio-temp-file')
+    vm.RemoveFile(posixpath.join(vm.GetScratchDir(), DEFAULT_TEMP_FILE_NAME))
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -65,17 +65,16 @@ SCENARIOS = {
 FLAGS = flags.FLAGS
 
 flags.DEFINE_string('fio_jobfile', None,
-                    'Job file that fio will use. Cannot use with '
-                    '--generate_scenarios.')
+                    'Job file that fio will use. If not given, use a job file '
+                    'bundled with PKB. Cannot use with --generate_scenarios.')
 flags.DEFINE_list('generate_scenarios', None,
-                  'Generate a job file with the given scenarios. If no '
-                  'scenarios are listed, generate all of them. Cannot use '
-                  'with --fio_jobfile.')
+                  'Generate a job file with the given scenarios. Special '
+                  'scenario \'all\' generates all scenarios. Cannot use with '
+                  '--fio_jobfile.')
 flags.DEFINE_boolean('against_device', False,
-                     'If --generate_scenarios is given, test against the raw '
-                     'block device. If --fio_jobfile is given, unmount the '
-                     'device\'s filesystem to enable testing against the '
-                     'device.')
+                     'Unmount the device\'s filesystem so we can test against '
+                     'the raw block device. If --generate-scenarios is given, '
+                     'will generate a job file that uses the block device.')
 flags.DEFINE_string('device_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'This flag is only valid when against_device=True. '
@@ -104,6 +103,24 @@ flags.RegisterValidator('io_depths',
                         IODepthsValidator,
                         message='--io_depths must be an integer '
                                 'or a range of integers, all > 0')
+
+
+def GenerateFillCommand(fio_path, fill_path, fill_size):
+  """Generate a command to sequentially write a device or file.
+
+  Args:
+    fio_path: path to the fio executable.
+    fill_path: path to the device or file to fill.
+    fill_size: amount of device or file to fill, in fio format.
+
+  Returns:
+    A string containing the command.
+  """
+
+  return (('sudo %s --filename=%s --ioengine=libaio '
+           '--name=fill-device --blocksize=512k --iodepth=64 '
+           '--rw=write --direct=1 --size=%s') %
+          (fio_path, fill_path, fill_size))
 
 
 BENCHMARK_INFO = {'name': 'fio',
@@ -198,10 +215,11 @@ def JobFileString(fio_jobfile, disk, against_device,
   """Get the contents of our job file.
 
   Args:
-    fio_jobfile: string or False. The path to the user's jobfile, if provided.
+    fio_jobfile: string or None. The path to the user's jobfile, if provided.
     disk: the disk.BaseDisk object we're benchmarking with.
     against_device: bool. True if we're using a raw disk.
-    scenario_strings: iterable of strings. The workload scenarios to generate.
+    scenario_strings: list of strings or None. The workload scenarios to
+      generate.
     io_depths: iterable. The IO queue depths to test.
 
     vm: the virtual_machine.BaseVirtualMachine that we will run on.
@@ -218,13 +236,13 @@ def JobFileString(fio_jobfile, disk, against_device,
     with open(fio_jobfile, 'r') as jobfile:
       return jobfile.read()
   else:
-    if scenario_strings:
+    if 'all' in scenario_strings:
+      scenarios = SCENARIOS.itervalues()
+    else:
       for name in scenario_strings:
         if name not in SCENARIOS:
           logging.error('Unknown scenario name %s', name)
       scenarios = (SCENARIOS[name] for name in scenario_strings)
-    else:
-      scenarios = SCENARIOS.itervalues()
 
     return GenerateJobFileString(disk, against_device, scenarios, io_depths)
 
@@ -272,24 +290,20 @@ def Prepare(benchmark_spec):
     if FLAGS.device_fill_size:
       device_path = disk.GetDevicePath()
       logging.info('Fill scratch disk on %s at %s', vm, device_path)
-      command = (
-          ('sudo %s --filename=%s --ioengine=libaio '
-           '--name=fill-device --blocksize=512k --iodepth=64 '
-           '--rw=write --direct=1 --size=%s') %
-          (fio.FIO_PATH, device_path, FLAGS.device_fill_size))
+      command = GenerateFillCommand(fio.FIO_PATH,
+                                    device_path,
+                                    FLAGS.device_fill_size)
       vm.RemoteCommand(command)
   else:
     file_path = posixpath.join(disk.mount_point, DEFAULT_TEMP_FILE_NAME)
     # We compute in MB to avoid rounding errors with 1GB disks.
-    size = str(
+    fill_size = str(
         min(MAX_FILE_SIZE_GB * 1000,
             int(DISK_USABLE_SPACE_FRACTION * 1000 * disk.disk_size))) + 'M'
     logging.info('Fill temp file on %s at %s', vm, file_path)
-    command = (
-        ('sudo %s --filename=%s --ioengine=libaio '
-         '--name=fill-file --blocksize=512k --iodepth=64 '
-         '--rw=write --direct=1 --size=%s') %
-        (fio.FIO_PATH, file_path, size))
+    command = GenerateFillCommand(fio.FIO_PATH,
+                                  file_path,
+                                  fill_size)
     vm.RemoteCommand(command)
 
   job_file_path = vm_util.PrependTempDir(LOCAL_JOB_FILE_NAME)
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -25,6 +25,7 @@ import re
 
 import jinja2
 
+from perfkitbenchmarker import data
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import fio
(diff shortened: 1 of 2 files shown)


noahl:
@yuyantingzero @stfeng @cmccoy Would you take a look?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,96 @@
...
+  Raises:
+    ValueError if FLAGS.io_depths doesn't follow a format it recognizes.
+  """
+
+  try:
+    return [int(io_depths)]
+  except ValueError:
+    pass
+
+  bounds = io_depths.split('-', 1)
```
yuyantingzero: Move this part inside except ValueError

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +155,37 @@ def GetInfo():
...
     benchmark_spec: The benchmark specification. Contains all data that is
         required to run the benchmark.
+
   """
-  vms = benchmark_spec.vms
-  vm = vms[0]
+
+  if (FLAGS.fio_jobfile and
+      (FLAGS.against_device or FLAGS.device_fill_size or FLAGS.io_depths)):
+    logging.warning('Fio job file specified. Ignoring other fio options.')
```
yuyantingzero: Give more detailed logging here, current message is a bit confusing.
eg, job_file= ..., against_device=...,

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +155,37 @@ def GetInfo():
...
-    job_file = fio.DeleteParameterFromJobFile(job_file, 'filename')
-    tmp_dir = vm_util.GetTempDir()
-    file_path = vm_util.PrependTempDir(FLAGS.fio_jobfile)
-    logging.info('Write modified job file to temp directory %s', tmp_dir)
-    with open(file_path, 'w') as modified_f:
-      modified_f.write(job_file)
-  vm.PushFile(file_path)
+
+  mount_point = vm.scratch_disks[0].mount_point
+  logging.info('Umount scratch disk on %s at %s', vm, mount_point)
```
yuyantingzero: We only umount the device when testing against raw device. If you are creating a fio test file at mount point (/scratch0). The device must be mounted.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -20,20 +20,21 @@
...
 import logging
+import os
 
-from perfkitbenchmarker import data
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import fio
 
+LOCAL_JOB_FILE_NAME = 'fio.job'  # used with vm_util.PrependTempDir()
+REMOTE_JOB_FILE_PATH = '~/fio.job'
```
cmccoy: Prefer something in `vm_util.VM_TMP_DIR`

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+
+"""
+
+SINGLE_JOB_TEMPLATE = """
+[{rwkind}-io-depth-{iodepth}]
+stonewall
+rw={rwkind}
+iodepth={iodepth}
+
+"""
```
cmccoy: We've been using Jinja2 templates for dynamically-generated config files. Something like:

```
[global]
ioengine=libaio
invalidate=1
blocksize=4k
direct=1
runtime=10m
time_based
filename={{filename}}
do_verify=0
verify_fatal=0
randrepeat=0
size={{size}}

{% for rwkind in ('randread', 'randwrite') %}  # or an input variable
{% for iodepth in iodepths %}
[{{rwkind}}-io-depth-{{iodepth}}]
stonewall
rw={{rwkind}}
iodepth={{iodepth}}
{% endfor %}
{% endfor %}
```

I think that communicates the intent of the configuration slightly better, since the template text and composition aren't separated.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+
+  Args:
+    vm: the virtual_machine.BaseVirtualMachine that we will run on.
+
+  Returns:
+    A string containing the user's job file.
+  """
+
+  if FLAGS.fio_jobfile:
+    return open(FLAGS.fio_jobfile, 'r').read()
```
cmccoy: Prefer using a context manager here:
```{python}
with open(FLAGS.fio_jobfile, 'r') as fp:
  return fp.read()
```
which guarantees that the file will be closed.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+
+  Returns:
+    The contents of a fio job file, as a string.
+  """
+
+  if FLAGS.against_device:
+    filename = mount_point
+    size = FLAGS.device_fill_size
+  else:
+    filename = os.path.join(mount_point, 'fio-temp-file')
```
cmccoy: Introduce a constant for `fio-temp-file`

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -152,8 +221,10 @@ def Cleanup(benchmark_spec):
...
-  vm = vms[0]
+  vm = benchmark_spec.vms[0]
   logging.info('FIO Cleanup up on %s', vm)
-  vm.RemoveFile(flags.FLAGS.fio_jobfile)
-  vm.RemoveFile(vm.GetScratchDir() + flags.FLAGS.fio_benchmark_filename)
+  vm.RemoveFile(REMOTE_JOB_FILE_PATH)
+  if not FLAGS.against_device and not FLAGS.fio_jobfile:
+    # If the user supplies their own job file, then they have to clean
+    # up after themselves, because we don't know their temp file name.
+    vm.RemoveFile(vm.GetScratchDir() + 'fio-temp-file')
```
cmccoy: `posixpath.join(vm.GetScratchDir(), 'fio-temp-file')`

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+  """Parse the io_depths parameter.
+
+  Args:
+    io_depths: a string in the format of the --io_depths flag.
+
+  Returns:
+    An iterable of integers.
+
+  Raises:
+    ValueError if FLAGS.io_depths doesn't follow a format it recognizes.
```
cmccoy: Consider [registering a validator](https://github.com/google/python-gflags/blob/master/gflags.py#L166).

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+  """
+
+  try:
+    return [int(io_depths)]
```
cmccoy: Consider checking that all io_depths are >= 0

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I added logging that tells you which options are being ignored. Does this satisfy your concern?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed. Thanks!

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -53,6 +57,95 @@
...
+  return (GLOBALS_TEMPLATE.format(filename=filename, size=size) +
+          '\n'.join((SINGLE_JOB_TEMPLATE.format(rwkind='randread',
+                                                iodepth=str(i))
+                     for i in GetIODepths(FLAGS.io_depths))) +
+          '\n'.join((SINGLE_JOB_TEMPLATE.format(rwkind='randwrite',
+                                                iodepth=str(i))
+                     for i in GetIODepths(FLAGS.io_depths))))
+
+
+def JobFileString(vm):
```
yuyantingzero: Method name seems confusing. Consider GetJobFileContent?

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -123,26 +200,18 @@ def Run(benchmark_spec):
...
-    fio_command = '--filename=%s %s' % (vm.scratch_disks[0].GetDevicePath(),
-                                        FLAGS.fio_jobfile)
-  else:
-    file_path = data.ResourcePath(FLAGS.fio_jobfile)
-    fio_command = '--directory=%s %s' % (
-        vm.GetScratchDir(), FLAGS.fio_jobfile)
-  fio_command = 'sudo %s --output-format=json %s' % (fio.FIO_PATH, fio_command)
+
+  fio_command = 'sudo %s --output-format=json %s' % (fio.FIO_PATH,
+                                                     REMOTE_JOB_FILE_PATH)
```
yuyantingzero: I am concerned about the REMOTE_JOB_FILE_PATH. When you push the path to vm, the user is perfkit. When you ran fio command, you are root(sudo). The home directory might be different, so please use absolute path in REMOTE_JOB_FILE_PATH.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: I am confused here, why we don't know their temp file name. PKB upload the job file to remote vm from data/ directory?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. Thank you!

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. Also changed other uses of os.path to posixpath.

perfkitbenchmarker/benchmarks/fio_benchmark.py
cmccoy: Also, please use `posixpath.join` to create remote file paths to maintain support for a windows controller.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done in the new validator.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I changed REMOTE_JOB_FILE_PATH to be in vm_util.VM_TMP_DIR, which is defined as an absolute path.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: We don't know their temp file name because we would have to parse their job file in order to know it. It would add a lot of complexity to handle all of the cases correctly, so I think it's best to just say that if the user supplies a file, they can handle cleanup. It shouldn't be too big a deal, because in the common case we just destroy the scratch disk after testing and the file goes away.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I changed it to GenerateJobFileString, so it matches the name JobFileString below it.

noahl:
I've uploaded a new version that I think addresses all of the comments above. I also wanted to make sure everyone knows that this PR removes some functionality from the fio benchmark. Specifically, I removed the memory_multiple option, because it was very closely tied to the default fio job file, and I don't think that made sense in a world where we're generating lots of different files. I also removed the ability to scan user-supplied job files and change the filename parameter. I did that because I think that scanning for filename correctly in all cases is hard (there might be multiple filenames, or there might be no filename parameter), and if we can't do it in all cases, then it seems better to just say to the user that we can't rather than trying and maybe silently failing.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +76,93 @@
...
+  """
+
+  if FLAGS.against_device:
+    filename = mount_point
+    size = FLAGS.device_fill_size
+  else:
+    filename = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
+    size = '100G'
+
+  return jinja2.Template(JOB_FILE_TEMPLATE).render(
```
cmccoy: Nit: no need to change, but [`VirtualMachine.RenderTemplate(...)`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/virtual_machine.py#L334) is similar.

If you leave as-is, could you set `undefined=jinja2.StrictUndefined` in the Template constructor? Otherwise missing variables will be left blank, which can be hard to debug.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +76,93 @@
...
+
+  match = IODEPTHS_REGEXP.match(io_depths)
+
+  if match.group(2) is None:
+    return [int(match.group(1))]
+  else:
+    return range(int(match.group(1)), int(match.group(3)) + 1)
+
+
+def WriteJobFile(mount_point):
```
cmccoy: To the extent you can, prefer taking arguments to accessing the FLAGS global - it makes testing easier.

The naming is also a bit confusing...this doesn't actually write the fio job.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +76,93 @@
...
+  Returns:
+    The contents of a fio job file, as a string.
+  """
+
+  if FLAGS.against_device:
+    filename = mount_point
+    size = FLAGS.device_fill_size
+  else:
+    filename = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
+    size = '100G'
```
cmccoy: Could you make this a constant?
Will this work for disks < 100G?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +170,47 @@ def GetInfo():
...
     benchmark_spec: The benchmark specification. Contains all data that is
         required to run the benchmark.
+
   """
-  vms = benchmark_spec.vms
-  vm = vms[0]
+
+  if FLAGS.fio_jobfile:
+    ignored_flags = []
+    if FLAGS.against_device:
```
cmccoy: Could you:

* Check `FLAGS[flag_name].present` instead of repeating the default values here?
* Convert these individual checks to a for loop?

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+from perfkitbenchmarker.benchmarks import fio_benchmark
+
+
+class TestGetIODepths(unittest.TestCase):
+  def testOneInteger(self):
+    self.assertEqual(list(fio_benchmark.GetIODepths('3')), [3])
+
+  def testIntegerRange(self):
+    self.assertEqual(list(fio_benchmark.GetIODepths('3-5')), [3, 4, 5])
```
cmccoy: Could you add a test for `WriteJobFile`? Even a sanity check that the template renders successfully would be helpful.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +170,47 @@ def GetInfo():
...
+      ignored_flags.append('--against_device')
+    if FLAGS.device_fill_size != '100%':
+      ignored_flags.append('--device_fill_size')
+    if FLAGS.io_depths != '1':
+      ignored_flags.append('--io_depths')
+
+    if ignored_flags:
+      logging.warning('Fio job file specified. Ignoring options "%s"',
+                      ', '.join(ignored_flags))
+  if FLAGS.device_fill_size != '100%' and not FLAGS.against_device:
```
yuyantingzero: This flag is actually used in Prepare stage if writing to raw device.
For example, to use AWS GP. They require prewarm, basically sequentially writting the whole device.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: Why use different size?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. I left it as jinja2.Template rather than VirtualMachine.RenderTemplate because that requires a path on the remote machine to render to, but we need the results on the local machine.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Changed to take arguments rather than using FLAGS. This function also has a different name now.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I've changed it so the size is the maximum of 100G or the size of the scratch disk. The maximum size is a constant. The size shouldn't actually affect our runs at all - since they're time-based, they will keep repeating their workload until they hit their time limit. I believe the only thing this affects is the size of the file we create when we're not running against a raw disk.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. I also have a new constant at the top of the file listing all of the arguments that will be ignored if the user supplies a job file.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I changed it this morning so we won't prewarm if the user supplies a job file. I think that makes sense to give the user the most flexibility. They can always put a prewarm job in their job file if they want to.

tests/benchmarks/fio_benchmark_test.py
noahl: Done.

noahl:
I signed it!

noahl:
The latest revision should fix Connor's points above. The one thing I haven't responded to is Yanting's request for a metadata label that would go on user-specified jobfile runs (or maybe on automatically-generated-jobfile runs) to let us easily see which tests have had the disk prewarmed and which have not. The goal is to prevent accidental apples-to-oranges comparisons when making dashboards.

I want to check what other people think before implementing, but I think a metadata tag like 'generated-jobfile' would be best. We don't know if user jobfiles prewarm or not, so it seems unfair to label them 'not-prewarmed'. Does that work for everyone?

noahl:
A quick note that I ran the current code with disk sizes 50GB and 100GB and it seemed to work, which should mean that scaling the size of the IO down with the size of the disk works properly.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,35 +188,40 @@ def GetInfo():
...
-    #    invalid. Once the benchmark results are returned from Run() an
-    #    invalid (or is that rather a 'valid' flag should be added.
-    exit(1)
-  if FLAGS.against_device:
-    device_path = vm.scratch_disks[0].GetDevicePath()
-    logging.info('Umount scratch disk on %s at %s', vm, device_path)
-    vm.RemoteCommand('sudo umount %s' % vm.GetScratchDir())
+
+  disk = vm.scratch_disks[0]
+  if FLAGS.against_device and not FLAGS.fio_jobfile:
```
yuyantingzero: A bit confused here. The only chance we test against device, is when using the autogenerated job file?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +80,107 @@
                   'num_machines': 1}
 
 
+JOB_FILE_TEMPLATE = """
```
yuyantingzero: We need be careful if replacing the old job file with this new template. This seems to be a completely new benchmark.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Yes, that's how it's set up. I realized after reading your comment that people using custom job files still probably need an option to test against the disk, because you can't unmount a disk from a fio jobfile. I can fix that.

yuyantingzero: Cool. Thanks.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: You're right, and the results are not comparable to the results from the old benchmark. We don't want it to be too easy to confuse the two benchmarks. I'm not sure how to handle this properly. I think the choices are either put something in the metadata to indicate what the benchmark is, make my code a new benchmark, or maybe have fio_benchmark record samples with multiple test names (I was thinking that each rw mode could be its own test, so "read", "randread", "write", and "randwrite" would all be different). I can argue for each of these options, but I can't convince myself that one of them is clearly right.

stfeng: @yuyantingzero @noahl 

I am curious to know why using a job file template here makes it a "completely new benchmark". It's still FIO benchmark. :)

Perhaps it's because we are no longer testing directly against device if we use the template? but we can still specify the filename to something like '/dev/sdb' to test directly against device, right? 

Let me know if I am missing something here...

noahl: @stfeng 

I agreed that it was a new benchmark because I thought the results from the new code wouldn't be comparable to results from the old code. I just checked the job files, and actually the old and new use the same options (I think I did this on purpose), so they are more comparable than I thought.

The new jobfile generation code will test directly against the device if you use the --against_device argument.

yuyantingzero: @stfeng @noahl The default behavior of the new code is doing random RW with varying queue depth, while the default behavior of the old code is doing seq RW, ran RW, which is more like a overview of the disk performance.
If user is using their own fio.job, this PR won't make any difference. However, if someone write some automation using the default old fio benchmark, the published metric will definitely be different and we might break their automation.
I suggest, we should at least mention in CHANGES.md that this is a breaking change that changes the default behavior of the fio benchmark, or go through the community, which is far more complex, since this benchmark is in standard_set.

stfeng: Sgtm. Yes these are comparable job files so the tests should be comparable
too.

Thanks for the explanation.

On Aug 20, 2015, at 9:42 PM, noahl <notifications@github.com> wrote:

In perfkitbenchmarker/benchmarks/fio_benchmark.py
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/421#discussion_r37606266>
:

> @@ -54,6 +80,107 @@
>                    'num_machines': 1}
>
>
> +JOB_FILE_TEMPLATE = """

@stfeng <https://github.com/stfeng>

I agreed that it was a new benchmark because I thought the results from the
new code wouldn't be comparable to results from the old code. I just
checked the job files, and actually the old and new use the same options (I
think I did this on purpose), so they are more comparable than I thought.

The new jobfile generation code will test directly against the device if
you use the --against_device argument.

—
Reply to this email directly or view it on GitHub
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/421/files#r37606266>
.

stfeng: I see. Thanks yanting!

Noah, is it possible to not change the default behavior with your new code?
That is, by default, still use the old job files. And opt-in to use the new
template-based job files generation logic (doing random rw with varying
queue depths).



On Aug 20, 2015, at 10:03 PM, yuyanting <notifications@github.com> wrote:

In perfkitbenchmarker/benchmarks/fio_benchmark.py
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/421#discussion_r37606815>
:

> @@ -54,6 +80,107 @@
>                    'num_machines': 1}
>
>
> +JOB_FILE_TEMPLATE = """

@stfeng <https://github.com/stfeng> @noahl <https://github.com/noahl> The
default behavior of the new code is doing random RW with varying queue
depth, while the default behavior of the old code is doing seq RW, ran RW,
which is more like a overview of the disk performance.
If user is using their own fio.job, this PR won't make any difference.
However, if someone write some automation using the default old fio
benchmark, the published metric will definitely be different and we might
break their automation.
I suggest, we should at least mention in CHANGES.md that this is a breaking
change that changes the default behavior of the fio benchmark, or go
through the community, which is far more complex, since this benchmark is
in standard_set.

—
Reply to this email directly or view it on GitHub
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/421/files#r37606815>
.

noahl: @stfeng @yuyantingzero 

I've been thinking about this, and I'd like to propose that we handle the change by generalizing the template-based code so it can generate something similar to the old fio.job file, and then making that the default. My vision for the new code is that it lets you essentially take the old job file and vary some parameters to see what happens. Right now it only lets you vary queue depth, but we could easily make it vary more parameters without changing the interface.

The parts of the old job file that will be difficult to generate exactly with my new code are the fact that the jobs use different sizes, the fact that the sizes depend on $mb_memory, and the fact that one job has end_fsync=1 and none of the others do. If you two think those are important things to preserve, then I can make it use the old job file by default. But I think it would be much cleaner to have the default be generating a test that happens to look like the old fio.job file, while giving the user the option to still run exactly the old fio.job file.

Also, since we're hoping to release 1.0 soon, I think we should be asking not just will this change break someone's automation, but also, what are the defaults that we want to keep for a long time? If we're going to change, I'd much rather do it before 1.0 than after it.

stfeng: We should decouple changing the default of FIO benchmarks from this pr. This PR is about introducing a way to dynamically generate a job file via a template and use it for various scenarios. 

If it's easy to use the new way to generate the same job file as the test was using before as default, great. Otherwise, we should just retain those job files and focus this PR on its original purpose.

We can address the default behavior of this benchmark in a separate discussion/PR.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -20,24 +20,62 @@
...
+
 
 FLAGS = flags.FLAGS
 
-flags.DEFINE_string('fio_benchmark_filename', 'fio_benchmark_file',
-                    'scratch file that fio will use')
-flags.DEFINE_string('fio_jobfile', 'fio.job', 'job file that fio will use')
-flags.DEFINE_integer('memory_multiple', 10,
-                     'size of fio scratch file compared to main memory size.')
+flags.DEFINE_string('fio_jobfile', None,
```
yuyantingzero: By default, we are still reading the fio.job, consider using "fio.job" as default value.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -20,24 +20,62 @@
...
+                    '--generate_scenarios.')
+flags.DEFINE_list('generate_scenarios', None,
+                  'Generate a job file with the given scenarios. If no '
+                  'scenarios are listed, generate all of them. Cannot use '
+                  'with --fio_jobfile.')
 flags.DEFINE_boolean('against_device', False,
-                     'Test direct against scratch disk. If set True, will '
-                     'create a modified job file in temp directory, '
-                     'which ignores directory and filename parameter.')
+                     'If --generate_scenarios is given, test against the raw '
```
yuyantingzero: I think the documentation here is confusing. From the code, whenever against_device is True, the disk is umounted.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +113,122 @@
...
+  """
+
+  # The default behavior is to run with the standard fio job file.
+  if scenario_strings is None and not fio_jobfile:
+    fio_jobfile = data.ResourcePath('fio.job')
+
+  if fio_jobfile:
+    with open(fio_jobfile, 'r') as jobfile:
+      return jobfile.read()
+  else:
```
yuyantingzero: else block will never be executed, right? Since you reset the fio_jobfile in line 214

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +236,72 @@ def GetInfo():
...
+      command = (
+          ('sudo %s --filename=%s --ioengine=libaio '
+           '--name=fill-device --blocksize=512k --iodepth=64 '
+           '--rw=write --direct=1 --size=%s') %
+          (fio.FIO_PATH, device_path, FLAGS.device_fill_size))
+      vm.RemoteCommand(command)
+  else:
+    file_path = posixpath.join(disk.mount_point, DEFAULT_TEMP_FILE_NAME)
+    # We compute in MB to avoid rounding errors with 1GB disks.
+    size = str(
```
yuyantingzero: You can directly specify G in job file, for example size=100G. Also consider adding a flag for size.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: If scenario_strings is None, else block will never be executed.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +236,72 @@ def GetInfo():
...
-    logging.info('Fill scratch disk on %s at %s', vm, device_path)
+    mount_point = disk.mount_point
+    logging.info('Umount scratch disk on %s at %s', vm, mount_point)
+    vm.RemoteCommand('sudo umount %s' % mount_point)
+
+    if FLAGS.device_fill_size:
+      device_path = disk.GetDevicePath()
+      logging.info('Fill scratch disk on %s at %s', vm, device_path)
+      command = (
+          ('sudo %s --filename=%s --ioengine=libaio '
```
yuyantingzero: Consider making command string a constant

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: This patch changes the semantics of the flag so the fio_jobfile is a path in the user's file system, not a path in our data directory. I think it is confusing now to make the default be 'fio.job' when we don't look for a file of that name. I will add a note in the help string saying what the default is.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I agree - thanks for catching it! I'll fix the documentation.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: The else block can be executed if the user passes --generate_scenarios and does not pass --fio_jobfile. In that case, scenario_strings will either be [] or a list of strings and line 214 will not reset fio_jobfile, so it is possible to hit the else case. I will make the code more clear.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I like your idea a lot, but I ended up factoring the entire process of generating the command out to a global function. It feels like a function already.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: This size only applies to the prewarming, not the job file. I will make that more clear.

The regular size parameter doesn't apply to the generated files, because they use time-based runs, but I do plan to add a flag for working set size. You can expect that patch very soon.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: Thanks. Please also add a note saying it is looking for absolute directory.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: This works fine for the generated fio job. But if user provides a customized fio job with different size and different filename. Current warmup step won't writie the blocks we want to test. @stfeng , what is your suggestion about doing warming up with disk mounted?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I don't think it should look for an absolute directory, actually. I'd like it to behave the same way as other command-line utilities, so the path is relative to your working directory. However, when you pass it a job file, it always copies it to its temporary directory.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +264,69 @@ def GetInfo():
...
+      device_path = disk.GetDevicePath()
+      logging.info('Fill scratch disk on %s at %s', vm, device_path)
+      command = GenerateFillCommand(fio.FIO_PATH,
+                                    device_path,
+                                    FLAGS.device_fill_size)
+      vm.RemoteCommand(command)
+  else:
+    file_path = posixpath.join(disk.mount_point, DEFAULT_TEMP_FILE_NAME)
+    # We compute in MB to avoid rounding errors with 1GB disks.
+    fill_size = str(
```
yuyantingzero: Sorry to make this complicated. To make warmup more reliable:
If fill_device=True, umount the disk, write the device. If not against_device, format and mount the device back.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -20,24 +20,61 @@
...
-flags.DEFINE_string('fio_jobfile', 'fio.job', 'job file that fio will use')
-flags.DEFINE_integer('memory_multiple', 10,
-                     'size of fio scratch file compared to main memory size.')
+flags.DEFINE_string('fio_jobfile', None,
+                    'Job file that fio will use. If not given, use a job file '
+                    'bundled with PKB. Cannot use with --generate_scenarios.')
+flags.DEFINE_list('generate_scenarios', None,
+                  'Generate a job file with the given scenarios. Special '
+                  'scenario \'all\' generates all scenarios. Cannot use with '
+                  '--fio_jobfile.')
```
yuyantingzero: List available scenarios in help string.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +135,128 @@
...
+
+  file_string = str(globals_template.render(filename=filename))
+  size_string = str(working_set_size) + 'G' if working_set_size else '100%'
+  for scenario in scenarios:
+    file_string = file_string + str(job_template.render(
+        name=scenario['name'],
+        rwkind=scenario['rwkind'],
+        blocksize=scenario['blocksize'],
+        iodepths=io_depths,
+        size=size_string))
```
yuyantingzero: If I remember it correctly, if you are testing against file system, size cannot be percentage.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I just tested, and this appears to work.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I tried this and ran into trouble. Eventually Eric helped me figure out why, and I think this will not work.

The reason is that when you fill the unmounted block device, you also get rid of the filesystem structure, so you can't re-mount the device. You could initialize a new filesystem, but then when you're reading from a large file, the file system will consider it unwritten, which I believe means it might just return zeros for untouched blocks without touching the block device. Because of this, I think we have to initialize through files. I've made the code for that cleaner and pushed a new revision.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -61,53 +264,68 @@ def GetInfo():
...
+        # Default to 90% of capacity because the file system will add
+        # some overhead.
+        fill_size = str(DISK_USABLE_SPACE_FRACTION *
+                        1000 * disk.disk_size) + 'M'
+
+    logging.info('Fill file %s on %s', fill_path, vm)
+    command = GenerateFillCommand(fio.FIO_PATH,
+                                  fill_path,
+                                  fill_size)
     vm.RemoteCommand(command)
```
yuyantingzero: The default behavior is changed to if we do nothing with the device_fill_size, we would fill the device? Please at least document the default behavior to avoid confusions.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +135,128 @@
...
+
+  if match.group(2) is None:
+    return [int(match.group(1))]
+  else:
+    return range(int(match.group(1)), int(match.group(3)) + 1)
+
+
+def GenerateJobFileString(disk, against_device,
+                          scenarios, io_depths, working_set_size):
+  """Write a fio job file.
```
yuyantingzero: The documentation for this method is not accurate. This method does not write a job file, it just generate the job file string.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +135,128 @@
...
+        blocksize=scenario['blocksize'],
+        iodepths=io_depths,
+        size=size_string))
+
+  return file_string
+
+
+def JobFileString(fio_jobfile, disk, against_device,
+                  scenario_strings, io_depths, working_set_size):
+  """Get the contents of our job file.
```
yuyantingzero: Be specific about the method. It does not simply get the contents, it also generate a new job file.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +135,128 @@
...
+        name=scenario['name'],
+        rwkind=scenario['rwkind'],
+        blocksize=scenario['blocksize'],
+        iodepths=io_depths,
+        size=size_string))
+
+  return file_string
+
+
+def JobFileString(fio_jobfile, disk, against_device,
```
yuyantingzero: Consider rename the method name to be something more specific.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +135,128 @@
...
+
+  Args:
+    fio_jobfile: string or None. The path to the user's jobfile, if provided.
+    disk: the disk.BaseDisk object we're benchmarking with.
+    against_device: bool. True if we're using a raw disk.
+    scenario_strings: list of strings or None. The workload scenarios to
+      generate.
+    io_depths: iterable. The IO queue depths to test.
+    working_set_size: int or None. If int, the size of the working set in GB.
+    vm: the virtual_machine.BaseVirtualMachine that we will run on.
```
yuyantingzero: the vm doesn't appear in the parameter list of the method.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -123,26 +341,25 @@ def Run(benchmark_spec):
         If a 4th element is included, it is a dictionary of sample
```
yuyantingzero: Return section is outdated, it returns Sample.sample object.

yuyantingzero:
A few comments about readability, otherwise LGTM

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: The default behavior is the same if you use --against_device, because the default value of --device_fill_size is '100%'. It's different if you don't use --against_device, because now we actually know how to fill file systems. I'll make sure the help text says what happens.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I changed it. Thank you for spotting this!

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed. Thank you!

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I've been thinking about this, but I don't know what to call it. Do you have any ideas?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed. Thank you.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed.

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -0,0 +1,59 @@
...
+  def setUp(self):
+    # self.disk = mock.Mock(mount_point='/foo', disk_size=100)
+    # self.disk.GetDevicePath = mock.MagicMock(return_value='/bar')
+
+    self.disk_spec = disk.BaseDiskSpec(100, 'remote_ssd', '/scratch0')
+    self.disk = gce_disk.GceDisk(self.disk_spec, 'foo', 'us-central1-a', 'proj')
+
+  def testAgainstDevice(self):
+    # This just checks that the template renders.
+    fio_benchmark.GenerateJobFileString(
```
yuyantingzero: Could you verify the tests here by self.assertEqual(expected_job, generated_job)?

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -0,0 +1,59 @@
...
+        self.disk,
+        True,
+        [fio_benchmark.SCENARIOS['sequential_read']],
+        range(1, 5),
+        None)
+
+  def testAgainstFile(self):
+    fio_benchmark.GenerateJobFileString(
```
yuyantingzero: Same here.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +136,130 @@
...
+                                 undefined=jinja2.StrictUndefined)
+
+  file_string = str(globals_template.render(filename=filename))
+  size_string = str(working_set_size) + 'G' if working_set_size else '100%'
+  for scenario in scenarios:
+    file_string = file_string + str(job_template.render(
+        name=scenario['name'],
+        rwkind=scenario['rwkind'],
+        blocksize=scenario['blocksize'],
+        iodepths=io_depths,
```
yuyantingzero: When io_depths is a list, the generated job file will be iodepths=[1,2,3,4]? Correct me if I am wrong about how jinja2 works

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: What about GetOrGenerateJobFileString? Just a suggestion, feel free to ignore.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +136,130 @@
...
+    filename = posixpath.join(disk.mount_point, DEFAULT_TEMP_FILE_NAME)
+
+  globals_template = jinja2.Template(GLOBALS_TEMPLATE,
+                                     undefined=jinja2.StrictUndefined)
+  job_template = jinja2.Template(JOB_TEMPLATE,
+                                 undefined=jinja2.StrictUndefined)
+
+  file_string = str(globals_template.render(filename=filename))
+  size_string = str(working_set_size) + 'G' if working_set_size else '100%'
+  for scenario in scenarios:
```
cmccoy: Could you move all this into the template? Mixing logic between this function and the template seems confusing.
At least prefer building up a list and calling `str.join(iterable)` to repeatedly concatenating.

cmccoy: As does having multiple templates which are combined to create a single job file.

tests/benchmarks/fio_benchmark_test.py
noahl: Done.

tests/benchmarks/fio_benchmark_test.py
noahl: Done.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I don't use the iodepths variable directly, there's a for loop in the template. So iodepths will be [1,2,3,4], and then that will produce four jobs in the job file, each with one of those numbers as its iodepth.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I used your suggestion. Thank you!

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I've moved it all into the template except for the filename and the size string logic. I think those would clutter the template too much, but I'm happy to change it if you still think they should be inside.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -54,6 +136,124 @@
...
+invalidate=1
+direct=1
+runtime=10m
+time_based
+filename={{filename}}
+do_verify=0
+verify_fatal=0
+randrepeat=0
+
+{% for scenario in scenarios %}
```
cmccoy: Nice. Thanks for doing this.

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -0,0 +1,119 @@
...
+runtime=10m
+time_based
+filename=/scratch0/fio-temp-file
+do_verify=0
+verify_fatal=0
+randrepeat=0
+
+
+
+[sequential_read-io-depth-1]
```
cmccoy: Nit: consider adding a test with multiple scenarios.

noahl: Done.

noahl:
@yuyantingzero @cmccoy Could you take a look when you have a moment?

yuyantingzero:
LGTM

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -20,31 +20,113 @@
...
+FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
+    'generate_scenarios', 'io_depths'}
+
+
+IODEPTHS_REGEXP = re.compile(r'(\d+)(-(\d+))?')
+
+
+def IODepthsValidator(string):
+  match = IODEPTHS_REGEXP.match(string)
+  return match and match.end() == len(string) and int(match.group(1)) > 0
```
cmccoy: Nit: adding `$` to the regex above will be equivalent to `match.end() == len(string)`.

cmccoy:
LGTM. Thanks!

perfkitbenchmarker/benchmarks/fio_benchmark.py

Issue #169, opened 2015-03-16
Proposal: Use Corrected YCSB for Coordinated-Omission Problem


meteorfox:
As you might be aware, the coordinated-omission problem (unintentionally) can cause biased samples, which can skew your results to be better than what's actually are.

Here's a short description taken from [Nitsan Wakart's blog](http://psy-lob-saw.blogspot.com/2015/03/fixing-ycsb-coordinated-omission.html):

<blockquote>
This issue manifests in many load generating frameworks where the call into the system under test is done synchronously and the measurement thread holds up the next sample while the call is ongoing. This enables the system under test to delay requests that would have been made during the synchronous call thus skewing the sample set
</blockquote>

This has been "cautioned" before by the research community, this is essentially the problem of Open vs Closed  system models as described by this [CMU's paper](http://repository.cmu.edu/cgi/viewcontent.cgi?article=1872&context=compsci)

The same author of the blog shared above has posted a solution for this problem, and a [great detailed post](http://psy-lob-saw.blogspot.com/2015/03/fixing-ycsb-coordinated-omission.html) explaining how it was solved.

The corrected YCSB version is here:
https://github.com/LatencyUtils/YCSB2

How should we approach the update to this benchmark? This is essentially a different workload, even the author renamed it to YCSB2, should this be treated as its own benchmark, or similar to the UnixBench, should we use flags to use the different versions?

More info on Coordinated-Omission:
[How NOT to measure latency](http://www.infoq.com/presentations/latency-pitfalls)
[Mechanical Sympathy's Discussion](https://groups.google.com/forum/#!msg/mechanical-sympathy/icNZJejUHfE/BfDekfBEs_sJ)

voellm:
We should get Connor to comment on this one...

meteorfox:
@voellm During the community meeting, you mentioned there's a Googler maintaining YCSB, I was wondering if we should be looking at the efforts started by others in this [repo](https://github.com/YCSB/YCSB), and if it should be evaluated to be included as the default YCSB in PKB. That repo includes the fixes I mentioned on the comments above plus others improvements by other contributors.

voellm:
+Connor to followup.

On Wed, May 20, 2015 at 9:49 AM, Carlos Torres <notifications@github.com>
wrote:

> @voellm <https://github.com/voellm> During the community meeting, you
> mentioned there's a Googler maintaining YCSB, I was wondering if we should
[...]

PR #428, opened 2015-08-19
Update YCSB to v0.3.0.

cmccoy:
* Now downloading a pre-compiled YCSB release, rather than building with maven.
* Using the new -cp and -jvm-args flags in place of gnarly sed hacks for
  Cloud Bigtable test.

Tested HBase, Cloud Bigtable, and Cassandra.


cmccoy:
This allows for coordinated-omission issues to be addressed, as described in #169, though note that our YCSB usage does not set a target throughput by default.

ehankland:
LGTM

PR #437, opened 2015-08-27
Fio sample fixes

noahl:
This change makes it so that all samples from the same fio run have the same timestamp (which makes sense since they were actually collected at the same time, even though they were recorded at different times), and adds a metadata label 'fio_job', which identifies which job in a fio job file a particular sample is from. The overall goal here is to make it easier to pick out all of the samples from a particular fio job.

I considered making different jobs within a fio job file have different timestamps, but I decided not to because it seemed too much like making claims about timing that we don't actually know. This change makes it clear that the samples are together, but doesn't do more than that.

@yuyantingzero @stfeng Please take a look when you have a chance.


perfkitbenchmarker/publisher.py
```diff
@@ -554,7 +554,11 @@ def AddSamples(self, samples, benchmark, benchmark_spec):
       sample['product_name'] = FLAGS.product_name
       sample['official'] = FLAGS.official
       sample['owner'] = FLAGS.owner
-      sample['timestamp'] = time.time()
+      if 'timestamp' in sample['metadata']:
```
cmccoy: Could we make the timestamp a proper member of the sample.Sample class?

noahl: We could, but it would complicate testing, because a lot of the benchmark tests compare the results of the benchmark against a static list of Sample objects. It wouldn't be too hard to mock the timestamp, but it would mean that all tests have to care about the time, even ones where the benchmark just uses the default time.time() option.

On the other hand, a timestamp feels like a very natural part of a sample to me. What do you think?

cmccoy: I think including it would be helpful. For backwards compatibility, it could be optional, defaulting to the value of `time.time()` at serialization-time.

noahl: @cmccoy I just uploaded a new version that adds the timestamp to the Sample class.

perfkitbenchmarker/sample.py
```diff
@@ -57,11 +58,18 @@ class Sample(collections.namedtuple('Sample', _SAMPLE_FIELDS)):
...
     unit: string. Units for 'value'.
     metadata: dict. Additional metadata to include with the sample.
+    timestamp: float. Unix timestamp.
   """
 
-  def __new__(cls, metric, value, unit, metadata=None, **kwargs):
+  def __new__(cls, metric, value, unit, metadata=None, timestamp=None,
+              **kwargs):
+    if timestamp is None:
+      timestamp = time.time()
```
cmccoy: Making this change in the sample class will alter current behavior (samples from the same benchmark could have fairly different timestamps, whereas currently they are clustered around a single point in time).  I'm not sure that's a bad thing, but maybe worth a discussion?

tests/benchmarks/bonnie_benchmark_test.py
```diff
@@ -30,7 +33,8 @@ def setUp(self):
       self.contents = fp.read()
 
   def testParseCSVResults(self):
-    result = bonnie_benchmark.ParseCSVResults(self.contents)
+    with mock.patch(time.__name__ + '.time', return_value=1.0):
```
cmccoy: Maybe consider:
* adding a [custom type equality function](https://docs.python.org/2/library/unittest.html#unittest.TestCase.addTypeEqualityFunc) which ignores timestamp
* patching time.time in setUp
* a custom assert method.

perfkitbenchmarker/publisher.py
```diff
@@ -538,7 +538,7 @@ def AddSamples(self, samples, benchmark, benchmark_spec):
     for s in samples:
       # Convert input in deprecated format to Sample objects.
       if isinstance(s, (list, tuple)):
-        if len(s) not in (3, 4):
+        if len(s) not in (3, 4, 5):
```
skschneider: Documentation for this function is outdated now.

noahl: Fixed! Thanks for pointing this out.

perfkitbenchmarker/sample.py
noahl: I sent out emails more than a day ago and haven't heard anyone object. I think it's safe to make the change.

tests/benchmarks/bonnie_benchmark_test.py
noahl: Done.

perfkitbenchmarker/sample.py
```diff
@@ -57,12 +58,57 @@ class Sample(collections.namedtuple('Sample', _SAMPLE_FIELDS)):
...
+                                      metadata=metadata or {},
+                                      timestamp=timestamp,
+                                      **kwargs)
 
   def asdict(self):
     """Converts the Sample to a dictionary."""
     return self._asdict()
+
+
+class SamplesTestMixin(object):
```
cmccoy: Please move this to `tests/`

perfkitbenchmarker/sample.py
```diff
@@ -57,12 +58,57 @@ class Sample(collections.namedtuple('Sample', _SAMPLE_FIELDS)):
...
+  """A mixin for unittest.TestCase that adds a type-specific equality
+  predicate for samples.
+  """
+
+  def __init__(self, *args, **kwargs):
+    super(SamplesTestMixin, self).__init__(self, *args, **kwargs)
+
+    self.addTypeEqualityFunc(Sample, self.assertSamplesEqual)
+
+  def assertSamplesEqual(self, a, b, msg=None):
```
cmccoy: Could you add a docstring? At minimum this method should state that timestamp is ignored. Consider also renaming it to make it clear that the timestamp is being ignored.

perfkitbenchmarker/sample.py
noahl: I'm working on this, but I don't think tests/ is in the load path for our test modules when they run. Is there some place in perfkitbenchmarker/ that would work for this? Maybe a new folder called test_util/ ?

perfkitbenchmarker/sample.py
noahl: Done.

perfkitbenchmarker/sample.py
noahl: @cmccoy I just uploaded a new version in which I create a file perfkitbenchmarker/test_util.py and put SamplesTestMixin in that file. I thought about it, and I don't see a reason to make a folder called "test_util" when there would only be one file in it. This way is also parallel to log_util.py and timing_util.py.

perfkitbenchmarker/publisher.py
```diff
@@ -528,17 +528,18 @@ def AddSamples(self, samples, benchmark, benchmark_spec):
...
       benchmark: string. The name of the benchmark.
       benchmark_spec: BenchmarkSpec. Benchmark specification.
     """
     for s in samples:
       # Convert input in deprecated format to Sample objects.
       if isinstance(s, (list, tuple)):
-        if len(s) not in (3, 4):
+        if len(s) not in (3, 4, 5):
           raise ValueError(
               'Invalid sample "{0}": should be 3- or 4-tuple.'.format(s))
```
cmccoy: This doesn't match the check above (3, 4, or 5).  I don't think that extending the legacy format to handle manually specified timestamps is required (it's no longer supported).

perfkitbenchmarker/test_util.py
```diff
@@ -0,0 +1,56 @@
...
+    built-in equality predicate for testing the equality of elements
+    in a list. Since we compare lists of samples a lot, we need a
+    custom test for that.
+    """
+
+    self.assertEqual(len(a), len(b), msg=msg)
+    for i in xrange(len(a)):
+      self.assertIsInstance(a[i], sample.Sample)
+      self.assertIsInstance(b[i], sample.Sample)
+      self.assertSamplesEqualUpToTimestamp(a[i], b[i], msg=msg)
```
cmccoy: This will be a bit confusing if it fails - any way to annotate the message with the index of the sample that failed?

cmccoy:
Two small thins, otherwise LGTM.

perfkitbenchmarker/publisher.py
noahl: I did this because as long as we have the legacy code in here, I felt like I should keep it up-to-date. I will update this PR to fix the message.

However, I've also been looking for users of the old code, and I believe there are none left. Once this lands, I will send a PR to remove it.

perfkitbenchmarker/test_util.py
noahl: Unfortunately the assert* functions have to follow a set interface, so I can't do it that way. I'm uploading a pretty hackish solution instead, but at least it works.

noahl:
@cmccoy I believe this version addresses both of your concerns. Still LGTM?

cmccoy:
@noahl - Yup.

PR #440, opened 2015-08-27
add support for affinity and anti-affinity groups in OpenStack driver

kivio:
Simple add possibility to support OpenStack affinity and ant-affinity policies on booting VMs

### DIFF
--- a/perfkitbenchmarker/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/openstack/os_virtual_machine.py
@@ -34,6 +34,9 @@ flags.DEFINE_boolean('openstack_boot_from_volume', False,
 flags.DEFINE_integer('openstack_volume_size', 20,
                      'Size of the volume (GB)')
 
+flags.DEFINE_enum('openstack_scheduler_policy', None, [None, 'affinity', 'anti-affinity'],
+                  'Add possibility to use affinity or anti-affinity policy in scheduling process')
+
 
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/openstack/os_virtual_machine.py
```diff
@@ -34,6 +35,11 @@
 flags.DEFINE_integer('openstack_volume_size', 20,
                      'Size of the volume (GB)')
 
+flags.DEFINE_enum('openstack_scheduler_policy', NONE,
```
skschneider: Instead of a string that says 'None', could you change this to the Python constant None? Then the NONE variable can be removed.

skschneider: Sorry, I saw your commit history. If the point of this change was because None is incompatible with DEFINE_enum, then how about changing the casing to match the 'affinity' and 'anti-affinity' options?

skschneider:
Adding LGTM. Let us know when you want this merged.

perfkitbenchmarker/openstack/os_virtual_machine.py
kivio: Afinity and ant-Afinity are conflicting options then i think just one option is to use it as a switch.

kivio:
For me in any time. Code is ready and tested then it can be merged right now.

PR #444, opened 2015-08-31
Update Cloud Bigtable ALPN and client jars

cmccoy:
The Cloud Bigtable client is now v0.2.1, and requires an updated ALPN jar.

Tested manually.


yuyantingzero:
LGTM

PR #445, opened 2015-08-31
Add a large object scenario for the CLI based tests in object storage…

stfeng:
… service benchmark. (the default is still old behavior)

### DIFF
--- a/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
@@ -657,7 +657,7 @@ class AzureBlobStorageBenchmark(object):
                            _MakeAzureCommandSuffix(vm.azure_account,
                                                    vm.azure_key,
                                                    False)))
-    if FLAGS.cli_test_size = 'normal':
+    if FLAGS.cli_test_size == 'normal':
       upload_cmd = ('time for i in {0..99}; do azure storage blob upload '
                     '%s/run/data/file-$i.dat %s %s; done' %
                     (scratch_dir,
@@ -672,11 +672,11 @@ class AzureBlobStorageBenchmark(object):
                      self.bucket_name,
                      _MakeAzureCommandSuffix(vm.azure_account,
                                              vm.azure_key,
-                                             True)))      
+                                             True)))
 
     cleanup_local_temp_cmd = 'rm %s/run/temp/*' % scratch_dir
 
-    if FLAGS.cli_test_size = 'normal':
+    if FLAGS.cli_test_size == 'normal':
       download_cmd = ('time for i in {0..99}; do azure storage blob download '
                       '%s file-$i.dat %s/run/temp/file-$i.dat %s; done' % (
                           self.bucket_name,
--- a/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
@@ -59,10 +59,12 @@ flags.DEFINE_enum('object_storage_scenario', 'all',
                   'api_namespace: runs API based benchmarking for namespace '
                   'operations.')
 
-flags.DEFINE_enum('cli_size', 'normal',
+flags.DEFINE_enum('cli_test_size', 'normal',
                   ['normal', 'large'],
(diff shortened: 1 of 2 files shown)


stfeng:
@yuyantingzero please take a look. Thanks!

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -59,6 +59,11 @@
                   'api_namespace: runs API based benchmarking for namespace '
                   'operations.')
 
+flags.DEFINE_enum('cli_size', 'normal',
+                  ['normal', 'large'],
+                  'size of the cli tests. Normal means half small and half \n'
+                  'large files. large means all large files each with 1GB.')
```
thobrla: "large files" has two fairly different meanings here, yes?

For 'normal', a "large file" is 32MB.  For 'large', a "large file" is 1GiB.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -59,6 +59,11 @@
                   'api_namespace: runs API based benchmarking for namespace '
                   'operations.')
 
+flags.DEFINE_enum('cli_size', 'normal',
```
yuyantingzero: This name is a bit confusing. Maybe cli_test_size.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -477,14 +493,18 @@ def _CliBasedTests(output_results, metadata, vm, iteration_count,
                                 'iterations.')
 
   # Report various percentiles.
+  metrics_prefix = ''
+  if FLAGS.cli_size != 'normal':
+    metrics_prefix = '%s ' % FLAGS.cli_size
```
yuyantingzero: Seems like you want to keep the default results format compatible with old ones. I suggest adding a field in metadata cli_test_size: normal/large, which is more consistent. Up to you.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -107,7 +114,10 @@
 
 UPLOAD_THROUGHPUT_VIA_CLI = 'upload throughput via cli Mbps'
 DOWNLOAD_THROUGHPUT_VIA_CLI = 'download throughput via cli Mbps'
+
 CLI_TEST_ITERATION_COUNT = 100
+LARGE_CLI_TEST_ITERATION_COUNT = 20
+
 CLI_TEST_FAILURE_TOLERANCE = 0.05
 # Azure does not parallelize operations in its CLI tools. We have to
```
yuyantingzero: Someone mentioned a while ago Azure supports parallelize operations nowadays. Maybe add that in next PR?

yuyantingzero:
I noticed there is no large cli test for Azure, any reason for that?

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: sure, that would be a separate PR.

stfeng:
Added support for Azure with large cli scenario. Missed it in previous CL. Thanks for pointing this out @yuyantingzero !

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: I thought about this, but that means if existing dashboards based on the metrics names would be broken unless we change those dashboards to also look for the new metadata field (cli_test_size). I'd prefer to add new metric names here so those existing dashboards continue to work.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: Done.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: Good point, clarified the help text here.

yuyantingzero:
LGTM

PR #446, opened 2015-09-01
Azure's latest breaking change broke object stroage benchmark scripts…

stfeng:
…. We are making an adjustment here to our script so it works with Azure's latest Python package again.


yuyantingzero:
LGTM

voellm:
Can you also comment in the README what version of Azure blogs this supports.

PR #450, opened 2015-09-02
Added more aerospike flags

ehankland:


perfkitbenchmarker/benchmarks/aerospike_benchmark.py
```diff
@@ -40,6 +40,19 @@
 flags.DEFINE_enum('aerospike_storage_type', MEMORY, [MEMORY, DISK],
                   'The type of storage to use for Aerospike data. The type of '
                   'disk is controlled by the "scratch_disk_type" flag.')
+flags.DEFINE_integer('aerospike_min_client_threads', 8,
```
yuyantingzero: Should these also be written into metadata?

ehankland: The number of client threads is already written into metadata - but the read percent should be and I dont think that it is - I'll fix that.

ehankland: Done

yuyantingzero:
LGTM

PR #454, opened 2015-09-08
[OpenStack] use volume size from flavor instead of parameter

kivio:
Better and easier option to specify size of volume for VM disk

### DIFF
--- a/perfkitbenchmarker/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/openstack/os_virtual_machine.py
@@ -31,7 +31,7 @@ flags.DEFINE_boolean('openstack_config_drive', False,
 flags.DEFINE_boolean('openstack_boot_from_volume', False,
                      'Boot from volume instead of an image')
 
-flags.DEFINE_integer('openstack_volume_size', -1,
+flags.DEFINE_integer('openstack_volume_size', None,
                      'Size of the volume (GB)')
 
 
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/openstack/os_virtual_machine.py
```diff
@@ -75,10 +75,16 @@ def _Create(self):
         image_id = image.id
         boot_from_vol = []
         if FLAGS.openstack_boot_from_volume:
+
+            if FLAGS.openstack_volume_size > 0:
+                volume_size = FLAGS.openstack_volume_size
+            else:
+                volume_size = flavor.disk
+
```
voellm: I did not look through the rest of the code but it might be worth logging the size chosen here.

voellm:
LGTM

voellm:
Please let me know if you want to merge.

perfkitbenchmarker/openstack/os_virtual_machine.py
kivio: I think logging is required to add to whole OpenStack driver then we can make separated story for that.

perfkitbenchmarker/openstack/os_virtual_machine.py
```diff
@@ -31,7 +31,7 @@
 flags.DEFINE_boolean('openstack_boot_from_volume', False,
                      'Boot from volume instead of an image')
 
-flags.DEFINE_integer('openstack_volume_size', 20,
+flags.DEFINE_integer('openstack_volume_size', -1,
```
skschneider: DEFINE_integer should support None as the default value if you want to use that. I recommend explaining the what the None/-1 means in the flag description string.

kivio:
It can be merged now in my opinion.

PR #463, opened 2015-09-09
Add version to help and log. Add a -v flag.

yuyantingzero:


yuyantingzero:
I signed it!

perfkitbenchmarker/pkb.py
```diff
@@ -468,9 +470,13 @@ def Main(argv=sys.argv):
       '%s:  %s' %
       (set_name, benchmark_sets.BENCHMARK_SETS[set_name]['message'])
       for set_name in benchmark_sets.BENCHMARK_SETS]
-  sys.modules['__main__'].__doc__ = __doc__ + (
-      '\nBenchmarks (default requirements):\n\t%s' %
-      _GenerateBenchmarkDocumentation())
+  sys.modules['__main__'].__doc__ = (
```
skschneider: Looks like the point of this string is to be printed on the local machine. Would os.linesep be more appropriate?

yuyantingzero: AFAIK, Pkb can only run on unix, linux machine, both using \n to separate lines. Any platform you know \n will be broken?

skschneider: Not currently, though I imagine we may eventually want to get things to work on Windows as well. Considering we already have \n scattered throughout the rest of the code, it's not a blocking issue.

skschneider:
LGTM

PR #464, opened 2015-09-10
[OpenStack] use zone from flags

kivio:
make possible to specify custom zone in parameter instead of using 'nova' zone


perfkitbenchmarker/openstack/os_virtual_machine.py
```diff
@@ -112,7 +112,7 @@ def _Create(self):
             key_name=self.key_name,
             security_groups=['perfkit_sc_group'],
             nics=nics,
-            availability_zone='nova',
+            availability_zone=self.zone,
```
voellm: I dont see this set anywhere... did you mean to set this in __init?

cmccoy: It's set in the base class.

PR #467, opened 2015-09-10
Update unixbench repository.

yuyantingzero:


voellm:
LGTM... the next set of changes in UnixBench will include what @meteorfox  did.  There are pull requests there I need to complete and then we can update the version here.

Thanks for doing this.

PR #470, opened 2015-09-11
Add function that can execute parallel calls in new processes.

skschneider:
Useful for dodging GIL or forcing cleanup of a memory-intensive task.

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -276,34 +276,15 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
         '\n'.join([stacktrace for stacktrace in exceptions]))
 
 
-def _ExecuteProcCall(target_arg_tuple):
-  """Function invoked in another process by RunParallelProcesses.
-
-  Executes a specified function call.
-
-  Args:
-    target_arg_tuple: (target, args, kwargs) tuple containing the function to
-        call and the arguments to pass it.
-
-  Returns:
-    (result, exception) tuple. The first element is the return value from the
-    called function, or None if the function raised an exception. The second
-    element is the raised exception, or None if the function succeeded.
-  """
-  target, args, kwargs = target_arg_tuple
-  try:
-    return target(*args, **kwargs), None
-  except:
-    return None, traceback.format_exc()
-
-
 def RunParallelProcesses(target_arg_tuples, max_concurrency=None):
   """Executes function calls concurrently in separate processes.
 
   Args:
     target_arg_tuples: list of (target, args, kwargs) tuples. Each tuple
         contains the function to call and the arguments to pass it.
-    max_concurrency: int. The maximum number of concurrent new processes.
+    max_concurrency: int or None. The maximum number of concurrent new
+        processes. If None, it will default to the number of processors on the
+        machine.
 
   Returns:
     list of function return values in the order corresponding to the order of
@@ -313,16 +294,24 @@ def RunParallelProcesses(target_arg_tuples, max_concurrency=None):
     errors.VmUtil.CalledProcessException: When an exception occurred in any
         of the called functions.
   """
+  call_futures = []
   results = []
-  exceptions = []
+  error_strings = []
   with futures.ProcessPoolExecutor(max_workers=max_concurrency) as executor:
-    for result, exception in executor.map(_ExecuteProcCall, target_arg_tuples):
-      results.append(result)
-      if exception:
-        exceptions.append(exception)
-  if exceptions:
+    for target, args, kwargs in target_arg_tuples:
+      call_futures.append(executor.submit(target, *args, **kwargs))
+    for index, future in enumerate(call_futures):
+      try:
+        results.append(future.result())
+      except:
+        results.append(None)
+        msg = 'Exception occured in call {0}:{1}{2}'.format(
+            index, os.linesep, traceback.format_exc())
+        logging.error(msg)
+        error_strings.append(msg)
+  if error_strings:
     msg = ('The following exceptions occurred during parallel execution: '
-           '{0}'.format('\n'.join([stacktrace for stacktrace in exceptions])))
+           '{0}'.format(os.linesep.join([e for e in error_strings])))
     raise errors.VmUtil.CalledProcessException(msg)
   return results
 
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -15,7 +15,9 @@
 """Set of utility functions for working with virtual machines."""
 
 import contextlib
+import copy
 import logging
+import multiprocessing
 import os
 import random
 import re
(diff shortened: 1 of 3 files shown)


cmccoy:
IMO the [`futures`](http://pythonhosted.org/futures/) package (backport from Python >= 3.2) does a really nice job of this.

skschneider:
Thanks, much simpler this way.

perfkitbenchmarker/vm_util.py
```diff
@@ -276,6 +277,57 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+    return target(*args, **kwargs), None
+  except:
+    return None, traceback.format_exc()
+
+
+def RunParallelProcesses(target_arg_tuples, max_concurrency=None):
+  """Executes function calls concurrently in separate processes.
+
+  Args:
+    target_arg_tuples: list of (target, args, kwargs) tuples. Each tuple
```
cmccoy: Does this interface need to differ from RunThreaded (which takes `(args, kwargs)` tuples)?

perfkitbenchmarker/vm_util.py
```diff
@@ -276,6 +277,57 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+        contains the function to call and the arguments to pass it.
+    max_concurrency: int. The maximum number of concurrent new processes.
```
cmccoy: RunThreaded starts a thread per input argument - this starts a process per core - at minimum document this. How are you planning on using this method?

perfkitbenchmarker/vm_util.py
```diff
@@ -276,6 +277,57 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+    target_arg_tuples.
+
+  Raises:
+    errors.VmUtil.CalledProcessException: When an exception occurred in any
+        of the called functions.
+  """
+  results = []
+  exceptions = []
+  with futures.ProcessPoolExecutor(max_workers=max_concurrency) as executor:
+    for result, exception in executor.map(_ExecuteProcCall, target_arg_tuples):
```
cmccoy: Nit: I think _ExecuteProcCall can be removed
```{python}
futures = []
tracebacks = []
results = []
for target, args, kwargs in target_arg_tuples:
  futures.append(executor.submit(target, args, **kwargs))
for future in futures:
  try:
    results.append(future.get())
  except:
    results.append(None)
    tracebacks.append(traceback.format_exc())
```

perfkitbenchmarker/vm_util.py
```diff
@@ -276,6 +277,57 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+      results.append(result)
+      if exception:
+        exceptions.append(exception)
+  if exceptions:
+    msg = ('The following exceptions occurred during parallel execution: '
+           '{0}'.format('\n'.join([stacktrace for stacktrace in exceptions])))
```
cmccoy: It would be helpful to include the indices and `target_arg_tuples` entries that failed.

tests/vm_util_test.py
```diff
@@ -120,6 +123,44 @@ def run(self):
...
+
+
+def _IncrementCounter(lock, counter):
+  with lock:
+    counter.value += 1
+
+
+class RunParallelProcessesTestCase(unittest.TestCase):
+
+  def testLessThreadsThanConcurrencyLimit(self):
```
cmccoy: Grammar: FewerThreads

perfkitbenchmarker/vm_util.py
skschneider: The difference here is that it is easy to execute different functions concurrently without needing to define a helper function that forwards to the real target. If we want this to be consistent with RunThreaded, I'd prefer to enhance RunThreaded to also support different targets, since we've already encountered PRs that could benefit from it.

perfkitbenchmarker/vm_util.py
```diff
@@ -276,6 +277,57 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+    errors.VmUtil.CalledProcessException: When an exception occurred in any
+        of the called functions.
+  """
+  results = []
+  exceptions = []
+  with futures.ProcessPoolExecutor(max_workers=max_concurrency) as executor:
+    for result, exception in executor.map(_ExecuteProcCall, target_arg_tuples):
+      results.append(result)
+      if exception:
+        exceptions.append(exception)
```
cmccoy: Could you log the exception here? Immediate feedback would be helpful if the function(s) are long-running.

perfkitbenchmarker/vm_util.py
skschneider: Per core? The docstring already uses the term "process" - do you want a comment in the docstring comparing it to RunThreaded?
I use this method to parallelize memory-intensive post-processing of data collected in a benchmark I'm working on.

perfkitbenchmarker/vm_util.py
skschneider: Will do.

perfkitbenchmarker/vm_util.py
skschneider: Ack

tests/vm_util_test.py
skschneider: Ack

perfkitbenchmarker/vm_util.py
skschneider: Ack

perfkitbenchmarker/vm_util.py
cmccoy: I mean to document the maximum number of concurrent processes when `max_concurrency` is not specified. The default behavior of `RunThreaded` creates a thread for each input argument. `ProcessPoolExecutor` launches a process [per cpu cure](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). This could cause confusion if someone expects all the tasks in `target_arg_tuples` to be launched at the same time.

skschneider: Got it, will do.

skschneider: Done.

perfkitbenchmarker/vm_util.py
skschneider: Done.

perfkitbenchmarker/vm_util.py
skschneider: Done.

tests/vm_util_test.py
skschneider: Done.

perfkitbenchmarker/vm_util.py
skschneider: Done.

cmccoy:
Two small things, otherwise LGTM.

skschneider:
Removed [e for e in error_strings]. Reintroduced helper function to preserve traceback and make debugging easier.

cmccoy:
Thanks for fixing that. LGTM.

skschneider:
Rebased.

PR #477, opened 2015-09-16
Collect fio data over time

noahl:
This PR adds a flag --run_for_minutes to the fio benchmark. When supplied, the user can specify a length of time (in minutes) to run for. Fio will then do repeated 10-minute jobs until it hits the user's requested time, reporting statistics for each job individually, and annotating each job's data with its job number and how many minutes into the overall run it started.

This PR also includes a bug fix in the case where fio is run against a formatted disk using a non-default device_fill_size parameter.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -18,6 +18,7 @@ Man: http://manpages.ubuntu.com/manpages/natty/man1/fio.1.html
 Quick howto: http://www.bluestop.org/fio/HOWTO.txt
 """
 
+import datetime
 import json
 import logging
 import posixpath
@@ -90,11 +91,15 @@ flags.DEFINE_integer('working_set_size', None,
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device.',
                      lower_bound=0)
+flags.DEFINE_integer('run_for_minutes', 10,
+                     'Repeat the job scenario(s) for the given number of '
+                     'minutes. Only valid when using --generate_scenarios.',
+                     lower_bound=0)
 
 
 
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
-    'generate_scenarios', 'io_depths'}
+    'generate_scenarios', 'io_depths', 'run_for_minutes'}
 
 
 IODEPTHS_REGEXP = re.compile(r'(\d+)(-(\d+))?$')
@@ -159,6 +164,9 @@ size={{size}}
 {% endfor %}
 {% endfor %}
 """
+
+MINUTES_PER_JOB = 10  # Must match the JOB_FILE_TEMPLATE above.
+SECONDS_PER_MINUTE = 60
 
 
 def GetIODepths(io_depths):
@@ -280,6 +288,10 @@ def Prepare(benchmark_spec):
       logging.warning('Fio job file specified. Ignoring options "%s"',
                       ', '.join(ignored_flags))
 
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
+    logging.warning('Runtime %s will be rounded up to the next multiple of %s '
+                    'minutes.' % (MINUTES_PER_JOB,))
+
   vm = benchmark_spec.vms[0]
   logging.info('FIO prepare on %s', vm)
   vm.Install('fio')
@@ -340,21 +352,46 @@ def Run(benchmark_spec):
 
   fio_command = 'sudo %s --output-format=json %s' % (fio.FIO_PATH,
                                                      REMOTE_JOB_FILE_PATH)
+
+  run_reps = FLAGS.run_for_minutes // MINUTES_PER_JOB
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
+    run_reps += 1
+    # We already warned the user about rounding during the prepare
+    # phase. No need to warn them again here.
+
+  disk = vm.scratch_disks[0]
+
   # TODO(user): This only gives results at the end of a job run
   #      so the program pauses here with no feedback to the user.
   #      This is a pretty lousy experience.
   logging.info('FIO Results:')
-  stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
 
-  disk = vm.scratch_disks[0]
-  return fio.ParseResults(
-      GetOrGenerateJobFileString(FLAGS.fio_jobfile,
-                                 disk,
-                                 FLAGS.against_device,
-                                 FLAGS.generate_scenarios,
-                                 GetIODepths(FLAGS.io_depths),
-                                 FLAGS.working_set_size),
-      json.loads(stdout))
+  samples = []
+  start_time = datetime.datetime.now()
+  for rep_num in xrange(run_reps):
+    logging.info('**** Repetition number %s of %s ****' % (rep_num, run_reps))
+    run_start = datetime.datetime.now()
+    stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
+
+    seconds_since_start = int(round((run_start - start_time).total_seconds()))
+    minutes_since_start = seconds_since_start // SECONDS_PER_MINUTE
+    if seconds_since_start % SECONDS_PER_MINUTE > (SECONDS_PER_MINUTE // 2):
+      minutes_since_start += 1
+    base_metadata = {
+        'repeat_number': rep_num,
+        'minutes_since_start': minutes_since_start
+    }
+    samples.extend(fio.ParseResults(
+        GetOrGenerateJobFileString(FLAGS.fio_jobfile,
+                                   disk,
+                                   FLAGS.against_device,
+                                   FLAGS.generate_scenarios,
+                                   GetIODepths(FLAGS.io_depths),
+                                   FLAGS.working_set_size),
+        json.loads(stdout),
+        base_metadata=base_metadata))
+
+  return samples
 
 
 def Cleanup(benchmark_spec):
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -298,9 +298,8 @@ def Prepare(benchmark_spec):
       fill_size = FLAGS.device_fill_size
     else:
       fill_path = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
-
-      if FLAGS.FlagValuesDict().get('disk_fill_size') is not None:
-        fill_size = FLAGS.disk_fill_size
+      if FLAGS['device_fill_size'].present:
+        fill_size = FLAGS.device_fill_size
       else:
         # Default to 90% of capacity because the file system will add
         # some overhead.
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -35,6 +35,7 @@ LOCAL_JOB_FILE_NAME = 'fio.job'  # used with vm_util.PrependTempDir()
 REMOTE_JOB_FILE_PATH = posixpath.join(vm_util.VM_TMP_DIR, 'fio.job')
 DEFAULT_TEMP_FILE_NAME = 'fio-temp-file'
 DISK_USABLE_SPACE_FRACTION = 0.9
+MINUTES_PER_JOB = 10
 
 
 # This dictionary maps scenario names to dictionaries of fio settings.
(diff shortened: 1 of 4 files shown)


noahl:
@yuyantingzero @cmccoy Will you take a look?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -280,6 +288,10 @@ def Prepare(benchmark_spec):
       logging.warning('Fio job file specified. Ignoring options "%s"',
                       ', '.join(ignored_flags))
 
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
+    logging.warning('Runtime %s will be rounded up to the next multiple of %s '
+                    'minutes.' % (MINUTES_PER_JOB,))
```
yuyantingzero: Missing one argument?

noahl: Yes, thanks for catching! Fixed.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -160,6 +165,9 @@ def GenerateFillCommand(fio_path, fill_path, fill_size):
 {% endfor %}
 """
 
+MINUTES_PER_JOB = 10  # Must match the JOB_FILE_TEMPLATE above.
```
yuyantingzero: Use this number in the job template, so we do not accidentally break the logic here.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -340,21 +352,46 @@ def Run(benchmark_spec):
 
   fio_command = 'sudo %s --output-format=json %s' % (fio.FIO_PATH,
                                                      REMOTE_JOB_FILE_PATH)
+
+  run_reps = FLAGS.run_for_minutes // MINUTES_PER_JOB
```
yuyantingzero: This is not always correct. What if there are more than 1 scenarios in generated job file? Each scenario takes 10 minutes.

yuyantingzero:
Could you add a test for fio benchmark itself on the repeated flow?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -340,21 +352,46 @@ def Run(benchmark_spec):
...
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
```
yuyantingzero: If users gives a customized job file and also gives FLAGS.run_for_minutes, the logic is broken and leads to unexpected behavior. The job file will be executed until run_reps repetitions. We can either make FLAGS.run_for_minutes more generic and work with customized job file, or we check fio_jobfile flag.

perfkitbenchmarker/benchmarks/fio_benchmark.py
cmccoy: Use the logging module's built-in substitution here:
```{python}
logging.warning('Runtime %s will be ...', arg1, ...)
```

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -340,21 +352,46 @@ def Run(benchmark_spec):
...
-                                 disk,
-                                 FLAGS.against_device,
-                                 FLAGS.generate_scenarios,
-                                 GetIODepths(FLAGS.io_depths),
-                                 FLAGS.working_set_size),
-      json.loads(stdout))
+  samples = []
+  start_time = datetime.datetime.now()
+  for rep_num in xrange(run_reps):
+    logging.info('**** Repetition number %s of %s ****' % (rep_num, run_reps))
```
cmccoy: As above, no need for `%`

perfkitbenchmarker/packages/fio.py
```diff
@@ -115,12 +115,13 @@ def FioParametersToJob(fio_parameters):
                                 JOB_STONEWALL_PARAMETER)
 
 
-def ParseResults(job_file, fio_json_result):
+def ParseResults(job_file, fio_json_result, base_metadata={}):
```
cmccoy: [Don't use mutable objects for argument defaults](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments).

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -298,13 +310,13 @@ def Prepare(benchmark_spec):
       fill_size = FLAGS.device_fill_size
     else:
       fill_path = posixpath.join(mount_point, DEFAULT_TEMP_FILE_NAME)
-      if FLAGS['disk_fill_size'].present:
-        fill_size = FLAGS.disk_fill_size
+      if FLAGS['device_fill_size'].present:
```
cmccoy: Is this a bug fix?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed to use built-in substitution. Thanks!

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: The behavior I want is that each scenario is run for the given number of minutes, so I think this is actually the correct logic, but a confusing interface. I updated the help message for the --run_for_minutes option and added a comment here. Thanks for catching this!

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Wow, this is bad. Thanks for catching it! I made the new logic run only for generated job files.

I thought about extending this flag to custom job files, but I don't see a good way to do it. We don't know how long a custom job file will last, so we can't figure out how many repeats we need. I could have made the flag number of repeats instead of minutes to run, but I think minutes are what we really care about - the number of minutes per job is just an implementation detail.

The only way I can see it working is if we start the custom job file and just keep running it until the total accumulated time passes the user's request, which might work. But I'm worried about measuring overhead somewhere in the system rather than actual fio run time, and fixing that would be very complex. I think that could be a separate project.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Fixed. Thanks!

perfkitbenchmarker/packages/fio.py
noahl: Fixed.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Yes, it is. There's another one in this PR too. I can move them to a separate PR if you want.

noahl:
@yuyantingzero @cmccoy I added a test checking that the fio benchmark runs multiple times with the --run_for_minutes flag, and I believe I've addressed all of your other comments with the changes I just pushed.

perfkitbenchmarker/benchmarks/fio_benchmark.py
cmccoy: Please do, or at least note the additional fixes in the PR description. Was the bug present in the last release?

voellm:
I plan to roll the next release today.  Is this ready to integrate?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I don't think the bugs were in the last release. I'd like to keep the bugs in this PR, but I'll make a point of separating distinct changes in future PRs. Sorry for mixing the two this time.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -90,11 +91,17 @@
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device.',
                      lower_bound=0)
+flags.DEFINE_integer('run_for_minutes', 10,
+                     'Repeat the job scenario(s) for the given number of '
+                     'minutes. Only valid when using --generate_scenarios. '
+                     'When using multiple scenarios, each one is run for the '
+                     'given number of minutes.',
+                     lower_bound=0)
```
cmccoy: Should this be 1? What happens when `fio` is set to run for 0 minutes?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -90,11 +91,17 @@
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device.',
                      lower_bound=0)
+flags.DEFINE_integer('run_for_minutes', 10,
+                     'Repeat the job scenario(s) for the given number of '
+                     'minutes. Only valid when using --generate_scenarios. '
+                     'When using multiple scenarios, each one is run for the '
```
cmccoy: Could you document here the interaction between `--run_for_minutes` and `MINUTES_PER_JOB`?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -340,21 +355,59 @@ def Run(benchmark_spec):
...
+
+    samples = []
+    start_time = datetime.datetime.now()
+    for rep_num in xrange(run_reps):
+      logging.info('**** Repetition number %s of %s ****', rep_num, run_reps)
+      run_start = datetime.datetime.now()
+      stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
+
+      seconds_since_start = int(round((run_start - start_time).total_seconds()))
+      minutes_since_start = seconds_since_start // SECONDS_PER_MINUTE
```
cmccoy: Could you  `round(...)` here too?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: If you set it to run for 0 minutes, it should do a dry run. Everything looks exactly like a run, except that it never runs the fio command (and never tries to parse the results). I made this 0 because the behavior at 0 fits the pattern that happens at positive integers, so I didn't feel like I had a reason to forbid it. I can make it 1 - I don't have strong feelings about this.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. I think the `round(...)` version is cleaner, but see what you think. I had to add `float(...)` calls to make it round properly.

perfkitbenchmarker/benchmarks/fio_benchmark.py
cmccoy: Could you update the description to note this fix?

cmccoy:
+1

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Updated. Sorry, I forgot to do this before.

noahl:
@yuyantingzero @cmccoy This branch is ready for review again.

cmccoy:
LGTM

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -90,11 +92,18 @@
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device.',
                      lower_bound=0)
+flags.DEFINE_integer('run_for_minutes', 10,
+                     'Repeat the job scenario(s) for the given number of '
+                     'minutes. Only valid when using --generate_scenarios. '
+                     'When using multiple scenarios, each one is run for the '
+                     'given number of minutes. Time will be rounded up to the '
+                     'next multiple of %s minutes.' % (MINUTES_PER_JOB,),
```
yuyantingzero: extra comma here.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -341,21 +356,58 @@ def Run(benchmark_spec):
...
+      # phase. No need to warn them again here.
+
+    samples = []
+    start_time = datetime.datetime.now()
+    for rep_num in xrange(run_reps):
+      logging.info('**** Repetition number %s of %s ****', rep_num, run_reps)
+      run_start = datetime.datetime.now()
+      stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
+
+      seconds_since_start = int(round((run_start - start_time).total_seconds()))
```
yuyantingzero: Not necessary to create a new variable here if it's only used once.

yuyantingzero:
2 small comments. LGTM

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Removed the extra comma and the unnecessary parentheses.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I'd like to leave the variable there just for readability. If I get rid of it, there will be a very large expression.

PR #479, opened 2015-09-17
Web search: update flag names, minor tweaks.

cmccoy:
For the CloudSuite Web Search benchmark:

* Adds a `cs_websearch_` prefix to all flag names, to prevent confusion
  with other benchmarks that use a JVM.
* Adds a validation function for heap sizes.
* Renames cloudsuite_web_search_benchmark to cloudsuite_websearch_benchmark

### DIFF
--- a/perfkitbenchmarker/benchmarks/cloudsuite_web_search_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/cloudsuite_web_search_benchmark.py
@@ -46,7 +46,7 @@ CLOUDSUITE_WEB_SEARCH_DIR = posixpath.join(vm_util.VM_TMP_DIR,
 
 FABAN_OUTPUT_DIR = posixpath.join(vm_util.VM_TMP_DIR, 'outputFaban')
 
-PACKAGES_URL = ('http://lsi-www.epfl.ch/parsa')
+PACKAGES_URL = 'http://lsi-www.epfl.ch/parsa'
 INDEX_URL = posixpath.join(PACKAGES_URL, 'index')
 SOLR_CONFIG_URL = posixpath.join(PACKAGES_URL, 'solrconfig.xml')
 SCHEMA_URL = posixpath.join(PACKAGES_URL, 'schema.xml')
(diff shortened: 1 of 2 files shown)


cmccoy:
@Vukasin92: Would you mind taking a look at your convenience? Thanks!

perfkitbenchmarker/benchmarks/cloudsuite_web_search_benchmark.py
```diff
@@ -59,27 +59,27 @@
              '$FABAN_HOME/lib/fabandriver.jar:$JAVA_HOME/lib/tools.jar:'
              '$FABAN_HOME/search/build/lib/search.jar')
 
-flags.DEFINE_string('client_heap_size', '2g',
+flags.DEFINE_string('websearch_client_heap_size', '2g',
```
voellm: We still have some naming inconsistency here...

websearch vs cloudsuite_web_search

I've ok if we want to shorten this but we should be consistent on websearch or web_search and either rename the flag or benchmark.

My preference is websearch

cmccoy: Ah, thanks for pointing that out. @Vukasin92: opinions?

Vukasin92: I'm gonna check again, but the guys at EPFL once told me that they would want the benchmark to have cloudsuite in name, but I think it is ok for parameters not to have cloudsuite in name.
I see that cloudsuite_web_search_client_heap_size is too long. What about faban_client_heap_size and solr_server_heap_size? Or just faban_heap_size and solr_heap_size? That would distinguish those parameters from other benchmarks (although not the ones which also use solr or faban and have the same option), but still leaves cloudsuite_web_search_num_clients too long. This also requires user to know what components benchmark uses instead of just benchmark name, so maybe it is not a good idea.
What do you think about making an abbreviation for flag names only, like cs_websearch, would that be consistent enough?

cmccoy: Sounds good. I definitely don't plan to change the benchmark name, just want to find a balance between:
* Making it clear that changing the flag will only affect cloudsuite web search.
* Making the flag too long.

`cs_websearch` or `websearch` are both fine with me.  Using `faban_heap_size` and `solr_heap_size` is fine, but we would then also need a prefix for `distr`, `ramp_up`, etc.

Vukasin92: I guess you could rename the benchmark to cloudsuite_websearch, that way, when you abbreviate flags to cs_websearch, it would be more consistent than previously.
If you go with faban/solr flag names, then ramp_up, steady_state and scale are also faban parameters, num_clients is number of machines the faban runs it's agents, but distribution belongs to benchmark configuration.
If cs_websearch for flags work for you, I think it is better option among those two options.

cmccoy: `cs_websearch` sounds great to me. I'll make the change.

Vukasin92: Great, go ahead :)

cmccoy: Done - mind taking another look? Thanks again.

voellm:
LGTM

perfkitbenchmarker/benchmarks/cloudsuite_websearch_benchmark.py
```diff
@@ -150,13 +140,13 @@ def _BuildIndex(solr_nodes, fw):
   def DownloadIndex(vm):
     solr_core_dir = posixpath.join(vm.GetScratchDir(), 'solr_cores')
     vm.RobustRemoteCommand('cd {0} && '
-                           'wget {1} && '
-                           'tar zxvf index -C {2}'.format(
+                           'curl -L {1} | '
```
Vukasin92: Do we need to switch to vm.Install('curl') in Prepare function instead of 'wget' now? Otherwise looks good to me too.

cmccoy: Good catch - thanks. Done.

PR #480, opened 2015-09-18
RunThreaded changes.

skschneider:
Returns function return values.
Schedules calls in the order of the args lists (previously reverse).
Preserves interruptability.

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -190,8 +190,8 @@ def _ExecuteThreadCall(target_arg_tuple, call_id, queue, parent_log_context):
     parent_log_context: ThreadLogContext of the parent thread.
   """
   target, args, kwargs = target_arg_tuple
-  log_context = log_util.ThreadLogContext(parent_log_context)
   try:
+    log_context = log_util.ThreadLogContext(parent_log_context)
     log_util.SetThreadLogContext(log_context)
     queue.put((call_id, target(*args, **kwargs), None))
   except:
@@ -209,6 +209,10 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
         in the list can either be a singleton or a (args, kwargs) tuple/list.
         Usually this is a list of VMs.
     max_concurrent_threads: The maximum number of concurrent threads to allow.
+
+  Returns:
+    List of the same length as thread_params. Contains the return value from
+    each threaded function call in the corresponding order as thread_params.
 
   Raises:
     ValueError: when thread_params is not valid.
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -18,6 +18,7 @@ from concurrent import futures
 import contextlib
 import logging
 import os
+import Queue
 import random
 import re
 import string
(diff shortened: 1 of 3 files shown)


cmccoy:
Looks nice to have. Any chance of using `ThreadPoolExecutor` from `concurrent.futures`? If possible, it would be nice to have a single function to execute tasks in parallel (that takes an `Executor`).

skschneider:
I encountered two problems with ThreadPoolExecutor: doesn't support daemon threads / custom thread types, and isn't interruptable in Python 2.7. If you know a solution for these, I agree it would be preferable to use that class.

cmccoy:
I believe the threads spawned [are daemons](https://hg.python.org/cpython/file/3.5/Lib/concurrent/futures/thread.py#l128), but I don't know of any way to handle custom thread types / interruption in 2.7. Thanks for the explanation.

perfkitbenchmarker/vm_util.py
```diff
@@ -172,32 +173,29 @@ def GetSshOptions(ssh_key_filename):
...
-    """Modified join to raise an exception if needed."""
-    super(ThreadWithExceptions, self).join(*args, **kwargs)
-    if self.exception:
-      raise errors.VmUtil.ThreadException(self.exception)
+  Args:
+    target_arg_tuple: (target, args, kwargs) tuple containing the function to
+        call and the arguments to pass it.
+    call_id: int. Index corresponding to the call in the thread_params argument
+        of RunThreaded.
+    queue: Queue. Receives a (call_id, result, stacktrace) tuple. If the called
```
cmccoy: Maybe introduce a namedtuple for these?

perfkitbenchmarker/vm_util.py
```diff
@@ -238,42 +240,55 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
-      args, kwargs = thread_params.pop()
-      thread = ThreadWithExceptions(target=target, args=args, kwargs=kwargs)
-      threads.append(thread)
+      # Using a timeout makes this wait interruptable.
+      call_id, result, stacktrace = queue.get(block=True, timeout=1000)
+    except Queue.Empty:
+      continue
+    results[call_id] = result
+    if stacktrace:
+      msg = 'Exception occured in call {0}:{1}{2}'.format(
```
cmccoy: Could you log the args?

perfkitbenchmarker/vm_util.py
```diff
@@ -238,42 +240,55 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
     return
 
   if not isinstance(thread_params[0], tuple):
-    thread_params = [((arg,), {}) for arg in thread_params]
+    target_arg_tuples = [(target, (arg,), {}) for arg in thread_params]
```
cmccoy: Consider moving the logic below into a function that accepts `target_arg_tuples`, so multiple functions can be run concurrently as in the multiprocess case.

perfkitbenchmarker/vm_util.py
```diff
@@ -238,42 +240,55 @@ def RunThreaded(target, thread_params, max_concurrent_threads=200):
...
+                         for args, kwargs in thread_params]
+
+  queue = Queue.Queue()
+  log_context = log_util.GetThreadLogContext()
+  max_concurrent_threads = min(max_concurrent_threads, len(target_arg_tuples))
+  results = [None] * len(target_arg_tuples)
+  error_strings = []
+  for call_id in xrange(max_concurrent_threads):
+    target_arg_tuple = target_arg_tuples[call_id]
+    thread = threading.Thread(
```
cmccoy: consider naming the thread.

perfkitbenchmarker/vm_util.py
skschneider: Would have been my next PR. Will add it to this one.

perfkitbenchmarker/vm_util.py
skschneider: I think we've used default naming so far. RunThreaded can be called from within a function invoked by RunThreaded, so a name based on the call_id in this function wouldn't be good. An ideal name might be based on what the thread's role is, but the implementation for that would be complicated. I think it's best to delay on that and add it as a separate feature.

cmccoy: SGTM.

perfkitbenchmarker/vm_util.py
skschneider: Done.

perfkitbenchmarker/vm_util.py
skschneider: Done.

perfkitbenchmarker/vm_util.py
skschneider: Done.

cmccoy:
test this please.

cmccoy:
test this please.

cmccoy:
LGTM.

PR #483, opened 2015-09-23
Add a timeout to netperf tests

cmccoy:


voellm:
LGTM

voellm:
This timeout is being added because of spurious netperf hangs.  iperf has already been updated.

PR #485, opened 2015-09-24
Updates PKB to work with the latest Azure CLI

ehankland:
I ran iperf and copy_throughput to exercise everything that might have changed - they both run fine now.

### DIFF
--- a/perfkitbenchmarker/azure/__init__.py
+++ b/perfkitbenchmarker/azure/__init__.py
@@ -11,3 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from perfkitbenchmarker import events
+from perfkitbenchmarker.azure import util
+
+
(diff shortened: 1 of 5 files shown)


perfkitbenchmarker/azure/__init__.py
```diff
@@ -11,3 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from perfkitbenchmarker.azure import util
+
+util.CheckAzureVersion()
```
cmccoy: I don't think we should warn when the cloud isn't Azure. Perhaps you could subscribe to [`initialization_complete`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/events.py#L29) and warn conditionally?

ehankland: Sure I can do that - checking FLAGS.cloud won't be sufficient in the future, but it should work for now.

cmccoy:
:+1:

PR #487, opened 2015-09-24
Update Maven version and download URL.

cmccoy:
The current mirror is down; hopefully us.apache.org is more reliable.


ehankland:
LGTM

PR #488, opened 2015-09-24
Remove the temporary status written by execute_command.py

cmccoy:
execute_command.py has been writing `-1` to the status file until the
command completes. If `execute_command.py` itself exits / is killed,
this has been the return code of wait_for_command. Since bash exit codes
are in the range `[0,255]`, -1 overflows to 255, is interpreted as a
temporary SSH failure, and the command retried. On the second try, the
process stdout / stderr files have been removed.

This also adds some logging for the wrapper script itself, which is
logged when RobustRemoteCommand has an exception.


voellm:
@ehankland - YI

voellm:
LGTM

PR #489, opened 2015-09-24
Archive run directory

cmccoy:
Add a flag to archive the run directory (with logs, etc) in GCS/S3.

### DIFF
--- a/perfkitbenchmarker/archive.py
+++ b/perfkitbenchmarker/archive.py
@@ -1,0 +1,68 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


voellm:
@ehankland - FYI

perfkitbenchmarker/archive.py
```diff
@@ -0,0 +1,68 @@
...
+
+  try:
+    cmd = (prefixes[target_bucket[:prefix_len]] +
+           ['-', posixpath.join(target_bucket, tar_file_name)])
+  except KeyError:
+    raise ValueError('Unsupported bucket name: {0}'.format(target_bucket))
+
+  logging.info('Streaming %s to %s\n%s', run_temp_directory, tar_file_name,
+               ' '.join(cmd))
+  p = subprocess.Popen(cmd, stdin=subprocess.PIPE)
```
voellm: Why not call RemoteProcess for this on 127.0.0.1 or have a local RunProcess command?  Seems like we should also wrap local calls for better Linux/Windows isolation.

cmccoy: We have a local command runner: [`vm_util.IssueCommand`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L463). I didn't use it here because it doesn't support piped input.

This should work on Windows as well.

ehankland:
LGTM

PR #495, opened 2015-09-28
Let fio benchmark accept list of IO depths

noahl:
This lets the user use any set of IO depths they want, which will make
benchmarking high IOPS devices easier.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -129,9 +129,31 @@ def GetIODepths(io_depths):
   return result
 
 
+def WarnIODepths(depths_list):
+  """Given a list of IO depths, log warnings if it seems "weird".
+
+  Args:
+    depths_list: a list of IO depths.
+  """
+
+  for i in range(len(depths_list) - 1):
+    if depths_list[i] >= depths_list[i + 1]:
+      logging.warning('IO depths list is not monotonically increasing: '
+                      '%s comes before %s.', depths_list[i], depths_list[i + 1])
+
+  values = set()
+  warned_on = set()
+  for val in depths_list:
+    if val in values and val not in warned_on:
+      logging.warning('IO depths list contains duplicate entry %s.', val)
+      warned_on.add(val)
+    else:
+      values.add(val)
+
+
 def IODepthsValidator(string):
   try:
-    GetIODepths(string)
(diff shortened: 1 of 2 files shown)


noahl:
@yuyantingzero @cmccoy Will you take a look?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -100,15 +101,46 @@
...
+  result = []
+
+  for group in groups:
+    match = IODEPTHS_REGEXP.match(group)
+    if match is None:
+      raise ValueError('Invalid io_depths expression %s', io_depths)
+    elif match.group(2) is None:
+      result.append(int(match.group(1)))
+    else:
+      result.extend(range(int(match.group(1)), int(match.group(3)) + 1))
```
cmccoy: Worth checking if the range specified is monotonically increasing?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -100,15 +101,46 @@
...
+    return True
+  except ValueError:
+    return False
 
 
 flags.RegisterValidator('io_depths',
                         IODepthsValidator,
                         message='--io_depths must be an integer '
-                                'or a range of integers, all > 0')
+                                'or a list of integers, all > 0')
```
cmccoy: It seems like it can be an integer, range, or mix thereof.

cmccoy:
A few minor things, otherwise LGTM.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: How do we handle duplicates?
eg. 1-4, 2-5. Do we need to run 2,3,4 twice or sort the final list and only grep uniq values?

noahl: I added a function to print warnings if the IO depths list looks wrong. I could make it fail if it prints any warnings. I don't feel strongly about doing that or not.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Changed, thanks.

noahl:
@cmccoy @yuyantingzero This branch is ready for review again.

yuyantingzero:
LGTM

Issue #497, opened 2015-09-29
Fio: pre-filling the attached disk does not work


mateusz-blaszkowski:
Look at the fio_benchmark.Prepare() method:

`fill_size = str(DISK_USABLE_SPACE_FRACTION *
                        1000 * disk.disk_size) + 'M'`

Let's say that `disk.disk_size` is 3 and `DISK_USABLE_SPACE_FRACTION` is 0.9.
`fill_size` will be exactly `2700.0M`. This format is not accepted by fio and as a result, pre-filling the scratch volume will not work. What is worse, fio command will fail with exit_code=0 and we will not even see that something went wrong. 

My proposition would be:
1. convert `fill_size` to int before converting it to str (format like `2700M` is accepted by fio)
2. add more logging around fio command as it sometimes exits without error code.

yuyantingzero:
@stfeng

stfeng:
I think we should get rid of the DISK_USABLE_SPACE_FRACTION parameter, it's pretty flaky (and setting it to 90% is also just a rough estimate that don't really work for all file systems). 

If someone really wants to test against the file system (which isn't that encouraged for fio command anyway), and if they want to pre-warm the device before the testing, then the implementation here in FIO should reuse the device filling logic in the "against device case", fill the device first using /dev/*** first, then format the device again using a file system, and then mount it for subsequent testing.

noahl:
Hi Mateusz,

This issue is already fixed in the latest fio_benchmark.py in the master
branch. It looks like the fix was committed 13 days ago, so it may not have
made it into the last release, but it will be in the next one.

Thanks,
Noah

noahl:
If you want a fix right away, what you suggested with `int( ... )` is exactly what I implemented. You can patch your copy of fio_benchmark.py with that change, or cherry-pick commit 84dfd924fb6a4cd51485e8ae615527b9beb06a39 . I'm closing the issue, but please re-open if there's anything else I can do to help.

stfeng:
I think to properly fix this issue, we should get rid off the DISK_USABLE_SPACE_FRACTION and merge the fill the file code with fill the device code.

mateusz-blaszkowski:
@noahl - I have tested with a pretty fresh master version. PR with the fix was merged 2 days ago, so I was unlucky ;) anyway, thanks!

noahl:
Fixed in #502.

Issue #496, opened 2015-09-29
Fio: when no jobfile is specified, fio is not running against scratch disk


mateusz-blaszkowski:
Let's say that scratch disk is by default mounted in /scratch0 directory.

Current simplified fio benchmark flow:
1. Fill scratch disk with 'fio-temp-file' - look at the `GenerateFillCommand()` method. It uses `fill_path` which is retrieved directly from disk object (either `mount_point` or by `GetDevicePath()` method). As a result, the following file is created: /**scratch0**/fio-temp-file . This part is OK - scratch disk is pre-filled with some data.
2. Run fio with the default fio.job file from perfkitbenchmarker/data directory. 
  - command which is being executed: `ssh root@... sudo /tmp/pkb/fio/fio --output-format=json /tmp/pkb/fio.job` (look at the Run() method)
  - fio will not be run against scratch because it produces 'fio_test_file' in /root directory (which is caused by logging as root. When logging as different user, another directory from /home will be used)
  
Expected behaviour:
fio should create a file in /scratch0 directory. In the past either --filename or --directory flag was passed to fio command, but since https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/421 it has been changed.

yuyantingzero:
@stfeng

noahl:
I'm testing a fix right now. PR should be coming soon.

noahl:
Fixed in #502.

PR #502, opened 2015-09-29
Add templating in fio jobfiles

noahl:
When fio_benchmark.py gets a user-supplied job file, replace the
string {{filename}} with the path to the file it pre-filled. Use this
ability to fix a bug introduced recently where the built-in fio.job
file no longer used the pre-filled disk or file.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -67,9 +67,7 @@ FLAGS = flags.FLAGS
 
 flags.DEFINE_string('fio_jobfile', None,
                     'Job file that fio will use. If not given, use a job file '
-                    'bundled with PKB. {{filename}} in the job file will be '
-                    'replaced by the path of the disk or file we pre-filled, '
-                    'if requested. Cannot use with --generate_scenarios.')
+                    'bundled with PKB. Cannot use with --generate_scenarios.')
 flags.DEFINE_list('generate_scenarios', None,
                   'Generate a job file with the given scenarios. Special '
                   'scenario \'all\' generates all scenarios. Available '
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -490,7 +490,7 @@ def Run(benchmark_spec):
   #      This is a pretty lousy experience.
   logging.info('FIO Results:')
 
-  if not FLAGS.run_for_minutes:
+  if not FLAGS['run_for_minutes'].present:
     RunIt()
   else:
     RunForMinutes(RunIt, FLAGS.run_for_minutes, MINUTES_PER_JOB)
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -27,6 +27,7 @@ import re
 import jinja2
 
 from perfkitbenchmarker import data
+from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import fio
@@ -353,7 +354,7 @@ def WarnOnBadFlags():
       not FLAGS.working_set_size and
       not FLAGS.against_device):
     logging.error(NEED_SIZE_MESSAGE)
-    raise AssertionError(NEED_SIZE_MESSAGE)
+    raise errors.Benchmarks.PrepareException(NEED_SIZE_MESSAGE)
 
 
 def RunForMinutes(proc, mins_to_run, mins_per_call):
@@ -458,10 +459,17 @@ def Run(benchmark_spec):
     fio_command = 'sudo %s --output-format=json --directory=%s %s' % (
         fio.FIO_PATH, mount_point, REMOTE_JOB_FILE_PATH)
 
-
   samples = []
 
   def RunIt(repeat_number=None, minutes_since_start=None, total_repeats=None):
+    """Run the actual fio command on the VM and save the results.
+
+    Args:
+      repeat_number: if given, our number in a sequence of repetitions.
+      minutes_since_start: if given, minutes since the start of repetition.
+      total_repeats: if given, the total number of repetitions to do.
+    """
+
     if repeat_number:
       logging.info('**** Repetition number %s of %s ****',
                    repeat_number, total_repeats)
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -226,23 +226,34 @@ size={{size}}
 SECONDS_PER_MINUTE = 60
 
 
-def GenerateJobFileString(filename, scenarios, io_depths, working_set_size):
+def GenerateJobFileString(filename, scenario_strings,
+                          io_depths_string, working_set_size):
   """Make a string with our fio job file.
 
   Args:
     filename: the file or disk we pre-filled, if any.
-    scenarios: iterable of dicts, taken from SCENARIOS.
-    io_depths: iterable. The IO queue depths to test.
+    scenario_strings: list of strings with names in SCENARIOS.
+    io_depths_string: string. The IO queue depths to test.
     working_set_size: int or None. If int, the size of the working set in GB.
 
   Returns:
     The contents of a fio job file, as a string.
   """
 
+  if 'all' in scenario_strings:
+    scenarios = SCENARIOS.itervalues()
+  else:
+    for name in scenario_strings:
+      if name not in SCENARIOS:
+        logging.error('Unknown scenario name %s', name)
+    scenarios = (SCENARIOS[name] for name in scenario_strings)
+
   size_string = str(working_set_size) + 'G' if working_set_size else '100%'
 
   job_file_template = jinja2.Template(JOB_FILE_TEMPLATE,
                                       undefined=jinja2.StrictUndefined)
+
+  io_depths = GetIODepths(io_depths_string)
 
   return str(job_file_template.render(
       minutes_per_job=MINUTES_PER_JOB,
@@ -255,14 +266,16 @@ def GenerateJobFileString(filename, scenarios, io_depths, working_set_size):
 FILENAME_PARAM_REGEXP = re.compile('filename\s*=.*$', re.MULTILINE)
 
 
-def ProcessUserJobFile(fio_jobfile, remove_filename):
-  """Get the contents of a user job file, slightly edited.
+def ProcessedJobFileString(fio_jobfile, remove_filename):
+  """Get the contents of a job file as a string, slightly edited.
 
   Args:
     fio_jobfile: the path to a fio job file.
+    remove_filename: bool. If true, remove the filename parameter from
+      the job file.
 
   Returns:
-    The user's job file as a string, possibly without filename parameters.
+    The job file as a string, possibly without filename parameters.
   """
 
   with open(fio_jobfile, 'r') as jobfile:
@@ -272,8 +285,8 @@ def ProcessUserJobFile(fio_jobfile, remove_filename):
       return jobfile.read()
 
 
-def GetOrGenerateJobFileString(fio_jobfile, remove_filename,
-                               filename, scenario_strings,
+def GetOrGenerateJobFileString(job_file_path, scenario_strings,
+                               against_device, disk,
                                io_depths, working_set_size):
   """Get the contents of the fio job file we're working with.
 
@@ -281,35 +294,36 @@ def GetOrGenerateJobFileString(fio_jobfile, remove_filename,
   new one.
 
   Args:
-    fio_jobfile: string or None. The path to the user's jobfile, if provided.
-    remove_filename: bool. If true, remove the filename from a user job file.
-    filename: the file or disk we pre-filled, if any.
-    scenario_strings: list of strings or None. The workload scenarios to
-      generate.
-    io_depths: iterable. The IO queue depths to test.
-    working_set_size: int or None. If int, the size of the working set in GB.
+    job_file_path: string or None. The path to the user's job file, if
+      provided.
+    scenario_strings: list of strings or None. The workload scenarios
+      to generate.
+    against_device: bool. True if testing against a raw device, False
+      if testing against a filesystem.
+    disk: the disk.BaseDisk object to test against.
+    io_depths: string. The IO queue depths to test.
+    working_set_size: int or None. If int, the size of the working set
+      in GB.
 
   Returns:
     A string containing a fio job file.
   """
 
-  # The default behavior is to run with the standard fio job file.
-  if scenario_strings is None and not fio_jobfile:
-    fio_jobfile = data.ResourcePath('fio.job')
+  use_user_jobfile = job_file_path or not scenario_strings
 
-  if fio_jobfile:
-    return ProcessUserJobFile(fio_jobfile,
-                              remove_filename)
+  if use_user_jobfile:
+    remove_filename = against_device
+    return ProcessedJobFileString(job_file_path or data.ResourcePath('fio.job'),
+                                  remove_filename)
   else:
-    if 'all' in scenario_strings:
-      scenarios = SCENARIOS.itervalues()
+    if against_device:
+      filename = disk.GetDevicePath()
     else:
-      for name in scenario_strings:
-        if name not in SCENARIOS:
-          logging.error('Unknown scenario name %s', name)
-      scenarios = (SCENARIOS[name] for name in scenario_strings)
+      # Since we pass --directory to fio, we must use relative file
+      # paths or get an error.
+      filename = DEFAULT_TEMP_FILE_NAME
 
-    return GenerateJobFileString(filename, scenarios,
+    return GenerateJobFileString(filename, scenario_strings,
                                  io_depths, working_set_size)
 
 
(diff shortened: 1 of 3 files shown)


noahl:
@yuyantingzero @stfeng @cmccoy Will you take a look? This fixes #496 by adding a templating feature that I think will be generally useful.

stfeng:
If user has specified a job file with a specific filename, then we should just use that, we shouldn't try to be smart here and replace that file name with something else.

The issue #496 is about when the user does *not* specify an fio job file, and for that case, fio benchmark uses the default job file and we just need to make sure we use scratch drive for the default job file.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -356,17 +356,16 @@ def Prepare(benchmark_spec):
         fill_size = str(int(DISK_USABLE_SPACE_FRACTION *
                             1000 * disk.disk_size)) + 'M'
 
-    logging.info('Fill file %s on %s', fill_path, vm)
+    logging.info('Fill file %s on %s', filename, vm)
```
yuyantingzero: Remove the comments on line 354-355.

noahl:
If the user specifies a job file, we will not replace it with our own. (I think we used to do that in an old version of the benchmark, but I changed that because I agree that we shouldn't try to be smart.) The templating feature will only be activated if the user puts "{{filename}}" as their file name, which I think is extremely unlikely as the name of an actual file.

I'm also happy about the templating feature because it will make it easy to write a custom job file that uses a pre-filled disk or file even across providers, where disks may have different device names.

stfeng:
The token "{{filename}}" is not a standard or correct token for fio job file. No one would use it. If we want to have the capability to help users override the "filename" field of the job file on the fly, then we should introduce a parameter to do so. )At this point I don't see a need)

The issue at hand is to make our default fio job file work again. It used to work on the scratch disk, but after recent changes, it started to work on home directories instead. I'm not sure why this new templating feature is required to fix this particular issue. maybe I'm missing something here?

noahl:
You're right that {{filename}} is not a standard token. I am happy to
change that. But it seems like a useful feature to be able to use a
provided filename. I know that I have wanted this feature in the past, and
in fact I had a private branch that did templating that I never published.

The reason the default fio job file does not work is that it uses a
relative path for its filename. I think we used to cd to the right
directory before running, but now we don't. I see three ways to fix this
bug:

1. Keep the templating
2. Change the built-in job file to use /scratch0/fio-temp-file
3. Make the remote host cd before running fio like before

I prefer 1, but since you don't, do you prefer 2 or 3?

Noah

yuyantingzero:
Just FYI,
to test against file system, we used to add --directory=/scratch0 in the fio command.
to test against raw device, we used to delete filename/directory in job and use --filename=device_path in fio command.

noahl:
We talked about this and decided to fix it by removing templating and using the --filename and --directory options to fio. I'm in the middle of implementing that right now.

noahl:
I just pushed revisions that implement the agreed-upon fix for this bug. They pass all of my tests, but I hope to do some code cleanups tomorrow.

stfeng:
@noahl - thanks for the fix! Please let us know when the code is ready for review.

noahl:
@stfeng @yuyantingzero As part of cleaning up the code, I went ahead and started to fix #497 by removing the DISK_USABLE_SPACE_FRACTION as discussed on that thread. However, I have run into an issue. Now when running fio against a filesystem, PKB unmounts the drive, prefills the disk, recreates a filesystem, and runs fio. Since there is no pre-created file in the new filesystem to work with, fio needs to know what size file it should create. The question is, what should the default value be if the user doesn't pass anything?

I'm debating between using fio's --fill_device mode, where it makes a file bigger and bigger until it gets ENOSPC and then uses that as its working file, or just requiring the user to specify a file size. The reason I'm unsure is that I don't know if it really makes sense to fill a filesystem until it's full. That doesn't seem like a real use-case, and also it might cause weird kernel behavior (like trouble swapping) that would defeat the purpose of using a filesystem in the first place. What do you think?

yuyantingzero:
A valid fio job file testing on filesystem should contain size parameter, unless the data file already exists. I don't see any problem ask user to specify a file size here.

stfeng:
There is a DEFAULT_TEMP_FILE_NAME for the case where the user does not want to test against device, the question is: how did the old code generate this DEFAULT_TEMP_FILE_NAME and how does it determine the size?

noahl:
In the old code, we multiplied the disk size by DISK_USABLE_SPACE_FRACTION,
and that became the size of DEFAULT_TEMP_FILE_NAME. That's why we have to
find a new way of determining the size.

stfeng:
By "old code", I meant code before your recent refactoring. 

Looking closer, I realized DEFAULT_TEMP_FILE_NAME was added as part of the support to generate job files dynamically based on scenarios, right? DEFAULT_TEMP_FILE_NAME would NOT be needed if the user specifies a job file, or if the default/PKB bundled job file is used, right?

If above is correct, then this discussion/issue of determining size only comes into play when the user wants to dynamically generate a job file by specifying a scenario, right? If so, I'm fine with requiring the user to specify working_set_size (just like they would need to do if they invoke FIO directly)

noahl:
I just pushed several commits that make `fio_benchmark.py` a lot more readable. This branch is now ready for review.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -260,51 +266,124 @@ def GenerateJobFileString(disk, against_device,
...
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
+    logging.warning('Runtime %s will be rounded up to the next multiple of %s '
+                    'minutes.', FLAGS.run_for_minutes, MINUTES_PER_JOB)
+
+  if (FLAGS.fio_jobfile is None and
+      FLAGS.generate_scenarios and
+      not FLAGS.working_set_size):
+    logging.error('You must specify the working set size when using generated '
+                  'scenarios against a filesystem.')
+    sys.exit(1)
```
yuyantingzero: We should not exit within the benchmark. Instead we should raise an exception and let the upper level to decide.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -317,59 +396,38 @@ def Prepare(benchmark_spec):
...
+    # and changes its permissions, which we have already done.
+    vm.RemoteCommand('sudo mount %s %s' % (disk.GetDevicePath(), mount_point))
+
+  job_file_string = GetOrGenerateJobFileString(FLAGS.fio_jobfile,
+                                               FLAGS.generate_scenarios,
+                                               FLAGS.against_device,
+                                               disk,
+                                               FLAGS.io_depths,
+                                               FLAGS.working_set_size)
   job_file_path = vm_util.PrependTempDir(LOCAL_JOB_FILE_NAME)
```
yuyantingzero: I suggest move job file generation into run stage. In this way, we are able to change job file every run and you dont need to generate job_file_string multiple times.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -175,22 +177,22 @@ def IODepthsValidator(string):
                                 'integers and ranges, all > 0')
 
 
-def GenerateFillCommand(fio_path, fill_path, fill_size):
-  """Generate a command to sequentially write a device or file.
+def FillDevice(fio_path, vm, disk, fill_size):
+  """Fill the given disk on the given vm up to fill_size.
 
   Args:
     fio_path: path to the fio executable.
```
yuyantingzero: this is accessible from fio package, no need to use an parameter here.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: If you exit here, the cleanup won't be invoked and we are leaking resources.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -260,51 +266,124 @@ def GenerateJobFileString(disk, against_device,
...
-    against_device: bool. True if we're using a raw disk.
-    scenario_strings: list of strings or None. The workload scenarios to
-      generate.
-    io_depths: iterable. The IO queue depths to test.
-    working_set_size: int or None. If int, the size of the working set in GB.
+    job_file_path: string or None. The path to the user's job file, if
+      provided.
+    scenario_strings: list of strings or None. The workload scenarios
+      to generate.
+    against_device: bool. True if testing against a raw device, False
```
yuyantingzero: If this is already a flag, any reason it needs to use a parameter here?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -317,59 +396,38 @@ def Prepare(benchmark_spec):
...
+    FillDevice(fio.FIO_PATH, vm, disk, FLAGS.device_fill_size)
+
+  if not FLAGS.against_device:
+    vm.FormatDisk(disk.GetDevicePath())
+    # Don't use vm.MountDisk() because that also creates the mount point
+    # and changes its permissions, which we have already done.
+    vm.RemoteCommand('sudo mount %s %s' % (disk.GetDevicePath(), mount_point))
+
+  job_file_string = GetOrGenerateJobFileString(FLAGS.fio_jobfile,
+                                               FLAGS.generate_scenarios,
```
yuyantingzero: Any reason to pass flags as parameters?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Changed, thanks!

noahl: Interestingly, we actually do clean up when I use `sys.exit`, but I still agree that using an exception is much better.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Changed, thanks.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I passed them as parameters because I think it makes the functions easier to understand and because I think it is much easier to mock parameters than flags, which makes these functions easier to test. Do you think it's worth it to use flags rather than parameters?

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Same answer as above.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I can eliminate the extra job file generation by loading the job file from the LOCAL_JOB_FILE_PATH in the Run() method. I thought about doing that, but I thought it might actually be faster to generate again since we wouldn't touch the disk. However, probably simplicity is more important than speed here. What do you think of this solution?

As for changing the job file every run, I see that it might be useful, but I don't know how we would use it right now. Maybe a future feature?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -388,60 +446,55 @@ def Run(benchmark_spec):
...
+
+  # TODO(user): This only gives results at the end of a job run
+  #      so the program pauses here with no feedback to the user.
+  #      This is a pretty lousy experience.
+  logging.info('FIO Results:')
+
+  if not FLAGS.run_for_minutes:
+    RunIt()
+  else:
+    RunForMinutes(RunIt, FLAGS.run_for_minutes, MINUTES_PER_JOB)
```
yuyantingzero: I do not think there is a need to add FLAGS.run_for_minutes and MINUTES_PER_JOB as parameters. Unless you want to make RunForMinutes more generic and work for customized job file.

yuyantingzero: What happens if I use a customized job file and also specify a run_for_minutes flag

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -388,60 +446,55 @@ def Run(benchmark_spec):
...
+    samples.extend(fio.ParseResults(job_file_string,
+                                    json.loads(stdout),
+                                    base_metadata=base_metadata))
+
+  # TODO(user): This only gives results at the end of a job run
+  #      so the program pauses here with no feedback to the user.
+  #      This is a pretty lousy experience.
+  logging.info('FIO Results:')
+
+  if not FLAGS.run_for_minutes:
```
yuyantingzero: Default value for run_for_minutes is 10.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: The reason to make FLAGS.run_for_minutes and MINUTES_PER_JOB be parameters was partly to make them more generic as you said (I don't want MINUTES_PER_JOB to always be 10), and partly to make testing easy. In fio_benchmark_test.py you'll find a bunch of tests for RunForMinutes that would have been much more annoying and difficult to make if I had to mock FLAGS for each of them.

noahl: If you use a custom job flie and pass run_for_minutes, then it will print a warning and ignore your run_for_minutes flag. This is because run_for_minutes is in FLAGS_IGNORED_FOR_CUSTOM_JOBFILE at the top of the file.

perfkitbenchmarker/benchmarks/fio_benchmark.py
yuyantingzero: What I mean is we can completely eliminate the fob generation in prepare stage, including writing to tmp directory and upload to vm.

I think it is quite useful, actually I am about to send a PR for this feature. But it would be great if your PR includes this. And it also make code more concise.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -388,60 +447,55 @@ def Run(benchmark_spec):
...
+  job_file_string = GetOrGenerateJobFileString(FLAGS.fio_jobfile,
+                                               FLAGS.generate_scenarios,
+                                               FLAGS.against_device,
+                                               disk,
+                                               FLAGS.io_depths,
+                                               FLAGS.working_set_size)
+
+  samples = []
+
+  def RunIt(repeat_number=None, minutes_since_start=None, total_repeats=None):
```
yuyantingzero: Documentations.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Wow, that's bad. I fixed it. Thanks a lot!

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -388,60 +435,62 @@ def Run(benchmark_spec):
...
-      logging.info('**** Repetition number %s of %s ****', rep_num, run_reps)
-      run_start = datetime.datetime.now()
-      stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
-
-      seconds_since_start = int(round((run_start - start_time).total_seconds()))
-      minutes_since_start = int(round(float(seconds_since_start)
-                                      / float(SECONDS_PER_MINUTE)))
+    fio_command = 'sudo %s --output-format=json --directory=%s %s' % (
+        fio.FIO_PATH, mount_point, REMOTE_JOB_FILE_PATH)
+
```
yuyantingzero: Remove extra blank line.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -388,60 +435,62 @@ def Run(benchmark_spec):
...
+
+  samples = []
+
+  def RunIt(repeat_number=None, minutes_since_start=None, total_repeats=None):
```
yuyantingzero: Document this method.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -260,51 +264,128 @@ def GenerateJobFileString(disk, against_device,
...
+  if FLAGS.run_for_minutes % MINUTES_PER_JOB != 0:
+    logging.warning('Runtime %s will be rounded up to the next multiple of %s '
+                    'minutes.', FLAGS.run_for_minutes, MINUTES_PER_JOB)
+
+  if (FLAGS.fio_jobfile is None and
+      FLAGS.generate_scenarios and
+      not FLAGS.working_set_size and
+      not FLAGS.against_device):
+    logging.error(NEED_SIZE_MESSAGE)
+    raise AssertionError(NEED_SIZE_MESSAGE)
```
yuyantingzero: Use Errors from https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/errors.py

yuyantingzero:
A few small comments. LGTM otherwise.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. Thanks for showing me that!

noahl:
I just pushed the commits that fix Yanting's comments. I've also tested the built-in job file, user-supplied job files, and generated job files against filesystems and block devices. I think this branch is ready to merge.

PR #515, opened 2015-10-07
Default to Not Pre-filling Filesystems in Fio

noahl:
Don't automatically pre-fill filesystems. This restores the old
default behavior.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -77,14 +77,16 @@ flags.DEFINE_list('generate_scenarios', None,
 flags.DEFINE_boolean('against_device', False,
                      'Unmount the device\'s filesystem so we can test against '
                      'the raw block device. If --generate-scenarios is given, '
-                     'will generate a job file that uses the block device.')
+                     'will generate a job file that uses the block device. '
+                     'When using --against_device, default to pre-filling the '
+                     'device to 100% unless modified with --device_fill_size.')
 flags.DEFINE_string('device_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'The valid value can either be an integer, which '
                     'represents the number of bytes to fill or a '
                     'percentage, which represents the percentage '
                     'of the device. A filesystem will be unmounted before '
-                    'filling and remounted afterwards. Default is 100%')
+                    'filling and remounted afterwards. Default is not to fill.')
 flags.DEFINE_string('io_depths', '1',
                     'IO queue depths to run on. Can specify a single number, '
                     'like --io_depths=1, a range, like --io_depths=1-4, or a '
@@ -411,7 +413,7 @@ def Prepare(benchmark_spec):
   logging.info('Umount scratch disk on %s at %s', vm, mount_point)
   vm.RemoteCommand('sudo umount %s' % mount_point)
 
-  if FLAGS.device_fill_size is not '0':
+  if FLAGS['device_fill_size'].present or FLAGS.against_device:
     logging.info('Fill device %s on %s', disk.GetDevicePath(), vm)
     FillDevice(vm, disk, FLAGS.device_fill_size)
 
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -80,13 +80,18 @@ flags.DEFINE_boolean('against_device', False,
                      'will generate a job file that uses the block device. '
                      'When using --against_device, default to pre-filling the '
                      'device to 100% unless modified with --device_fill_size.')
+flags.DEFINE_boolean('prefill_device', False,
+                     'If true, write on the device before starting the '
+                     'benchmark. The amount to write is given by '
+                     '--device_fill_size.')
 flags.DEFINE_string('device_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'The valid value can either be an integer, which '
                     'represents the number of bytes to fill or a '
                     'percentage, which represents the percentage '
                     'of the device. A filesystem will be unmounted before '
-                    'filling and remounted afterwards. Default is not to fill.')
+                    'filling and remounted afterwards. Only valid when '
+                    'used with --prefill_device.')
 flags.DEFINE_string('io_depths', '1',
                     'IO queue depths to run on. Can specify a single number, '
                     'like --io_depths=1, a range, like --io_depths=1-4, or a '
@@ -413,7 +418,7 @@ def Prepare(benchmark_spec):
   logging.info('Umount scratch disk on %s at %s', vm, mount_point)
   vm.RemoteCommand('sudo umount %s' % mount_point)
 
-  if FLAGS['device_fill_size'].present or FLAGS.against_device:
+  if FLAGS.prefill_device:
     logging.info('Fill device %s on %s', disk.GetDevicePath(), vm)
     FillDevice(vm, disk, FLAGS.device_fill_size)
 
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -77,9 +77,7 @@ flags.DEFINE_list('generate_scenarios', None,
 flags.DEFINE_boolean('against_device', False,
                      'Unmount the device\'s filesystem so we can test against '
                      'the raw block device. If --generate-scenarios is given, '
-                     'will generate a job file that uses the block device. '
-                     'When using --against_device, default to pre-filling the '
-                     'device to 100% unless modified with --device_fill_size.')
+                     'will generate a job file that uses the block device.')
 flags.DEFINE_boolean('prefill_device', False,
                      'If true, write on the device before starting the '
                      'benchmark. The amount to write is given by '


noahl:
@stfeng @yuyantingzero Will you take a look?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -411,7 +413,7 @@ def Prepare(benchmark_spec):
   logging.info('Umount scratch disk on %s at %s', vm, mount_point)
   vm.RemoteCommand('sudo umount %s' % mount_point)
 
-  if FLAGS.device_fill_size is not '0':
```
yuyantingzero: If against_device is true, and device_fill_size is 0, we still want to skip the device fill part.
The logic should be:
if device_fill_size is not 0:
  if device_fill_size.present or against_device:
    fill_device

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -77,14 +77,16 @@
 flags.DEFINE_boolean('against_device', False,
                      'Unmount the device\'s filesystem so we can test against '
                      'the raw block device. If --generate-scenarios is given, '
-                     'will generate a job file that uses the block device.')
+                     'will generate a job file that uses the block device. '
+                     'When using --against_device, default to pre-filling the '
+                     'device to 100% unless modified with --device_fill_size.')
 flags.DEFINE_string('device_fill_size', '100%',
```
voellm: It would be better to make this 0 to be clear there is no fill.

stfeng: Tony, actually it's tricker than it appears... What we want is:

1) if the user is testing against the device directly, the default value of --device_fill_size should be 100% because in this mode, the user should fill the device entirely before the fio run.

2) If the user is testing using a file system, then the default behavior should be no-fill (0%) because that is unnecessary as fio will create the test file first and use that file for testing. However, on some providers like AWS, they require pre-warming of the device no matter what, so we still need to provide the user a choice to specify --device_fill_size to be 100% when they run fio on those providers in file system mode.

To sum it up: the desired default value of this flag changes as the scenario changes. 

After some discussion, we figured out a smart way to handle this: we will simply check for the presence of the flag in the test-with-file-system mode case, if the flag is present, then we will follow user's order, otherwise, we won't fill the device. In the test-against-device mode, we will just use the flag as is with the default value being 100%. (so if the flag is not present in the test-against-device mode, the default value of 100% will apply).

voellm: >> After some discussion, we figured out a smart way to handle this: we
will simply check for the presence of the flag in the test-with-file-system
mode case, if the flag is present, then we will follow user's order,
otherwise, we won't fill the device. In the test-against-device mode, we
will just use the flag as is with the default value being 100%. (so if the
flag is not present in the test-against-device mode, the default value of
100% will apply).

This is how I read the code.  However its not clear.  Perhaps there needs
to be a third flag.

against_device

device_fill_size

prefill_device_when_using_filesystem

By adding the third flag you make it clear the behavior without having to
read the code and deduce special settings.

Please add the additional flag so this is clear in the code.


On Wed, Oct 7, 2015 at 1:00 PM, stfeng <notifications@github.com> wrote:

> In perfkitbenchmarker/benchmarks/fio_benchmark.py
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/515#discussion_r41438756>
> :
>
> > @@ -77,14 +77,16 @@
> >  flags.DEFINE_boolean('against_device', False,
> >                       'Unmount the device\'s filesystem so we can test against '
> >                       'the raw block device. If --generate-scenarios is given, '
> > -                     'will generate a job file that uses the block device.')
> > +                     'will generate a job file that uses the block device. '
> > +                     'When using --against_device, default to pre-filling the '
> > +                     'device to 100% unless modified with --device_fill_size.')
> >  flags.DEFINE_string('device_fill_size', '100%',
>
> Tony, actually it's tricker than it appears... What we want is:
>
> 1) if the user is testing against the device directly, the default value
> of --device_fill_size should be 100% because in this mode, the user should
> fill the device entirely before the fio run.
>
> 2) If the user is testing using a file system, then the default behavior
> should be no-fill (0%) because that is unnecessary as fio will create the
> test file first and use that file for testing. However, on some providers
> like AWS, they require pre-warming of the device no matter what, so we
> still need to provide the user a choice to specify --device_fill_size to be
> 100% when they run fio on those providers in file system mode.
>
> To sum it up: the desired default value of this flag changes as the
> scenario changes.
>
> After some discussion, we figured out a smart way to handle this: we will
> simply check for the presence of the flag in the test-with-file-system mode
> case, if the flag is present, then we will follow user's order, otherwise,
> we won't fill the device. In the test-against-device mode, we will just use
> the flag as is with the default value being 100%. (so if the flag is not
> present in the test-against-device mode, the default value of 100% will
> apply).
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/515/files#r41438756>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I took out the if device_fill_size is 0 check for two reasons. First, it wasn't very well done: it only caught a literal "0", not "0%", or "0G", or anything like that. Second, it didn't gain us very much. In my testing, if you run fio and tell it to fill with size 0, it returns very quickly. We didn't gain very much, and it made the code more complex, so I think it is not worth it.

I would be open to adding the 0 check back, but I think it should be a separate change. If we do that change, I'd also like to address the issue that if we run against a filesystem with no fill (which is now the default), we have an unnecessary unmount-reformat-remount step. That will  be a bigger change, but should also produce a more significant speedup.

perfkitbenchmarker/benchmarks/fio_benchmark.py
stfeng: A new flag Sounds good to me.

perfkitbenchmarker/benchmarks/fio_benchmark.py
voellm: You should also simplify this and just accept a floating point number.  Not
reason to include or accept the percent.

So move from DEFINE_string to DEFINE_double


On Wed, Oct 7, 2015 at 1:35 PM, noahl <notifications@github.com> wrote:

> In perfkitbenchmarker/benchmarks/fio_benchmark.py
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/515#discussion_r41443204>
> :
>
> > @@ -411,7 +413,7 @@ def Prepare(benchmark_spec):
> >    logging.info('Umount scratch disk on %s at %s', vm, mount_point)
> >    vm.RemoteCommand('sudo umount %s' % mount_point)
> >
> > -  if FLAGS.device_fill_size is not '0':
>
> I took out the if device_fill_size is 0 check for two reasons. First, it
> wasn't very well done: it only caught a literal "0", not "0%", or "0G", or
> anything like that. Second, it didn't gain us very much. In my testing, if
> you run fio and tell it to fill with size 0, it returns very quickly. We
> didn't gain very much, and it made the code more complex, so I think it is
> not worth it.
>
> I would be open to adding the 0 check back, but I think it should be a
> separate change. If we do that change, I'd also like to address the issue
> that if we run against a filesystem with no fill (which is now the
> default), we have an unnecessary unmount-reformat-remount step. That will
> be a bigger change, but should also produce a more significant speedup.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/515/files#r41443204>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

noahl:
Sounds good. Can we call the flag --prefill_device, and default to our semi-smart behavior, rather than --prefill_device_when_using_filesystem? I think it's obvious on reading it what --prefill_device=True or --prefill_device=False mean, but --prefill_device_when_using_filesystem is more complicated.

perfkitbenchmarker/benchmarks/fio_benchmark.py
stfeng: Percentage is still necessary because the caller does not always know the
size of the disk (PKB provisions the disk and picked the size in the
default case)

noahl:
I just pushed a new commit adding a `--prefill_device` flag. Tony and I also talked about just not prefilling at all when using a filesystem, but I remembered after our conversation that that would take away a lot of the functionality of the `--working_set_size` flag, because fio will only write a file as large as our working set, but we might want to have a lot of cold data with a small working set.

voellm:
LGTM

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -411,7 +416,7 @@ def Prepare(benchmark_spec):
   logging.info('Umount scratch disk on %s at %s', vm, mount_point)
   vm.RemoteCommand('sudo umount %s' % mount_point)
 
-  if FLAGS.device_fill_size is not '0':
+  if FLAGS.prefill_device:
```
stfeng: I believe this should be:

if FLAGS.against_device or FLAGS.prefill_device:

That is, if the user wants to test against device, then we will always prefill the device (igoring the FLAGs.prefill_device); however, if the user wants to test using file system, then we read the prefill_device flag to check if the user still wants to fill the device even when they are testing against the file system.

Without my proposed fix, any caller who sets against_device to true would also *always* set prefill_device to true which is redundant. 

Also all our current automation that runs FIO tests against device, they would be broken after this code (because existing automations aren't aware of the prefill_device flag)

stfeng: I made a PR to fix this: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/516

PR #516, opened 2015-10-08
Fix the device filling behavior when testing directly against the dev…

stfeng:
…ice. (we should always fill the device in the directly-against-device mode to ensure correctness of the test!)

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -76,11 +76,14 @@ flags.DEFINE_list('generate_scenarios', None,
                   '--fio_jobfile.')
 flags.DEFINE_boolean('against_device', False,
                      'Unmount the device\'s filesystem so we can test against '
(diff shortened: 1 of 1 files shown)


stfeng:
@noahl please take a look. This PR fixes the issue I last commented in PR #515  

Thanks.

stfeng:
@yuyantingzero FYI, if you have a chance, please take a look as well. Thanks!

noahl:
I think that if prefill_device is explicitly given, then it should override against_device. A check like `if FLAGS.prefill_device or (not FLAGS['prefill_device'].present and FLAGS.against_device)` should work.

However, I don't think @voellm agrees that this should be the default behavior. We'll have to talk about it.

voellm:
LGTM... the plan is to take this change to avoid a breaking change this time around.  That said there are a couple of improvement for fio next time;
- namespace all the flags
- merge the against_device and prefill_device into a enum like DEVICE_WITH_PREFILL, DEVICE_WITHOUT_PREFILL, FILE_WITH_PREFILL, FILE_WITHOUT_PREFILL.

PR #524, opened 2015-10-09
Pre-split YCSB table for Cloud Bigtable benchmark.

cmccoy:
This is recommended, and is now supported by the Bigtable client.

### DIFF
--- a/perfkitbenchmarker/benchmarks/hbase_ycsb_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/hbase_ycsb_benchmark.py
@@ -93,14 +93,16 @@ def CheckPrerequisites():
   ycsb.CheckPrerequisites()
 
 
-def _CreateYCSBTable(vm, table_name=TABLE_NAME, family=COLUMN_FAMILY,
-                     n_splits=TABLE_SPLIT_COUNT):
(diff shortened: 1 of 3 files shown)


perfkitbenchmarker/benchmarks/hbase_ycsb_benchmark.py
```diff
@@ -93,14 +93,16 @@ def CheckPrerequisites():
   ycsb.CheckPrerequisites()
 
 
-def _CreateYCSBTable(vm, table_name=TABLE_NAME, family=COLUMN_FAMILY,
-                     n_splits=TABLE_SPLIT_COUNT):
+def CreateYCSBTable(vm, table_name=TABLE_NAME, family=COLUMN_FAMILY,
```
ehankland: If this is being used across benchmarks, should it be pulled into one of the packages (hbase.py or ycsb.py)?

cmccoy: I thought about that.
I don't want it in `hbase.py`, since it's specific to HBase with YCSB, and putting database-specific implementations in `ycsb.py` seems like it will make that (already large) file larger.
Since Cloud Bigtable here is essentially HBase with a different backend, I thought I'd reuse some code from the HBase test.

ehankland:
LGTM

PR #526, opened 2015-10-09
Fix String Comparison in Object Storage Benchmark

noahl:
The benchmark is meant to give an error when the user wants to store
data in GCS or S3 but does not have a boto file. This commit fixes a
bug where it would also throw an error if the user wants to store data
in the Azure Blob Service but does not have a boto file.


noahl:
@stfeng Please take a look.

voellm:
LGTM

PR #530, opened 2015-10-12
Fix # of VMs in specification of aerospike_ycsb

cmccoy:


cmccoy:
@ehankland : could you take a look? I know #510 will also fix this, but I'd like the issue fixed in master for now.

ehankland:
LGTM - https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/510 should be in tomorrow barring any issues with the test I'm running

PR #531, opened 2015-10-13
rackspace: Remove extra parameter in constructor for firewall implementation

meteorfox:
Removes vestigial parameter on the constructor for Rackspace's Firewall
implementation, which made PKB exit with __init__ error.

This is a fix for the following error that seems to have sneaked in due to lack of testing on Rackspace.

```
    firewalls_dict[cls] = cls()
TypeError: __init__() takes exactly 2 arguments (1 given)
```


cmccoy:
Thanks!

Issue #532, opened 2015-10-13
Fio benchmark should use RobustRemoteCommand()


noahl:
This will let us continue when the ssh connection is reset.

noahl:
Fixed by #534

PR #201, opened 2015-04-01
Add support for configurable SSH port

meteorfox:
Adds the ability for a user to specify a port to be used for SSH and SCP
commmands when using static VMs. A port is specified using the key
'ssh_port' in the JSON static_vm file. If no port is specify, it defaults
to port 22.

Also, it makes the SSH port an instance variable for BaseVirtualMachine
instances, and removes it from the methods that took an SSH port as an
argument. Instead, it now uses the instance variable set at the object
initialization.

PR #534, opened 2015-10-13
Fio robustremotecommand

noahl:
Add a should_log option to vm.RobustRemoteCommand, which is passed through to the vm.RemoteCommand call. Then use RobustRemoteCommand in fio_benchmark.py to hopefully be more robust to ssh timeouts.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -188,7 +188,7 @@ def FillDevice(vm, disk, fill_size):
   """Fill the given disk on the given vm up to fill_size.
 
   Args:
-    vm: a virtual_machine.VirtualMachine object.
+    vm: a linux_virtual_machine.BaseLinuxMixin object.
     disk: a disk.BaseDisk attached to the given vm.
     fill_size: amount of device to fill, in fio format.
   """
@@ -198,7 +198,7 @@ def FillDevice(vm, disk, fill_size):
               '--rw=write --direct=1 --size=%s') %
              (fio.FIO_PATH, disk.GetDevicePath(), fill_size))
 
-  vm.RemoteCommand(command)
+  vm.RobustRemoteCommand(command)
 
 
 BENCHMARK_INFO = {'name': 'fio',
@@ -481,7 +481,7 @@ def Run(benchmark_spec):
       logging.info('**** Repetition number %s of %s ****',
                    repeat_number, total_repeats)
 
-    stdout, stderr = vm.RemoteCommand(fio_command, should_log=True)
+    stdout, stderr = vm.RobustRemoteCommand(fio_command, should_log=True)
 
     if repeat_number:
       base_metadata = {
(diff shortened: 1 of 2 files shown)


noahl:
@stfeng @yuyantingzero @cmccoy Please take a look.

noahl:
This will fix #532.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -151,7 +154,8 @@ def RobustRemoteCommand(self, command):
       return self.RemoteCommand(' '.join(wait_command), should_log=False)
     except:
       # In case the error was with the wrapper script itself, print the log.
-      stdout, _ = self.RemoteCommand('cat %s' % wrapper_log, should_log=False)
+      stdout, _ = self.RemoteCommand('cat %s' % wrapper_log,
+                                     should_log=should_log)
```
cmccoy: I don't think this should log - if it's interesting it's logged immediately below. Rather, I think the call to "wait_command" on L151 should use the passed in value of `should_log`.

noahl: Yep, my mistake. I just pushed a fix.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -198,7 +198,7 @@ def FillDevice(vm, disk, fill_size):
               '--rw=write --direct=1 --size=%s') %
              (fio.FIO_PATH, disk.GetDevicePath(), fill_size))
 
-  vm.RemoteCommand(command)
+  vm.RobustRemoteCommand(command)
```
voellm: did you want to pass (command. true) to turn on logging?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -103,7 +103,7 @@ def _PushRobustCommandScripts(self):
                                             os.path.basename(f)))
         self._has_remote_command_script = True
 
-  def RobustRemoteCommand(self, command):
+  def RobustRemoteCommand(self, command, should_log=False):
```
voellm: Depending on the use here you might also want to support a timeout like other remote commands.

voellm:
Once you address the comment for line #201 (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/534#discussion-diff-42001328L201)  LGTM.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I didn't turn on logging here because this is a pretty boring command that just pre-fills the disk. I do use logging in the more interesting call to RobustRemoteCommand later.

perfkitbenchmarker/linux_virtual_machine.py
noahl: I didn't add timeout to this change just because I don't need it right now and I wanted to keep this PR as simple as possible. If you want, I'll start a new PR and add timeouts, and maybe other features of RemoteCommand().

PR #536, opened 2015-10-14
Fix bug with running the same benchmark multiple times in the same 'run'

ehankland:


voellm:
Is the problem the following ... 
./pkb.py --benchmarks=iperf,iperf ?

Not sure what issue you are actually trying to address.  Can you explain the problem?  I want to understand this before reviewing.

Thanks!

ehankland:
If you run with a config, you can specify benchmarks to run:
```
benchmarks:
  - iperf: null
  - iperf: null
```
This would run iperf twice with the default config. Because of the way we pickle benchmark specs, both would be written out to the same file, which is obviously a problem. This PR fixes this.

voellm:
Ok ... this is a result of the new config files vs command line options.
Got it.

On Wed, Oct 14, 2015 at 9:15 AM, ehankland <notifications@github.com> wrote:

> If you run with a config, you can specify benchmarks to run:
>
> benchmarks:
>   - iperf: null
>   - iperf: null
>
> This would run iperf twice with the default config. Because of the way we
> pickle benchmark specs, both would be written out to the same file, which
> is obviously a problem. This PR fixes this.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/536#issuecomment-148103494>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -176,23 +176,24 @@ def _GetDiskSpecClass(cloud):
...
       benchmark_config: A Python dictionary representation of the configuration
         for the benchmark. For a complete explanation, see
         perfkitbenchmarker/configs/__init__.py.
-      benchmark_name: The name of the benchmark.
+      benchmark_uid: An identifier unique to this run of the benchmark even
+        if the same benchmark is run multiple times with different configs.
     """
     self.config = benchmark_config
-    self.name = benchmark_name
+    self.uid = benchmark_uid
```
voellm: For purposes of debug it might be useful to leave the name as well.

voellm:
Other than the comment on line 190 LGTM

perfkitbenchmarker/benchmark_spec.py
ehankland: uid is the name of the benchmark + a number

Issue #538, opened 2015-10-14
rackspace: SSH key is not uploaded to provider when using a multi-zone benchmark configuration


meteorfox:
Given that now PKB allows benchmarks to be executed across zones, and even providers, there's a bug in the Rackspace implementation where it assumes that the public SSH key was uploaded once to one region.

PR #539, opened 2015-10-14
rackspace: Fix SSH key upload for multi-zone benchmarks

meteorfox:
Remove unnecessary interaction with Rackspace API to upload an SSH public key.
Uploading a key using the Rackspace API required coordination to avoid a name
conflict errors, hence a lock was implemented that uploaded the key once per
benchmark. This is not necessary since the public SSH key is already uploaded
to the VM using cloud-init.

Previously, the Rackspace implementation assumed that one public SSH key
was uploaded per benchmark. This assumption is no longer valid since
PKB supports benchmarks configuration that spans multiple zones.

Fixes issue #538.

### DIFF
--- a/perfkitbenchmarker/rackspace/rackspace_virtual_machine.py
+++ b/perfkitbenchmarker/rackspace/rackspace_virtual_machine.py
@@ -33,7 +33,6 @@ import logging
 import os
 import re
 import tempfile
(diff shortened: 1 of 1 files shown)


cmccoy:
Looks good, thanks!

Issue #537, opened 2015-10-14
Cassandra_stress benchmark fails in prepare stage


mateusz-blaszkowski:
I'm attaching the traceback:
```
2015-10-14 13:31:33,303 e337fc01 Thread-93 cassandra_stress(1/1) INFO     Running: ssh -A -p 30734 root@10.102.17.37 -2 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o IdentitiesOnly=yes -o PreferredAuthentications=publickey -o PasswordAuthentication=no -o ConnectTimeout=5 -o GSSAPIAuthentication=no -o ServerAliveInterval=30 -o ServerAliveCountMax=10 -i /tmp/perfkitbenchmarker/run_e337fc01/perfkitbenchmarker_keyfile cd /tmp/pkb/apache-cassandra/lib && curl -LJO https://maven.java.net/content/repositories/releases/net/java/dev/jna/jna/4.1.0/jna-4.1.0.jar
2015-10-14 13:31:36,612 e337fc01 MainThread cassandra_stress(1/1) ERROR    Error during benchmark cassandra_stress
Traceback (most recent call last):
  File "PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 281, in RunBenchmark
    DoPreparePhase(benchmark, benchmark_name, spec, detailed_timer)
  File "PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 198, in DoPreparePhase
    benchmark.Prepare(spec)
  File "PerfKitBenchmarker/perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py", line 123, in Prepare
    vm_util.RunThreaded(configure, vm_dict[DATA_NODE])
  File "PerfKitBenchmarker/perfkitbenchmarker/vm_util.py", line 336, in RunThreaded
    return RunParallelThreads(target_arg_tuples, max_concurrent_threads)
  File "PerfKitBenchmarker/perfkitbenchmarker/vm_util.py", line 264, in RunParallelThreads
    _GetCallString(target_arg_tuples[call_id]), os.linesep, stacktrace))
  File "PerfKitBenchmarker/perfkitbenchmarker/vm_util.py", line 186, in _GetCallString
    return '{0}({1})'.format(target.__name__, ', '.join(arg_strings))
AttributeError: 'functools.partial' object has no attribute '__name__'
```

cmccoy:
[...]

PR #540, opened 2015-10-14
Fix _GetCallString when called with partial.

cmccoy:
See #537.


cmccoy:
Closing in favor of #542

PR #542, opened 2015-10-14
Add support for partials to _GetCallString.

skschneider:
Fix for the exception referenced in #537


skschneider:
Bit of confusion here about issue numbers. Original bug was in #537. Updated this PR's description to reference it.

cmccoy:
LGTM.

PR #544, opened 2015-10-14
Move integer list parsing to new module

noahl:
Remove GetIODepths and related functions from fio_benchmark.py, and move
them to a new module, flag_util.py. They are not connected to fio at
all, and should be more generally useful.

### DIFF
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -30,6 +30,7 @@ from perfkitbenchmarker import configs
 from perfkitbenchmarker import data
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
+from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.packages import fio
 
@@ -94,10 +95,12 @@ flags.DEFINE_string('device_fill_size', '100%',
                     'of the device. A filesystem will be unmounted before '
                     'filling and remounted afterwards. Only valid when '
                     'used with --prefill_device.')
-flags.DEFINE_string('io_depths', '1',
-                    'IO queue depths to run on. Can specify a single number, '
-                    'like --io_depths=1, a range, like --io_depths=1-4, or a '
-                    'list, like --io_depths=1-4,6-8')
+flag_util.DEFINE_integerlist('io_depths', [1],
+                             'IO queue depths to run on. Can specify a single '
+                             'number, like --io_depths=1, a range, like '
+                             '--io_depths=1-4, or a list, like '
+                             '--io_depths=1-4,6-8',
+                             on_nonincreasing=flag_util.IntegerListParser.WARN)
 flags.DEFINE_integer('working_set_size', None,
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device. If using '
@@ -115,74 +118,6 @@ flags.DEFINE_integer('run_for_minutes', 10,
 
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
     'generate_scenarios', 'io_depths', 'run_for_minutes'}
-
-
-IODEPTHS_REGEXP = re.compile(r'(\d+)(-(\d+))?$')
-
-
-def GetIODepths(io_depths):
-  """Parse the io_depths parameter.
-
-  Args:
-    io_depths: a string in the format of the --io_depths flag.
-
-  Returns:
-    A list of integers.
-
-  Raises:
-    ValueError if io_depths doesn't follow a format it recognizes.
-  """
-
-  groups = io_depths.split(',')
-  result = []
-
-  for group in groups:
-    match = IODEPTHS_REGEXP.match(group)
-    if match is None:
-      raise ValueError('Invalid io_depths expression %s', io_depths)
-    elif match.group(2) is None:
-      result.append(int(match.group(1)))
-    else:
-      result.extend(range(int(match.group(1)), int(match.group(3)) + 1))
-
-  return result
-
-
-def WarnIODepths(depths_list):
-  """Given a list of IO depths, log warnings if it seems "weird".
-
-  Args:
-    depths_list: a list of IO depths.
-  """
-
-  for i in range(len(depths_list) - 1):
-    if depths_list[i] >= depths_list[i + 1]:
-      logging.warning('IO depths list is not monotonically increasing: '
-                      '%s comes before %s.', depths_list[i], depths_list[i + 1])
-
-  values = set()
-  warned_on = set()
-  for val in depths_list:
-    if val in values and val not in warned_on:
-      logging.warning('IO depths list contains duplicate entry %s.', val)
-      warned_on.add(val)
-    else:
-      values.add(val)
-
-
-def IODepthsValidator(string):
-  try:
-    WarnIODepths(GetIODepths(string))
-    return True
-  except ValueError:
-    return False
-
-
-flags.RegisterValidator('io_depths',
-                        IODepthsValidator,
-                        message='--io_depths must be an integer, '
-                                'range of integers, or a list of '
-                                'integers and ranges, all > 0')
 
 
 def FillDevice(vm, disk, fill_size):
@@ -241,13 +176,13 @@ SECONDS_PER_MINUTE = 60
 
 
 def GenerateJobFileString(filename, scenario_strings,
-                          io_depths_string, working_set_size):
+                          io_depths, working_set_size):
   """Make a string with our fio job file.
 
   Args:
     filename: the file or disk we pre-filled, if any.
     scenario_strings: list of strings with names in SCENARIOS.
-    io_depths_string: string. The IO queue depths to test.
+    io_depths: iterable of integers. The IO queue depths to test.
     working_set_size: int or None. If int, the size of the working set in GB.
 
   Returns:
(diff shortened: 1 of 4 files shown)


noahl:
@cmccoy @yuyantingzero Please take a look.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,94 @@
...
+
+"""Utility functions for working with user-supplied flags."""
+
+import logging
+import re
+
+INTEGER_GROUP_REGEXP = re.compile(r'(\d+)(-(\d+))?$')
+
+
+def ParseIntegerList(inp, on_nonincreasing=None):
```
cmccoy: How about making a flag-defining function here, a la [how gflags does it](https://github.com/gflags/python-gflags/blob/master/gflags.py#L2698)?

noahl: The only reason I didn't is that I don't think that's a public interface, but I guess with the config stuff we're way past that point with gflags. I'll add a flag-defining function.

noahl: After looking at it for a bit longer, I can write a DEFINE_-style function like the ones in gflags, but I don't think I can implement a serializer in the way I'd want to. I think the right way to do it is to parse the string input into an intermediate list that stores either single integers or tuples giving the bounds on ranges. You can easily go from this to either an iterable over all of the integers, equivalent to the list the ParseIntegerList function returns, or back to the original string representation for your serializer.

When PKB first starts up, it prints out all of its command-line flags using serializers (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/pkb.py#L326). This would break that logging, in a way that having people use DEFINE_string and register validators would not.

I could make a function called DEFINE_integerlist that looks like a flags function but doesn't really work the same way, but I'm afraid that would break things that depend on the flags interface. However, I don't see any other options. What do you think?

noahl: I just implemented the solution we discussed yesterday.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+      match = INTEGER_GROUP_REGEXP.match(group)
+      if match is None:
+        raise ValueError('Invalid integer list %s', inp)
+      elif match.group(2) is None:
+        val = int(match.group(1))
+
+        if (self.on_nonincreasing and
+            len(result) > 0 and
+            val <= result[-1]):
+          if self.on_nonincreasing == "Warn":
```
cmccoy: Introduce a private function to handle this.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+          return self.groups[group][0] + idx
+
+    if isinstance(self.groups[group], tuple):
+      return self.groups[group][0]
+    else:
+      return self.groups[group]
+
+  def __iter__(self):
+    return IntegerList.IntegerListIterator(self)
```
cmccoy: Nit: I think this would be a bit cleaner with an inline generator function here.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+        length += 1
+      if isinstance(elt, tuple):
+        length += elt[1] - elt[0] + 1
+
+    self.length = length
+
+  def __len__(self):
+    return self.length
+
+  def __getitem__(self, idx):
```
cmccoy: Do you need to support `__getitem__`?

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+import logging
+import re
+
+from perfkitbenchmarker import flags
+
+INTEGER_GROUP_REGEXP = re.compile(r'(\d+)(-(\d+))?$')
+
+
+class IntegerList(object):
+  """An immutable list of nonnegative integers stored with
```
cmccoy: Nit: reduce this to a summary that can fit on a single line.

perfkitbenchmarker/flag_util.py
noahl: I did that, then realized I could make it even shorter (and hopefully clearer).

perfkitbenchmarker/flag_util.py
noahl: Done.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+    return IntegerList(result)
+
+  def Type(self):
+    return 'integer list'
+
+
+class IntegerListSerializer(flags.ArgumentSerializer):
+  def Serialize(self, il):
+    return ','.join([str(val) if isinstance(val, int) or isinstance(val, long)
+                     else '%s-%s' % (val[0], val[1])
```
cmccoy: Consider introducing namedtuple subclasses that support serialization to string and iteration to represent the two cases (int or range) that can be stored in the list. Probably overkill.

perfkitbenchmarker/flag_util.py
noahl: I did it because it's part of the immutable sequence interface (https://docs.python.org/2/reference/datamodel.html#special-method-names). I don't use it in my own code, but I would have to implement basically the same thing anyway to implement __iter__, which I do use.

(It doesn't quite implement the sequence interface, because __getitem__ doesn't accept slices, but it's very close.)

perfkitbenchmarker/flag_util.py
noahl: Done.

perfkitbenchmarker/flag_util.py
noahl: I agree that the integer list representation is not great, but I thought that adding more machinery for that was probably more effort than this deserved. I'm happy to do it if you want it, though.

noahl:
Oops! Yes, that was O(n^2). I implemented a better iter() method that will yield the next element in O(1) time and iterate through the whole list in O(n) time.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+    Raises:
+      ValueError if inp doesn't follow a format it recognizes.
+    """
+
+    if isinstance(inp, IntegerList):
+      return inp
+    elif isinstance(inp, list):
+      return IntegerList(inp)
+
+    def NonIncreasing():
```
cmccoy: Nit: `HandleNonIncreasing`?

perfkitbenchmarker/flag_util.py
cmccoy: No, SGTM.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,198 @@
...
+    if isinstance(self.groups[group_idx], tuple):
+      return self.groups[group_idx][0]
+    else:
+      return self.groups[group_idx]
+
+  def __iter__(self):
+    group_idx = 0
+    in_group_idx = 0
+
+    while group_idx < len(self.groups):
```
cmccoy: All of the index tracking muddles the intent a bit here. Could you simplify by iterating directly? Something like (untested):

```{python}
for group in self.groups:
  if (single value):
    yield group
  else:
    low, high = group
    for x in xrange(low, high+1):
      yield x
```

perfkitbenchmarker/flag_util.py
noahl: Done.

perfkitbenchmarker/flag_util.py
noahl: Done. I like your version much better. :)

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -94,10 +95,12 @@
...
-flags.DEFINE_string('io_depths', '1',
-                    'IO queue depths to run on. Can specify a single number, '
-                    'like --io_depths=1, a range, like --io_depths=1-4, or a '
-                    'list, like --io_depths=1-4,6-8')
+flag_util.DEFINE_integerlist('io_depths', [1],
+                             'IO queue depths to run on. Can specify a single '
+                             'number, like --io_depths=1, a range, like '
+                             '--io_depths=1-4, or a list, like '
+                             '--io_depths=1-4,6-8',
+                             on_nonincreasing="Warn")
```
cmccoy: Prefer single quotes.

perfkitbenchmarker/flag_util.py
```diff
@@ -0,0 +1,186 @@
...
+class IntegerListParser(flags.ArgumentParser):
+  """Parse a string containing a comma-separated list of nonnegative integers.
+
+  The list may contain single integers and dash-separated ranges. For
+  example, "1,3,5-7" parses to [1,3,5,6,7].
+
+  Can pass the flag on_nonincreasing to the constructor to tell it
+  what to do if the list is nonincreasing. Options are
+    - None: do nothing.
+    - "Warn": log a warning.
```
cmccoy: Nit: Consider constants for these. Also, most of our enum values are lower case.

cmccoy:
Two more very small things, otherwise LGTM. Consider squashing these commits before merging.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/flag_util.py
noahl: Made constants and changed to lower case.

noahl:
Just made the two small changes and squashed everything together.

PR #545, opened 2015-10-14
Fix reference to BenchmarkSpec.benchmark_name in dstat

cmccoy:


perfkitbenchmarker/traces/dstat.py
```diff
@@ -110,7 +110,7 @@ def _StopOnVm(self, vm):
 
   def Start(self, sender, benchmark_spec):
     """Install and start dstat on all VMs in 'benchmark_spec'."""
-    suffix = '-{0}-{1}-dstat'.format(benchmark_spec.benchmark_name,
+    suffix = '-{0}-{1}-dstat'.format(benchmark_spec.name,
```
ehankland: benchmark_spec.name was changed to benchmark_spec.uid to allow for pickling multiple runs of the same benchmark. The uid is just the name of the benchmark followed by a number. See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/536

ehankland:
Other than replacing name with uid, LGTM

perfkitbenchmarker/traces/dstat.py
cmccoy: Done.

Issue #512, opened 2015-10-06
Write configs for benchmarks


ehankland:
As soon as the configs refactor is in, we need to write reasonable configs for all of the benchmarks (especially the client/server ones where the clients dont need disks, but the servers do). 

At some point we should also discuss changing the defaults on a per benchmark basis - for example using disks that have the same number of advertised IOPS rather than just using 500 GB disks.

ehankland:
Considering this closed thanks to @cmccoy 's PRs. We still will need to address better defaults, but multi node benchmarks have sane configs now

PR #549, opened 2015-10-15
Specify cassandra-stress VMs via configs.

### LINKED ISSUE #490: Cassandra stress advertises that it doesn't require disks
But it does:

https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py#L97-L108

See #475.

cmccoy:
Fixes #490.
Partial work toward #512.

### DIFF
--- a/perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
@@ -29,7 +29,6 @@ import time
 
 
 from perfkitbenchmarker import configs
-from perfkitbenchmarker import disk
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import sample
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -55,18 +54,20 @@
 cassandra_stress:
   description: Benchmark Cassandra using cassandra-stress
   vm_groups:
-    default:
+    cassandra:
+      vm_spec: *default_single_core
+      disk_spec: *default_500_gb
```
voellm: Why has the size been hard coded for hte default data disk?  Why not call this "*default_data_disk"?

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -55,18 +54,20 @@
 cassandra_stress:
   description: Benchmark Cassandra using cassandra-stress
   vm_groups:
-    default:
+    cassandra:
```
voellm: cassandra_cluster?

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
cmccoy: This was discussed in #510; the decision there was to leave the current default in place and be explicit about the characteristics of the disks involved. @noahl - could you open an issue for your proposed changes to this?

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
cmccoy: How about `cassandra_nodes`? That matches `hbase_ycsb` fairly well.

cmccoy:
Note that this changes the metadata reported slightly - there will be separate entries for the client and server node groups.

ehankland:
LGTM

PR #557, opened 2015-10-16
Avoid formatting disk multiple times in fio

ehankland:
@noahl @yuyantingzero  Could you please take a look?

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -157,7 +157,7 @@ flags.DEFINE_enum(
     'os_type is "rhel". In general if two OS\'s use the same package manager, '
     'and are otherwise very similar, the same os_type should work on both of '
     'them.')
-flags.DEFINE_string('scratch_dir', '/scratch',
+flags.DEFINE_string('scratch_dir', None,
                     'Base name for all scratch disk directories in the VM.'
                     'Upon creation, these directories will have numbers'
                     'appended to them (for example /scratch0, /scratch1, etc).')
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/benchmark_spec.py
```diff
@@ -157,7 +157,7 @@ def UnPickleLock(locked, *args):
     'os_type is "rhel". In general if two OS\'s use the same package manager, '
     'and are otherwise very similar, the same os_type should work on both of '
     'them.')
-flags.DEFINE_string('scratch_dir', '/scratch',
+flags.DEFINE_string('scratch_dir', None,
```
noahl: Does this mean that *every* benchmark now gets its disk unmounted by default?

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -352,20 +358,15 @@ def Prepare(benchmark_spec):
...
-
-  logging.info('Umount scratch disk on %s at %s', vm, mount_point)
-  vm.RemoteCommand('sudo umount %s' % mount_point)
 
   if FLAGS.against_device or FLAGS.prefill_device:
     logging.info('Fill device %s on %s', disk.GetDevicePath(), vm)
     FillDevice(vm, disk, FLAGS.device_fill_size)
 
-  if not FLAGS.against_device:
+  if not FLAGS.against_device and FLAGS.prefill_device:
```
noahl: We need to mount the disk if we're not running against the device, regardless of whether we're prefilling it.

perfkitbenchmarker/benchmark_spec.py
ehankland: No - the default is to use the value in default_config_constants.yaml, which is still /scratch.

perfkitbenchmarker/benchmarks/fio_benchmark.py
ehankland: If we are not prefilling it, then it is already mounted (it is only not mounted if the against_device or prefill_device flags are specified).

noahl: Got it.

noahl:
LGTM.

PR #571, opened 2015-10-20
Introduce GCE disk integration tests

noahl:
Refactor pkb.py slightly to make it easy to initialize PKB without
running benchmarks. Use this ability to add a GCE disk integration test,
which uses configs to actually allocate a GCE VM with a disk attached
and check that everything worked properly. Add a convenience function in
test_util.py to make this easier.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -365,7 +365,7 @@ class BenchmarkSpec(object):
     vm.AddMetadata(benchmark=self.uid)
     vm.WaitForBootCompletion()
     vm.OnStartup()
-    if FLAGS.scratch_disk_type == disk.LOCAL:
+    if any((d.disk_type == disk.LOCAL for d in vm.disk_specs)):
       vm.SetupLocalDisks()
     for disk_spec in vm.disk_specs:
       vm.CreateScratchDisk(disk_spec)
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -393,7 +393,7 @@ class BenchmarkSpec(object):
     vm.AddMetadata(benchmark=self.uid, perfkit_uuid=self.uuid)
     vm.WaitForBootCompletion()
     vm.OnStartup()
-    if any((d.disk_type == disk.LOCAL for d in vm.disk_specs)):
+    if any((spec.disk_type == disk.LOCAL for spec in vm.disk_specs)):
       vm.SetupLocalDisks()
     for disk_spec in vm.disk_specs:
       vm.CreateScratchDisk(disk_spec)
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -65,6 +65,7 @@ from perfkitbenchmarker import benchmark_sets
 from perfkitbenchmarker import benchmark_spec
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import disk
+from perfkitbenchmarker import errors
 from perfkitbenchmarker import events
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import log_util
@@ -331,23 +332,17 @@ def _LogCommandLineFlags():
   logging.info('Flag values:\n%s', '\n'.join(result))
 
 
-def RunBenchmarks(publish=True):
-  """Runs all benchmarks in PerfKitBenchmarker.
+def SetUpPKB():
+  """Set globals and environment variables for PKB.
 
-  Args:
-    publish: A boolean indicating whether results should be published.
-
-  Returns:
-    Exit status for the process.
+  Having this function separate from functions that do any work lets
+  us call this to set up integration tests.
   """
-  if FLAGS.version:
-    print version.VERSION
-    return
 
   for executable in REQUIRED_EXECUTABLES:
     if not vm_util.ExecutableOnPath(executable):
       logging.error('Could not find required executable "%s".' % executable)
-      return 1
+      raise errors.Error()
 
   if FLAGS.run_uri is None:
     if FLAGS.run_stage not in [STAGE_ALL, STAGE_PREPARE]:
@@ -361,13 +356,13 @@ def RunBenchmarks(publish=True):
       else:
         logging.error(
             'No run_uri specified. Could not run "%s".', FLAGS.run_stage)
-        return 1
+        raise errors.Error()
     else:
       FLAGS.run_uri = str(uuid.uuid4())[-8:]
   elif not FLAGS.run_uri.isalnum() or len(FLAGS.run_uri) > MAX_RUN_URI_LENGTH:
     logging.error('run_uri must be alphanumeric and less than or equal '
                   'to 8 characters in length.')
-    return 1
+    raise errors.Error()
 
   vm_util.GenTempDir()
   log_util.ConfigureLogging(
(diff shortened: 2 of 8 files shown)


noahl:
@cmccoy @stfeng @yuyantingzero Please take a look.

noahl:
@skschneider PTAL

perfkitbenchmarker/pkb.py
```diff
@@ -331,23 +332,17 @@ def _LogCommandLineFlags():
...
   """
-  if FLAGS.version:
-    print version.VERSION
-    return
 
   for executable in REQUIRED_EXECUTABLES:
     if not vm_util.ExecutableOnPath(executable):
       logging.error('Could not find required executable "%s".' % executable)
-      return 1
+      raise errors.Error()
```
skschneider: Maybe UserSetupError or some other more specific type. Also, add message to exception. Same with others below.

perfkitbenchmarker/pkb.py
```diff
@@ -331,23 +332,17 @@ def _LogCommandLineFlags():
...
-
-  Args:
-    publish: A boolean indicating whether results should be published.
+def SetUpPKB():
+  """Set globals and environment variables for PKB.
 
-  Returns:
-    Exit status for the process.
+  Having this function separate from functions that do any work lets
+  us call this to set up integration tests.
```
skschneider: I don't think this comment is necessary in the code itself, but it might be nice to elaborate here on what this function does (for example, I see it creates a temp dir on the local OS, which isn't mentioned yet)

perfkitbenchmarker/pkb.py
```diff
@@ -375,6 +370,26 @@ def RunBenchmarks(publish=True):
...
       run_uri=FLAGS.run_uri)
   logging.info('PerfKitBenchmarker version: %s', version.VERSION)
+
+  vm_util.SSHKeyGen()
+
+  events.initialization_complete.send(parsed_flags=FLAGS)
+
+
+def RunBenchmarks(publish=True):
+
```
skschneider: There's a blank line here.

perfkitbenchmarker/pkb.py
```diff
@@ -491,4 +504,10 @@ def Main(argv=sys.argv):
         '%s\nUsage: %s ARGS\n%s', e, sys.argv[0], FLAGS)
     sys.exit(1)
 
+  try:
+    SetUpPKB()
+  except Exception as e:
+    logging.error('Got exception %s', e)
+    sys.exit(1)
```
skschneider: I don't think this try/except is necessary. If an exception is raised, python will exit with a non-zero status.

perfkitbenchmarker/test_util.py
```diff
@@ -66,3 +69,23 @@ def assertSampleListsEqualUpToTimestamp(self, a, b, msg=None):
         ex.message = ex.message + (' (was item %s in list)' % i)
         ex.args = (ex.message,)
         raise ex
+
+
+def testDiskMounts(benchmark_config, mount_point):
+  spec = benchmark_spec.BenchmarkSpec(benchmark_config, 'uid')
```
skschneider: Please add a docstring.

tests/gce_disk_integration_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+from perfkitbenchmarker import pkb
+from perfkitbenchmarker import test_util
+
+MOUNT_POINT = '/scratch'
+
+
+@unittest.skipUnless('PERFKIT_INTEGRATION' in os.environ,
+                     'PERFKIT_INTEGRATION not in environment')
+class GcpScratchDiskIntegrationTest(unittest.TestCase):
```
skschneider: Please add a docstring explaining what these tests are doing. Since they involve actually interacting with a machine and may require a certain setup prior to running the test, there should be some resource for a user who wants to run all tests but doesn't know what these ones should do but then sees them failing.

skschneider:
Are there going to be README updates in a follow-up PR? There should be instructions about how to run these tests, since they aren't run by default.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -365,7 +365,7 @@ def PrepareVm(self, vm):
     vm.AddMetadata(benchmark=self.uid)
     vm.WaitForBootCompletion()
     vm.OnStartup()
-    if FLAGS.scratch_disk_type == disk.LOCAL:
+    if any((d.disk_type == disk.LOCAL for d in vm.disk_specs)):
```
yuyantingzero: Consider using more verbose variable name.

noahl: Changed to `spec`.

perfkitbenchmarker/pkb.py
noahl: Changed the description to be more clear. I mentioned what it does, but I think the most important part is the postcondition - after running `SetUpPKB()`, you should be able to run any other PKB function and get useful results.

perfkitbenchmarker/pkb.py
noahl: Added new errors in `errors.py` and used them here and in the places below. Done.

perfkitbenchmarker/pkb.py
noahl: Fixed. Thanks for catching it.

perfkitbenchmarker/pkb.py
noahl: I did it this way to imitate the way the old code returned. However, you're completely right. Changed.

perfkitbenchmarker/test_util.py
noahl: Done. Also renamed the function to `assertDiskMounts` to be more clear about what it does.

tests/gce_disk_integration_test.py
noahl: After talking with Skyler in person, I added a section to the README discussing the integration tests, and put small docstrings in each of the integration tests classes pointing to the README. This lets me avoid repeating the same information in the docstring for each integration test, but still keeps it visible and available to users.

noahl:
I believe I have now fixed all the issues found in review comments. @skschneider @yuyantingzero please take a look.

skschneider:
The latest I'm seeing is that merge from master to the disk-integration-tests branch. Make sure you've pushed your local changes to github.

noahl:
@skschneider Oops, looks like I managed to push to Github before committing locally. I really uploaded them this time.

perfkitbenchmarker/pkb.py
```diff
@@ -335,23 +336,20 @@ def _LogCommandLineFlags():
...
 
-def RunBenchmarks(publish=True):
-  """Runs all benchmarks in PerfKitBenchmarker.
+def SetUpPKB():
+  """Set globals and environment variables for PKB.
 
-  Args:
-    publish: A boolean indicating whether results should be published.
+  After SetUpPKB() returns, it should be possible to call PKB
+  functions, like benchark_spec.Prepare() or benchmark_spec.Run().
```
skschneider: Spelling "benchark_spec"

perfkitbenchmarker/test_util.py
```diff
@@ -66,3 +69,39 @@ def assertSampleListsEqualUpToTimestamp(self, a, b, msg=None):
         ex.message = ex.message + (' (was item %s in list)' % i)
         ex.args = (ex.message,)
         raise ex
+
+
+def assertDiskMounts(benchmark_config, mount_point):
+  """Test whether a disk mounts in a given configuration.
+
+  Sets up a virtual machine following benchmark_config and then tests
```
skschneider: benchmark_config may specify more than one VM, so I think it should be clear that only one VM from the config will be used. I think the best approach here is to force this function to only accept a spec with one VM (doc & assert) because the order of VMs in the spec.vms list appears to be random, so it's unclear which settings are used if more than one VM are specified.

skschneider:
Just the spelling and the one doc thing. Otherwise LGTM.

perfkitbenchmarker/pkb.py
noahl: Fixed.

perfkitbenchmarker/test_util.py
noahl: Thanks for pointing this out. I implemented your doc & assert idea.

noahl:
@skschneider Fixed the spelling and doc things you found. Okay to merge?

perfkitbenchmarker/test_util.py
```diff
@@ -66,3 +69,48 @@ def assertSampleListsEqualUpToTimestamp(self, a, b, msg=None):
...
+  whether the path mount_point contains a working disk by trying to
+  create a file there. Returns nothing if file creation works;
+  otherwise raises an exception.
+
+  The given benchmark_config must specify exactly one virtual machine.
```
skschneider: This could go under the benchmark_config arg description.

skschneider:
Another nit comment, but yes, looks fine. Merge whenever ready.

perfkitbenchmarker/test_util.py
noahl: Done.

PR #574, opened 2015-10-21
Kubernetes: adapting provider to work with different versions of kubectl.

mateusz-blaszkowski:
I got rid of 'jsonpath' output format which is not available in older
versions of kubectl. On the other hand, '--validate=false' flag was used
in order to work with the latest kubectl.

I also removed the unnecessary selectors.

### DIFF
--- a/perfkitbenchmarker/kubernetes/kubernetes_virtual_machine.py
+++ b/perfkitbenchmarker/kubernetes/kubernetes_virtual_machine.py
@@ -138,7 +138,7 @@ class KubernetesVirtualMachine(virtual_machine.BaseVirtualMachine):
     Creates a Service (POD accessor).
     """
     create_cmd = [FLAGS.kubectl, '--kubeconfig=%s' % FLAGS.kubeconfig,
-                  'create', '-f', '-']
+                  'create', '--validate=false', '-f', '-']
     create_svc_body = self._BuildServiceBody()
     output = vm_util.IssueCommand(create_cmd, input=create_svc_body)
     if output[EXIT_CODE]:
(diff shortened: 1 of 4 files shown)


cmccoy:
What affect does `--validate=false` have? If it's related to a version change, could you add a comment somewhere noting when it can be removed?

perfkitbenchmarker/kubernetes/kubernetes_virtual_machine.py
```diff
@@ -215,14 +205,11 @@ def _Exists(self):
...
-                  'pod', '-o=json', '-l', '%s=%s' % (SELECTOR_PREFIX,
-                                                     self.name)]
+                  'pod', '-o=json', self.name]
     pod_info, _, _ = vm_util.IssueCommand(exists_cmd, suppress_warning=True)
-    pod_info = json.loads(pod_info)
-
-    if len(pod_info['items']) == 0:
-      return False
-    return True
+    if pod_info:
```
cmccoy: So there will be no output if the pod doesn't exist, rather than an empty JSON document?

mateusz-blaszkowski: Yes. '-l' parameter is removed and now I will get no output instead of empty JSON document

mateusz-blaszkowski:
`--validate` flag has to be passed when creating a Service because kubectl drops the request if `nodePort` parameter is not provided. However, it is not needed - in such case `nodePort` will be automatically drawn from the pool of ports. Looks like a bug in the v1.2.0-alpha.2 version of kubectl which I was using lately. 

There was a similar issue when creating a POD with Ceph volume. Kubectl requires to pass keyring and user. Thanks to your inquiry I realized that `--validate` flag is not needed in this case because I bypassed this by providing missing parameters all the time (with some default values).

I will push the improvements soon. Thanks

voellm:
Can you please update the README on how to install the matched version of kubctl that is expected to work. 

At some point real soon we should have code to check CLI versions to give the user warning when things are out of sync.

mateusz-blaszkowski:
@cmccoy - since now I'm able to merge the PR when it's ready and reviewed - how can I know that you (or someone else) accept (and reviewed) the changes? 'LGTM' is enough? :) I would merge this PR but I wanted to be clear on this. Thanks!

cmccoy:
@mateusz-blaszkowski: Right - we've been merging once someone else has give an LGTM. Thanks!

mateusz-blaszkowski:
Closing this PR as it has been already merged into master together with https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/593

PR #579, opened 2015-10-22
Move published run_uri to benchmark spec, add to VM metadata

cmccoy:
Publishes a per-benchmark UUID as VM metadata, includes is as run_uid in the output samples.


cmccoy:
@ehankland Could you PTAL? This changes the published run_uri behavior a bit, but it was already disconnected from the `--run_uri` flag..

ehankland:
LGTM

Issue #553, opened 2015-10-15
OpenStack: Race condition when creating floating ips for benchmarks with multiple VMs


meteorfox:
When using a benchmark that requires multiple VMs, such as netperf, ocassionally the Openstack implementation hits a race condition, and gets the same floating IP address for both VMs, which makes the benchmark fail.

Note below that both VMs got the IPv4 address 10.50.48.11

For the output below, the following flags were used:
```
2015-10-15 20:29:26,985 e047c792 MainThread INFO     PerfKitBenchmarker version: v0.21.0-52-g3a986bc
2015-10-15 20:29:26,985 e047c792 MainThread INFO     Flag values:
--machine_type=m1.medium
--openstack_public_network=public
--image=Ubuntu 14.04
--cloud=OpenStack
--benchmarks=netperf
--openstack_private_network=service
```

```bash
2015-10-15 20:29:44,797 e047c792 Thread-1 netperf(1/1) INFO     Starting new HTTP connection (1): 172.29.236.10
2015-10-15 20:29:44,829 e047c792 Thread-2 netperf(1/1) INFO     VM: 10.50.48.11
2015-10-15 20:29:44,830 e047c792 Thread-2 netperf(1/1) INFO     Waiting for boot completion.
2015-10-15 20:29:44,830 e047c792 Thread-2 netperf(1/1) INFO     Starting new HTTP connection (1): 172.29.236.10
2015-10-15 20:29:44,836 e047c792 Thread-1 netperf(1/1) INFO     VM: 10.50.48.11
2015-10-15 20:29:44,836 e047c792 Thread-1 netperf(1/1) INFO     Waiting for boot completion.
2015-10-15 20:29:44,837 e047c792 Thread-1 netperf(1/1) INFO     Starting new HTTP connection (1): 172.29.236.10
2015-10-15 20:29:59,867 e047c792 Thread-2 netperf(1/1) INFO     Running: ssh -A -p 22 ubuntu@10.50.48.11 -2 -o UserKnownHostsFile=/de
v/null -o StrictHostKeyChecking=no -o IdentitiesOnly=yes -o PreferredAuthentications=publickey -o PasswordAuthentication=no -o Connec
tTimeout=5 -o GSSAPIAuthentication=no -o ServerAliveInterval=30 -o ServerAliveCountMax=10 -i /tmp/perfkitbenchmarker/run_e047c792/per
fkitbenchmarker_keyfile hostname
[...]

PR #580, opened 2015-10-22
fix race condition in getting floating ip

kivio:
Fix for issue https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/553.
I canno't have environment to test that. Can you do it @meteorfox

### DIFF
--- a/perfkitbenchmarker/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/openstack/os_virtual_machine.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import threading
 import time
 
 from perfkitbenchmarker import virtual_machine, linux_virtual_machine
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/openstack/os_virtual_machine.py
```diff
@@ -131,8 +133,10 @@ def _PostCreate(self):
             instance = self.client.servers.get(self.id)
             status = instance.status
 
-        self.floating_ip = self.public_network.get_or_create()
-        instance.add_floating_ip(self.floating_ip)
+        with self._floating_ip_lock:
```
voellm: This CL looks fine but is this a Workaround for an Openstack issue?  If so can you add a comment here with the "os" bug?

kivio:
@voellm it's not OpenStack bug. My first method create and get info about some IP and in next method try to add it to instance, if another thread do this same with same IP in the meantime we have race condition. Getting information about IP doesn't block it or remove from pool of available IPs.

Issue #518, opened 2015-10-08
Clean up fio benchmark flag logic


stfeng:
1. We need to rename all fio flags to start with the "fio_" prefix.
2. We need to merge the against_device and prefill_device flags into one flag of the enum type, perhaps with the name fio_test_mode and it has the following four values: against_device_with_fill, against_device_without_fill, against_file_with_fill, against_file_without_fill
3. then we need to change the rest of code that currently consumes the against_device flag and the prefill_device flag to properly understand and process the new fio_test_mode flag.

Above is pretty much the code change involved, the more timing consuming part of this work is testing, we need to have complete test coverage of the newly named flags, including: 1) all 4 modes mentioned above, 2) for each renamed flags, test they can be used properly with their new names.

stfeng:
@noahl, I've discussed this with @voellm. We are going to fix this issue (and make a breaking change) for the next release. (It's a p0 for the next release.)

Expected code change should be simple (search and replace), the more time consuming part is testing. I'd budget about 2 days on this. Let me or Yanting know once you are done, we'd be happy to CR. Thanks!

noahl:
This is fixed by #581

PR #581, opened 2015-10-22
Update fio flags

noahl:
Fix GitHub issue #518 by renaming fio flags and changing the way fio
expresses its target and fill mode.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-# v0.23.0
+# v0.24.0
 
 * Benchmark-specific changes:
   * **BREAKING CHANGE** fio flags are prefixed with "fio", and new flag
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# v0.23.0
+
+* Benchmark-specific changes:
+  * **BREAKING CHANGE** fio flags are prefixed with "fio", and new flag
+    --fio_target_mode replaces --against_device and --prefill_device.
+
 # v0.22.1
 
 * Update the default image used in Azure tests to Ubuntu 14.04.3 (runs failed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,3 @@
-# v0.24.0
-
-* Benchmark-specific changes:
-  * **BREAKING CHANGE** fio flags are prefixed with "fio", and new flag
-    --fio_target_mode replaces --against_device and --prefill_device.
-
 # v0.22.1
 
 * Update the default image used in Azure tests to Ubuntu 14.04.3 (runs failed
--- a/perfkitbenchmarker/benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/fio_benchmark.py
@@ -70,47 +70,42 @@ FLAGS = flags.FLAGS
 
 flags.DEFINE_string('fio_jobfile', None,
                     'Job file that fio will use. If not given, use a job file '
-                    'bundled with PKB. Cannot use with --generate_scenarios.')
-flags.DEFINE_list('generate_scenarios', None,
+                    'bundled with PKB. Cannot use with '
+                    '--fio_generate_scenarios.')
+flags.DEFINE_list('fio_generate_scenarios', None,
                   'Generate a job file with the given scenarios. Special '
                   'scenario \'all\' generates all scenarios. Available '
                   'scenarios are sequential_write, sequential_read, '
                   'random_write, and random_read. Cannot use with '
                   '--fio_jobfile.')
-flags.DEFINE_boolean('against_device', False,
-                     'Unmount the device\'s filesystem so we can test against '
-                     'the raw block device; will always prefill the device to '
-                     'device_fill_size. '
-                     'If --generate-scenarios is given, '
-                     'will generate a job file that uses the block device.')
-flags.DEFINE_boolean('prefill_device', False,
-                     'Used when against_device is false: decides if it is '
-                     'required to prefill the device before the fio run. '
-                     'The amount to prefill is given by '
-                     '--device_fill_size.')
-flags.DEFINE_string('device_fill_size', '100%',
+flags.DEFINE_enum('fio_target_mode', 'against_file_without_fill',
+                  ['against_device_with_fill', 'against_device_without_fill',
+                   'against_file_with_fill', 'against_file_without_fill'],
+                  'Whether to run against a raw device or a file, and whether '
+                  'to prefill.')
+flags.DEFINE_string('fio_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'The valid value can either be an integer, which '
                     'represents the number of bytes to fill or a '
                     'percentage, which represents the percentage '
                     'of the device. A filesystem will be unmounted before '
                     'filling and remounted afterwards. Only valid when '
-                    'used with --prefill_device.')
-flag_util.DEFINE_integerlist('io_depths', [1],
+                    '--fio_target_mode specifies prefilling')
+flag_util.DEFINE_integerlist('fio_io_depths', [1],
                              'IO queue depths to run on. Can specify a single '
-                             'number, like --io_depths=1, a range, like '
-                             '--io_depths=1-4, or a list, like '
-                             '--io_depths=1-4,6-8',
+                             'number, like --fio_io_depths=1, a range, like '
+                             '--fio_io_depths=1-4, or a list, like '
+                             '--fio_io_depths=1-4,6-8',
                              on_nonincreasing=flag_util.IntegerListParser.WARN)
-flags.DEFINE_integer('working_set_size', None,
+flags.DEFINE_integer('fio_working_set_size', None,
                      'The size of the working set, in GB. If not given, use '
                      'the full size of the device. If using '
-                     '--generate_scenarios and not --against_device, you must '
-                     'pass --working_set_size.',
+                     '--fio_generate_scenarios and not running against a raw '
+                     'device, you must pass --fio_working_set_size.',
                      lower_bound=0)
-flags.DEFINE_integer('run_for_minutes', 10,
+flags.DEFINE_integer('fio_run_for_minutes', 10,
                      'Repeat the job scenario(s) for the given number of '
-                     'minutes. Only valid when using --generate_scenarios. '
+                     'minutes. Only valid when using --fio_generate_scenarios. '
                      'When using multiple scenarios, each one is run for the '
                      'given number of minutes. Time will be rounded up to the '
                      'next multiple of %s minutes.' % MINUTES_PER_JOB,
@@ -118,7 +113,13 @@ flags.DEFINE_integer('run_for_minutes', 10,
 
 
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
-    'generate_scenarios', 'io_depths', 'run_for_minutes'}
+    'fio_generate_scenarios', 'fio_io_depths', 'fio_run_for_minutes'}
+
+
+# Modes from --fio_target_mode
+AGAINST_DEVICE_MODES = {'against_device_with_fill',
+                        'against_device_without_fill'}
+FILL_TARGET_MODES = {'against_device_with_fill', 'against_target_with_fill'}
 
 
 def FillDevice(vm, disk, fill_size):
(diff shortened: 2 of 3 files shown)


noahl:
@stfeng @yuyantingzero @cmccoy Please take a look.

stfeng:
I will take a look.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -70,47 +70,42 @@
...
+flags.DEFINE_string('fio_fill_size', '100%',
                     'The amount of device to fill in prepare stage. '
                     'The valid value can either be an integer, which '
                     'represents the number of bytes to fill or a '
                     'percentage, which represents the percentage '
                     'of the device. A filesystem will be unmounted before '
                     'filling and remounted afterwards. Only valid when '
-                    'used with --prefill_device.')
-flag_util.DEFINE_integerlist('io_depths', [1],
+                    '--fio_target_mode specifies prefilling')
```
stfeng: be specific here: "when --fio_target_mode is in one of these modes [blah] "

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -118,7 +113,13 @@
...
 
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
-    'generate_scenarios', 'io_depths', 'run_for_minutes'}
+    'fio_generate_scenarios', 'fio_io_depths', 'fio_run_for_minutes'}
+
+
+# Modes from --fio_target_mode
+AGAINST_DEVICE_MODES = {'against_device_with_fill',
+                        'against_device_without_fill'}
+FILL_TARGET_MODES = {'against_device_with_fill', 'against_target_with_fill'}
```
stfeng: what is this "against_target_with_fill'?

stfeng: I think you meant "against_file_with_fill".

Please make sure you have COMPLETE coverage of this change which is relatively simple but more risky. (Things like this should have been caught by your testing effort)

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -359,11 +360,11 @@ def Prepare(benchmark_spec):
...
   disk = vm.scratch_disks[0]
 
-  if FLAGS.against_device or FLAGS.prefill_device:
+  if FLAGS.fio_target_mode in FILL_TARGET_MODES:
     logging.info('Fill device %s on %s', disk.GetDevicePath(), vm)
-    FillDevice(vm, disk, FLAGS.device_fill_size)
+    FillDevice(vm, disk, FLAGS.fio_fill_size)
 
-  if not FLAGS.against_device and FLAGS.prefill_device:
+  if FLAGS.fio_target_mode == 'against_file_with_fill':
```
stfeng: Put a comment here to clarify why we only need to mount the FS here in the against_file_with_fill mode, because:

1) if it's in any of those two against_device modes, we should not mount FS.
2) if it's in against_file_without_fill mode, we should not need to mount the FS here, because the FS was never unmounted in this mode - see the code in GetConfig() which does not clears the mount point when we are in against_file_without_fill mode.

It took me a bit of time to read this code, adding a comment like this helps future readers to figure this out quickly, saving their time.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -398,7 +400,7 @@ def Run(benchmark_spec):
 
   vm.PushFile(job_file_path, REMOTE_JOB_FILE_PATH)
 
-  if FLAGS.against_device:
+  if FLAGS.fio_target_mode in AGAINST_DEVICE_MODES:
```
stfeng: There are many repeat patterns like this: "FLAGS.fio_target_mode in AGAINST_DEVICE_MODES"
or "FLAGS.fio_target_mode not in AGAINST_DEVICE_MODES", consider making a private utility function that is simple one liner that returns a boolean for this.

e.g., 

def AgainstDevice() {
  return (blah)
}

Then in all these 'if' statements can be simplifed.

stfeng:
Finished first round of review. The code change itself is relatively straightforward but needs careful testing. (unfortunately) testing coverage of this change is much more time consuming than coding...

Please reply with a description of your test plan that will provide solid coverage for this change. Thanks!

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -329,7 +330,7 @@ def RunForMinutes(proc, mins_to_run, mins_per_call):
 
 def GetConfig(user_config):
   config = configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
-  if FLAGS.against_device or FLAGS.prefill_device:
+  if FLAGS.fio_target_mode != 'against_file_without_fill':
```
yuyantingzero: Use constants for fio_target_mode

cmccoy:
Ping. Is this on track for v0.23.0? If not, please update the milestone on #518.

perfkitbenchmarker/benchmarks/fio_benchmark.py

noahl:
@stfeng Testing the change has two parts. For renaming the flags that are now prefixed, I think manual testing (already done) plus linting is sufficient. For the --fio_target_mode flag, manual testing clearly isn't sufficient. For this flag, I will have two sets of tests - one verifying that Prepare() calls FillDevice() on the fill modes and not on the non-fill modes, and another verifying that Run() calls GetOrGenerateJobFileString() with the proper arguments in against device and against file modes.

noahl:
@cmccoy I don't know; it's going to be close to the line. I'm hoping to have the testing set up soon, and I should know more after that gets reviewed.

stfeng:
@noahl, thanks for the update on testing plan. Look forward to the updated code.

@cmccoy I think v0.23.0 might be tight (is the deadline early next week?). In general, quality trumps speed, especially for this one: no one else is blocked on this PR.

cmccoy:
@stfeng @noahl - the cutoff is today for code changes, release is scheduled for early next week. I'll update the milestone on #518 to the next release.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Just added testing.

perfkitbenchmarker/benchmarks/fio_benchmark.py

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. I also added a parallel function for checking if we fill the target device, called `FillTarget`.

perfkitbenchmarker/benchmarks/fio_benchmark.py

noahl:
@stfeng I just uploaded a new version with testing. I forgot to mention it earlier, but I also test whether or not we reformat the filesystem, because that also depends on the target mode.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -118,7 +114,32 @@
...
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
-    'generate_scenarios', 'io_depths', 'run_for_minutes'}
+    'fio_generate_scenarios', 'fio_io_depths', 'fio_run_for_minutes'}
+
+
+# Modes from --fio_target_mode
+AGAINST_DEVICE_MODES = {'against_device_with_fill',
+                        'against_device_without_fill'}
+FILL_TARGET_MODES = {'against_device_with_fill', 'against_file_with_fill'}
+AGAINST_DEVICE_WITH_FILL_MODE = 'against_device_with_fill'
```
stfeng: Please define this constant at the top of this function, and then use this constant in all the places where "against_device_with_fill" string is referenced, e.g., in the AGAINST_DEVICE_MODES array.

Also, define constants for all similar strings like: "against_device_without_fill", "against_file_with_fill", etc. This will provide a pretty effective guard against miss spelling of these strings. In all places where these strings are referenced in this code, they should be changed to refer to the global variables instead.

perfkitbenchmarker/benchmarks/fio_benchmark.py
stfeng: Not done? Need to define the string 'against_file_without_fill' as a constant at the top of this file and only refer to the constant in subsequent code.

perfkitbenchmarker/benchmarks/fio_benchmark.py
```diff
@@ -359,11 +380,15 @@ def Prepare(benchmark_spec):
...
     logging.info('Fill device %s on %s', disk.GetDevicePath(), vm)
-    FillDevice(vm, disk, FLAGS.device_fill_size)
+    FillDevice(vm, disk, FLAGS.fio_fill_size)
 
-  if not FLAGS.against_device and FLAGS.prefill_device:
+  # We only need to format and mount if the target mode is against
+  # file with fill because 1) if we're running against the device, we
+  # don't want it mounted and 2) if we're running against a file
+  # without fill, it was never unmounted (see GetConfig()).
+  if FLAGS.fio_target_mode == AGAINST_DEVICE_WITH_FILL_MODE:
```
stfeng: Your comment above is correct, but code is wrong - doing exactly the opposite of your intention. Follow-up questions:

1. Did you run your newly added integration/unit tests after you made this change? 

2. If so, why wasn't this bug caught by your tests? (I skimmed through your new unit tests, I *think* this should have been caught, please investigate further).

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -153,5 +154,62 @@ def TestZeroMinutes(self):
     self.assertEquals(proc.call_count, 0)
 
 
+class TestFioTargetModeFlag(unittest.TestCase):
+  def doTargetModeTest(self, mode, fill_device=None, against_device=None,
```
stfeng: Consider renaming: fill_device to "expect_fill_device", against_device to "expect_against_device" and format_disk to "expect_format_disk"

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -153,5 +154,62 @@ def TestZeroMinutes(self):
...
+                       format_disk=None):
+    with mock.patch(fio_benchmark.__name__ + '.FillDevice') as fill_device, \
```
stfeng: Consider renaming "fill_device" here to FillDevice (retain its original method name) so the reader knows this refers to a method/function rather than a local variable.

Similar comment to all the names below that are extracted from the fio_benchmark module. ("get_job_string", etc)

tests/benchmarks/fio_benchmark_test.py
```diff
@@ -153,5 +154,62 @@ def TestZeroMinutes(self):
...
+        self.assertEquals(against_device_arg, True)
+      elif against_device is False:
+        self.assertEquals(against_device_arg, False)
+
+      if format_disk is True:
+        self.assertEquals(benchmark_spec.vms[0].FormatDisk.call_count, 1)
+      elif format_disk is False:
+        self.assertEquals(benchmark_spec.vms[0].FormatDisk.call_count, 0)
+
+    def testAgainstFileWithFill(self):
```
stfeng: I'd expect most of the test cases below to have failed with your current code, that is, the error in:

line 391: "if FLAGS.fio_target_mode == AGAINST_DEVICE_WITH_FILL_MODE: then format disk", 

should have failed 3 of the cases here.

If you see these are all passing even with your current bug, then it's actually a more serious problem, something is wrong with the tests, please dig deeper.

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# v0.23.0
```
stfeng: since this is not going to make into v0.23.0, need to update this to v0.24.0.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Sorry, I fixed it in another part of the file and missed this one. This is done now.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: Done. I also used the constants in defining the actual flag enum values.

perfkitbenchmarker/benchmarks/fio_benchmark.py
noahl: I thought I was running the tests, but I discovered just now that none of tests tests were actually being run when I did `hooks/check-everything`. I can explain why tomorrow, but I have fixed it, validated that tests are actually run using a test designed to fail, and fixed the errors.

tests/benchmarks/fio_benchmark_test.py
noahl: Done.

tests/benchmarks/fio_benchmark_test.py
noahl: Done. I kept the `fio_` prefix on FLAGS because there are many different variables named FLAGS in different modules and I wanted to be clear which I was changing.

tests/benchmarks/fio_benchmark_test.py
noahl: The issue here was the bug I mentioned above where my tests weren't being run. When I fixed that, I saw 3 test failures, as expected (which I then fixed).

CHANGES.md
noahl: Done.

perfkitbenchmarker/benchmarks/fio_benchmark.py
stfeng: Thanks for the follow-up on this. Yes, I'm curious to know the reason...

CHANGES.md
stfeng: Looks like release v0.23.0 might get delayed. Let's see how our CR and testing go today (10/26), we might be able to make it.

As to updates to this CHANGES.md file, it's nice that you are doing it here, but usually this is done by the build master when they cut the build (and decide which changes are in). So you can remove this change from changes.md file and communicate the text to the build master (and he can decide if he wants to include this change in v0.23 or punt it to next release).

stfeng:
LGTM! Thanks for the follow-up on the unit-test issues.

Please coordinate with the current release master regarding the timing of the release and your check-in.

PR #585, opened 2015-10-24
Always use the latest version of the azure package. (If we only use v…

stfeng:
…ersion 1.0.0 we won't be able to pickup fixes and some times substantial perf improvements of the azure python package, e.g., the v1.0.2 has a 10x improvement in 1 byte request latency due to connection reuse)

### DIFF
--- a/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
@@ -898,7 +898,7 @@ def Prepare(benchmark_spec):
 
   vms[0].Install('pip')
   vms[0].RemoteCommand('sudo pip install python-gflags==2.0')
-  vms[0].RemoteCommand('sudo pip install azure==1.0.0')
+  vms[0].RemoteCommand('sudo pip install azure')
   vms[0].Install('gcs_boto_plugin')
 
   OBJECT_STORAGE_BENCHMARK_DICTIONARY[FLAGS.storage].Prepare(vms[0])


stfeng:
@cmccoy @yuyantingzero @voellm 

This is a one-liner but has significant impact in latency numbers measured on Azure. 

The summary of the issue is:

1. PKB (with versions <=0.21.0) used to always grab the latest version of Azure python package.
2. Due to a breaking change of Azure when they released 1.0 in early September, our PKB code was broken (namespace was changed).
3. As part of fixing 2, we decided to just always install azure = v1.0.0 because we know it works.
4. Sure, it works, but after v1.0.0, Azure folks made an important fix in their client side that would reuse HTTP connection across multiple requests, and that fix is available in v1.0.1 and later, and that fix has significant performance benefit.

Here is the CL in Azure python client that made the improvement in their v1.0.1
https://github.com/Azure/azure-storage-python/commit/20d6c240e3ced45f48eb8b5900c8783c4e1f8dc6

So, the situation on the PKB side is now:

1. If you have an old PKB binary (<=v.0.21), that PKB will actually always download the LATEST version of Azure and therefore get GOOD latency numbers. (that's why we have been seeing good numbers in Azure in mid-to-late September!!)

2. However, if you have the latest PKB binary (>=v.0.22) which was shipped in early Oct, that PKB will always download Azure client v1.0.0 which does not have the aforementioned perf fix Azure folks made in their client package.

Therefore, I believe we should always download the latest Azure client package from this point on so we always pickup their latest and greatest fix!

Here is the difference between Azure client v1.0.0 and v1.0.2:

v1.0.0, one byte upload/download latency:
Upload p50: 0.0673, Download p50: 0.0327

v1.0.2, one byte upload/download latency:
Upload p50: 0.005462, Download p50: 0.006987

10x difference!, and if you take a closer look at v1.0.2 logs, you will see message for the initial connnection establishment, and the time to do the first request is about the same as the old binary, but subsequent requests will use the same connection and take much less time:

INFO:root:Storage provider is AZURE, bucket is pkbb4be8938, scenario is OneByteR
W
INFO:urllib3.connectionpool:Starting new HTTPS connection (1): pkbb4be8938.blob.
core.windows.net
INFO:root:Individual results of one byte upload:
INFO:root:0.051287
INFO:root:0.005961
INFO:root:0.005789
INFO:root:0.005494

stfeng:
@cmccoy - I believe this is a p0 must have for the next release (which starts early next week)...

yuyantingzero:
lgtm

PR #588, opened 2015-10-26
Add the 'zones' flag back

ehankland:


voellm:
LGTM... thanks for putting this back.  While "ill-defined" it is a bit more usable for things like networking benchmarks.  For sure it can be abused but that's a tradeoff.

perfkitbenchmarker/pkb.py
```diff
@@ -97,10 +97,13 @@
...
-    'zone', None,
-    'A zone within which to run PerfKitBenchmarker. '
-    'This is specific to the cloud provider you are running on.')
+flags.DEFINE_list(
+    'zones', None,
+    'A list of zones within which to run PerfKitBenchmarker.'
+    ' This is specific to the cloud provider you are running on. '
+    'If multiple zones are given, PerfKitBenchmarker will create 1 VM in '
+    'zone, until enough VMs are created as specified in each '
+    'benchmark.')
```
cmccoy: Consider noting that the order is not defined.

ehankland: Done

Issue #572, opened 2015-10-21
Passing project (or other param) to Network and Disk class when using *VmSpec


syed:
the `project` field is optional in Cloudstack. Currently I am checking the `flags` to get its value but I wanted to add that as a `vm_spec` so that I can run the same benchmark on multiple clouds using the config. However, I don't find any way to pass `project` to the network or disk class apart from the global flags.

syed:
I checked and it seems that I can modify my `Disk` class as that gets created in my `VM` class. However, network is not. Can we add a callback similar to `CreateScratchDisk()` for network?

ehankland:
Yes we can - I'll work on a fix.

ehankland:
This is possible now that the above PR is merged. You need to define a NetworkSpec object and implement  _GetKeyFromNetworkSpec as well as _GetNetworkSpecFromVm. If you need help, let me know.

PR #591, opened 2015-10-27
Allow Networks and Firewalls to take arbitrary arguments

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/572

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -32,14 +32,6 @@ from perfkitbenchmarker import static_virtual_machine as static_vm
 from perfkitbenchmarker import virtual_machine
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker import providers  # NOQA
-from perfkitbenchmarker.providers.alicloud import ali_network
-from perfkitbenchmarker.providers.aws import aws_network
-from perfkitbenchmarker.providers.azure import azure_network
-from perfkitbenchmarker.providers.cloudstack import cloudstack_network as cs_nw
-from perfkitbenchmarker.providers.gcp import gce_network
-from perfkitbenchmarker.providers.openstack import \
-    os_network as openstack_network
-from perfkitbenchmarker.providers.rackspace import rackspace_network as rax_net
 
 
 def PickleLock(lock):
@@ -80,40 +72,6 @@ DEBIAN = 'debian'
 RHEL = 'rhel'
 WINDOWS = 'windows'
 UBUNTU_CONTAINER = 'ubuntu_container'
-VIRTUAL_MACHINE = 'virtual_machine'
-NETWORK = 'network'
-FIREWALL = 'firewall'
-CLASSES = {
-    GCP: {
-        FIREWALL: gce_network.GceFirewall,
-        NETWORK: gce_network.GceNetwork,
-    },
-    AZURE: {
-        NETWORK: azure_network.AzureNetwork,
-        FIREWALL: azure_network.AzureFirewall,
-    },
-    AWS: {
-        FIREWALL: aws_network.AwsFirewall,
-        NETWORK: aws_network.AwsNetwork,
-    },
-    ALICLOUD: {
-        FIREWALL: ali_network.AliFirewall,
-        NETWORK: ali_network.AliNetwork,
-    },
-    DIGITALOCEAN: {
-    },
-    KUBERNETES: {
-    },
-    OPENSTACK: {
-        FIREWALL: openstack_network.OpenStackFirewall
-    },
-    CLOUDSTACK: {
-        NETWORK: cs_nw.CloudStackNetwork
-    },
-    RACKSPACE: {
-        FIREWALL: rax_net.RackspaceSecurityGroup
-    }
-}
 
 FLAGS = flags.FLAGS
 
@@ -154,6 +112,8 @@ class BenchmarkSpec(object):
     self.vms = []
     self.networks = {}
     self.firewalls = {}
+    self.networks_lock = threading.Lock()
+    self.firewalls_lock = threading.Lock()
     self.vm_groups = {}
     self.deleted = False
     self.file_name = os.path.join(vm_util.GetTempDir(), self.uid)
@@ -275,12 +235,14 @@ class BenchmarkSpec(object):
       except Exception:
         logging.exception('Got an exception deleting VMs. '
                           'Attempting to continue tearing down.')
+
     for firewall in self.firewalls.itervalues():
       try:
         firewall.DisallowAllPorts()
       except Exception:
         logging.exception('Got an exception disabling firewalls. '
                           'Attempting to continue tearing down.')
+
     for net in self.networks.itervalues():
       try:
         net.Delete()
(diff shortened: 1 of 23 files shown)


skschneider:
Here's what I see:

Old way:
* There is a big CLASSES dict in benchmark_spec.py
* The benchmark spec's _CreateVirtualMachine calls the provider-specific GetNetwork and GetFirewall functions (this is why CLASSES is needed) with no args and passes the returned instances to the constructor of the VM class.
* GetFirewall and GetNetwork take a dict (owned and passed in by the benchmark spec) and add the new instances to it. The dict maps class to instance, so the benchmark spec only has one firewall or network per provider.

New way:
* The benchmark spec's _CreateVirtualMachine just calls the provider-specific VM class constructor.
* The VM constructor calls GetNetwork and GetFirewall with whatever args it wants as appropriate to the provider's implementation.
* GetFirewall and GetNetwork grab the benchmark spec from the thread context and add the new instances to the benchmark spec's dicts. The dict maps a tuple of class and args to instance, so more than one firewall or network per provider is possible.

My impressions:
* I like that the calls to GetFirewall and GetNetwork have been moved into the provider-specific VM classes. Conceptually, one part of configuring a benchmark run is configuring each VM used by that benchmark, and things like which network a VM belongs to should be part of a VM's config/spec. From the spec alone, the VM should be able to fetch the appropriate network object and construct itself to be in that network.
* I'm not a fan of the way the benchmark spec's dict is implemented. Rather than a class & args tuple for the key, I think it would be better to allow each resource to have a string key/ID/name. The yaml configs could allow configuring networks, similar to how we can currently configure VMs. The config for a VM group or individual VM could then allow picking a particular network. Without configuring any networks explicitly, we could resort to the current behavior of having a single default network per benchmark run.
* The method by which GetFirewall and GetNetwork reach into the benchmark spec to add themselves to the dicts is also a bit weird and conflicts with the way VMs are created. Perhaps create dedicated methods for this (if we want to continue to allow this type of create-upon-first-get behavior), or stick with a model consistent with how we create VM objects: benchmark spec creates them during resource provisioning (but then the VM class can fetch them by the info in the VM spec). I wouldn't expect significant time saving with the delayed-creation approach, since the benchmark run config shouldn't be specifying resources that won't be used, but it would have the benefit of resolving the order in which resources need to be created, provided there are no complicated circular dependencies to consider. Either way, I think it would be good to remove the need to pass variable-shape args directly from the VM constructor to the GetNetwork function. Using a named resource and bundled config object would match what we're currently doing with VMs. BTW, I think this is a good opportunity to think about what other named resources might need to be created and interacted with aside from networks and firewalls - for example, AWS subnets and security groups, which may not have a direct equivalent in other providers.

skschneider:
And BTW, this particular PR doesn't need to solve everything. What you've already written is an improvement in functionality from the old way and doesn't lock us out from future changes. If you'd like to get the new functionality in quickly and consider future steps later, the primary thing to focus on before merging this PR would be if there's a better system for those keys in those firewall and network dicts.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -83,37 +75,6 @@ def UnPickleLock(locked, *args):
 VIRTUAL_MACHINE = 'virtual_machine'
 NETWORK = 'network'
 FIREWALL = 'firewall'
```
skschneider: Some of these other constants may no longer be in use.

perfkitbenchmarker/network.py
```diff
@@ -22,20 +22,41 @@
...
 import threading
 
+from perfkitbenchmarker import context
+from perfkitbenchmarker import errors
+
 
 class BaseFirewall(object):
   """An object representing the Base Firewall."""
 
+
```
skschneider: There's an extra blank line here.

perfkitbenchmarker/network.py
```diff
@@ -22,20 +22,41 @@
...
+    will be returned rather than creating a new one. This enables multiple
+    VMs to call this method and all share the same BaseFirewall object.
+
+    Args:
+      args: The args to pass to the class's constructor.
+      kwargs: The kwargs to pass to the class's constructor.
+    """
+    benchmark_spec = context.GetThreadBenchmarkSpec()
+    if benchmark_spec is None:
+      raise errors.Error('GetFirewall called in a thread without a '
```
skschneider: Perhaps roll this error into a version of GetThreadBenchmarkSpec to be used when one is expected to exist, since that may be a common thing to complain about.

tests/gce_virtual_machine_test.py
```diff
@@ -17,11 +17,20 @@
...
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.providers.gcp import gce_virtual_machine
 
 
 class GCEPreemptibleVMFlagTestCase(unittest.TestCase):
+
+  def setUp(self):
+    # VM Creation depends on there being a BenchmarkSpec.
+    self.spec = benchmark_spec.BenchmarkSpec({}, 'name', 'benchmark_uid')
+    self.addCleanup(context.SetThreadBenchmarkSpec, None)
```
skschneider: This pattern looks asymmetric. Might be out of scope for this PR, but I suggest removing the auto-register from the spec constructor or adding a check in the constructor to make sure there isn't already a registered spec for the thread when the new one is constructed.

perfkitbenchmarker/network.py
```diff
@@ -22,20 +22,38 @@
...
 
+from perfkitbenchmarker import context
+from perfkitbenchmarker import errors
+
 
 class BaseFirewall(object):
   """An object representing the Base Firewall."""
 
   # Lock used to serialize the instantiation of new BaseFirewall objects.
   _class_lock = threading.Lock()
```
skschneider: If these class locks are only used for the dictionaries in the benchmark spec, it would be nice to move them out of this class.

perfkitbenchmarker/network.py
```diff
@@ -51,27 +69,63 @@ def DisallowAllPorts(self):
...
-      # This probably will never happen, but we want to ensure that
-      # networks from different clouds never share the same entry, so
-      # in addition to using the zone, we also use the class as part
-      # of the key.
-      key = (cls, zone)
-      if key not in networks_dict:
-        networks_dict[key] = cls(zone)
-      return networks_dict[key]
+      spec = cls._GetNetworkSpecFromVm(vm)
+      key = cls._GetKeyFromNetworkSpec(spec)
```
skschneider: Can creating the spec & key happen outside the lock?

perfkitbenchmarker/providers/alicloud/ali_network.py
```diff
@@ -138,11 +140,13 @@ def DisallowAllPorts(self):
 class AliNetwork(network.BaseNetwork):
   """Object representing a AliCloud Network."""
 
-  def __init__(self, zone):
-    super(AliNetwork, self).__init__(zone)
+  CLOUD = 'AliCloud'
```
skschneider: Eventually these constants should be combined to avoid duplication.

ehankland: Done

perfkitbenchmarker/network.py
ehankland: Done

perfkitbenchmarker/network.py
ehankland: Done

ehankland:
PTAL

skschneider:
Thanks, LGTM.

PR #592, opened 2015-10-28
Add couter_write/user command to cassandra-stress.

yuyantingzero:


perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -157,19 +191,31 @@ def RunTestOnLoader(vm, loader_index, keys_per_vm, data_node_ips):
...
   """
+  cassandra_stress_command = FLAGS.cassandra_stress_command
+
+  if FLAGS.cassandra_stress_command == USER_COMMAND:
+    cassandra_stress_command += ' profile={profile} ops\({ops}\)'.format(
+        profile=TEMP_PROFILE_PATH,
+        ops=FLAGS.cassandra_stress_ops)
+    schema_option = ''
+  else:
+    schema_option = '-schema replication\(factor={replication_factor}\)'.format(
```
ehankland: It would be nice if we could also specify the replication strategy in addition to the replication factor (so that it would do intelligent things if running in multiple zones), but to take advantage of that, the Cassandra configuration would have to be smarter (i.e. we need to fill out the cassandra-rackdc.properties file based off of VM zones).

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -157,19 +191,31 @@ def RunTestOnLoader(vm, loader_index, keys_per_vm, data_node_ips):
     keys_per_vm: The number of keys per loader vm need to insert.
     data_node_ips: List of IP addresses for all data nodes.
   """
+  cassandra_stress_command = FLAGS.cassandra_stress_command
+
+  if FLAGS.cassandra_stress_command == USER_COMMAND:
```
ehankland: nit: s/FLAGS.cassandra_stress_command/cassandra_stress_command/

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
```diff
@@ -40,15 +41,49 @@
...
+                  ['ONE', 'QUORUM', 'LOCAL_ONE', 'LOCAL_QUORUM',
+                   'EACH_QUORUM', 'ALL', 'ANY'],
+                  'Set the consistency level to use during cassandra-stress.')
+
+flags.DEFINE_integer('cassandra_stress_retries', 1000,
+                     'Number of retries when error encountered during stress.')
+
+# Options to use with cassandra-stress user mode.
+# http://www.datastax.com/dev/blog/improved-cassandra-2-1-stress-tool-benchmark-any-schema
+flags.DEFINE_string('cassandra_stress_profile', '',
```
ehankland: I see you mentioned this in a comment above, but might be nice to explicitly mention that these flags only matter if --cassandra_stress_command is "user" in the flag help text so users using --help wont get confused

ehankland:
LGTM - a few nits and a note that it might be nice to support replication strategy in addition to replication factor - that probably can just get a TODO since it means a fair amount of work

perfkitbenchmarker/benchmarks/cassandra_stress_benchmark.py
yuyantingzero: Done.

Issue #529, opened 2015-10-12
Cleanup phase is not executed when running with --parallelism flag


mateusz-blaszkowski:
Run at least two benchmarks in parallel (using --parallelism flag). Interrupt the execution with Ctrl + C. Cleanup phase is not triggered at all.

voellm:
Skyler - Can you look into this?

skschneider:
Off the bat, I'd suspect that the KeyboardInterrupt is being caught by the main thread's RunThreaded that it uses to launch all the other threads that run benchmarks, whereas when running single-threaded it would be caught by the more general exception handler that makes sure Cleanup is executed if there's any exception during a benchmark's execution. This has probably been the behavior for a long time. Solution would probably be to add functionality to RunThreaded to handle KeyboardInterrupt. I'll look into it later this week.

skschneider:
Ideally, we'd want Ctrl + C to trigger the cleanup behavior in benchmarks and to stop scheduling new benchmarks, so I considered if it was possible to get the main thread to have the benchmark threads raise a KeyboardInterrupt. There does exist an API PyThreadState_SetAsyncExc that could do this, but it looks like it can have unpredictable side effects depending on what the worker thread is doing. Alternatively, we could switch the call of RunThreaded into a RunParallelProcesses. That currently uses a ProcessPoolExecutor, which also reportedly has problems with interrupt signal handling, but there does exist a way to launch worker processes and control interrupt signals: http://noswap.com/blog/python-multiprocessing-keyboardinterrupt.

I would like if someone more experienced in Python weighed in on this one. I don't think there's a convenient model for redesigning the benchmarks themselves to poll for problems, and between the thread-based injected exception solution and process-based signal solution, I'm inclined toward a process approach with a modified RunParallelProcesses.

voellm:
@skschneider  @mateusz-blaszkowski What is the next step here?  This is marked as a P0 bug so I'd like to either downgrade because its not important or work on a fix.

Thanks!

skschneider:
Spoke with cmccoy. We're agreed that replacing the call to RunThreaded with RunParallelProcesses and then enhancing RunParallelProcesses is a good way to go. I'll work on that PR.

skschneider:
Working on this now and almost done with implementation but need some more time for testing everything and then iterating with reviews. May not make it into the current release but will definitely be in by the release after that.

skschneider:
[...]

Issue #573, opened 2015-10-21
Don't stop the run if one of the benchmarks failed.


mateusz-blaszkowski:
When you run multiple benchmarks (i.e. `--benchmarks=mongodb_ycsb,cassandra_ycsb`) and the first one fails, all the subsequent benchmarks will not be run. This is particularly unwanted when we want to run tests at night and it turns out that most of the benchmarks were not executed at all because of the first one failing.

skschneider:
I believe this is also a current discrepancy between the single-threaded and parallel behavior. If you set --parallelism=2, it should run all benchmarks and then raise a single exception at the end that reports failures raised during any of the runs. But with the current default of --parallelism=1, you get the fail-and-stop behavior you mentioned.

Unless there's any reason to preserve the old behavior, I think continuing after a failure should be the new default.

voellm:
I'm ok with continuing but we need to make sure the failed runs are
reported at the end clearly.  Not sure that happened now.

We should also have a flag like --continue_runs_on_failure and default to
true


On Wed, Oct 21, 2015 at 11:02 AM, skschneider <notifications@github.com>
wrote:

> I believe this is also a current discrepancy between the single-threaded
> and parallel behavior. If you set --parallelism=2, it should run all
> benchmarks and then raise a single exception at the end that reports
> failures raised during any of the runs. But with the current default of
> --parallelism=1, you get the fail-and-stop behavior you mentioned.
>
> Unless there's any reason to preserve the old behavior, I think continuing
> after a failure should be the new default.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/573#issuecomment-149979050>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382
<https://www.google.com/voice/b/0?pli=1#phones>
Blog: http://perfguy.blogspot.com/
Benchmark like a pro... PerfKitBenchmarker
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker>

mateusz-blaszkowski:
Having something like failure rate at the end of the run would be great. If I run dozens of benchmarks (either sequentially or in parallel) I would like to know how many benchmarks failed (and what were the reasons of failures).

voellm:
Skyler - assigning to you.  Feel free to ask Noah to do it if you dont have time.

[...]

PR #597, opened 2015-10-29
Use RunParallelProcesses to execute benchmarks, add interrupt support

skschneider:
Addresses #529, #573

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -54,8 +54,11 @@ resources
 """
 
 import collections
+import functools
 import getpass
 import logging
+import multiprocessing.managers
+import signal
 import sys
 import uuid
 
@@ -257,6 +260,9 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks,
     benchmark_config: The config to run the benchmark with.
     benchmark_uid: An identifier unique to this run of the benchmark even
       if the same benchmark is run multiple times with different configs.
+
+  Returns:
+    True.
   """
   benchmark_name = benchmark.BENCHMARK_NAME
 
@@ -326,6 +332,7 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks,
           spec.Delete()
         # Pickle spec to save final resource state.
         spec.PickleSpec()
+  return True
 
 
 def _LogCommandLineFlags():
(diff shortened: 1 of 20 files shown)


perfkitbenchmarker/pkb.py
```diff
@@ -257,6 +260,9 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks,
     benchmark_config: The config to run the benchmark with.
     benchmark_uid: An identifier unique to this run of the benchmark even
       if the same benchmark is run multiple times with different configs.
+
+  Returns:
+    True.
```
cmccoy: Why return anything then?

perfkitbenchmarker/publisher.py
```diff
@@ -487,8 +487,11 @@ class SampleCollector(object):
       SampleCollector._DefaultPublishers.
     run_uri: A unique tag for the run.
   """
-  def __init__(self, metadata_providers=None, publishers=None):
-    self.samples = []
+  def __init__(self, samples=None, metadata_providers=None, publishers=None):
+    if samples is None:
```
cmccoy: nit: `self.samples = samples or []`

skschneider: Changed this to
```
self.samples = [] if samples is None else samples
```
reason being that passing in a Falsey value like an empty list shouldn't create a new list.

perfkitbenchmarker/pkb.py
skschneider: In order to support a request in #573, I added an option to suppress exceptions to RunParallelThreads, in which case the returned list, which would otherwise be a list of return values from the called functions, is filled in with None for functions that failed. There are other ways around it (e.g. defining a new sentinel value to put in the list or changing the shape of the returned object if suppress_exceptions is set). At the moment, this True is here because it's not None.

perfkitbenchmarker/publisher.py
cmccoy: I see. Are you explicitly trying to avoid a copy? Or expecting the
constructor caller to continue modifying their "samples" list?

On Wed, Oct 28, 2015 at 10:31 PM, skschneider <notifications@github.com>
wrote:

> In perfkitbenchmarker/publisher.py
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/597#discussion_r43350524>
> :
>
> > @@ -487,8 +487,11 @@ class SampleCollector(object):
> >        SampleCollector._DefaultPublishers.
> >      run_uri: A unique tag for the run.
> >    """
> > -  def __init__(self, metadata_providers=None, publishers=None):
> > -    self.samples = []
> > +  def __init__(self, samples=None, metadata_providers=None, publishers=None):
> > +    if samples is None:
>
> Changed this to
>
> self.samples = [] if samples is None else samples
>
> reason being that passing in a Falsey value like an empty list shouldn't
> create a new list.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/597/files#r43350524>
> .
>



-- 
Connor McCoy | Software Engineer | connormccoy@google.com | 206-876-1657

skschneider: The very specific reason is because we used to use RunThreaded for benchmarks, passing them all the same SampleCollector and expecting all samples to be aggregated in its samples list. Now that we're using RunParallelProcesses, the samples list in the SampleCollector received by each process needs to refer to the same list. So instead of relying on the default serializing/deserializing behavior, which would give each process a different list (even though originally only a single SampleCollector is passed in), I give the SampleCollector a proxy to a list managed by a SyncManager. That way, the deserialized SampleCollector in every process will append to the same list. Take a look at the modified RunBenchmarks in pkb.py to see how the managers are used.

skschneider:
Looks like there are some static variables in misc modules that need to be managed and kept in sync across processes.  I'm going to make a separate module to host one manager and have those other modules init their values from there.

skschneider:
Fixed the VM instance count collision.

skschneider:
Some of the functionality in this PR has been merged into master via #614, #615, #620. The need to fix issue #529 has been removed with our decision to remove the --parallelism flag (see #616). Closing this PR.

PR #593, opened 2015-10-28
Release version 0.23.0

cmccoy:

Issue #427, opened 2015-08-19
Magento Web workload Benchmark


kivio:
My proposition is to add real life benchmark for web services. This proposition was tested by our team using Magento (PHP based free to use web shop). 

To simulate with you workload which you are interested in this benchmark should share some options:
1.	Number of DB machines
2.	Number of Cache machines (default 0 – without cache)
3.	Number of Web machines (Front-End)
4.	All-in-one option (run all services on one VM)
5.	Number of clients at ones (number of clients machines is counted automatically based of that number)
6.	Shop size (number of products)
7.	Workload time (default 600s)

Technical spec:
1.	Magento
2.	MySQL
3.	Memcached
4.	Apache
5.	Gatling – for traffic simulation

Data are automatically generated on provision machines with DB.
Whole test return us:
1.	Request per second
2.	Median time of request
3.	Distribution characteristic

voellm:
I like what you are proposing and its important to have a solid web workload before we delare v1.0 on PKB :)

Right now EPFL has a pull request for a websearch benchmark here https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/422

Can you check this out an see if it matches your needs?

voellm:
@kivio - Is this something you are going to work on or should I try and find someone?

kivio:
i try to do something in this subject but in very limited time

kozyraki:
@kivio Sounds great. Is there a dataset and a query trace or generator to use with Magento?

kivio:
We have it done a little bit manually but it's simple to automize. But my biggest problem currently is lack of time to do it.

PR #598, opened 2015-10-29
Tomcat/wrk tests.

cmccoy:
This is a continuation of #593. Using `nginx` stressed the client more than the server.
To get a (relatively) simple web-serving workload, I've switched to Tomcat on the server side.
The server eventually becomes CPU-bound.

There is also some overlap with #427 and #451, both of which simulate a more realistic, layered web service deployment. This aims to be more of a microbenchmark: single client, single server, no database, with a small amount of server-side work per request (generating an SVG with rotated text).

We could also omit this in favor of #427 or #451 if there's too much overlap.
Comments appreciated (cc: @kivio, @rmend016).

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,9 @@ In its current release these are the benchmarks that are executed:
   - `scimark2`: public domain (http://math.nist.gov/scimark2/credits.html)
   - `speccpu2006`: Spec CPU2006 (http://www.spec.org/cpu2006/)
   - `sysbench_oltp`: GPL v2 (https://github.com/akopytov/sysbench)
+  * `tomcat`: Apache v2. (http://tomcat.apache.org)
   - `unixbench`: GPL v2 (https://code.google.com/p/byte-unixbench/)
+  * `wrk`: Apache v2 (https://github.com/wg/wrk)
   - `ycsb` (used by `mongodb`, `hbase_ycsb`, and others): Apache V2 (https://github.com/brianfrankcooper/YCSB/blob/master/LICENSE.txt)
 
 Some of the benchmarks invoked require Java. You must also agree with the following license:
--- a/perfkitbenchmarker/benchmark_sets.py
+++ b/perfkitbenchmarker/benchmark_sets.py
@@ -78,7 +78,7 @@ BENCHMARK_SETS = {
     'google_set': {
         MESSAGE: ('This benchmark set is maintained by Google Cloud Platform '
                   'Performance Team.'),
-        BENCHMARK_LIST: [STANDARD_SET, 'oldisim']
+        BENCHMARK_LIST: [STANDARD_SET, 'oldisim', 'tomcat_wrk']
     },
     'intel_set': {
         MESSAGE: 'Intel benchmark set.',
--- a/perfkitbenchmarker/data/wrk_latency.lua
+++ b/perfkitbenchmarker/data/wrk_latency.lua
@@ -1,0 +1,36 @@
+-- Copyright 2015 Google Inc. All rights reserved.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Write stats to stdout in CSV format, for simpler parsing by PKB.
+function done(summary, latency, requests)
+        io.write("==CSV==\n")
+        io.write("variable,value,unit\n")
+        for _, p in pairs({ 5, 25, 50, 75, 90, 99, 99.9 }) do
+                n = latency:percentile(p)
+                io.write(string.format("p%g latency,%g,ms\n", p, n / 1000))
+        end
+        local errors = summary.errors.connect + summary.errors.read +
+            summary.errors.write + summary.errors.timeout
+        local data = {
+                {'bytes transferred', summary.bytes, 'bytes'},
+                {'errors', errors, 'n'},
+                {'requests', summary.requests, 'n'},
+                {'throughput',
+                 summary.requests / summary.duration * 1e6,
+                 'requests/sec'}
+        }
+        for _, p in pairs(data) do
+                io.write(string.format("%s,%g,%s\n", p[1], p[2], p[3]))
+        end
+end
--- a/perfkitbenchmarker/linux_benchmarks/tomcat_wrk_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/tomcat_wrk_benchmark.py
@@ -1,0 +1,193 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Run wrk against a simple Tomcat web server.
+
+This is close to HTTP-RR:
+
+  * Connections are reused.
+  * The server does very little work.
+
+Doubles connections up to a fixed count, reports single connection latency and
+maximum error-free throughput.
+
+`wrk` is a scalable web load generator.
+`tomcat` is a popular Java web server.
+"""
+
+import functools
+import logging
+import operator
+import urlparse
+
+from perfkitbenchmarker import configs
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import vm_util
+from perfkitbenchmarker.linux_packages import tomcat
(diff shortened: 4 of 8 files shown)


cmccoy:
So far the 90th percentile seems like it has a nice performance cliff:

tomcat_wrk_n1-highcpu-8

And the server (right) has very high CPU usage:
dstat_cpu

meteorfox:
@cmccoy Nice work. Is that R's ggplot2 library you used for the charts? I always like how ggplot charts look for some reason, probably is the colors.

Anyway, I was thinking about this, I still think that having a cutoff of 10x increase of 90%ile might be too sensitive to noise. Say for example, a network hiccup, a GC event, and/or some JIT optimization/de-optimization kicks in, that can easily inflate your 90%ile. I personally would want to see this effects. There could be a periodic issue that might slow down the requests, periodically, but then recover. I want to be able to see that, and not have the benchmark exit immediately the 90%ile increase by some factor.

I think the complexity comes from trying to manipulate multiple variables at the same time. IMHO, we should focus on one metric, especially that we want this benchmark to be simple.

We could measure max throughput, without conditioning on latency. The client would go as fast possible to complete N requests with M connections. At each iteration M is increased, and the test is repeated. The benchmark finishes after either MAX_CONNECTIONS, or some time limit is reached. Latency will still be reported for each iteration, but it will not abort the test. This case is similar to a CPU throughput benchmark, which just measures how long it takes to complete a number of instructions, but it ignores the 'latency' to retire each instruction that might be cause by other processes/threads running on the same CPUs. 

Alternatively, if latency is the metric we care about, then we could fix the request rate, execute the test for a given duration (i.e. 100 rps * 1000 seconds = 100K reqs), and after each iteration increase the request rate, until we hit some max request rate limit. For this case, I'd also suggest fixing the number of connections to a scale factor of the number of threads available.

wrk might not have all the capabilities to do what I suggest above, these tools could be use to achieve those workloads.

[wrk2](https://github.com/giltene/wrk2) - is a modified version of wrk, corrected for coordinated ommission, let's you control request rate as well.

[Gatling](http://gatling.io) - Scala, Netty, Akka based tool, supports open- and closed-loop workload generation, as well as stochastic request injection (exponential, uniform, etc). Allows you to control number of requests, and request rates as well.

@cmccoy @voellm What do you think?

voellm:
This is good feedback and I agree for a simpler benchmark one variable is
enough.  In this cause your suggestion of ramping up connections to a fixed
maximum is a good one.

On Thu, Oct 29, 2015 at 3:43 PM, Carlos Torres <notifications@github.com>
wrote:

> @cmccoy <https://github.com/cmccoy> Nice work. Is that R's ggplot2
> library you used for the charts? I always like how ggplot charts look for
> some reason, probably is the colors.
>
> Anyway, I was thinking about this, I still think that having a cutoff of
> 10x increase of 90%ile might be too sensitive to noise. Say for example, a
> network hiccup, a GC event, and/or some JIT optimization/de-optimization
> kicks in, that can easily inflate your 90%ile. I personally would want to
> see this effects. There could be a periodic issue that might slow down the
> requests, periodically, but then recover. I want to be able to see that,
> and not have the benchmark exit immediately the 90%ile increase by some
> factor.
>
> I think the complexity comes from trying to manipulate multiple variables
> at the same time. IMHO, we should focus on one metric, especially that we
> want this benchmark to be simple.
>
> We could measure max throughput, without conditioning on latency. The
> client would go as fast possible to complete N requests with M connections.
> At each iteration M is increased, and the test is repeated. The benchmark
> finishes after either MAX_CONNECTIONS, or some time limit is reached.
> Latency will still be reported for each iteration, but it will not abort
> the test. This case is similar to a CPU throughput benchmark, which just
> measures how long it takes to complete a number of instructions, but it
> ignores the 'latency' to retire each instruction that might be cause by
> other processes/threads running on the same CPUs.
>
> Alternatively, if latency is the metric we care about, then we could fix
> the request rate, execute the test for a given duration (i.e. 100 rps *
> 1000 seconds = 100K reqs), and after each iteration increase the request
> rate, until we hit some max request rate limit. For this case, I'd also
> suggest fixing the number of connections to a scale factor of the number of
> threads available.
>
> wrk might not have all the capabilities to do what I suggest above, these
> tools could be use to achieve those workloads.
>
> wrk2 <https://github.com/giltene/wrk2> - is a modified version of wrk,
> corrected for coordinated ommission, let's you control request rate as well.
>
> Gatling <http://gatling.io> - Scala, Netty, Akka based tool, supports
> open- and closed-loop workload generation, as well as stochastic request
> injection (exponential, uniform, etc). Allows you to control number of
> requests, and request rates as well.
>
> @cmccoy <https://github.com/cmccoy> @voellm <https://github.com/voellm>
> What do you think?
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/598#issuecomment-152348386>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

cmccoy:
@meteorfox, thanks for the comments.

> @cmccoy Nice work. Is that R's ggplot2 library you used for the charts? I always like how ggplot charts look for some reason, probably is the colors.

Exactly! Amazing tool for data exploration.

Good points.

> Alternatively, if latency is the metric we care about, then we could fix the request rate, execute the test for a given duration (i.e. 100 rps * 1000 seconds = 100K reqs), and after each iteration increase the request rate, until we hit some max request rate limit. For this case, I'd also suggest fixing the number of connections to a scale factor of the number of threads available.

My main focus is latency, but I'd like to test scalability with VM size to some extent. This could be done by fixing the number of connections to a scale of the number of cores, as you suggest. I'm going to check in on Gatling - it looks pretty powerful.

cmccoy:
OK, sorry for the delay here. I simplified the test: it now reports single connection latency and maximum throughput up to a fixed number of concurrent connections.
@meteorfox: I looked at Gatling, but ended up sticking with WRK here, as Gatling was a bit more complicated to run and only reports to millisecond resolution. Aiming for simple here, something closer to netperf TCP/RR than real traffic.

perfkitbenchmarker/benchmarks/tomcat_wrk_benchmark.py
```diff
@@ -0,0 +1,181 @@
...
+  description: Run wrk against tomcat.
+  vm_groups:
+    server:
+      vm_spec: *default_single_core
+    client:
+      vm_spec: *default_single_core
+"""
+
+# Stop when >= 1% of requests have errors
+MAX_ERROR_RATE = 0.01
```
meteorfox: `MAX_ERROR_RATE` is never used

meteorfox:
@cmccoy Yes, that's true. Gatling collect timestamps at milliseconds resolution, and it's definitely more complicated to execute than wrk. Gatling shines at expressing more complex workloads that might include data dependency between requests/responses, and wrk shines at hitting one URL *very* fast.

The PR looks good to me. :+1: 

It'd be awesome if you could post the same data, with graphs, for the new changes just for comparison.

perfkitbenchmarker/benchmarks/tomcat_wrk_benchmark.py
cmccoy: I just retried with/without this, and got more consistent results allowing up to 1% errors. Re-added the error-rate check.

cmccoy:
Thanks for helping on this @meteorfox.

Here's raw data for 5 replicates

tomcat_wrk_raw

And coefficient of variation for the same.
tomcat_wrk_cov

 Throughput and p50 latency are very consistent, tail latency quite a bit less so.

cmccoy:
There were some conflicts with #640. Rebased and fixed.

noahl:
I talked with Connor. Seeing as we have an LGTM, I am going to merge this and pull it into release 0.24.0.

meteorfox:
@cmccoy Thanks for generating these. Do you know why the throughput was smaller compared to the previous run?

Also, I think you mentioned it, but I was in the impression the benchmark reported these metrics at every step the connections were increased, but it looks like it only reports latency for the single connection case and the max throughput case. Do you think we should extend it to report these for all the intermediate steps? if not, maybe the latency at the max throughput case?

On a separate note:

It'd be cool to extend this benchmark (or wrk) to print a CSV of all the requests/responses sorted by timestamp, or maybe it could be a separate benchmark.

I'm not sure if it's possible in wrk, but if we could collect timestamps for these events:
  * first byte sent from the request (wrk takes timestamp here already)
  * last byte sent from the request
  * first byte received from the response
  * last byte received from the response (wrk takes timestamp here already)

Then with these we can create time series of both latency and throughput, separate request rate from response rate, and generate the aggregated statistics for the whole run as well. I find that plotting throughput or latency over time is a good way to study the performance consistency.

I think right now we could get all the e2e response times, and the Avg rates over time using that same `done()` Lua function in the script, but since they don't have timestamps we won't be able to generate the time series. It is still useful for generating histograms, though.

I'm going to add a TODO for me on this one, it seems I could get this data from wrk with a few modifications. 

I would probably need to capture the extra time stamps from here:
  * for last byte of the request, I think here should be good enough https://github.com/wg/wrk/blob/master/src/wrk.c#L411-L414
  * for when the first byte of response is received probably here but I'd need a variable to only measure the first time it enters: https://github.com/wg/wrk/blob/master/src/wrk.c#L427-L433

Then I'd need to extend the stats struct to store timestamps for each request.

cmccoy:
@meteorfox:

It did, but after your comments, I opted to simplify and report latency with low load, and maximum throughput. I'd be happy to add the per-connection-count throughput/latency back, but maybe under a flag to keep the results brief by default?

Re: more detailed stats: SGTM. I know that Gatling collects first/last byte sent/received, but has the millisecond resolution issue.

meteorfox:
@cmccoy ah, I see. What do you think if we add the throughput metric for the single connection case, and the latency metrics for the max throughput case, then with a flag report all the intermediate results?

IMHO, it just seems incomplete that we report the single connection latency results without the throughput result, and the max throughput case without the latency results. I'm not sure if others feel the same.

cmccoy:
@meteorfox: that makes sense to me. Opened #650.

PR #599, opened 2015-10-29
Disk types refactor

noahl:
Rename flags scratch_disk_type to disk_type, scratch_disk_size to
disk_size, and scratch_disk_iops to aws_provisioned_iops. Also change
the possible disk types from standard, remote_ssd, piops, and local to
ephemeral_hdd, ephemeral_ssd, building_replicated_hdd, and
building_replicated_ssd.


noahl:
This is the disk types refactor that was discussed at the last community meeting. The design doc was https://docs.google.com/document/d/10jWz9vSyVWIwBFwUP1mVpr-idCZl05TFt3VPjMcngy4 .

A few notes:
1. I updated the GCP, AWS, and Azure providers. Most of the other providers don't offer distinct disk types, but Rackspace still needs to be updated. @meteorfox, can you help with this? Ideally it would be done by adding a rackspace_disk_integration_test, following the model of the gcp, aws, and azure integration tests I made.
2. The design doc didn't consider what happens to the samples we record, which contain 'scratch_disk_type' and 'scratch_disk_size' as labels. The solution I implemented was to keep the label names the same, but change the values to correspond to the new disk type names. I also added new keys 'disk_type' and 'disk_size', with the same information. My tentative plan is to remove the 'scratch_disk_*' keys at the same time as we remove the deprecated flags with those names. I would specifically like feedback on this proposal, because it was not part of the design.

@voellm @stfeng @yuyantingzero @meteorfox Please take a look.

perfkitbenchmarker/benchmarks/aerospike_ycsb_benchmark.py
```diff
@@ -128,8 +128,10 @@ def Run(benchmark_spec):
 
   metadata = {'ycsb_client_vms': FLAGS.ycsb_client_vms,
               'num_vms': len(aerospike_vms),
-              'scratch_disk_type': FLAGS.scratch_disk_type,
-              'scratch_disk_size': FLAGS.scratch_disk_size}
+              'scratch_disk_type': FLAGS.disk_type,
```
ehankland: Since the disk type and size can be specified by configs, FLAGS.disk_type may not actually be the disk type used by the benchmark (same for size etc).

noahl:
Another note: I have updated various benchmarks (Aerospike, Aerospike YCSB, Cassandra YCSB, and HBase YCSB), but I have not tested them. I would appreciate it if someone who is familiar with those benchmarks would take a look at the changes.

perfkitbenchmarker/aws/aws_disk.py
```diff
@@ -34,9 +34,8 @@
 VOLUME_DELETED_STATUSES = frozenset(['deleting', 'deleted'])
 VOLUME_KNOWN_STATUSES = VOLUME_EXISTS_STATUSES | VOLUME_DELETED_STATUSES
 DISK_TYPE = {
-    disk.STANDARD: 'standard',
-    disk.REMOTE_SSD: 'gp2',
-    disk.PIOPS: 'io1'
```
stfeng: You still need this constant because 'io1' is referred by your new code.

perfkitbenchmarker/aws/aws_disk.py
```diff
@@ -56,7 +55,7 @@ def __init__(self, iops=None, **kwargs):
   def ApplyFlags(self, flags):
     """Apply flags to the DiskSpec."""
     super(AwsDiskSpec, self).ApplyFlags(flags)
-    self.iops = flags.scratch_disk_iops or self.iops
+    self.iops = flags.aws_provisioned_iops or self.iops
```
stfeng: Please retain existing flags at the first version of this work. 

We should only deprecate existing flags after new flags are checked-in and become well-understood and usable by the users.

meteorfox:
@noahl thanks for working on these. I'll take a look. I apologize I haven't been as engaged with this refactor, but I have been under the gun with a release.

stfeng:
@noahl, three high level comments:

1. As we discussed a couple of days ago, we should retain existing flags at the first iteration of this work, add support for new flags, and print out "deprecation warning" for existing flags. Once the new flags (and associated functionality) are made well-known and proven working for a while, we can then officially remove existing flags.

2. The new disk types will need to be plumbed into the config framework too.

3. Please complete at least one round of complete testing, including all the benchmarks you changed here, before proceeding to the code review step. (Also, it'd be nice if you could post testing logs to a gist repo and linked that to this PR)

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,100 @@
...
+    logging.warning('Disk type name %s is deprecated and will be removed. '
+                    'Translating to %s for now.', name, new_name)
+    return new_name
+  else:
+    raise ValueError('%s is not a disk type name', name)
+
+
+def WarnAndCopyFlag(old_name, new_name, translator=None):
+  """Copy a value from an old flag to a new one, warning the user.
+  """
```
yuyantingzero: Document this method

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,100 @@
...
+      # Mark the new flag as present so we'll print it out in our list
+      # of flag values.
+      FLAGS[new_name].present = True
+    # Remove the old flag so we can't accidentally use it.
+    del FLAGS.FlagDict()[old_name]
+
+
+def WarnAndTranslateDiskFlags():
+  """Translate old disk-related flags to new disk-related flags.
+  """
```
yuyantingzero: Documentatiton

noahl:
We at Google have talked about this internally, and I've come up with a modified proposed design. Anyone who is interested, please comment on this doc and let's discuss: https://docs.google.com/document/d/1ruXqIEg_QAGAD9kpH5rskFVltLtM6gZLqAR67UQg9Es/edit?usp=sharing

noahl:
I just pushed a rework of this refactor based on the v3 design doc.

My testing strategy has two parts. In automatic testing,
  1. Disk integration tests make sure we can still provision and use disks
  2. The name translation test verifies that the old->new disk type name mapping works
  3. The publish test verifies that the new metadata gets stored
  4. The disk metadata test verifies that disks are annotated with the correct metadata keys.

I used manual testing to verify that the old flags and the names passed to them get translated (as opposed to only translating configs) and that we can actually still run benchmarks. There were two test matrices:
  1. I tested GCP, AWS, and Azure running the fio benchmark, because it's fast and touches the disk. I inspected the results to make sure the metadata looked right. I did one run with the old-style flags and type names on each provider, and tested each disk type they supported with new-style flags and type names.
  2. I tested the benchmarks I modified (Aerospike, Aerospike YCSB, Cassandra YCSB, and hBase YCSB) on a GCP VM just to make sure they still worked after the modifications.

My code differs from the design doc in these ways (that I know of):
  1. I made a special disk type called "local", declared in disk.py and shared across all providers, because the Aerospike benchmarks need it.
  2. Instead of implementing a global mapping from providers and disk types to metadata, I had each provider's disk class annotate itself with the proper metadata. This seemed like an easier way to accomplish the same thing.
  3. After talking with Stanley and Eric about the boot_disk_size flag, we decided to postpone the implementation so we could talk about it more. (Questions: how do boot disks work with configs? How do we say which disk is boot in the metadata?)

@stfeng @yuyantingzero @ehankland please take a look.

perfkitbenchmarker/azure/azure_disk.py
```diff
@@ -44,7 +44,33 @@
...
+AZURE_REPLICATION_MAP = {
+    azure_network.LRS: disk.ZONE,
+    azure_network.ZRS: disk.REGION,
+    # Deliberately omitting PLRS, because that is handled by
+    # PREMIUM_STORAGE_METADATA, and (RA)GRS, because those are
+    # asynchronously replicated.
+}
+
+LOCAL_SSD_PREFIXES = {
+    'Standard_DS',
```
yuyantingzero: D series also has local ssds.
@skschneider , please confirm.

noahl:
I just pushed an update that rebases on the providers refactor. I tested it by running one benchmark on each of the three providers I changed, plus making sure all unit tests pass.

I also ran a test on DigitalOcean, a provider I have *not* updated, to verify that it still works. (It does.) If a provider has not been updated, the behavior is that:
  1. The old commands with the old flags and old disk names still work like they did.
  2. All of the old metadata will still be published. Some of the new metadata keys will appear as well, but not `media' and `replication', because those depend on the provider annotating their disk objects with metadata.

perfkitbenchmarker/azure/azure_disk.py
skschneider: Yes. I'll add a comment on the latest version of this file.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -44,7 +44,33 @@
...
+
+AZURE_REPLICATION_MAP = {
+    azure_network.LRS: disk.ZONE,
+    azure_network.ZRS: disk.REGION,
+    # Deliberately omitting PLRS, because that is handled by
+    # PREMIUM_STORAGE_METADATA, and (RA)GRS, because those are
+    # asynchronously replicated.
+}
+
+LOCAL_SSD_PREFIXES = {
```
skschneider: References for Azure machine types:
https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-size-specs/
https://azure.microsoft.com/en-us/pricing/details/virtual-machines/#Linux
All D, DS, D v2, G, and GS machines have local SSD. Only the Basic A and Standard A series have HDD..

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -229,7 +229,19 @@ def ConstructVirtualMachines(self):
 
         if DISK_SPEC in group_spec:
           disk_spec_class = disk.GetDiskSpecClass(cloud)
-          disk_spec = disk_spec_class(**group_spec[DISK_SPEC][cloud])
+          disk_spec_dict = group_spec[DISK_SPEC][cloud]
+          # We have to translate the data_disk_type flag here because
+          # this is where we know for sure what cloud we're using, and
+          # therefore we're able to pick the right translation table.
+          if FLAGS['data_disk_type'].present:
```
stfeng: I am seeing a bunch of references to the literal string 'data_disk_type' here, I think it'd be more robust and useful if we can define a constant representing the literal string and use the constant here. (and preferrably the constant should be defined right at the place where the flag is defined).

Or is it possible to do FLAGS.data_disk_type.GetFlagName() sort of thing here?

ehankland: A simpler flow for this would be doing this check after the flags get applied (on line 245).
You would only need one line (and wouldn't need to do all this flag presence checking):
disk_spec.disk_type = disk.WarnAndTranslateDiskTypes(disk_spec.disk_type, cloud)

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -229,7 +229,19 @@ def ConstructVirtualMachines(self):
 
         if DISK_SPEC in group_spec:
           disk_spec_class = disk.GetDiskSpecClass(cloud)
-          disk_spec = disk_spec_class(**group_spec[DISK_SPEC][cloud])
+          disk_spec_dict = group_spec[DISK_SPEC][cloud]
+          # We have to translate the data_disk_type flag here because
```
stfeng: Add a note here to indicate that data_disk_type flag may contain legacy values that were copied from the scratch_disk_type flag at the beginning of the run.

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,45 @@
...
+
+
+# These are the (deprecated) old disk type names
 STANDARD = 'standard'
 REMOTE_SSD = 'remote_ssd'
 PIOPS = 'piops'  # Provisioned IOPS (SSD) in AWS
+
+# 'local' refers to disks that come attached to VMs. It is the only
+# "universal" disk type that is not associated with a provider. It
+# exists because we can provision a local disk without creating a disk
```
stfeng: would be nice to enforce the requirement that a disk spec is needed to provision any disk! then we can get rid off this special "local" thing. 

Not in this CL, but would be good to file a issue to track this.

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,45 @@
...
+REPLICATION = 'replication'
+# And some possible values
+HDD = 'hdd'
+SSD = 'ssd'
+NONE = 'none'
+ZONE = 'zone'
+REGION = 'region'
+
+
+def RegisterDiskTypeMap(provider_name, map):
```
stfeng: add some comments to this function.

perfkitbenchmarker/disk.py
```diff
@@ -45,6 +76,64 @@ def __init__(cls, name, bases, dct):
...
+    disk_type_map = DISK_TYPE_MAPS[cloud]
+    if name in disk_type_map:
+      new_name = disk_type_map[name]
+      logging.warning('Disk type name %s is deprecated and will be removed. '
+                      'Translating to %s for now.', name, new_name)
+      return new_name
+    else:
+      return name
+  else:
+    logging.warning('No disk type map for provider %s', cloud)
```
stfeng: Not necessarily a warning... for a new provider that is added after your code, they don't have to have this map, right? ;)

Logging.info() should do, IMO. Also, change the message to indicate it's a disk type map for legacy to new flag mapping.

perfkitbenchmarker/disk.py
```diff
@@ -45,6 +76,64 @@ def __init__(cls, name, bases, dct):
...
+  if FLAGS[old_name].present:
+    logging.warning('Flag --%s is deprecated and will be removed. Please '
+                    'switch to --%s.' % (old_name, new_name))
+    if not FLAGS[new_name].present:
+      FLAGS[new_name].value = FLAGS[old_name].value
+
+      # Mark the new flag as present so we'll print it out in our list
+      # of flag values.
+      FLAGS[new_name].present = True
+  # We keep the old flag around so that providers that haven't been
```
stfeng: If both the new flag and old flag exist, consider print out a warning here that the old flags are ignored.

perfkitbenchmarker/disk.py
```diff
@@ -73,8 +162,8 @@ def __init__(self, disk_size=None, disk_type=None,
 
   def ApplyFlags(self, flags):
     """Applies flags to the DiskSpec."""
-    self.disk_size = flags.scratch_disk_size or self.disk_size
-    self.disk_type = flags.scratch_disk_type or self.disk_type
+    self.disk_size = flags.data_disk_size or self.disk_size
+    self.disk_type = flags.data_disk_type or self.disk_type
     self.num_striped_disks = flags.num_striped_disks or self.num_striped_disks
```
stfeng: Let's add a comment in the description of the flag --num_striped_disks to emphasize this is: 1) for striping DATA disks, 2) all stripped data disk will show up as one single logical data disk and is indicated as data_disk_0 in the metadata.

Ideally I'd like to rename the flag to --num_striped_data_disks but I am ok if this is out of scope for this CL.

perfkitbenchmarker/pkb.py
```diff
@@ -143,8 +141,16 @@
     'scratch_disk_type', None,
     [disk.STANDARD, disk.REMOTE_SSD, disk.PIOPS, disk.LOCAL],
     'Type for all scratch disks. The default is standard')
+flags.DEFINE_string(
+    'data_disk_type', None,
+    'Type for all benchmark disks.')
```
stfeng: Define "data_disk" more crispy here, (or even consider provide a link here as code comment pointing to your design doc which should be uploaded as a wiki page on PKB). 

Also mention how we are going to handle the case where the provider does not support multiple disks beyond the root disk.

perfkitbenchmarker/pkb.py
```diff
@@ -143,8 +141,16 @@
...
+flags.DEFINE_integer('scratch_disk_size', None, 'Size, in gb, for all scratch '
+                     'disks.')
+flags.DEFINE_integer('data_disk_size', None, 'Size, in gb, for all disks.')
 flags.DEFINE_integer('scratch_disk_iops', None,
                      'IOPS for Provisioned IOPS (SSD) volumes in AWS.')
+flags.DEFINE_integer('aws_provisioned_iops', None,
+                     'IOPS for Provisioned IOPS (SSD) volumes in AWS.')
 flags.DEFINE_integer('num_striped_disks', None,
                      'The number of disks to stripe together to form one '
                      '"logical" scratch disk. This defaults to 1 '
```
stfeng: s/scratch disk/data disk/

perfkitbenchmarker/providers/aws/aws_disk.py
```diff
@@ -33,12 +33,49 @@
...
+
+STANDARD = 'standard'
+GP2 = 'gp2'
+IO1 = 'io1'
+
 DISK_TYPE = {
-    disk.STANDARD: 'standard',
-    disk.REMOTE_SSD: 'gp2',
-    disk.PIOPS: 'io1'
+    # no need to translate 'standard' because the old and new names
```
stfeng: I don't think this map is defined for translation purpose, right? It's used to ensure the disk type specified makes sense for AWS. Perhaps it's still better that we put disk.STANDARD here so 'standard' is explicitly accepted as a good terminology for AWS disks?

perfkitbenchmarker/providers/aws/aws_disk.py
```diff
@@ -76,6 +113,14 @@ def __init__(self, disk_spec, zone):
     self.device_letter = None
     self.attached_vm_id = None
 
+    if self.disk_type != disk.LOCAL:
+      self.metadata = DISK_METADATA[self.disk_type]
+    else:
+      if machine_type[:2].lower() in LOCAL_HDD_PREFIXES:
```
stfeng: please make this if check into a function and define the function right at the place where LOCAL_HDD_PREFIXES are defined.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -53,18 +79,36 @@ class AzureDisk(disk.BaseDisk):
...
 
+    if self.disk_type == PREMIUM_STORAGE:
+      self.metadata = PREMIUM_STORAGE_METADATA
+    elif self.disk_type == STANDARD_DISK:
+      self.metadata = {
+          disk.MEDIA: disk.HDD,
+          disk.REPLICATION: AZURE_REPLICATION_MAP[FLAGS.azure_storage_type]
+      }
+    elif self.disk_type == disk.LOCAL:
+      if any((machine_type.startswith(prefix)
```
stfeng: please make this into a utility routine (isDiskLocal())

perfkitbenchmarker/publisher.py
```diff
@@ -144,12 +143,24 @@ def AddMetadata(self, metadata, benchmark_spec):
...
-        if data_disk.disk_type == disk.PIOPS:
+        if getattr(data_disk, 'iops', None) is not None:
           metadata[name_prefix + 'scratch_disk_iops'] = data_disk.iops
+        # Modern metadata keys
+        metadata[name_prefix + 'data_disk_0_type'] = data_disk.disk_type
+        metadata[name_prefix + 'data_disk_0_size'] = (
+            data_disk.disk_size * data_disk.num_striped_disks)
+        metadata[name_prefix + 'data_disk_0_num_stripes'] = (
+            data_disk.num_striped_disks)
+        if getattr(data_disk, 'iops', None) is not None:
```
stfeng: consider merge this if statement to the identical one just a few lines above....

stfeng:
I finished my first round of CR. Mostly minor comments. 

Super nice job on the testing front, thanks!

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: Fixed, thanks!

perfkitbenchmarker/benchmark_spec.py
noahl: @stfeng As far as I know, all you can do with FLAGS.data_disk_type is get its value. For anything else, you have to use FLAGS['data_disk_type'].

@ehankland I didn't do that because I didn't want to leave around an untranslated value in FLAGS.data_disk_type in case someone used it. Your comment made me realize that nobody uses it because it's brand new, and they shouldn't use it in any case. I will use your version.

This also resolves the issue of using `'data_disk_type'` instead of making a constant.

perfkitbenchmarker/benchmark_spec.py
noahl: Added the note.

perfkitbenchmarker/disk.py
noahl: Filed #631 to track this.

perfkitbenchmarker/disk.py
noahl: Added docstring explaining what it's used for.

perfkitbenchmarker/disk.py
noahl: Yes, you're right! Changed to logging.info, and changed the message to indicate what the mapping is for.

perfkitbenchmarker/disk.py
noahl: Added a logging.warning().

perfkitbenchmarker/disk.py
noahl: I added a comment covering your points 1 and 2. I would be happy to rename this flag -- I already have the flag-renaming machinery, so it's very easy to rename more flags -- but it wasn't in the design doc, so I think we'd need to have another round of discussion for this. What would you like to do?

perfkitbenchmarker/pkb.py
noahl: I added the info in the help text here. I think a comment would make the most sense in disk.py, since that has information on disks. I can add one once I upload to the wiki.

perfkitbenchmarker/pkb.py
noahl: Done.

perfkitbenchmarker/providers/aws/aws_disk.py
noahl: The DISK_TYPE map was used for that in the Azure provider (it will not be after this PR is applied), but in AWS this map has only been used to translate disk type names.

In the new system, the DISK_METADATA map is the one that checks the validity of the disk type, because you have to look up the disk name in the DISK_METADATA map in order to create a disk spec. I wasn't really trying to set up a validity check, though, so there are cases you can get a bad disk type past it. I thought it would make more sense to leave the checking up to the provider CLIs. If you want, I can implement a real validity check for disk type names.

perfkitbenchmarker/providers/aws/aws_disk.py
noahl: Done.

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: Done. I called it `SdbIsSSD`, because I think it makes sense here to keep the `disk.LOCAL` check where it is and move the check about what kind of disk it is to where LOCAL_SSD_PREFIXES is defined.

perfkitbenchmarker/publisher.py
noahl: I duplicated the statements so I could keep the legacy and modern metadata handling completely separate. I think the duplication is worth it to make it very clear what is legacy, and also to make it very easy to remove the legacy code once we're ready to do that. If you still disagree, I'll merge the if statements, but that's why I did this.

perfkitbenchmarker/providers/azure/azure_disk.py
ehankland: You may want to consider detecting the disk type at run time instead of depending on these maps which may be prone to breaking over time:
http://unix.stackexchange.com/questions/65595/how-to-know-if-a-disk-is-an-ssd-or-an-hdd

perfkitbenchmarker/providers/aws/aws_disk.py
stfeng: As discussed, let's retain the default behavior of provisioning a "standard" scratch disk. i.e., ./pkb.py --benchmarks=xyz --scratch_disk_size=100G will provision a 100GB data disk for the user with a default "standard" disk type without having to ask the user to figure out what exactly the "standard" disk type string is for the provider they want to run on.

perfkitbenchmarker/providers/azure/azure_disk.py
stfeng: naming: IsLocalDiskSSD()

perfkitbenchmarker/providers/aws/aws_disk.py
noahl: I just talked about this with Eric. I misunderstood how defaults work. The *benchmark* is what determines whether or not we create a disk spec, not the flags. All the flags do is modify properties of the benchmark's config.

Right now, you can create disks without passing any arguments if the benchmark wants them. But there will be a warning, because the disk types are specified using the old names. I will update the file `default_config_constants.yaml` to fix this.

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: Changed the name of the function for now.

Auto-detecting would be nice, but is a bigger issue than this PR. I filed #634 to track it.

noahl:
@stfeng @ehankland @yuyantingzero  I just pushed an update which I believe fixes all of the issues you three found in the previous version. Please take a look.

Here's how I tested the update:
1. Ran fio on GCP with the old and new argument styles to verify that disk type translation still works.
2. Ran local disk tests on AWS checking that SSDs and HDDs work and record correct metadata.
3. Ran local disk tests on Azure also checking SSDs and HDDs. They work and record correct metadata.
4. Ran pkb with only a benchmark specified, nothing about disks, to verify that the default disk behavior works. I watched it get to the prepare stage with no warnings on GCP and AWS, but didn't wait for the entire benchmark to finish. (It does produce warnings on Azure, but this is because Azure depends on the `--azure_storage_type` flag. I think you needed to specify that in the old version too, so we're not making it any harder.)

perfkitbenchmarker/providers/aws/aws_disk.py
stfeng: Understood. Quick question then: Today can a benchmark spec create a scratch disk without specifying the scratch_disk_type or equivalent in their benchmark spec? If so, we need to continue to support that behavior.

Going back to this particular line, I think we need to put back the disk.STANDARD entry here in the mapping. The goal of this mapping is to map those disk type constants as defined in the base class (disk.py) to provider-specific literal strings that can be passed to provider CLI tool to create disk. It's merely a "happy coincidence" that the disk.STANDARD constant is the same as the literal string "standard" accepted by AWS and we can use as is here. However, it'd be semantically much cleaner if we still maintain the mapping here so we don't break the layers of abstractions and make this class make assumptions about some constants defined by some other class. Doing the mapping here will also avoid future maintenance problems: e.g., if someone decides to change the disk.STANDARD  constant in disk.py in the future to some other literal string, they would have no idea to come to this class to add a mapping here.

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,53 @@
...
+def RegisterDiskTypeMap(provider_name, map):
+  """Register a map from legacy disk type names to modern ones.
+
+  The translation machinery looks here to find the map corresponding
+  to the chosen provider and translates the user's flags and configs
+  to the new naming system. This function should be removed once the
+  (deprecated) legacy flags are removed.
+  """
+
+  DISK_TYPE_MAPS[provider_name] = map
```
meteorfox: Use of the parameter name `map` shadows the `map()` function in Python. Please change parameter name.

perfkitbenchmarker/disk.py
meteorfox: I agree with @stfeng as well. The intention would be clearer.

meteorfox:
Some minor stylistic issue. Thanks for the tests we sure need more of them, especially at the provider level.

I'll test this on Rackspace as well.

meteorfox:
1) Ran fio on Rackspace with old- and new-style flags. Works as expected
2) Omitting flags --data_disk_type, or --scratch_disk_type still defaults to original behavior.

stfeng:
Two high level comments (the code as it stands now looks good to me, with one coding comment pending):

1. About local disk provisioning and usage before and after this change:
[@yuyantingzero - since you have been dealing with local SSD a lot these days, I'd like to loop you into this particular discussion.]

Can you please provide specific commands that a user can use today (before this CL) to run PKB on a local SSD or HDD disk on GCP, AWS, Azure? (3 commands, one each). Then, let's see what will happen to these commands after this CL? My understanding is they will continue to run just fine, but I'd like to confirm. Also, after this CL, how will a user use the new "--data_disk_type" flag to specify local disks for GCP, AWS, and Azure? (looks like this question is similar to what we need to discuss in the new issue you filed about "local").

2. So far we have been focused on the user's angle, that is, how does a user specifies various flags to provision and use disks via PKB, and what should the values of these flags be. But the other angle of looking at this is from the benchmark's: 

how does a benchmark (and its developer) specify what disk type they want to use in a benchmark? 

To do that, they have to use provider-independent terms such as "standard", "Remote_SSD", etc unless they want their benchmark to run only on a particular provider (most of benchmarks want to be able to run on all providers).

My understanding is: today, they do that by using those constants as defined in disk.py, such as disk.STANDARD, disk.LOCAL, disk.REMOTE_SSD in their benchmark config (or spec). I think your CL will be backward compatible for these, but since we plan to deprecate these constants, my question to us is: what is our plan for disk types and names that can be specified in benchmarks? Not for this CL though but it should be part of this disk type refactoring work as well.

noahl:
@stfeng 

1) You asked for three command lines that users can use before this change to run benchmarks against local SSD. These should work:

1. On AWS: `./pkb --cloud=AWS --scratch_disk_type=local --machine_type=m3.2xlarge --benchmarks=fio`
2. On GCP: `./pkb --scratch_disk_type=local --gcp_num_local_ssds=1 --benchmarks=fio`
3. On Azure: `./pkb --cloud=Azure --scratch_disk_type=local  --zone="East US 2" --machine_type=Standard_D2 --benchmarks=fio`

After the change, the commands still work fine. (I just tested them on PKB *with* my PR.) The same thing with new-style flags is

1. On AWS: `./pkb --cloud=AWS --data_disk_type=local --machine_type=m3.2xlarge --benchmarks=fio`
2. On GCP: `./pkb --data_disk_type=local --gcp_num_local_ssds=1 --benchmarks=fio`
3. On Azure: `./pkb --cloud=Azure --data_disk_type=local --machine_type=Standard_D2 --benchmarks=fio`

All I did was change `--scratch_disk_type` to `--data_disk_type`.

2) From the benchmark developer's perspective, life is very easy because of the config change. While they could technically specify the disk they want using 'standard', 'remote_ssd', etc., I just looked through our benchmarks, and in practice they do not. The benchmarks that want disks all use the `default_500_gb` disk definition from `perfkitbenchmarker/configs/default_config_constants.yaml`, which this PR updates to use the new terminology. Therefore, all current benchmarks will be updated automatically.

When we add more benchmarks, whether they use constants from `default_config_constants.yaml` or specify disks themselves, they can use the config infrastructure to specify different disks on different providers.

noahl:
@meteorfox Thanks a lot for testing!

perfkitbenchmarker/providers/aws/aws_disk.py
noahl: @stfeng I don't think it is possible for a benchmark to create a disk without specifying scratch_disk_type. However, I just read through the configs for all of the benchmarks, and I am *sure* that none of them actually do it. :-)

I added disk.STANDARD to the mapping. That raised another issue, which is that since STANDARD is now in the mapping, the translation machinery will warn the user to switch names. I added a check in `WarnAndTranslateDiskTypes` to prevent that. I tested that using `--scratch_disk_type=standard` produces a warning about the flag name, but not about the disk type name.

perfkitbenchmarker/disk.py
noahl: Changed, thanks.

yuyantingzero:
@noahl , @stfeng , the local ssd commands seem good.
From benchmarks angle, we can define local, remote_ssd disk types with config files. Which should be able to hide naming differences between different cloud: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/configs/default_config_constants.yaml#L41

noahl:
@stfeng @meteorfox I just pushed changes that I believe fixed the issues you commented on; please take a look.

meteorfox:
@noahl LGTM

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,53 @@
...
+ZONE = 'zone'
+REGION = 'region'
+
+
+def RegisterDiskTypeMap(provider_name, type_map):
+  """Register a map from legacy disk type names to modern ones.
+
+  The translation machinery looks here to find the map corresponding
+  to the chosen provider and translates the user's flags and configs
+  to the new naming system. This function should be removed once the
```
yuyantingzero: Add a todo for removing this functioon

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,53 @@
...
+REPLICATION = 'replication'
+# And some possible values
+HDD = 'hdd'
+SSD = 'ssd'
+NONE = 'none'
+ZONE = 'zone'
+REGION = 'region'
+
+
+def RegisterDiskTypeMap(provider_name, type_map):
```
yuyantingzero: Document parameters.

perfkitbenchmarker/publisher.py
```diff
@@ -144,12 +143,23 @@ def AddMetadata(self, metadata, benchmark_spec):
...
         metadata[name_prefix + 'scratch_disk_type'] = data_disk.disk_type
         metadata[name_prefix + 'scratch_disk_size'] = data_disk.disk_size
         metadata[name_prefix + 'num_striped_disks'] = (
             data_disk.num_striped_disks)
-        if data_disk.disk_type == disk.PIOPS:
+        if getattr(data_disk, 'iops', None) is not None:
           metadata[name_prefix + 'scratch_disk_iops'] = data_disk.iops
+          metadata[name_prefix + 'aws_provisioned_iops'] = data_disk.iops
+        # Modern metadata keys
+        metadata[name_prefix + 'data_disk_0_type'] = data_disk.disk_type
```
yuyantingzero: Copy benchmark can have multiple data disks. How about second disk?

perfkitbenchmarker/configs/default_config_constants.yaml
```diff
@@ -40,11 +40,11 @@ default_single_core: &default_single_core
 
 default_500_gb: &default_500_gb
   GCP:
-    disk_type: standard
+    disk_type: pd-standard
```
yuyantingzero: What about other providers?

stfeng:
@noahl, LGTM. Please also wait for Yanting and Eric's CR before check-in. Thanks for the hard work!

perfkitbenchmarker/disk.py
noahl: Added the todo.

perfkitbenchmarker/disk.py
noahl: Done.

perfkitbenchmarker/publisher.py
noahl: We decided not to handle multiple disks right now but to label the disk metadata with `data_disk_0_` for future compatibility. This matches the current behavior.

perfkitbenchmarker/configs/default_config_constants.yaml
noahl: AWS is also handled by this update, and Azure is handled to the extent that we can handle it (you still need to pass the `--azure_storage_type` flag, but that doesn't change the current behavior). For the others, passing `standard` is the right thing to do right now, because those providers will translate `standard` to some appropriate provider-specific disk name. As they're updated for the refactor, we will have to also update the entries in this table.

yuyantingzero: Add a TODO please.

noahl:
@yuyantingzero I just updated in response to your comments. Please take another look.

perfkitbenchmarker/disk.py
```diff
@@ -45,6 +93,67 @@ def __init__(cls, name, bases, dct):
...
+  if cloud in DISK_TYPE_MAPS:
+    disk_type_map = DISK_TYPE_MAPS[cloud]
+    if name in disk_type_map and disk_type_map[name] != name:
+      new_name = disk_type_map[name]
+      logging.warning('Disk type name %s is deprecated and will be removed. '
+                      'Translating to %s for now.', name, new_name)
+      return new_name
+    else:
+      return name
+  else:
```
yuyantingzero: Will this else block ever be executed?

perfkitbenchmarker/disk.py
```diff
@@ -45,6 +93,67 @@ def __init__(cls, name, bases, dct):
...
+
+  Args:
+    name: a string specifying a disk type, either new or old.
+    cloud: the cloud we're running on.
+
+  Returns:
+    The new-style disk type name (i.e. the provider's name for the type).
+
+  Raises:
+    ValueError, if name is not a disk type name.
```
yuyantingzero: I didn't see this exception raised anywhere

perfkitbenchmarker/pkb.py
```diff
@@ -142,14 +140,26 @@
...
     'Type for all scratch disks. The default is standard')
+flags.DEFINE_string(
+    'data_disk_type', None,
+    'Type for all data disks. If a provider keeps the operating system and '
+    'user data on separate disks, this only affects the user data disk(s).'
+    'If the provider has OS and user data on the same disk, this flag affects'
+    'that disk.')
+flags.DEFINE_integer('scratch_disk_size', None, 'Size, in gb, for all scratch '
+                     'disks.')
+flags.DEFINE_integer('data_disk_size', None, 'Size, in gb, for all disks.')
```
yuyantingzero: For all data disks?

perfkitbenchmarker/pkb.py
```diff
@@ -142,14 +140,26 @@
...
 flags.DEFINE_integer('scratch_disk_iops', None,
                      'IOPS for Provisioned IOPS (SSD) volumes in AWS.')
+flags.DEFINE_integer('aws_provisioned_iops', None,
```
yuyantingzero: Since this flag is aws specific, can we move to aws/flags.py?

ehankland: This flag should be moved to perfkitbenchmarker/providers/aws/flags.py

perfkitbenchmarker/disk.py
```diff
@@ -45,6 +93,67 @@ def __init__(cls, name, bases, dct):
...
+                      old_name, new_name)
+  # We keep the old flag around so that providers that haven't been
+  # updated yet will continue to work.
+
+
+def WarnAndTranslateDiskFlags():
+  """Translate old disk-related flags to new disk-related flags.
+  """
+
+  WarnAndCopyFlag('scratch_disk_type', 'data_disk_type')
```
yuyantingzero: I think it is cleaner to put old flag, new flag into a constant dict. And loop through the dict to copy old flags into new flags.

perfkitbenchmarker/disk.py
noahl: Yes, for providers that have not been updated for the disk types refactor. This lets us port providers gradually rather than all at once.

perfkitbenchmarker/providers/aws/aws_disk.py
```diff
@@ -33,12 +33,55 @@
...
+    disk.MEDIA: disk.SSD,
+    disk.REPLICATION: disk.NONE
+}
+
+LOCAL_HDD_METADATA = {
+    disk.MEDIA: disk.HDD,
+    disk.REPLICATION: disk.NONE
 }
 
+LOCAL_HDD_PREFIXES = ['d2']
```
ehankland: It's currently deprecated, but the hs1 type also has local hdd

perfkitbenchmarker/disk.py
noahl: You're right, this is left over from an older version of the code. I will remove it.

perfkitbenchmarker/pkb.py
noahl: Fixed, thanks.

perfkitbenchmarker/pkb.py
noahl: Moved.

perfkitbenchmarker/disk.py
noahl: Done, thanks.

perfkitbenchmarker/providers/aws/aws_disk.py
noahl: Added `hs` to the list (they have to be two-character prefixes).

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,62 @@
...
+from perfkitbenchmarker import flags
 from perfkitbenchmarker import resource
 
+FLAGS = flags.FLAGS
+
+
+# These are the (deprecated) old disk type names
 STANDARD = 'standard'
 REMOTE_SSD = 'remote_ssd'
 PIOPS = 'piops'  # Provisioned IOPS (SSD) in AWS
```
yuyantingzero: Can we move PIOPS constant to aws/flags?

noahl:
@yuyantingzero @ehankland I just updated again.

perfkitbenchmarker/disk.py
noahl: It looks like Alicloud uses it too, so it's not just AWS any more. I was not expecting this.

perfkitbenchmarker/disk.py
```diff
@@ -18,14 +18,62 @@
...
+  to the new naming system. This function should be removed once the
+  (deprecated) legacy flags are removed.
+
+  Args:
+    provider_name: a string. The name of the provider. Must match
+      the names we give to providers in benchmark_spec.py.
+    type_map: a dict. Maps generic disk type names (STANDARD,
+      REMOTE_SSD, PIOPS) to provider-specific names.
+
+  TODO(nlavine): remove this function when we remove the deprecated
```
yuyantingzero: Move todo before the definition of the method:
TODO()... \n
def Register...:

perfkitbenchmarker/disk.py
yuyantingzero: Update the comment to be more general in this case.

yuyantingzero:
A few small comments, LGTM otherwise.

perfkitbenchmarker/disk.py
noahl: Updated.

perfkitbenchmarker/disk.py
noahl: Done.

noahl:
@yuyantingzero @ehankland I just pushed some small code cleanups that you suggested, and also one more change that we discussed last night - there's now a special disk metadata attribute called 'legacy_disk_type', and if this attribute is set, it becomes the metadata value for 'scratch_disk_type'. This lets us keep our dashboard queries working until we update them. Can you please review this change too? Thanks for reviewing such a big change.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -44,21 +78,37 @@ class AzureDisk(disk.BaseDisk):
...
     super(AzureDisk, self).__init__(disk_spec)
     self.host_caching = FLAGS.azure_host_caching
     self.name = None
     self.vm_name = vm_name
     self.lun = None
 
+    if self.disk_type == PREMIUM_STORAGE:
+      self.metadata = PREMIUM_STORAGE_METADATA
+    elif self.disk_type == STANDARD_DISK:
+      self.metadata = {
```
yuyantingzero: Why AWS and GCP disk metadata is factored out, Azure is not?

noahl: Azure's standard-disk can't be a constant dictionary because you need to look at FLAGS.azure_storage_type to find the replication strategy, and the local SSD metadata can't be a constant dictionary because you need to look at the machine type to find the media. I factored out premium storage because that actually is constant.

yuyantingzero: Thanks, LGTM

PR #600, opened 2015-10-30
Automatically register VM classes (and specs)

ehankland:


perfkitbenchmarker/disk.py
```diff
@@ -26,10 +26,29 @@
...
+def GetDiskSpecClass(cloud):
+  """Get the DiskSpec class corresponding to 'cloud'."""
+  return _DISK_SPEC_REGISTRY.get(cloud, BaseDiskSpec)
+
+
+class AutoRegisterDiskSpecMeta(type):
+  """Metaclass which automatically registers DiskSpecs."""
+
+  def __init__(cls, name, bases, dct):
+    if cls.CLOUD:
```
noahl: Should we raise an exception if cls.CLOUD is None? That seems like an almost-certain indication of a bug (except possibly for static VMs).

perfkitbenchmarker/virtual_machine.py
```diff
@@ -34,6 +34,38 @@
...
+    if cls.CLOUD:
+      _VM_SPEC_REGISTRY[cls.CLOUD] = cls
+    super(AutoRegisterVmSpecMeta, cls).__init__(name, bases, dct)
+
+
+class AutoRegisterVmMeta(abc.ABCMeta):
+  """Metaclass which allows VMs to automatically be registered."""
+
+  def __init__(cls, name, bases, dct):
+    if (hasattr(cls, 'CLOUD') and cls.CLOUD and
```
noahl: Same comment as for disk specs - should it be an error *not* to have CLOUD and OS_TYPE attributes?

perfkitbenchmarker/disk.py
ehankland: I now raise an exception if cls.CLOUD has already been registered (which is equivalent to what you asked for because BaseDiskSpec will be registered first with CLOUD == None).

perfkitbenchmarker/virtual_machine.py
ehankland: We now raise if they haven't defined those attributes

ehankland:
PTAL

noahl:
LGTM

PR #608, opened 2015-11-03
Return static VMs to the pool when deleted.

cmccoy:
Right now a static VM can be used in at most one benchmark.
This PR returns the VM to the pool of available static VMs when
`VirtualMachine.Delete()` is called.


voellm:
LGTM

ehankland:
I don't think VMs should get added back to the pool if they were never part of the pool to begin with (e.g. if they were created from the config file rather than from the json file).

PR #609, opened 2015-11-03
Specify storage region in object storage benchmark

noahl:
Let the user specify which storage region to use in the object storage
benchmark.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -51,6 +51,9 @@ flags.DEFINE_enum('storage', providers.GCP,
                    providers.AZURE, providers.OPENSTACK],
                   'storage provider (GCP/AZURE/AWS/OPENSTACK) to use.')
 
+flags.DEFINE_string('object_storage_region', None,
+                    'Storage region for object storage benchmark.')
+
 flags.DEFINE_enum('object_storage_scenario', 'all',
                   ['all', 'cli', 'api_data', 'api_namespace'],
                   'select all, or one particular scenario to run: \n'
@@ -168,7 +171,26 @@ CONTENT_REMOVAL_RETRY_LIMIT = 5
 BUCKET_REMOVAL_RETRY_LIMIT = 120
 RETRY_WAIT_INTERVAL_SECONDS = 30
 
-DEFAULT_GCS_REGION = 'US-CENTRAL1'
+DEFAULT_GCP_REGION = 'us-central1'
+DEFAULT_AWS_REGION = 'us-east-1'
+DEFAULT_AZURE_REGION = 'East US'
+
+# The endpoints in this table are subdomains of 'amazonaws.com'. So
+# where the table says 's3-us-west-2', you should connect to
+# 's3-us-west-2.amazonaws.com'. This table comes from
+# http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
+AWS_S3_REGION_TO_ENDPOINT_TABLE = {
+    'us-east-1': 's3-external-1',
+    'us-west-2': 's3-us-west-2',
+    'us-west-1': 's3-us-west-1',
+    'eu-west-1': 's3-eu-west-1',
+    'eu-central-1': 's3-eu-central-1',
+    'ap-southeast-1': 's3-ap-southeast-1',
+    'ap-southeast-2': 's3-ap-southeast-2',
+    'ap-northeast-1': 's3-ap-northeast-1',
+    'sa-east-1': 's3-sa-east-1'
+}
+AWS_S3_ENDPOINT_SUFFIX = '.amazonaws.com'
 
 
 def GetConfig(user_config):
@@ -263,7 +285,7 @@ def _MakeSwiftCommandPrefix(auth_url, tenant_name, username, password):
 
 def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
                        bucket_name, regional_bucket_name=None,
-                       azure_command_suffix=None):
+                       azure_command_suffix=None, host_to_connect=None):
     """This function contains all api-based benchmarks.
        It uses the value of the global flag "object_storage_scenario" to
        decide which scenario to run inside this function. The caller simply
@@ -278,6 +300,7 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
       bucket_name: The name of the bucket caller has created for this test.
       regional_bucket_name: The name of the "regional" bucket, if applicable.
       azure_command_suffix: A suffix for all Azure related test commands.
+      host_to_connect: An optional endpoint string to connect to.
 
     Raises:
       ValueError: unexpected test outcome is found from the API test script.
@@ -285,6 +308,28 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
     if FLAGS.object_storage_scenario == 'cli':
       # User only wants to run the CLI based tests, do nothing here:
       return
+
+    def BuildBenchmarkScriptCommand(args):
+      """Build a command string for the API test script.
+
+      Args:
+        A list of strings. These will become space-separated arguments to the
+          test script.
+
+      Returns:
+        A string that can be passed to vm.RemoteCommand.
+      """
+
+      cmd_parts = [
+          test_script_path,
+          '--storage_provider=%s' % storage
+      ] + args
+      if azure_command_suffix is not None:
+        cmd_parts += [azure_command_suffix]
+      if host_to_connect is not None:
+        cmd_parts += ['--host', host_to_connect]
+
+      return ' '.join(cmd_parts)
 
     if (FLAGS.object_storage_scenario == 'all' or
         FLAGS.object_storage_scenario == 'api_data'):
@@ -294,12 +339,9 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
         buckets.append(regional_bucket_name)
 
       for bucket in buckets:
-        one_byte_rw_cmd = ('%s --bucket=%s --storage_provider=%s '
-                           '--scenario=OneByteRW') % (
-                               test_script_path, bucket, storage)
-
-        if azure_command_suffix is not None:
-          one_byte_rw_cmd = ('%s %s') % (one_byte_rw_cmd, azure_command_suffix)
+        one_byte_rw_cmd = BuildBenchmarkScriptCommand([
+            '--bucket=%s' % bucket,
+            '--scenario=OneByteRW'])
 
         _, raw_result = vm.RemoteCommand(one_byte_rw_cmd)
         logging.info('OneByteRW raw result is %s', raw_result)
@@ -322,14 +364,9 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
                              '%s.' % raw_result)
 
       # Single stream large object throughput metrics
-      single_stream_throughput_cmd = ('%s --bucket=%s --storage_provider=%s '
-                                      '--scenario=SingleStreamThroughput') % (
-                                          test_script_path,
-                                          bucket_name,
-                                          storage)
-      if azure_command_suffix is not None:
-        single_stream_throughput_cmd = ('%s %s') % (
-            single_stream_throughput_cmd, azure_command_suffix)
+      single_stream_throughput_cmd = BuildBenchmarkScriptCommand([
+          '--bucket=%s' % bucket_name,
+          '--scenario=SingleStreamThroughput'])
 
       _, raw_result = vm.RemoteCommand(single_stream_throughput_cmd)
       logging.info('SingleStreamThroughput raw result is %s', raw_result)
@@ -357,17 +394,10 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
     if (FLAGS.object_storage_scenario == 'all' or
         FLAGS.object_storage_scenario == 'api_namespace'):
       # list-after-write consistency metrics
-      list_consistency_cmd = ('%s --bucket=%s --storage_provider=%s '
-                              '--iterations=%d --scenario=ListConsistency') % (
-                                  test_script_path,
-                                  bucket_name,
-                                  storage,
-                                  LIST_CONSISTENCY_ITERATIONS)
-
-      if azure_command_suffix is not None:
-        list_consistency_cmd = ('%s %s') % (list_consistency_cmd,
-                                            azure_command_suffix)
-
+      list_consistency_cmd = BuildBenchmarkScriptCommand([
+          '--bucket=%s' % bucket_name,
+          '--iterations=%d' % LIST_CONSISTENCY_ITERATIONS,
+          '--scenario=ListConsistency'])
 
       _, raw_result = vm.RemoteCommand(list_consistency_cmd)
       logging.info('ListConsistency raw result is %s', raw_result)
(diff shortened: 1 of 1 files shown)


noahl:
@yuyantingzero @stfeng Please take a look.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -865,6 +878,24 @@ def Cleanup(self, vm):
     benchmark_spec_class.AZURE: AzureBlobStorageBenchmark()}
 
 
+WRONG_REGION_FLAGS = {
```
voellm: What are you trying to guard against here?  Running from GCP VMs to AWS S3?

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -45,11 +45,21 @@
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.sample import PercentileCalculator  # noqa
 
+
 flags.DEFINE_enum('storage', benchmark_spec_class.GCP,
                   [benchmark_spec_class.GCP, benchmark_spec_class.AWS,
                    benchmark_spec_class.AZURE],
                   'storage provider (GCP/AZURE/AWS) to use.')
 
+flags.DEFINE_string('object_storage_gcp_region', 'us-central1',
```
voellm: The fact there are three very similar flags here seems to indicate a level of abstraction is missing.

I think we should treat object stores like other resources and have them in the GCP, AWS, .. .directory.  Then these flags could be there.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -45,11 +45,21 @@
...
+                    'Region for GCP object storage.')
+
+flags.DEFINE_string('object_storage_aws_region', 'us-east-1',
```
yuyantingzero: The only difference between these 3 flags is their default values. Could you merge these 3 flags and use a dict to map object storage providers to regions?

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -865,6 +878,24 @@ def Cleanup(self, vm):
...
+WRONG_REGION_FLAGS = {
+    benchmark_spec_class.GCP: ['object_storage_aws_region',
+                               'object_storage_azure_region'],
+    benchmark_spec_class.AWS: ['object_storage_gcp_region',
+                               'object_storage_azure_region'],
+    benchmark_spec_class.AZURE: ['object_storage_gcp_region',
+                                 'object_storage_aws_region']}
+
+
+def CheckWrongRegionFlags():
```
yuyantingzero: Documentation.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -529,8 +537,10 @@ def Prepare(self, vm):
     vm.PushFile(FLAGS.boto_file_location, DEFAULT_BOTO_LOCATION)
 
     self.bucket_name = 'pkb%s' % FLAGS.run_uri
+
     vm.RemoteCommand(
-        'aws s3 mb s3://%s --region=us-east-1' % self.bucket_name)
+        'aws s3 mb s3://%s --region=%s' % (
```
stfeng: It's more complicated than this, you will also need to change the actual test script to use the correct endpoint for the region where the storage bucket is created in.

Here is a official table listing all the regions and endpoints:
http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

Then in our code, we need to pass the region name to the actual test script which will then lookup the endpoint to use using the table above.

Similar logic is necessary for Azure, perhaps for Google (perhaps not).

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: I'm trying to guard against the situation where you run the benchmark with one storage provider, then switch to another storage provider, but forget to say what region you want to use with the new provider. This is a "hey, you probably haven't updated your command line all the way" message.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: I just talked with Stanley. Making object stores act like other resources is separate from the rest of this change, so I filed #621 to discuss. I'd like to submit this PR without that and maybe do that as a separate change.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: I could. I made them three separate flags because it makes the default values clear and also because it lets me print a warning if the user passes the wrong flag for their storage provider. I'm open to merging them if you think it's more clear.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Documented. Thanks.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: I've done this for AWS. I looked for a while on the Azure site, but I couldn't find any indication that we need to use different endpoints for different regions. I think they do DNS magic to always route you to a nearby endpoint from the same URL (and hopefully your object is in a bucket near you...).

noahl:
@stfeng @yuyantingzero I think I've fixed all of the issues you found. Please take a look.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: Yes please merge them into one flag, just like the --zones flag is.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -143,7 +153,20 @@
 BUCKET_REMOVAL_RETRY_LIMIT = 120
 RETRY_WAIT_INTERVAL_SECONDS = 30
 
-DEFAULT_GCS_REGION = 'US-CENTRAL1'
+# The endpoints in this table are subdomains of 'amazonaws.com'. So
+# where the table says 's3-us-west-2', you should connect to
+# 's3-us-west-2.amazonaws.com'.
```
stfeng: please provide a link pointing to the official documentation of this table.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -143,7 +153,20 @@
...
+AWS_REGION_TO_ENDPOINT_TABLE = {
```
stfeng: Rename to: AWS_S3_REGION_TO_ENDPOINT_TABLE

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -226,6 +250,28 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
       # User only wants to run the CLI based tests, do nothing here:
       return
 
+    def BuildCommandString(args):
```
stfeng: The way the current diff shows this function "BuildCommandString()" is embeded inside the ApiBasedBenchmark() function. Why is it so? Can you move this outside as a generic utility routine.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -226,6 +250,28 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
...
+        A list of strings. These will become space-separated arguments to the
+          test script.
+
+      Returns:
+        A string that can be passed to vm.RemoteCommand.
+      """
+
+      cmd_parts = [
+          test_script_path,
+          '--storage_provider=%s' % storage
```
stfeng: can this be passed in as a parameter?

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -572,8 +604,10 @@ def Run(self, vm, metadata):
 
     # Now tests the storage provider via APIs
     test_script_path = '%s/run/%s' % (scratch_dir, API_TEST_SCRIPT)
+    hostname = AWS_REGION_TO_ENDPOINT_TABLE[FLAGS.object_storage_aws_region]
     ApiBasedBenchmarks(results, metadata, vm, 'S3', test_script_path,
-                       self.bucket_name)
+                       self.bucket_name,
+                       host_to_connect=hostname + '.amazonaws.com')
```
stfeng: make the .amazonaws.com part as a constant.

In general, as we discussed in previous CRs, consider define all literal strings as constants. (to avoid typos, and easy for lookup)

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: It seems to be an unnecessary restriction to me. e.g., We allow testing from a GCE VM to S3 bucket, so it's actually a legitimate case here.

stfeng:
I added some comments. One additional question here: did you test and verify the new code? And what was the test matrix?

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Removed.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Done.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Added link.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Renamed.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: I made this function inside of ApiBasedBenchmarks() so that it could access the variables test_script_path, storage, azure_command_suffix, and host_to_connect without having to explicitly pass them as parameters. I think that makes it smaller and easier to read. However, I can certainly move it outside of ApiBasedBenchmarks() if you prefer that.

As for the name, I'll change it to something more descriptive.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Yes. If we decide to move the function outside of ApiBasedBenchmarks(), then test_script_path, storage, azure_command_suffix, and host_to_connect will become parameters.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
noahl: Changed to a constant.

noahl:
I think I've responded to all of your comments. I'm waiting for your decision on whether to move BuildCommandString() outside of its containing function, but otherwise, I've addressed everything.

I forgot to mention before how I tested this. I did two kinds of tests. First, I tested each benchmark using GCP as the provider, to verify that they still worked after the BuildCommandString() refactor. Then I tested each region in AWS S3 to verify that the region-to-endpoint mapping worked (I did them all in parallel. I believe two of the processes hung, but I think that is because I was out of quota rather than because there was a bug that only appeared in two regions).

I think I also tested GCP and Azure storage just to make sure it worked, but I'm not sure because it was a while ago. In any case, I will do more testing with the updated code before I push it.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: If we don't foresee a reason for other code to call BuildCommandString() outside ApiBasedBenchmarks() then I think it's OK to keep it in here.  

As to naming, since it's building a command string specifically for calling our object_storage_test_script, so let's make sure the name reflects that. ;)

stfeng:
Talked with Noah about this. Looks like we need some more testing here for this feature, specifically: 1) validate that bucket metadata shows the bucket is created in the region as expected. 2) For AWS S3 where we need to connect to a specific endpoint given a region, validate (via tcpdump) that the connect request is actually sent to the expected endpoint.

It will take some time to do such testing, which will likely happen at a later time (after Noah finishes some other higher priority work). 

Once testing is done, code review will resume.

noahl:
I did some more testing on this. I have observed by watching the consoles/portals that this creates buckets in the requested regions for AWS, Azure, and GCP. I also have a tcpdump file that shows the client connecting to s3-us-west-2.amazonaws.com when using a bucket in the us-west-2 region, as desired. Would you like me to check other regions, or is this enough verification?

stfeng:
Yep, the verification looks good to me! Once you update the code with merges and incorporate existing CR comments, please ping me, I will take another look. 

Thanks.

noahl:
@stfeng I updated with all of the comments above and rebased on master. Please take a look.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -589,9 +618,10 @@ def Prepare(self, vm):
     vm.PushFile(FLAGS.boto_file_location, DEFAULT_BOTO_LOCATION)
 
     vm.bucket_name = 'pkb%s' % FLAGS.run_uri
+    region = FLAGS.object_storage_region or DEFAULT_AWS_REGION
 
     vm.RemoteCommand(
-        'aws s3 mb s3://%s --region=us-east-1' % vm.bucket_name)
+        'aws s3 mb s3://%s --region=%s' % (vm.bucket_name, region))
```
stfeng: what happens if the region is invalid?

I presume an exception will be thrown (vm.RemoteCommand would fail?) and prepare stage would fail.

can you try it quickly to confirm this...

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -634,8 +664,11 @@ def Run(self, vm, metadata):
 
     # Now tests the storage provider via APIs
     test_script_path = '%s/run/%s' % (scratch_dir, API_TEST_SCRIPT)
+    region = FLAGS.object_storage_region or DEFAULT_AWS_REGION
```
stfeng: Since this is in the run phase, we should try to use exactly the same region as that is used in the prepare stage. Would it be possible to first save the region in the VM object (so it gets persisted at the end of the prepare stage), and then read it from the VM object at here (rather than reading from the flag).

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Confirmed. When I use an invalid region, the vm.RemoteCommand raises an exception, and PKB cleans up the resources.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Changed to save the region in the VM.

noahl:
@stfeng I think I've addressed all of your concerns. Please take a look.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -831,10 +867,11 @@ def Prepare(self, vm):
 
     vm.RemoteCommand('%s mb gs://%s' % (vm.gsutil_path, vm.bucket_name))
 
+    region = FLAGS.object_storage_region or DEFAULT_GCP_REGION
     vm.regional_bucket_name = '%s-%s' % (vm.bucket_name,
-                                         DEFAULT_GCS_REGION.lower())
+                                         region.lower())
     vm.RemoteCommand('%s mb -c DRA -l %s gs://%s' % (vm.gsutil_path,
-                                                     DEFAULT_GCS_REGION,
+                                                     region.upper(),
```
zbjornson: Is there a reason this is `upper` and L872 is `lower`?

PR #612, opened 2015-11-04
update version of nova API

kivio:
Update from deprecated version 1.1 to version  2 witch is currently proposed to use.


voellm:
LGTM... please just double check.cluster_boot and iperf work.  For changes like this those usually catch and issues.

kivio:
@voellm i was checking it!

voellm:
Cool... please let me know when to hit merge.

Optionally Mateusz from Intel and Carlos from Rackspace can now also
complete merges.

kivio:
@voellm i think it can be merged now :) thx :)

PR #614, opened 2015-11-05
Change behavior after a benchmark fails.

skschneider:
Make default action to continue executing benchmarks.
Add flag to allow stopping after failure.
Print failed benchmark rate at end of PKB run.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -157,6 +157,13 @@ flags.DEFINE_bool('install_packages', True,
                   'on any VMs. This option should probably only ever be used '
                   'if you have already created an image with all relevant '
                   'packages installed.')
+flags.DEFINE_bool(
+    'stop_after_benchmark_failure', False,
+    'Determines response when running multiple benchmarks serially and a '
+    'benchmark run fails. When True, no further benchmarks are scheduled, and '
+    'execution ends. When False, benchmarks continue to be scheduled. Does not '
+    'apply to keyboard interrupts, which will always prevent further '
+    'benchmarks from being scheduled.')
 
 # Support for using a proxy in the cloud environment.
 flags.DEFINE_string('http_proxy', '',
@@ -409,12 +416,14 @@ def RunBenchmarks(publish=True):
       static_virtual_machine.StaticVirtualMachine.ReadStaticVirtualMachineFile(
           fp)
 
+  successful_benchmarks = []
+  failed_benchmarks = []
   try:
     benchmark_tuple_list = benchmark_sets.GetBenchmarksFromFlags()
     total_benchmarks = len(benchmark_tuple_list)
 
     benchmark_counts = collections.Counter()
-    args = []
+    args = collections.deque()
     for i, benchmark_tuple in enumerate(benchmark_tuple_list):
       benchmark_module, user_config = benchmark_tuple
       benchmark_uid = (benchmark_module.BENCHMARK_NAME +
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/pkb.py
```diff
@@ -428,13 +437,34 @@ def RunBenchmarks(publish=True):
...
-      for run_args, kwargs in args:
-        RunBenchmark(*run_args, **kwargs)
+      while args:
+        run_args, kwargs = args.popleft()
+        benchmark_name = run_args[0].BENCHMARK_NAME
+        try:
+          RunBenchmark(*run_args, **kwargs)
+          successful_benchmarks.append(benchmark_name)
+        except BaseException as e:
+          logging.exception(e)
```
cmccoy: Nit: `logging.exception(e)` is redundant - a stacktrace will already be printed with `e` in it.
Could you add some extra information to the message, like the benchmark name, index within the collection of benchmarks running, and whether execution will continue?

perfkitbenchmarker/pkb.py
```diff
@@ -409,12 +416,14 @@ def RunBenchmarks(publish=True):
       static_virtual_machine.StaticVirtualMachine.ReadStaticVirtualMachineFile(
           fp)
 
+  successful_benchmarks = []
+  failed_benchmarks = []
```
cmccoy: I think it would be helpful to preserve the order of execution and status (in case there was a temporal association).
What do you think about maintaining a list of `(benchmark, result)` pairs?

perfkitbenchmarker/pkb.py
```diff
@@ -428,13 +437,34 @@ def RunBenchmarks(publish=True):
...
+          if (isinstance(e, KeyboardInterrupt) or
+              FLAGS.stop_after_benchmark_failure):
+            break
+      failed_benchmarks.extend(a[0].BENCHMARK_NAME for a, _ in args)
 
   finally:
     if collector.samples:
       collector.PublishSamples()
 
+    if successful_benchmarks:
```
cmccoy: Nit: log this unconditionally.

perfkitbenchmarker/pkb.py
skschneider: Done.

perfkitbenchmarker/pkb.py
skschneider: Done.

perfkitbenchmarker/pkb.py
skschneider: Done.

perfkitbenchmarker/pkb.py
```diff
@@ -428,13 +436,46 @@ def RunBenchmarks(publish=True):
...
+            logging.error('%s Execution will not continue.', msg)
+            break
+          logging.error('%s Execution will continue.', msg)
+      benchmark_arg_succeeded_tuples.extend((a, False) for a, _ in args)
 
   finally:
     if collector.samples:
       collector.PublishSamples()
 
+    if benchmark_arg_succeeded_tuples:
```
cmccoy: This is equivalent to the last version, correct?

skschneider: No, this is different. benchmark_arg_succeeded_tuples is a list of (args tuple, succeeded boolean) tuples, one per benchmark. It includes results for successful and failed benchmarks in order of execution per the other comment suggestion. But it is not populated in some cases: if --parallelism is used (since that's pending deletion and may not be worth the trouble to implement) or if, for some reason, the user specifies --benchmarks="".

cmccoy: Is the output different from what you had before though?

I was thinking of something interleaved like:

```
index  benchmark    status
0      iperf        SUCCESS
1      netperf      SUCCESS
2      cassandra    SUCCESS
3      netperf      FAILED
4      speccpu2006  SUCCESS
5      speccpu2006  FAILED
6      iperf        SKIPPED
7      netperf      SKIPPED
```

But up to you. Code LGTM.

skschneider: Merging for flag, will open separate PR for this formatting.

Issue #590, opened 2015-10-26
Log the version of boto for GCP and S3 to metadata in object storage benchmark


stfeng:
As title.

PR #632, opened 2015-11-13
Add boto lib version to GCP and S3 object storage benchmark metadata

stfeng:
This is for issue #590. Tested on all 3 providers, here are the metadata from test output:

GCP: 
'metadata': {'boto_lib_version': u'Version: 2.38.0\n',
 'machine_type': 'n1-standard-1', 'zone': 'us-central1-a', 'scratch_disk_type': 
'standard',
<snip>

AWS:
'metadata': {'boto_lib_version': u'Version: 2.38
.0\n', 'machine_type': 'm3.medium', 'zone': 'us-east-1b', 'scratch_disk_type': '
standard', 
<snip>

Azure:
 'metadata': {'storage provider': 'Azure', 'zone'
: 'East US', 'scratch_disk_type': 'standard', 'perfkitbenchmarker_version': 'v0.
23.0-111-g47eaf58', 'azure_lib_version': u'Version: 1.0.2\n', <---it's here.
<snip>

### DIFF
--- a/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
@@ -106,6 +106,7 @@ GCE_CREDENTIAL_LOCATION = '.config/gcloud/credentials'
 AZURE_CREDENTIAL_LOCATION = '.azure'
 
 DEFAULT_BOTO_LOCATION = '~/.boto'
+BOTO_LIB_VERSION = 'boto_lib_version'
 
 OBJECT_STORAGE_CREDENTIAL_DEFAULT_LOCATION = {
     benchmark_spec_class.GCP: '~/' + GCE_CREDENTIAL_LOCATION,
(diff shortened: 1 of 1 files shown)


stfeng:
@noahl please take a look. Thanks.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
```diff
@@ -678,7 +681,7 @@ def Run(self, vm, metadata):
       the results of all scenarios run here.
 
     """
-    metadata['azure_lib_version'] = _GetAzureClientLibVersion(vm)
+    metadata['azure_lib_version'] = _GetClientLibVersion(vm, 'azure')
```
noahl: You used a constant for BOTO_LIB_VERSION but not for 'azure_lib_version'. Is that  because 'azure_lib_version' is only used once?

noahl:
It would be nice to remove the 'Version: ' and '\n' from the version metadata, so it's just the number. Other than that, this looks good to me.

perfkitbenchmarker/benchmarks/object_storage_service_benchmark.py
stfeng: right, azure_lib_version is only used once.

stfeng:
Thanks Noahl. will merge soon. As to the "Version" text string, it was there before this CL, we might remove that later but not in this one.

Issue #508, opened 2015-10-02
Apt-get update should not be run if install_packages is set to False


mateusz-blaszkowski:
`apt-get update` is run independently of the flag --install_packages, however it is useless (and unnecessary extends benchmarks execution time) when the flag is set to False.

ehankland:
Fixed by the above PR

PR #644, opened 2015-11-17
Only call SetupPackageManager if install_packages is True

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/508


yuyantingzero:
LGTM

Issue #109, opened 2015-02-09
cluster_boot benchmark should create a network on GCE.


cmccoy:
Right now we attach to the default network, but for parity we should create a new one.

ehankland:
Fixed by above PR

PR #648, opened 2015-11-18
Create GCP networks

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/109

### DIFF
--- a/perfkitbenchmarker/providers/gcp/gce_network.py
+++ b/perfkitbenchmarker/providers/gcp/gce_network.py
@@ -34,17 +34,19 @@ GCP = 'GCP'
 class GceFirewallRule(resource.BaseResource):
   """An object representing a GCE Firewall Rule."""
 
-  def __init__(self, name, project, port):
+  def __init__(self, name, project, port, network):
     super(GceFirewallRule, self).__init__()
     self.name = name
     self.project = project
     self.port = port
+    self.network_name = network.network_resource.name
 
   def __eq__(self, other):
     """Defines equality to make comparison easy."""
     return (self.name == other.name and
             self.port == other.port and
-            self.project == other.project)
+            self.project == other.project and
+            self.network_name == other.network_name)
 
   def _Create(self):
     """Creates the Firewall Rule."""
@@ -52,6 +54,7 @@ class GceFirewallRule(resource.BaseResource):
                              self.name)
     cmd.flags['allow'] = ','.join('{0}:{1}'.format(protocol, self.port)
                                   for protocol in ('tcp', 'udp'))
+    cmd.flags['network'] = self.network_name
     cmd.Issue()
 
   def _Delete(self):
@@ -65,9 +68,7 @@ class GceFirewallRule(resource.BaseResource):
     cmd = util.GcloudCommand(self, 'compute', 'firewall-rules', 'describe',
                              self.name)
     _, _, retcode = cmd.Issue(suppress_warning=True)
-    if retcode:
-      return False
-    return True
+    return not retcode
 
 
 class GceFirewall(network.BaseFirewall):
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -35,11 +35,12 @@
...
 
-  def __init__(self, name, project, port):
+  def __init__(self, name, project, port, network):
     super(GceFirewallRule, self).__init__()
     self.name = name
     self.project = project
     self.port = port
+    self.network = network
 
   def __eq__(self, other):
```
skschneider: Does this __eq__ need to be updated to include network comparison?

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -54,6 +55,7 @@ def _Create(self):
                   'firewall-rules',
                   'create',
                   self.name,
+                  '--network=%s' % self.network.network_resource.name,
                   '--allow', 'tcp:%d' % self.port, 'udp:%d' % self.port]
     create_cmd.extend(util.GetDefaultGcloudFlags(self))
```
skschneider: Because of a change a merged, the old process of build gcloud command -> call util.GetDefaultGcloudFlags -> call vm_util.IssueCommand is replaced by creating a GcloudCommand object and calling its Issue/IssueRetryable method. Sorry about the hassle.

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -121,15 +124,87 @@ def DisallowAllPorts(self):
...
+    describe_cmd = [FLAGS.gcloud_path,
+                    'compute',
+                    'networks',
+                    'describe',
+                    self.name]
+    describe_cmd.extend(util.GetDefaultGcloudFlags(self))
+    _, _, retcode = vm_util.IssueCommand(describe_cmd, suppress_warning=True)
+    if retcode:
+      return False
+    return True
```
skschneider: I think ```not retcode``` is equivalent.

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -121,15 +124,87 @@ def DisallowAllPorts(self):
...
 class GceNetwork(network.BaseNetwork):
   """Object representing a GCE Network."""
 
   CLOUD = GCP
 
+  def __init__(self, network_spec):
+    super(GceNetwork, self).__init__(network_spec)
+    self.project = network_spec.project
+    name = 'pkb-network-%s' % FLAGS.run_uri
+    self.network_resource = GceNetworkResource(name, self.project)
```
skschneider: Is there a reason the GceNetworkResource is separate from the GceNetwork?

perfkitbenchmarker/providers/gcp/gce_network.py
ehankland: It doesn't need it right now, because we already compare projects, and  we only have one Network per project, but that  could change. I'll add it.

perfkitbenchmarker/providers/gcp/gce_network.py
ehankland: Yeah - the GceNetwork class needs to inherit from the BaseNetwork class. The GceNetworkResource class needs to inherit from BaseResource. Making them a single class and having it inherit from both leads to problems (depending on the order of base classes, the BaseNetwork Create method may get called instead of the BaseResource one or the BaseNetwork __init__ method wont get called properly because their __init__ methods take different arguments). Rather than trying to make this work, it seemed simpler to have two separate classes.

perfkitbenchmarker/providers/gcp/gce_network.py
ehankland: Done

perfkitbenchmarker/providers/gcp/gce_network.py
ehankland: Fixed

perfkitbenchmarker/providers/gcp/gce_network.py
ehankland: Done

ehankland:
PTAL

skschneider:
LGTM

PR #649, opened 2015-11-18
Add GcloudCommand helper.

skschneider:


perfkitbenchmarker/providers/gcp/gce_disk.py
```diff
@@ -67,34 +66,27 @@ def __init__(self, disk_spec, name, zone, project, image=None):
...
 
   def _Exists(self):
     """Returns true if the disk exists."""
-    getdisk_cmd = [FLAGS.gcloud_path,
-                   'compute', 'disks',
-                   'describe', self.name]
-    getdisk_cmd.extend(util.GetDefaultGcloudFlags(self))
-    stdout, _, _ = vm_util.IssueCommand(getdisk_cmd, suppress_warning=True)
+    cmd = util.GcloudCommand('compute', 'disks', 'describe', self.name)
+    cmd.AddCommonFlags(self)
```
cmccoy: Are there any cases where this isn't called? If not, could the Command object take the object being created/described/deleted in the constructor?

skschneider: I have not found any cases where it isn't called. You're suggesting having the constructor add the common flags?

cmccoy: Or storing a reference to the calling Resource on the `GcloudCommand` instance, so common flags can be added at the appropriate moment. If the flags can be added immediately, doing it in the constructor sounds fine too.

voellm:
@skschneider Please add some tests as part of this CL.  For something like this it should be straight forward to have unit tests.

perfkitbenchmarker/providers/gcp/gce_disk.py
skschneider: Done.

skschneider:
Added unit tests for GcloudCommand.

skschneider:
PTAL

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -13,26 +13,101 @@
...
+      A tuple of stdout, stderr, and retcode from running the gcloud command.
+    """
+    return vm_util.IssueCommand(self._GetCommand(),
+                                suppress_warning=suppress_warning)
+
+  def IssueRetryable(self):
+    """Tries running the gcloud command until it succeeds or times out.
+
+    Returns:
+      A tuple of stdout, stderr, and retcode from running the gcloud command.
```
cmccoy: IssueRetryable returns `(stdout, stderr)`; no return code.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -13,26 +13,101 @@
...
+    return cmd
+
+  def __repr__(self):
+    return '{0}({1})'.format(type(self).__name__, ' '.join(self._GetCommand()))
+
+  def Issue(self, suppress_warning=False):
+    """Tries running the gcloud command once.
+
+    Args:
+      suppress_warning: boolean. Controls the log level of the command result
```
cmccoy: Consider taking `kwargs` here to pass to `vm_util.IssueCommand`.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -13,26 +13,101 @@
...
 
-  Returns:
-    A common set of gcloud options.
+  Attributes:
+    args: list of strings. Non-flag args to pass to gcloud, typically
+        specifying an operation to perform (e.g. ['compute', 'images', 'list']
+        to list available images).
+    flags: OrderedDict mapping flag name string to flag value. Flags to pass to
+        gcloud (e.g. {'project': 'my-project-id'}). If a provided value is
+        True, the flag is passed to gcloud without a value. If a provided value
```
cmccoy: Should `False` create `--no-FLAG`?

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -106,6 +106,72 @@ def __init__(self, vm_spec):
...
+    return util.GcloudCommand(self, 'compute', 'instances', 'describe',
+                              self.name)
+
+  def _GenerateDeleteCommand(self):
+    """Generates a command to delete the VM instance.
+
+    Returns:
+      GcloudCommand. gcloud command to issue in order to delete the VM instance.
+    """
+    return util.GcloudCommand(self, 'compute', 'instances', 'delete', self.name)
```
cmccoy: Nit: I think all but `Create` of these `_Generate*Command` functions could be inlined where used - they're ~7 lines of declaration / docstring for a few lines of actual code.

tests/providers/gcp/util_test.py
```diff
@@ -0,0 +1,113 @@
...
+        '--quiet', '--project', 'test-project', '--zone', 'test-zone'])
+
+  def testListValue(self):
+    gce_resource = GceResource(project=None)
+    cmd = util.GcloudCommand(gce_resource, 'compute', 'instances', 'create')
+    cmd.flags['local-ssd'] = ['interface=nvme', 'interface=SCSI']
+    self.assertEqual(cmd._GetCommand(), [
+        'path/gcloud', 'compute', 'instances', 'create', '--format', 'json',
+        '--quiet', '--local-ssd', 'interface=nvme', '--local-ssd',
+        'interface=SCSI'])
```
cmccoy: Could you add a boolean value test?

perfkitbenchmarker/providers/gcp/util.py
skschneider: It isn't common that there's a need for a program to support both a --flag and a --no-flag, since one of them will usually be the default behavior, so I don't think it's necessary to build that support in as a separate feature. This is more a matter of whether a flag has a value or not. I considered using None, rather than True, but it looks weird when you're setting options to explicitly provide a None value when you mean "enable this one". The way to set a --no-flag would be to pass {'no-flag': True} to this function.

tests/providers/gcp/util_test.py
skschneider: This is already handled by --quiet, which is now part of the default constructor.

cmccoy: Ah, OK. Thanks.

perfkitbenchmarker/providers/gcp/util.py
skschneider: Fixed.

perfkitbenchmarker/providers/gcp/util.py
skschneider: Done.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
skschneider: Done.

skschneider:
PTAL

cmccoy:
:+1:

PR #653, opened 2015-11-19
Add GetMachineTypeDict() to replace machine_type in Sample metadata.

skschneider:
Supports machine types that are not defined by a single name.

### DIFF
--- a/perfkitbenchmarker/virtual_machine.py
+++ b/perfkitbenchmarker/virtual_machine.py
@@ -235,6 +235,17 @@ class BaseVirtualMachine(resource.BaseResource):
     """
     pass
 
+  def GetMachineTypeDict(self):
+    """Returns a dict containing properties that specify the machine type.
+
+    Returns:
+      dict mapping string property key to value.
+    """
+    result = {}
+    if self.machine_type is not None:
+      result['machine_type'] = self.machine_type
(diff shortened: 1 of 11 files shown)


perfkitbenchmarker/virtual_machine.py
```diff
@@ -235,6 +235,17 @@ def AddMetadata(self, **kwargs):
     """
     pass
 
+  def GetMachineTypeDict(self):
```
yuyantingzero: Probably renaming to GetMachinePropertyDict?

yuyantingzero:
LGTM

perfkitbenchmarker/virtual_machine.py
skschneider: Yes, the idea here is that sometimes a "machine type" isn't defined by a single string like "n1-standard-8". When using a custom GCE VM, we could instead have, for example, {'cpus': 2, 'memory_mb': 4.0, 'preemptible': True} that collectively defines the type of machine that was provisioned. The reason I used "MachineType" is that this the equivalent of what our previous machine_type field did. "Properties" might imply that there could be other info in here, like number of disks or OS image, whereas those are separate properties that aren't related to our current understanding of a "machine type".

PR #663, opened 2015-11-20
Fix bug with GCP networks

ehankland:
Adds a firewall rule that allows all internal traffic - only the "default" network has this by default


yuyantingzero:
LGTM

PR #664, opened 2015-11-23
Add support for GCE custom VMs.

skschneider:
Add ConfigOptionDecoder classes.
Add 'cpus' and 'memory' options to the GCE VM spec.
Add 'cpus', 'memory_mb', and 'preemptible' to GCE VM Sample metadata.
Modify gcloud VM provisioning command to use custom VM flags.

### DIFF
--- a/perfkitbenchmarker/configs/option_decoders.py
+++ b/perfkitbenchmarker/configs/option_decoders.py
@@ -1,0 +1,182 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Classes for verifying and decoding config option values."""
+
+import abc
+
+from perfkitbenchmarker import errors
+
+
+class ConfigOptionDecoder(object):
+  """Verifies and decodes a config option value.
+
+  Attributes:
+    component: string. Description of the component to which the option applies.
+    option: string. Name of the config option.
+    required: boolean. True if the config option is required. False if not.
+  """
+
+  def __init__(self, component, option, **kwargs):
+    """Initializes a ConfigOptionDecoder.
+
+    Args:
+      component: string. Description of the component to which the option
+          applies.
+      option: string. Name of the config option.
+      **kwargs: May optionally contain a 'default' key mapping to a value or
+          callable object. If a value is provided, the config option is
+          optional, and the provided value is the default if the user does not
+          set a value for the config option. If a callable object is provided,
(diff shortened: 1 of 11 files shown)


perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -48,26 +49,95 @@
...
+    image: string or None. The disk image to boot from.
+    machine_type: string or None. The instance type name for non-custom VMs.
+        Example: 'n1-standard-8'.
+    memory: None or string. For custom VMs, a string representation of the size
+        of memory, expressed in MB or GB. Must be an integer number of MB (e.g.
+        "1280MB", "7.5GB").
+    num_local_ssds: int. The number of local SSDs to attach to the instance.
+    preemptible: boolean. True if the VM should be preemptible, False otherwise.
+    project: string or None. The project to create the VM in.
+    zone: string or None. The region / zone in which to launch the VM.
```
ehankland: Personally, I'd prefer to just document the attributes related to the GceVmSpec specifically here and document the other ones in the BaseVmSpec. It shouldn't be that hard to keep them in sync though since I don't expect the BaseVmSpec ones to change that much.

perfkitbenchmarker/configs/option_decoders.py
```diff
@@ -0,0 +1,182 @@
...
+      Default value of an optional config option.
+    """
+    assert not self.required, (
+        'Attempted to get the default value of {0} required config option '
+        '"{1}".'.format(self.component, self.option))
+    if hasattr(self._default, '__call__'):
+      return self._default()
+    return self._default
+
+  @abc.abstractmethod
```
ehankland: As far as I can see, ConfigOptionDecoder's metaclass isn't abc.ABCMeta, but I'm pretty sure it needs to be if you want this to work properly. See https://docs.python.org/2/library/abc.html#abc.abstractmethod

ehankland:
Can you add a TODO to move the DiskSpec classes over to using decoders as well?

ehankland:
LGTM

perfkitbenchmarker/configs/option_decoders.py
skschneider: Done.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
skschneider: Done.

voellm:
One thing missing from this commit is how to use it in the README.md file.

skschneider:
Perhaps this wiki page would be more appropriate: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/wiki/Adding-a-new-cloud-provider

I see the README as being more for users, and guides like this one being more for developers.

skschneider:
Actually, I see there's already a link to virtual_machine.BaseVmSpec and disk.BaseDiskSpec in the README. I'll probably add a common base class in #668 and then add a sentence to the README.md clarifying how the derived classes add options.

skschneider:
Doy, I realize now you were probably talking about how to use custom VMs. I'll add a change to the README for that.

voellm:
Ping ... checked the readme and did not see how to use this feature.

voellm:
Ok it looks like this is just in the config file.  Any plans for a command line option?

skschneider:
I have a change planned that makes it easier to use this from the command line. Waiting on #675 as prerequisite.

PR #652, opened 2015-11-19
Add provision and teardown run_stages

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/409

PR #665, opened 2015-11-23
Fix bug with run stages

ehankland:


skschneider:
LGTM

noahl:
Does this bug affect release 0.24.0 as well as master?

ehankland:
This affects anything with https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/652 - I don't think it's in 0.24.0 based on the change log, but I could be wrong

noahl:
It's not in 0.24.0. Thanks.

PR #666, opened 2015-11-24
gce_virtual_machine: Add support for image_project

wonderfly:
Although ```providers/gcp/flags.py``` claims it supports the flag ```image_project```,
when creating an instance, it is never used in ```gce_virtual_machine.py```, or passed
to the underlining ```gcloud``` cli.

A sample failure trying to creating an instance with an image in another
project:

```
$ ./pkb.py --project=wonderfly-playground --machine_type=f1-micro
--image=test-image --image_project=test-image-project --os_type=ubuntu_container
--benchmarks=cluster_boot
```
A subset of the error message I got:
>
2015-11-23 15:58:27,143 3fa6ceaf Thread-6 cluster_boot(1/1) INFO     Ran gcloud
compute instances create pkb-3fa6ceaf-0 --format json --quiet --project
wonderfly-playground --zone us-central1-a --network pkb-network-3fa6ceaf --image
test-image --boot-disk-size 10 --boot-disk-type pd-standard --machine-type
f1-micro --tags perfkitbenchmarker --no-restart-on-failure --metadata-from-file
sshKeys=/tmp/perfkitbenchmarker/run_3fa6ceaf/key-metadatanXAmr1 --metadata
owner=wonderfly --maintenance-policy TERMINATE.
Got return code (1).
STDOUT:
STDERR: WARNING: You have selected a disk size of under [200GB]. This may result
in poor I/O performance. For more information, see:
https://developers.google.com/compute/docs/disks/persistent-disks#pdperformance.
ERROR: (gcloud.compute.instances.create) Some requests did not succeed:
 - Invalid value for field 'resource.disk.initializeParams.sourceImage':
   'https://www.googleapis.com/compute/v1/projects/wonderfly-playground/global/images/test-image'.
   Referenced resource was not found.

This change actually enables the ```image_project``` flag, and adds a unit test to
verify it.

Tested with ```python test/gce_virtual_machine_test.py``` and the same command line as
shown above.

@voellm @cmccoy

### DIFF
--- a/tests/gce_virtual_machine_test.py
+++ b/tests/gce_virtual_machine_test.py
@@ -14,12 +14,14 @@
 
 """Tests for perfkitbenchmarker.providers.gcp.gce_virtual_machine"""
 
-import unittest
+import contextlib
 import mock
+import unittest
 
 from perfkitbenchmarker import benchmark_spec
 from perfkitbenchmarker import context
 from perfkitbenchmarker import errors
+from perfkitbenchmarker import pkb
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.providers.gcp import gce_virtual_machine
 
@@ -111,18 +113,27 @@ class GceVirtualMachineTestCase(unittest.TestCase):
                                                'preemptible': True})
 
 
-class GCEPreemptibleVMFlagTestCase(unittest.TestCase):
+class GCEVMFlagsTestCase(unittest.TestCase):
 
   def setUp(self):
     # VM Creation depends on there being a BenchmarkSpec.
     self.spec = benchmark_spec.BenchmarkSpec({}, 'name', 'benchmark_uid')
     self.addCleanup(context.SetThreadBenchmarkSpec, None)
 
-  def testPreemptibleVMFlag(self):
+  @contextlib.contextmanager
+  def _PatchCriticalObjects(self):
+    """A context manager that patches a few critical objects with mocks."""
     with mock.patch(vm_util.__name__ + '.IssueCommand') as issue_command, \
             mock.patch('__builtin__.open'), \
             mock.patch(vm_util.__name__ + '.NamedTemporaryFile'), \
+            mock.patch(pkb.__name__ + '.FLAGS') as pkb_flags, \
             mock.patch(gce_virtual_machine.__name__ + '.FLAGS') as gvm_flags:
+              yield issue_command, pkb_flags, gvm_flags
+
+  def testPreemptibleVMFlag(self):
+    with self._PatchCriticalObjects() as mocked_env:
+      issue_command, pkb_flags, gvm_flags = mocked_env
+      pkb_flags.run_uri = 'foo'
       gvm_flags.gce_preemptible_vms = True
       gvm_flags.gcloud_scopes = None
       vm_spec = gce_virtual_machine.GceVmSpec(image='image')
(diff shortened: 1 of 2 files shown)


voellm:
Eric - please review this.

yuyantingzero:
Image project is used when creating in boot disk:https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/aa81a791941301747c4f14f267e68f417b87802d/perfkitbenchmarker/providers/gcp/gce_disk.py#L74

wonderfly:
@yuyantingzero 
> Image project is used when creating boot disk:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/aa81a791941301747c4f14f267e68f417b87802d/perfkitbenchmarker/providers/gcp/gce_disk.py#L74

True. But that only works when you separate instance creation to _two steps_: create a boot disk, and create the instance with that boot disk. If you ever try to create an instance directly from an image in another project, you need to pass ```image-project``` to the ```gcloud compute instances create``` command line, which it doesn't today.

Please see the error message I pasted above for the full command line arguments.

voellm:
This is a recent break. @connormccoy moved to combine the steps.  This
looks to fix it.

voellm:
LGTM.

@skschneider  Please review also.

tests/gce_virtual_machine_test.py
```diff
@@ -44,5 +47,24 @@ def testPreemptibleVMFlag(self):
       self.assertEquals(issue_command.call_count, 1)
       self.assertIn('--preemptible', issue_command.call_args[0][0])
 
+  def testImageProjectFlag(self):
+    """Tests that custom image_project flag is supported."""
+    with mock.patch(vm_util.__name__ + '.IssueCommand') as issue_command, \
```
skschneider: Now that this lengthy ```with``` appears twice, it'll be good to separate it into a helper or the setUp method for reuse. The name of the class is also no longer accurate.

skschneider:
LGTM. Added a comment to the test code - let me know if you plan to address it or if we should merge this in as-is. If there's flag support, I also think we should build support into the configs. That doesn't need to be in this PR, but if not I'll create a separate issue for it when I merge this in.

wonderfly:
@skschneider 
> LGTM. Added a comment to the test code - let me know if you plan to address it or if we should merge this in as-is.

Thanks. I'll take a look at your comments tomorrow.
> If there's flag support, I also think we should build support into the configs.

What do you mean by "build into the configs" specifically? What are the configs? Sorry I'm less than 1% familiar with this codebase.

skschneider:
PerfKit Benchmarker allows you specify configurations for a benchmark run either through the command-line flags (gflags) or yaml config files. If you take a look at the code change you made the gce_virtual_machine.py, you'll see that the gcloud command's 'image-project' value is read directly from the FLAGS variable (the PerfKit Benchmarker gflags). On the other hand, the 'image' value is read from the GceVirtualMachine, which is transferred in the __init__ method from the GceVmSpec (a derived class of BaseVmSpec). The values of the GceVmSpec constructor ultimately come from properties of the corresponding yaml config object that is loaded from the file specified by FLAGS.benchmark_config_file. The benefits of the config file are that it allows customization of different VM groups within a single benchmark, so you could have one VM provisioned with one image and another VM provisioned with a different image. Image and image-project are highly related, so if we support the image config option, I think we should also support an image_project config option.

skschneider:
FYI, #664 just added config option verifiers/decoders to VM specs. You don't need to wrap config changes into this PR, but the way it would be added would require:
1) Modify GceVmSpec._GetOptionDecoderConstructions to add a StringDecoder for the image_project config option.
2) Modify GceVmSpec.ApplyFlags to allow FLAGS.image_project to override the config value.
3) Modify GceVirtualMachine.__init__ to read the image_project from the spec.
4) Use that value in GceVirtualMachine._GenerateCreateCommand.

wonderfly:
@skschneider 
I'd rather leave the _config_ changes to you guys then. :) I addressed your earlier comments. PTAL at commit 6155bd1

skschneider:
Thanks for the change! I'll merge it.

PR #667, opened 2015-11-24
Broaden exception caught during PKB run.

cmccoy:
The current version does not catch KeyboardInterrupt or SystemExit (see [here](https://docs.python.org/2/library/exceptions.html#exceptions.KeyboardInterrupt)),
and we want to clean up in the first case. The exception is re-raised regardless.


voellm:
LGTM

PR #670, opened 2015-11-24
Add explicit --boot-disk-auto-delete to GCE VM creation.

skschneider:


cmccoy:
LGTM. Is this not working right now? I thought `--boot-disk-auto-delete` was the default.

skschneider:
It is the default, so there's no problem right now. @voellm suggested we make it explicit.

cmccoy:
Sounds good. Thanks.

PR #676, opened 2015-11-25
Add helper for mocking FLAGS across all modules.

skschneider:
No need to patch module1.FLAGS and module2.FLAGS separately.


perfkitbenchmarker/mock_flags_test.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
```
cmccoy: Can this go in tests/?

skschneider: Oops. Hold on, I'll clobber this branch and put it in the correct place.

Issue #564, opened 2015-10-20
Add a concept of 'machine_type_spec'.


mateusz-blaszkowski:
In Kubernetes (and Marathon which is a framework for Mesos) we don't have flavors. Passing a flavor using `--machine_type` flag doesn't make sense. We do have a possibility of specifying resources limits though (like memory and cpu). I'd like to create a common flag for these resources which will be used by some of the providers (like Kubernetes and Mesos). I would like to know your opinion before I start the implementation. Here are the solutions I see:

* leave current `--machine_type` flag untouched and introduce a second flag: `--machine_type_spec`. Resource limits would be passed in the following way: `--machine_type_spec=|cpu:2|,|memory:1024|` - we would have to agree on the format. I have seen that you already use this format in the JSON with benchmarks results but I'm open for new ideas.
* similar to the previous one, but resources limits would be passed using current `--machine_type` flag, i.e. `--machine_type=|cpu:2|,|memory:1024|`
* introduce new flag for each resource type, i.e. `--cpus=2`, `--memory=1024`. 

Personally I would stick to the first option. Please let me know what you think about this. Thanks!

cmccoy:
For Kubernetes/Mesos - are there other container parameters beyond CPU/memory? If not, I'm tempted by the second, but the first sounds fine too.

mateusz-blaszkowski:
Currently only CPU and memory, but for Kubernetes there is a plan to add other resource types (see 'Future resource types' section in http://kubernetes.io/v1.0/docs/design/resources.html ).

In such case I will stick to the first option if that's OK for you too. Thanks :)

cmccoy:
Sounds great -  thanks.

skschneider:
I believe the existing vm_spec configs are a good place for new options like cpu and memory. For example, cluster_boot.vm_groups.default.vm_spec.Kubernetes.memory
I'm concerned about conflicts between new flags and the config system. If I understand the trend so far, flags can override the config, but the flags described in the first post do not target a particular VM group, which would make it difficult to use.

skschneider:
Spoke with ehankland, current behavior of the --machine_type flag also doesn't target a VM group and overrides settings of all VM groups in the config. Since that's the case, I'd be okay with it if the new flag behaves similarly, but I also recommend config support.

voellm:
Ok so it looks like we all agree this is a useful feature.  Anyone want to take it on?

voellm:
[...]

Issue #671, opened 2015-11-24
Maintain lists of supported benchmarks for each cloud provider


hildrum:
Each cloud provider should maintain a list of benchmarks supported.  

When a run of PerfKit Benchmarker is started, only the benchmarks supported by the provider will be run.  The skipped benchmarks will be reported.  

I don't currently see a natural place for this in the provider-specific files, so perhaps a new properties.py in the provider directory could contain this information.

voellm:
@mateusz-blaszkowski - Is there a set of "known" working benchmarks for OpenStack/Kubernetes?   Any not working as intended?

@syed -  Is there a set of "known" working benchmarks for CloudStack?   Any not working as intended?

@hicrazyboy  -  Is there a set of "known" working benchmarks for AliCloud?   Any not working as intended?

@meteorfox - Is there a set of "known" working benchmarks for Rackspace?   Any not working as intended?

mateusz-blaszkowski:
There is a `kubernetes_set` with a list of benchmarks that worked well on K8S:

```
    'kubernetes_set': {
        MESSAGE: 'Kubernetes benchmark set.',
        BENCHMARK_LIST: ['block_storage_workload', 'cassandra_ycsb',
                         'cassandra_stress', 'cluster_boot', 'fio', 'iperf',
                         'mesh_network', 'mongodb_ycsb', 'netperf', 'redis',
                         'sysbench_oltp']
    }
```

Unfortunately I can't verify which benchmarks were working on Openstack cloud. @kivio - do you have a list of working benchmarks for OS?

kivio:
@mateusz-blaszkowski @voellm - i have list of working benchmarks but from release 18.x, I haven't currently environment to test newest version of PKB

voellm:
We can go with "it works" until proven not to or err on know is known.  Any
preference?

kivio:
@voellm  Ok, we can assume "it works" for all. I want to test it in nearby future.

kivio:
Another thing - OpenStack driver is currently implemented and tested just for Ubuntu image.

voellm:
With the pull request complete is this feature complete?  If so please close.

hildrum:
Closing, as it is complete.  Please reopen if something is missing.

PR #679, opened 2015-11-27
Mesos-Marathon provider

mateusz-blaszkowski:
There is an assumption that Marathon is used as a Mesos framework in order
to spawn Docker instances. Currently only local disks are supported but
there is a plan to add a Ceph volume support (like in Kubernetes provider).

Both CPU and memory are specified as a separate flags but this will be
changed accordingly to #564.


mateusz-blaszkowski:
This is WIP - I plan to gather a list of working benchmarks and create a new benchmark set based on this.

I've got also a question regarding flags like `docker_in_privileged_mode` which are common for both K8S and Mesos. How should I treat such flags? Flag with the same name cannot be defined twice so currently I just added a 'mesos' prefix, but I don't consider it a good solution. I also don't want to leave Mesos without a flag (it would work because flag is defined in Kubernetes module) because it may confuse users who base on auto-generated CLI help - in such case no description related to `docker_in_privileged_mode` under Mesos section would be displayed.

voellm:
Google is off til Monday here so we will pick up the review then.

We working on a benchmark set like feature which maps working benchmarks to
providers. @hildrum might implement this.

Once we have it the mapping it will intersect the provider with and
benchmark set and only run those.

voellm:
test this please
ok to test

(messages to autobuilder)

hildrum:
Issue number tracking the providers to their supported benchmarks is #671

mateusz-blaszkowski:
FYI - I'll wait for PR #690 to be merged - than I will analogically add ProviderInfo for Mesos. Currently, the following benchmarks pass:
`aerospike_ycsb, block_storage_workload, cassandra_stress, cassandra_ycsb, cluster_boot, copy_throughput , fio, hbase_ycsb, hpcc, iperf, mongodb_ycsb, netperf, oldisim, ping, redis, redis_ycsb, scimark2, silo, sysbench_oltp`

voellm:
Do you need this in V1.0.0 or is the next release ok?  We are rolling up V1
now.

mateusz-blaszkowski:
When are you going to release v1.0.0 and the subsequent release?

voellm:
Release 1.0.0 on about a week.  Next release is likely early next year.

mateusz-blaszkowski:
In that case I would be grateful if we managed to merge this PR before v1.0.0. If there is anything I can do to help, just let me know.

README.md
```diff
@@ -278,6 +279,22 @@ You have two Ceph authentication options available (http://kubernetes.io/v1.0/ex
    ```
    You will have to pass the Secret name (using `--ceph_secret` flag) when running the benchmakrs. In this case it should be: `--ceph_secret=my-ceph-secret`.
 
+### Mesos configuration
+Mesos provider communicates with Marathon framework in order to manage Docker instances. Thus it is required to setup Marathon framework along with the Mesos cluster. In order to connect to Mesos you need to provide IP address and port to Marathon framework using `--marathon_address` flag.
+
+Currently only local disks are supported so you need to execute Perfkit with `--scratch_disk_type=LOCAL` flag.
```
ehankland: You can make the default disk type local (in default_config_constants.yaml) so that you don't always have to specify this flag

perfkitbenchmarker/providers/mesos/mesos_docker_instance.py
```diff
@@ -0,0 +1,206 @@
...
+  def _CreateVolumes(self):
+    """
+    Creates volumes for scratch disks.
+    """
+    for disk_num, disk_spec in enumerate(self.disk_specs):
+      if disk_spec.disk_type == disk.LOCAL:
+        scratch_disk = LocalDisk(disk_num, disk_spec, self.name)
+      else:
+        # TODO: support for Ceph
+        pass
```
ehankland: Maybe explicitly raise an exception here that mentions that Ceph isn't supported yet

perfkitbenchmarker/providers/mesos/mesos_docker_instance.py
```diff
@@ -0,0 +1,206 @@
...
+    API in order to create App.
+    """
+    cat_cmd = ['cat', vm_util.GetPublicKeyPath()]
+    key_file, _ = vm_util.IssueRetryableCommand(cat_cmd)
+    cmd = "/bin/mkdir /root/.ssh; echo '%s' >> /root/.ssh/authorized_keys; " \
+          "/usr/sbin/sshd -D" % key_file
+    body = {
+        'id': self.name,
+        'mem': FLAGS.docker_memory_mb,
+        'cpus': FLAGS.docker_cpus,
```
ehankland: Instead of directly using the docker_memory_mb and docker_cpus flags could you add a MesosVmSpec so that these can be set in configs? Here's an example: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/gcp/gce_virtual_machine.py#L96

ehankland:
Only had a few small comments other than making the cpus/memory options compatible with configs - LGTM once those are fixed

mateusz-blaszkowski:
@ehankland Thank you, I've addressed your comments and also adjusted to https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/690 PR.

ehankland:
LGTM

voellm:
Congrats!

PR #680, opened 2015-11-27
Fix for silo benchmark which fails when running behind proxy.

### LINKED ISSUE #491: silo benchmark fails behind proxy
From @mateusz-blaszkowski in #475: 

> silo - failed with Clone of 'git://github.com/kohler/masstree-beta.git' into submodule path 'masstree' failed. I run the test behind the proxy and this is the case. I would have changed the path to Git repository to https:// but it is hidden somewhere in 'dbtest' (look a the command which failed: cd /tmp/pkb/silo && MODE=perf DEBUG=0 CHECK_INVARIANTS=0 make -j80 dbtest). Oh, i found that the exact path is specified here: https://github.com/stephentu/silo/blob/cc11ca1ea949ef266ee12a9b1c310392519d9e3b/.gitmodules

We should switch it to `https://`.

mateusz-blaszkowski:
Resolves #491

### DIFF
--- a/perfkitbenchmarker/linux_packages/silo.py
+++ b/perfkitbenchmarker/linux_packages/silo.py
@@ -33,6 +33,9 @@ def _Install(vm):
   vm.RemoteCommand('git clone {0} {1}'.format(GIT_REPO, SILO_DIR))
   vm.RemoteCommand('cd {0} && git checkout {1}'.format(SILO_DIR,
                                                        GIT_TAG))
+  # This is due to a failing clone command when executing behind a proxy.
+  # Replacing the protocol to https instead of git fixes the issue.
+  vm.RemoteCommand('git config --global url."https://".insteadOf git://')
   vm.RemoteCommand('cd {0} && MODE=perf DEBUG=0 CHECK_INVARIANTS=0 make\
           -j{1} dbtest'.format(SILO_DIR, nthreads))
 


voellm:
test this please
ok to test 

(message to test bots)

perfkitbenchmarker/linux_packages/silo.py
```diff
@@ -33,6 +33,9 @@ def _Install(vm):
   vm.RemoteCommand('git clone {0} {1}'.format(GIT_REPO, SILO_DIR))
   vm.RemoteCommand('cd {0} && git checkout {1}'.format(SILO_DIR,
                                                        GIT_TAG))
+  # This is due to a failing clone command when executing behind a proxy.
+  # Replacing the protocol to https instead of git fixes the issue.
+  vm.RemoteCommand('git config --global url."https://".insteadOf git://')
```
voellm: Is the problem here the use of port 8080?  IT seems like other places curl is used to to proxy via http but here you are trying to move to ssh.

LGTM but just wanted to make sure I understood.

mateusz-blaszkowski: I switched to use https:// instead of git:// protocol - the latter one uses port 9418 which is blocked. Both http (80) and https (443) works fine. I'm quite confused with port 8080 and SSH - I don't mess up with these at all :)

voellm: Sounds good I had the direction wrong.

LGTM.

PR #681, opened 2015-11-27
Fix for not-installed curl package.

mateusz-blaszkowski:
Curl command was used without curl package installation.


voellm:
LGTM

PR #685, opened 2015-12-01
Make ping benchmark bi-directional and add a test

hildrum:
This makes the ping benchmark bi-directional and adds a test.  It also adds `sending_zone` and `receiving_zone` to the metadata.

Old results summary:
```
PING:
  ip_type="internal"
  Min Latency                          36.244000 ms                            
  Average Latency                      36.879000 ms                            
  Max Latency                          89.677000 ms                            
  Latency Std Dev                       5.310000 ms                            
  End to End Runtime                  290.041688 seconds  
```

New results summary:
```
PING:
  ip_type="internal"
  Min Latency                          36.198000 ms                             (receiving_zone="us-east1-b" sending_zone="us-central1-a")
  Average Latency                      37.434000 ms                             (receiving_zone="us-east1-b" sending_zone="us-central1-a")
  Max Latency                         149.869000 ms                             (receiving_zone="us-east1-b" sending_zone="us-central1-a")
  Latency Std Dev                      11.300000 ms                             (receiving_zone="us-east1-b" sending_zone="us-central1-a")
  Min Latency                          36.183000 ms                             (receiving_zone="us-central1-a" sending_zone="us-east1-b")
  Average Latency                      36.308000 ms                             (receiving_zone="us-central1-a" sending_zone="us-east1-b")
  Max Latency                          36.496000 ms                             (receiving_zone="us-central1-a" sending_zone="us-east1-b")
  Latency Std Dev                       0.188000 ms                             (receiving_zone="us-central1-a" sending_zone="us-east1-b")
  End to End Runtime                  409.803674 seconds                       

```


perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -62,17 +62,41 @@ def Run(benchmark_spec):
     A list of sample.Sample objects.
   """
   vms = benchmark_spec.vms
-  if not vms[0].IsReachable(vms[1]):
-    logging.warn('%s is not reachable from %s', vms[1], vms[0])
+  results = []
+  for origin_index in [0, 1]:
+    origin_vm = vms[origin_index]
```
voellm: How about callnig this sending_vm to match with the result label names.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -62,17 +62,41 @@ def Run(benchmark_spec):
...
+    receiving_vm = vms[origin_index ^ 1]
+    results = results + _RunPing(origin_vm,
```
voellm: give this is a loop of two it might be more clear to eliminate the loop.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -62,17 +62,41 @@ def Run(benchmark_spec):
...
+                                 receiving_vm,
+                                 receiving_vm.internal_ip,
```
voellm: While you are at it how about adding support for external ips as well?

There is code in iperf you can follow... https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py#L70

    if vm_util.ShouldRunOnExternalIpAddress():
      vm.AllowPort(IPERF_PORT)

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -62,17 +62,41 @@ def Run(benchmark_spec):
...
+                                 'internal')
+  return results
+
+
+def _RunPing(origin_vm, dest_vm, dest_ip, ip_type):
```
voellm: please change origin_vm to sending_vm to be consistent with the labels fields.

tests/linux_benchmarks/ping_benchmark_test.py
```diff
@@ -0,0 +1,43 @@
...
+    vm0 = mock.MagicMock()
+    vm1 = mock.MagicMock()
+    vm_spec.vms = [vm0, vm1]
+    path = os.path.join(os.path.dirname(__file__), '..', 'data', 'ping.out')
+    outfile = open(path, 'r')
+    pingstdout = outfile.read()
+    for vm in vm_spec.vms:
+      vm.RemoteCommand.side_effect = [(pingstdout, ' ')]
+    ping_benchmark.Prepare(vm_spec)
+    ping_benchmark.Run(vm_spec)
```
voellm: If you have Prepare how about adding Cleanup as well.  Both are a pass but I suspect you added it in case something changes in the future.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -62,17 +62,41 @@ def Run(benchmark_spec):
     A list of sample.Sample objects.
   """
   vms = benchmark_spec.vms
-  if not vms[0].IsReachable(vms[1]):
-    logging.warn('%s is not reachable from %s', vms[1], vms[0])
+  results = []
+  for origin_index in [0, 1]:
```
skschneider: Addendum: the brackets are unnecessary and generate a new list object. They can be omitted:
```
for sending_vm, receiving_vm in vms, reversed(vms):
```

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
hildrum: Allowing `ping` using external ips requires changes to the firewall (at least for gcp), and since ping uses ICMP, it didn't see how it could be done with `vm.AllowPort()`.  This makes a supporting external ips for ping a much bigger piece of work--either it requires changing the firewall for all benchmarks and (potentially) all providers, or adding some way to change the firewall specifically for the ping benchmark.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
hildrum: I made the change as you suggested.  I'm not sure how to add the check in CheckPrerequisites, though.  In the examples I see, CheckPrerequisites takes no arguments and it does check the value of the flag num_vms.   But if someone has a YAML config file with more more than two vms, will that cause the num_vms flags to change?

I could do the VM count check in the Prepare method instead of the CheckPrerequisites method.

skschneider: Ah, I guess CheckPrerequisites is outdated and only supports flags, not configs. In the long term, I think it makes sense to move the call of CheckPrerequisites after the benchmark spec is created but before resources are provisioned so that the benchmark has access to the config and can check conditions before anything is provisioned.

For a short term solution, the check in Prepare is fine. Alternately, you could set ```vms = benchmark_spec.vms[:2]``` to only use the first two VMs.

hildrum:
I had copied the for-loop syntax from the`iperf_benchmark.py` so I modified that benchmark to follow @skschneider's pattern as well.

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
```diff
@@ -65,6 +65,11 @@ def Prepare(benchmark_spec):
         required to run the benchmark.
   """
   vms = benchmark_spec.vms
+  if len(vms) != 2:
+        raise ValueError(
```
skschneider: Irregular spacing.

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
```diff
@@ -157,9 +162,7 @@ def Run(benchmark_spec):
   logging.info('Iperf Results:')
 
   # Send traffic in both directions
-  for originator in [0, 1]:
-    sending_vm = vms[originator]
-    receiving_vm = vms[originator ^ 1]
+  for (sending_vm, receiving_vm) in vms, reversed(vms):
```
skschneider: Nit: First parentheses not required. Same in other file.

tests/linux_benchmarks/ping_benchmark_test.py
```diff
@@ -0,0 +1,44 @@
...
+    ping_benchmark.Cleanup(vm_spec)
+
+    self.assertEquals(vm_spec.vms[0].RemoteCommand.call_count, 1)
+    self.assertEquals(vm_spec.vms[1].RemoteCommand.call_count, 1)
+
```
skschneider: Consider checking the generated Samples.

skschneider:
A few small comments, otherwise LGTM.

PR #688, opened 2015-12-01
openstack: Add support for using multiple data volumes

meteorfox:
Before this PR, data disks where not working. This PR fixes that and adds the ability to use multiple volumes as well.

- Fix Pickling error caused by use of generator in class attributes.
- Improve error handling of disk creation
- Add switch for enabling verbose http log

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_disk.py
+++ b/perfkitbenchmarker/providers/openstack/os_disk.py
@@ -15,9 +15,11 @@
 import logging
 import time
 
-from perfkitbenchmarker.providers.openstack import utils as os_utils
+
+from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import disk
+from perfkitbenchmarker.providers.openstack import utils as os_utils
 from perfkitbenchmarker.providers.openstack.utils import retry_authorization
 
 FLAGS = flags.FLAGS
@@ -34,7 +36,7 @@ class OpenStackDisk(disk.BaseDisk):
         self.name = name
         self.zone = zone
         self.device = ""
-        self.virtual_disks = (c for c in "cdefghijklmnopqrstuvwxyz")
+        self._disk = None
 
     def _Create(self):
         self._disk = self.__nclient.volumes.create(self.disk_size,
@@ -59,34 +61,32 @@ class OpenStackDisk(disk.BaseDisk):
             self.__nclient.volumes.delete(self._disk)
             is_deleted = False
             while not is_deleted:
-                is_deleted = len(self.__nclient.volumes.findall(
-                    display_name=self.name)) == 0
+                volume = self.__nclient.volumes.get(self._disk.id)
+                is_deleted = volume is None
                 time.sleep(sleep)
                 sleep_count += 1
                 if sleep_count == 10:
                     sleep = 5
-        except (os_utils.NotFound, os_utils.BadRequest):
+        except Exception:
             logging.info('Volume already deleted')
 
     def _Exists(self):
         try:
-            if len(self.__nclient.volumes.findall(display_name=self.name)):
-                return True
-            else:
-                return False
-        except (os_utils.NotFound, os_utils.BadRequest):
+            volume = self.__nclient.volumes.get(self._disk.id)
+            return volume and volume.status in ('available', 'in-use',
+                                                'deleting',)
+        except Exception:
             return False
 
     def Attach(self, vm):
         self.attached_vm_name = vm.name
         self.attached_vm_id = vm.id
-        device_hint_name = "/dev/vd" + self.virtual_disks.next()
-        result = self.__nclient.volumes.create_server_volume(vm.id,
-                                                             self._disk.id,
-                                                             device_hint_name)
-        self.attach_id = result.id
-        self.device = "/dev/disk/by-id/virtio-" + result.id[:20]
 
+        result = self.__nclient.volumes.create_server_volume(vm.id,
+                                                             self._disk.id,)
+        self.attach_id = result.id
+
+        volume = None
         is_unattached = True
         while is_unattached:
             time.sleep(1)
@@ -94,6 +94,13 @@ class OpenStackDisk(disk.BaseDisk):
             if volume:
                 is_unattached = not(volume.status == "in-use"
                                     and volume.attachments)
+
+        for attachment in volume.attachments:
+            if self.attach_id == attachment.get('volume_id'):
+                self.device = attachment.get('device')
+                return
+
+        raise errors.Error("Couldn't not attach volume to %s" % vm.name)
 
     def GetDevicePath(self):
         return self.device
--- a/perfkitbenchmarker/providers/openstack/os_disk.py
+++ b/perfkitbenchmarker/providers/openstack/os_disk.py
@@ -51,10 +51,11 @@ class OpenStackDisk(disk.BaseDisk):
             volume = self.__nclient.volumes.get(self._disk.id)
             if volume:
                 is_unavailable = not (volume.status == "available")
-        self._disk = volume
+                self._disk = volume
 
     @retry_authorization(max_retries=4)
     def _Delete(self):
+        from novaclient.exceptions import NotFound
         sleep = 1
         sleep_count = 0
         try:
@@ -67,15 +68,17 @@ class OpenStackDisk(disk.BaseDisk):
                 sleep_count += 1
                 if sleep_count == 10:
                     sleep = 5
-        except Exception:
-            logging.info('Volume already deleted')
+        except NotFound:
+            logging.info('Volume %s not found, might have been already deleted'
+                         % self._disk.id)
 
     def _Exists(self):
+        from novaclient.exceptions import NotFound
         try:
             volume = self.__nclient.volumes.get(self._disk.id)
             return volume and volume.status in ('available', 'in-use',
                                                 'deleting',)
-        except Exception:
+        except NotFound:
             return False
 
     def Attach(self, vm):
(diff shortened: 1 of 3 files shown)


meteorfox:
PTAL @kivio @cmccoy

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, disk_spec, name, zone, image=None):
         self.name = name
         self.zone = zone
         self.device = ""
-        self.virtual_disks = (c for c in "cdefghijklmnopqrstuvwxyz")
```
meteorfox: This was causing a PicklingError when the spec was being pickled. The error was caused by the generator.

cmccoy:
Looks good to me.

meteorfox:
@cmccoy Thanks for the CR.

I'll wait until end of day, or until tomorrow for @kivio 's CR, since he developed most of the Openstack bindings; I want to make sure it doesn't break any of his stuff.

voellm:
FYI we added a "Provider" contact list here -
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/wiki/Provider-contacts

This should help is getting reviews from the original developers.  Feel
free to update.

On Wed, Dec 2, 2015 at 10:52 AM, Carlos Torres <notifications@github.com>
wrote:

> @cmccoy <https://github.com/cmccoy> Thanks for the CR.
>
> I'll wait until end of day, or until tomorrow for @kivio
> <https://github.com/kivio> 's CR, since he developed most of the
> Openstack bindings; I want to make sure it doesn't break any of his stuff.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/688#issuecomment-161396470>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

kivio:
@meteorfox looks good for me :) I haven't currently environment to test it :(

meteorfox:
@kivio Thanks for reviewing.

@cmccoy Can I merge?

9723:
@meteorfox , Run error in my env version liberty. I don't know how to fix it yeah.

default_config_constants.yaml:

  OpenStack:
    machine_type: m1.small
    zone: nova
    image: ubuntu-pkb
  OpenStack:
    disk_type: standard
    disk_size: 10
    mount_point: /scratch

./pkb.py --cloud=OpenStack --benchmarks=fio --noinstall_packages \
--openstack_public_network=external_network --openstack_private_network=internal_network

2015-12-04 10:11:56,727 0c7659b8 Thread-1 fio(1/1) ERROR    Got exception running FormatDisk: Got non-zero return code (1) executing sudo mke2fs -F -E lazy_itable_init=0 -O ^has_journal -t ext4 -b 4096 /dev/disk/by-id/virtio-ef21dff6-2e72-4a4e-9

mke2fs 1.42.9 (4-Feb-2014)
mke2fs: No such file or directory while trying to determine filesystem size

meteorfox:
@9723 Thanks for testing!

Hmm.. what's your hypervisor backend, is it KVM?

Also, do you consistently hit the error every time you try to run it or is it occasionally?

9723:
@meteorfox ,  yes, I got the error every time when running fio benchmark and hypervisor backend use KVM.

meteorfox:
@9723 ok, thanks for catching the bug.

I'm investigating the issue, most likely the bug is because it assumes all device paths for attachments start with `/dev/disk/by-id/virtio-` which might not always be the case. We should instead be using the `device_path` returned by the `attachments` property.

meteorfox:
@9723 Can you pull again and try the latest commit and check if it fixed the problem?

Before this last commit, I was able to replicate the issue but not consistently. After this last commit, I ran several tests with fio using 1 and 4 volumes and it has been consistently working for me.

Let me know if it works for you. Thanks!

meteorfox:
I caught another bug. It's not deleting all the volumes when using more than one.

meteorfox:
@kivio Ok, turned out that the exception classes in `OpenStack/utils.py` were not really catching the errors thrown by the `novaclient` and it was aborting the thread earlier than expected, making it skip the delete part of the volumes. I couldn't import the exceptions from the `novaclient` package since those will break lazy loading, so I instead opted for catching the generic `Exception`. 

@9723 Ok, this time should work and also clean up the volumes.

9723:
@meteorfox :+1:  LGTM, Thanks. tested in liberty.

kivio:
@9723 @meteorfox i known this error on OpenStack Juno long since but this is little problematic. First thing, path returned by nova client is just suggestion not sure path. It's described in Nova Client documentation. Second thing when you start more than one volume in one time you can have race condition with /virtio- path. For me best try was getting next /dev/vd(bced..) on each machine. But it still is not that easy. This problem is based on KVM i don't known how other hypervisors on OpenStack works with it.

voellm:
Should we cherry pick this into v1.0.0?  About ready?

On Fri, Dec 4, 2015 at 12:09 AM, Marcin Karkocha <notifications@github.com>
wrote:

> @9723 <https://github.com/9723> @meteorfox <https://github.com/meteorfox>
> i known this error on OpenStack Juno long since but this is little
> problematic. First thing, path returned by nova client is just suggestion
> not sure path. It's described in Nova Client documentation. Second thing
> when you start more than one volume in one time you can have race condition
> with /virtio- path. For me best try was getting next /dev/vd(bced..) on
> each machine. But it still is not that easy. This problem is based on KVM i
> don't known how other hypervisors on OpenStack works with it.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/688#issuecomment-161904885>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -95,6 +95,13 @@ def Attach(self, vm):
                 is_unattached = not(volume.status == "in-use"
                                     and volume.attachments)
 
+        for attachment in volume.attachments:
+            if self.attach_id == attachment.get('volume_id'):
+                self.device = attachment.get('device')
```
meteorfox: @kivio I think using the 'device' property from the attachments list of dict objects we get back from os-volume_attachements should be OK.

If you noticed at line 85, instead of giving a device hint, I just let OpenStack return the device path it got, then I just retrieve the actual device path once the volume is attached. I believe this is reliable, and will work for other kind of hypervisors.

For later releases of OpenStack, we will need to introduce the python-cinderclient library for volume management since these operations have been deprecated out of the nova client.

meteorfox:
@voellm I'd say yes, it's working for both @9723 and me, but I want to wait for @kivio response since he's the maintainer.

kivio:
@meteorfox ok, sounds good to me. :+1:

meteorfox:
@kivio I think the code addresses those concerns, and it works for Kilo (tested by me) and Liberty (tested by @9723 ). I would like to merge this change since is very low risk, and we have confidence that it works. If that's OK with you

meteorfox:
@kivio Sorry, I didn't see your comment, disregard the comment above.

@voellm @cmccoy OK to merge ?

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -59,34 +61,32 @@ def _Delete(self):
...
-                is_deleted = len(self.__nclient.volumes.findall(
-                    display_name=self.name)) == 0
+                volume = self.__nclient.volumes.get(self._disk.id)
+                is_deleted = volume is None
                 time.sleep(sleep)
                 sleep_count += 1
                 if sleep_count == 10:
                     sleep = 5
-        except (os_utils.NotFound, os_utils.BadRequest):
+        except Exception:
```
cmccoy: Would it be worth including some information on this in the logging statement? This could swallow an exception unrelated to the volume not being found...

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -59,34 +61,32 @@ def _Delete(self):
...
         try:
-            if len(self.__nclient.volumes.findall(display_name=self.name)):
-                return True
-            else:
-                return False
-        except (os_utils.NotFound, os_utils.BadRequest):
+            volume = self.__nclient.volumes.get(self._disk.id)
+            return volume and volume.status in ('available', 'in-use',
+                                                'deleting',)
+        except Exception:
```
cmccoy: As above.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -95,6 +95,13 @@ def Attach(self, vm):
                 is_unattached = not(volume.status == "in-use"
                                     and volume.attachments)
 
+        for attachment in volume.attachments:
```
cmccoy: Is it ever acceptable for no attachment to match?

cmccoy:
@meteorfox: I just added a few questions.

perfkitbenchmarker/providers/openstack/os_disk.py
meteorfox: @cmccoy Good point. I can append the exception message to the log.

perfkitbenchmarker/providers/openstack/os_disk.py
meteorfox: Agree

perfkitbenchmarker/providers/openstack/os_disk.py
meteorfox: Well, at this point since the while loop above never times out until there is an attachment, it will always have a matching attachment.

Perhaps, I could improve it by making it timeout, and throwing an appropriate error message.

cmccoy: Oh, sorry, I didn't catch that it was part of a larger loop. Thanks!

meteorfox:
@cmccoy: I've narrowed down the exception scope to just the errors it expect to handle. I brought back the imports, directly in `os_disk.py` and `os_virtual_machine.py` but I'm lazy loading them. Tested it again and actually found a cause were the volume failed to create and was throwing a None Exception.

cmccoy:
@meterfox - Thanks. :+1:

meteorfox:
Thanks everyone for the CRs

PR #692, opened 2015-12-03
Enable trim during format and mount.

yuyantingzero:


yuyantingzero:
Should I enable discard for rackspace machines as well? @voellm @ehankland

meteorfox:
@yuyantingzero feel free to add `discard`. The reason the method are overridden (Format, Mount) is to handle the case where we use a directory in the boot partition as the scratch disk or data disk for a benchmark. Looking at it now, though, this has to be revised.

meteorfox:
@yuyantingzero What's the effect enabling `discard` on an HDD? Is it NOOP?

yuyantingzero:
@meteorfox , trim will be ignored on HDD.

voellm:
@yuyantingzero Please post performance comparative performance numbers and then we can LGTM.  Thanks.

voellm:
@mateusz-blaszkowski @kivio  - Any concerns on your part?

mateusz-blaszkowski:
@voellm I'm OK with the change

meteorfox:
@yuyantingzero @voellm I'm not sure if this was waiting for my reply, but just in case, I'm also OK with the change.

voellm:
Time to merge.

PR #694, opened 2015-12-04
Make bucket name a function in object storage benchmark

noahl:
Previously, the API-based object storage benchmarks could not be used when the run
stage was a separate PKB invocation than the prepare stage, because the run
stage depended on a variable defined in the prepare stage. This change
fixes that issue.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -561,9 +561,10 @@ class S3StorageBenchmark(object):
     vm.PushFile(FLAGS.object_storage_credential_file, AWS_CREDENTIAL_LOCATION)
     vm.PushFile(FLAGS.boto_file_location, DEFAULT_BOTO_LOCATION)
 
-    self.bucket_name = 'pkb%s' % FLAGS.run_uri
+    vm.bucket_name = 'pkb%s' % FLAGS.run_uri
+
     vm.RemoteCommand(
-        'aws s3 mb s3://%s --region=us-east-1' % self.bucket_name)
+        'aws s3 mb s3://%s --region=us-east-1' % vm.bucket_name)
 
   def Run(self, vm, metadata):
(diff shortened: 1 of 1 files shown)


noahl:

stfeng:
If you follow the original code, you will see the bucket name is assigned to self.bucket_name at the prepare stage, and then later referred to by the run stage by reading from "self.bucket_name". This is the right paradigm and layering: the prepare stage creates a resource and saves the name, and the run stage should strictly depend on the name of the resource created and saved by the prepare stage, instead of coming up a name on the fly. 

The real issue here is the benchmark spec "self" is not being pickled properly: looks like bucket_name is not being pickled properly to the JSON file representing this run, therefore if you invoke run() separately, you won't be able to read the bucket_name from the saved benchmark spec. We should either fix the pickling logic of benchmark spec so it saves bucket_name properly, or save the bucket_name to the "vm" object during the prepare stage (because every field of the VM object is pickled properly, that I'm pretty sure). 

Actually I think saving the bucket name to the VM object is a better design, because it' will enable us to prepare different VMs that can target different buckets (e.g., each in a different region targeting their own regional buckets).

Thoughts? We can chat more about this.

noahl:
I see your point about why generating bucket_name in the Prepare stage is good design. Therefore the issue is to pickle bucket_name properly.

The reason bucket_name is not being pickled right now is that it's not attached to the benchmark_spec object. It's attached to the provider-specific object that comes from OBJECT_STORAGE_BENCHMARK_DICTIONARY, which is not saved anywhere.
I think in the long run we might want a list of bucket_names attached to a VM, or even a more complicated state, because I can imagine wanting one VM to use more than one bucket, for instance to see the difference in latency to different providers. However, I think we can go ahead and attach the bucket name to the VM object now, and change it later if we need to. I'll push an updated PR soon.

noahl:
@stfeng Please take a look. I tested the current patch by running the api_data benchmarks on GCP (for storage and VMs), doing each stage individually. It worked.

stfeng:
LGTM. thanks for the effort in testing and validating this.

PR #703, opened 2015-12-09
Delete set-interrupts.sh

ehankland:
This file doesn't get used


yuyantingzero:
I think we should use set-interrupts.sh for gcp local ssds. https://cloud.google.com/compute/docs/disks/local-ssd#runscript
Ubuntu didn't set interrupts by default.

yuyantingzero:
Talked offline, LGTM

PR #707, opened 2015-12-10
Fix for benchmark_compatibility_checking flag.

mateusz-blaszkowski:
Comparing string values using 'is' always returned False because the string
objects were not identical (although they were assigned equal value).


hildrum:
LGTM, thanks.

ehankland:
Thanks for the fix - merging now so that we can cherry-pick this.

PR #726, opened 2015-12-15
Create AWS/GCE default access rules before creating instance.

cmccoy:
A better fix for #717.

Azure endpoints target individual instances, so they can't be created before the VM.
On AWS, `authorize-security-group-ingress` uses the default security group from the instance create command.

This PR just modifies GCP and AWS instance creation, creating remote access rules prior to creating the VM. Azure requires the VM to be created before authorizing ingress.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -340,9 +340,9 @@ class BenchmarkSpec(object):
     logging.info('Waiting for boot completion.')
     for port in vm.remote_access_ports:
       vm.AllowPort(port)
+    vm.WaitForBootCompletion()
     vm.AddMetadata(benchmark=self.name, perfkit_uuid=self.uuid,
                    benchmark_uid=self.uid)
-    vm.WaitForBootCompletion()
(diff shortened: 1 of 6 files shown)


cmccoy:
test this please.

cmccoy:
test this please.

cmccoy:
test this please.

cmccoy:
@ehankland: could you PTAL?

ehankland:
LGTM - I think both Rackspace and AliCloud also use firewalls - can you let their authors know about this change?

voellm:
We have provider contacts on this page...
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/wiki/Provider-contacts

cmccoy:
cc: @meteorfox, @hicrazyboy: 
This change is aimed at measuring cluster boot time more accurately. Generally when creating a VM, the requisite access rules would already be in place, so I've changed GCP and AWS to create access rules before starting the VM-to-SSH timer. I'm not set up with Rackspace or AliCloud, so no changes there (but welcome!).

PR #727, opened 2015-12-15
Move GCE custom VM YAML options.

skschneider:
Allow string or object for machine_type config option in GCE VM spec.
Make VM 'cpus' and 'memory' options members of the machine_type object.
Due to the mutually exclusive nature of machine_type string or
cpus&memory combination, this organization allows easier use via the
--machine_type or --config_override flags.
Update README.md to demonstrate use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,10 +588,11 @@ PerfKit Benchmaker.
 
 Flag | Notes
 -----|------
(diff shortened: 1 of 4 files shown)


ehankland:
LGTM

PR #729, opened 2015-12-15
Make quantiles symmetric. Fixes #718

### LINKED ISSUE #718: object_storage_service benchmark quantiles should be symmetrical
The summary output for object_storage_service lists p1, p5, p50, p90, p99 and p99.9, which is great for metrics where lower is better (e.g. latency) because you can see with good resolution how long the tail is, but for metrics where higher is better (e.g. throughput) there's less data around the "bad" side of the tail. Would be nice to have p0.1, p1, [p5,] p10, p50, p90, [p95,] p99, p99.9.

zbjornson:
Note that this also affects some other test reporting (at least the mysql benchmark).


skschneider:
Thank you, looks good to me. I'll merge it in.

PR #730, opened 2015-12-15
Add mock_flags.PatchTestCaseFlags.

skschneider:
New method can be called once from a TestCase test method or setUp
method to patch flags for the entire test.


hildrum:
LGTM

PR #675, opened 2015-11-25
Create BaseSpec. Move functionality from BaseVmSpec to BaseSpec.

skschneider:
Related to #668.
Diffbase #672 (will rebase & replace this branch after that PR is in).

PR #731, opened 2015-12-15
Fix for the FloatDecoder used in Mesos provider.

mateusz-blaszkowski:
I must have done the merge improperly.

### DIFF
--- a/perfkitbenchmarker/configs/option_decoders.py
+++ b/perfkitbenchmarker/configs/option_decoders.py
@@ -187,33 +187,38 @@ class FloatDecoder(TypeVerifier):
     min: None or float. If provided, it specifies the minimum accepted value.
   """
 
-  def __init__(self, component, option, max=None, min=None, **kwargs):
-    super(FloatDecoder, self).__init__(component, option, (float,), **kwargs)
+  def __init__(self, option, max=None, min=None, **kwargs):
+    super(FloatDecoder, self).__init__(option, (float,), **kwargs)
     self.max = max
     self.min = min
 
-  def Decode(self, value):
-    """Verifies that the provided value is an int.
+  def Decode(self, value, component_full_name, flag_values):
+    """Verifies that the provided value is a float.
 
     Args:
       value: The value specified in the config.
+      component_full_name: string. Fully qualified name of the configurable
+          component containing the config option.
+      flag_values: flags.FlagValues. Runtime flag values to be propagated to
+          BaseSpec constructors.
 
     Returns:
-      int. The valid value.
+      float. The valid value.
 
     Raises:
       errors.Config.InvalidValue upon invalid input value.
     """
-    value = super(FloatDecoder, self).Decode(value)
+    value = super(FloatDecoder, self).Decode(value, component_full_name,
+                                           flag_values)
     if value is not None:
       if self.max and value > self.max:
         raise errors.Config.InvalidValue(
-            'Invalid {0} "{1}" value: "{2}". Value must be at most '
-            '{3}.'.format(self.component, self.option, value, self.max))
+            'Invalid {0}.{1} value: "{2}". Value must be at most '
+            '{3}.'.format(component_full_name, self.option, value, self.max))
       if self.min and value < self.min:
         raise errors.Config.InvalidValue(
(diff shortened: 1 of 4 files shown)


perfkitbenchmarker/configs/option_decoders.py
```diff
@@ -187,33 +187,38 @@ class FloatDecoder(TypeVerifier):
...
     Returns:
-      int. The valid value.
+      float. The valid value.
 
     Raises:
       errors.Config.InvalidValue upon invalid input value.
     """
-    value = super(FloatDecoder, self).Decode(value)
+    value = super(FloatDecoder, self).Decode(value, component_full_name,
+                                           flag_values)
```
ehankland: Looks like flake8 is complaining about the indent level of this line

perfkitbenchmarker/configs/option_decoders.py
```diff
@@ -187,33 +187,38 @@ class FloatDecoder(TypeVerifier):
     min: None or float. If provided, it specifies the minimum accepted value.
   """
 
-  def __init__(self, component, option, max=None, min=None, **kwargs):
-    super(FloatDecoder, self).__init__(component, option, (float,), **kwargs)
+  def __init__(self, option, max=None, min=None, **kwargs):
+    super(FloatDecoder, self).__init__(option, (float,), **kwargs)
```
skschneider: Sorry, I wasn't aware when this class was introduced, so it was my changes that broke it.
I have a concern about the use of (float,) here. Because of the way yaml.load works, an input like 'docker_cpus: 2' will become an int that I don't think would be accepted by this FloatDecoder. It's probably best to accept int as well. Consider adding a test case to tests/configs/option_decoders_test.py

skschneider:
One more thing I found that was broken by #675:
MesosDockerSpec.ApplyFlags. This should now be _ApplyFlags to match the base class's _ApplyFlags (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/virtual_machine.py#L94).
A summary of why this was changed: prior to #675, ApplyFlags was called after the spec was constructed and therefore after the decoders' Decode method was called. After #675, _ApplyFlags is called during the spec construction before the decoder's Decode method is called. This way, Decode is called on the provided value regardless of whether it was found in the config file or overridden by the command-line flag, which is helpful for config options that need to do verification that is more complicated than just type checking.

mateusz-blaszkowski:
Sorry for the delay! I'm not able to fix this right now unfortunately. I'll work on this at the beginning of the next week. It shouldn't affect anyone though.

mateusz-blaszkowski:
@skschneider thanks for the explanation. I've improved the FloatDecoder as suggested.

perfkitbenchmarker/providers/mesos/mesos_docker_instance.py
```diff
@@ -36,6 +36,8 @@ class MesosDockerSpec(virtual_machine.BaseVmSpec):
   Attributes:
     docker_cpus: None or float. Number of CPUs for Docker instances.
     docker_memory_mb: None or int. Memory limit (in MB) for Docker instances.
+    mesos_privileged_docker: None of boolean. Indicates if Docker container
```
skschneider: "of" -> "or"

perfkitbenchmarker/configs/option_decoders.py
```diff
@@ -187,33 +187,38 @@ class FloatDecoder(TypeVerifier):
...
+      float. The valid value.
 
     Raises:
       errors.Config.InvalidValue upon invalid input value.
     """
-    value = super(FloatDecoder, self).Decode(value)
+    value = super(FloatDecoder, self).Decode(value, component_full_name,
+                                             flag_values)
     if value is not None:
       if self.max and value > self.max:
```
skschneider: This should probably be ```if self.max is not None and...``` to handle 0 properly (self.min and IntDecoder have this problem too). I can fix this and add tests as a later pull request.

skschneider:
One little spelling mistake. Otherwise LGTM. You can merge when ready.

skschneider:
Considering the current FloatDecoder is broken, I'm going to merge this in for the version 1.1.0 release.

PR #733, opened 2015-12-15
openstack: Use only available floating-ips

meteorfox:
Prior this patch when it searches for a floating ip to use, it
might return floating IPs that might have been already allocated
for another instance and causes a conflict error.

This patch make sures that only 'Down', or available, floating ips
are requested.


meteorfox:
@kivio @cmccoy PTAL thanks

cmccoy:
+1

skschneider:
Ping. Should this be merged for the current release (v1.1.0)?

kivio:
+1

meteorfox:
@skschneider Yes please

PR #734, opened 2015-12-15
openstack: Expose availability zone

meteorfox:
Allows to override availability zone value to place instances on
a specific host or host aggregate.

### DIFF
--- a/perfkitbenchmarker/providers/openstack/flags.py
+++ b/perfkitbenchmarker/providers/openstack/flags.py
@@ -63,3 +63,6 @@ flags.DEFINE_enum('openstack_scheduler_policy', NONE,
                   [NONE, 'affinity', 'anti-affinity'],
                   'Add possibility to use affinity or anti-affinity '
                   'policy in scheduling process')
+
+flags.DEFINE_string('openstack_availability_zone', None,
+                    'Name of availability zone to create instance(s) on.')
--- a/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
@@ -95,6 +95,8 @@ class OpenStackVirtualMachine(virtual_machine.BaseVirtualMachine):
                               'destination_type': 'volume',
                               'delete_on_termination': True}]
 
+        availability_zone = FLAGS.openstack_availability_zone or self.zone
+
         vm = self.client.servers.create(
             name=self.name,
             image=image_id,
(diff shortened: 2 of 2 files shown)


meteorfox:
@kivio @cmccoy PTAL

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -63,3 +63,6 @@
                   [NONE, 'affinity', 'anti-affinity'],
                   'Add possibility to use affinity or anti-affinity '
                   'policy in scheduling process')
+
+flags.DEFINE_string('openstack_availability_zone', None,
```
cmccoy: How does this interact with the `--zones` flag?

meteorfox: @cmccoy 

First, let me clarify the difference between an `availability zone` in OpenStack and a `zone` in PKB terms. 

In OpenStack, administrators can create what's called an Host Aggregrate. A Host Aggregate is a grouping of hosts. A host can be in multiple host aggregates. Optionally, OpenStack administrators can expose these host aggregates to their users as Availability Zones. By default,  most OpenStack deployment use a single region, and usually all hosts are in a single host aggregate that's called `nova`.  Which is why in the OpenStack PKB bindings we set `zone` to `nova` as the default, and essentially only treat it as zone, instead of a region. This has a problem though, an availabilitiy zone !=  zone in GCP, at least in OpenStack terms, where like in AWS we interpret zones as separate from regions.

One use-case in OpenStack, which is why I exposed this flag, is that users can potentially target a particular host within an Availability Zone, by using this format, `<availability_zone_name>:<host_name>`.  If we were to set the `zone` flag to this, it would try to apply it everywhere the `zone` flag is used but fail since when creating other resources such as volumes, networks, and floating ips, such format is not supported.

For example, it returns this error:
```
File "/opt/PerfKitBenchmarker/pkb/local/lib/python2.7/site-packages/novaclient/client.py", line 391, in request
    raise exceptions.from_response(resp, body, url, method)
BadRequest: Invalid input received: Availability zone 'nova:711465-compute007' is invalid (HTTP 400) (Request-ID: req-d53930ff-260f-4cef-968b-f98d034f9070)
```

The flag `openstack_availability_zone` overrides the argument  for the zone, only when the instance is being created, but for other uses cases it defaults to the original zone.

cmccoy:
Thanks for the explanation! :+1:

kivio:
:+1: LGTM

skschneider:
Ping. Should this be merged for the current release (v1.1.0)?

meteorfox:
@skschneider Not yet. I want to re-work this PR due issue #776

voellm:
Ping ... ready to integrate?

meteorfox:
@voellm This is related to #776 and #770 as well. Maybe we should close the PRs. Once I have the changes ready I'll push a new one.

noahl:
@meteorfox Given your last message, it sounds like we should close this. I'll close, but please reopen if I'm wrong.

PR #736, opened 2015-12-15
Fix GceWindowsVirtualMachine Boot Disk Type

noahl:
Boot disk type no longer works with new GCE boot disk creation
mechanism.


noahl:
This looks like a variable that wasn't translated during the disk types refactor, although I'm not entirely sure. Anyway, what happens is that the BOOT_DISK_TYPE member of a GceVirtualMachine is passed to gcloud to initialize a disk. Normally this is fine, but WindowsGceVirtualMachine overrides this method and makes it 'remote_ssd', which is not a valid GCE disk type. This patch fixes that.

@ehankland @yuyantingzero please take a look.

ehankland:
LGTM

PR #743, opened 2015-12-16
Let disk objects return Windows DeviceIds

noahl:


noahl:
@yuyantingzero @ehankland please take a look.

noahl:
@skschneider please take a look.

perfkitbenchmarker/disk.py
```diff
@@ -53,6 +53,8 @@
 REGION = 'region'
 LEGACY_DISK_TYPE = 'legacy_disk_type'
 
+WINDOWS = 'windows'
```
skschneider: This should be referencing the original constant to avoid re-declaration, but the original is in benchmark_spec.py, so that's probably impractical as-is. Ideally, we should drag all the os_type constants out of benchmark_spec.py into a separate module.

noahl: I actually tried referencing the one in windows_virtual_machine.WindowsMixin.OS_TYPE, but got a recursive import. I think benchmark_spec.py would have the same problem. I agree that we should move the os_type constants into a separate module.

skschneider:
LGTM. I'll open an issue about os_type problems.

skschneider:
Opened #745.

PR #751, opened 2015-12-17
Speed up Cassandra cluster boot.

cmccoy:
By setting "auto_bootstrap" to false (only acceptable for new clusters),
multiple nodes can join in a shorter period of time. This significantly speeds up startup.

### DIFF
--- a/perfkitbenchmarker/linux_packages/cassandra.py
+++ b/perfkitbenchmarker/linux_packages/cassandra.py
@@ -45,9 +45,9 @@ NODETOOL = posixpath.join(CASSANDRA_DIR, 'bin', 'nodetool')
 
 # Number of times to attempt to start the cluster.
 CLUSTER_START_TRIES = 10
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -193,7 +193,7 @@ def StartCluster(seed_vm, vms):
   logging.info('Starting seed VM %s', seed_vm)
   Start(seed_vm)
   logging.info('Waiting %ds for seed to start', NODE_START_SLEEP)
-  time.sleep(NODE_START_SLEEP)
+  time.sleep(NODE_START_SLEEP * 2)
```
skschneider: Why 2 here? Looks like it retries anyway, and this doesn't match the log in the previous line.

skschneider:
Just one comment. LGTM.

cmccoy:
Removed the `*2`. Thanks.

Issue #735, opened 2015-12-15
Windows installation got a lot harder recently


noahl:
Adding paramiko to the requirements.txt file means that we need to build pycrypto when installing dependencies, which means that the user needs Visual C++ installed to build PKB on Windows. We should either document this new requirement or change it. (paramiko is only required for AliCloud, and since we lazy-load providers, you don't actually need to install it unless you want to use AliCloud from Windows.)

voellm:
Its only harder if you want to use Alicloud.  It is not impacted otherwise.  So that feature and the complexity go together.

There are lots of threads on StackOverflow on how to install paramiko on Windows.  Here is one such example - http://stackoverflow.com/questions/2964658/installing-paramiko-on-windows

noahl:
The problem is that if you do 'pip install -r requirements.txt', it will automatically install paramiko. I had to figure out what the problem was and manually edit my requirements.txt to fix it, but I don't think that's reasonable for a non-PKB-developer. Ideally we would let you install PKB without paramiko and just complain if you tried to use Alicloud. I think it's doable.

cmccoy:
Let's split out `requirements.txt` into core requirements (required to run PerfKitBenchmarker) and individual files for cloud providers with additional dependencies.

cmccoy:
The default requirements.txt no longer introduces any dependencies that require a compiler toolchain.

PR #753, opened 2015-12-17
Add provider-specific requirements.txt files

cmccoy:
Adds a `requirements-<provider-name>.txt` file for cloud providers with additional dependencies.
Changes installation instructions to refer to those files, rather than installing packages manually with `pip`.
Some of these are trivial (a single package), but it seems worth it to me to a consistent way to install provider Python dependencies.

Addresses #735.

cc: @meteorfox, @hicrazyboy, @syed, @kivio

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -1,0 +1,31 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# All requirements for running PerfKitBenchmarker unit tests.
+
+# Standard PerfKitBenchmarker requirements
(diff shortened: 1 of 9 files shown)


requirements-testing.txt
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# All requirements for running PerfKitBenchmarker unit tests.
+-rrequirements.txt
+-rrequirements-alicloud.txt
```
noahl: Does this line include requirements-alicloud in requirements-testing? I don't think we want that behavior.

noahl:
Other than my single comment, I am very excited about this change. LGTM.

requirements-testing.txt
cmccoy: Yes - for `providers_test` to pass, we need AliCloud and CloudStack dependencies. We could make those conditional. Generally, I feel okay having the requirements to run unit tests be a bit higher than just running code.  Are you wanting to run unit tests on windows?

noahl: No, I don't want to run unit tests on windows. It seemed like a somewhat-unnecessary dependency, but I understand if we need it to pass `providers_test`. Why not add all of the provider-specific requirements files, then, to be consistent?

cmccoy: These are the only requirements files required to run unit tests - everything else installs a command-line tool or conditionally-imported module. I'll make a note.

noahl: Fine with me. LGTM.

PR #758, opened 2015-12-17
Move all provider names to provider_info and use those constants

hildrum:
Move the provider name constants from `benchmark_spec.py` to `provider_info.py` and adjust the references those names accordingly.

I added a test for object_storage_service_benchmark in this same pull request because all the tests passed even though I'd forgotten to make a change on `object_storage_service_benchmark.py`.

### DIFF
--- a/perfkitbenchmarker/provider_info.py
+++ b/perfkitbenchmarker/provider_info.py
@@ -14,10 +14,10 @@
 
 """Module containing class for provider data
 
-Currently, this is only used for IsBenchmarkSupported
+This contains the BaseProviderInfo class which is
+used for IsBenchmarkSupported
 
 """
-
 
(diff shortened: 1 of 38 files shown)


hildrum:
I moved the provider names to `providers/__init.py__` and adjusted the paths accordingly.

perfkitbenchmarker/provider_info.py
```diff
@@ -14,11 +14,12 @@
 
 """Module containing class for provider data
 
-Currently, this is only used for IsBenchmarkSupported
+This contains constants for names of all the providers,
```
ehankland: This isn't true now that you've moved the constants

ehankland:
LGTM once you've fixed the above

PR #759, opened 2015-12-18
openstack: Add support for using Root and Local disks as data disks

meteorfox:
This PR adds the ability to use local disks that come with a machine type or flavor, and allows a user to use an empty directory path in the root disk ('/') to be used as the "scratch disk".

Using `ROOT` disks for benchmarks that either need more than 1 data disk per VM, or that try to access the device path of the disk will trigger an exception notifying the user that such usage is not allowed.

The 3 data disk type supported as of this PR:
- `VOLUME` (default) -> OpenStack Block Storage Volumes
- `LOCAL`                  -> Ephemeral data disks included with the flavor
- `ROOT`                   -> Empty directory under `/`   e.g. `/scratch`

I didn't make a distinction regarding the media type of the disks since this can change between OpenStack deployments. A possible solution for this problem is to include an OpenStack specific flag, to specify which media type is desired when using VOLUME.

Here's an example on how to run using the root disk executing the `fio` benchmark.

```bash
./pkb.py --cloud=OpenStack --machine_type=m1.medium --image='Ubuntu 14.04' \
         --openstack_public_network=public --openstack_private_network=service \
         --benchmarks=fio \
         --data_disk_type=root --data_disk_size=5 --num_striped_disks=1
```

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_disk.py
+++ b/perfkitbenchmarker/providers/openstack/os_disk.py
@@ -24,96 +24,198 @@ from perfkitbenchmarker.providers.openstack.utils import retry_authorization
 
 FLAGS = flags.FLAGS
 
+ROOT = 'root'
+LOCAL = 'local'
+VOLUME = 'volume'
+
+DISK_TYPE = {disk.STANDARD: VOLUME, disk.REMOTE_SSD: VOLUME, disk.LOCAL: LOCAL}
+
+DISK_METADATA = {
+    VOLUME: {
+        disk.LEGACY_DISK_TYPE: disk.STANDARD,
+        disk.REPLICATION: disk.REGION
+    },
+    ROOT: {
+        disk.REPLICATION: None
+    },
+    LOCAL: {
+        disk.LEGACY_DISK_TYPE: disk.LOCAL
+    }
+}
+
+OPENSTACK = 'OpenStack'
+disk.RegisterDiskTypeMap(OPENSTACK, DISK_TYPE)
+
 
 class OpenStackDisk(disk.BaseDisk):
+  """
+  Base class for OpenStack disks.
+  """
+  def __init__(self, disk_spec, name):
+    super(OpenStackDisk, self).__init__(disk_spec)
+    self.name = name
 
-    def __init__(self, disk_spec, name, zone, image=None):
-        super(OpenStackDisk, self).__init__(disk_spec)
-        self.__nclient = os_utils.NovaClient()
-        self.attached_vm_name = None
-        self.attached_vm_id = -1
-        self.image = image
-        self.name = name
-        self.zone = zone
-        self.device = ""
-        self._disk = None
+  def _Create(self):
+    pass
 
-    def _Create(self):
-        self._disk = self.__nclient.volumes.create(self.disk_size,
-                                                   display_name=self.name,
-                                                   availability_zone=self.zone,
-                                                   imageRef=self.image,
-                                                   )
+  def _Delete(self):
+    pass
 
-        is_unavailable = True
-        while is_unavailable:
-            time.sleep(1)
-            volume = self.__nclient.volumes.get(self._disk.id)
-            if volume:
-                is_unavailable = not (volume.status == "available")
-                self._disk = volume
+  def Attach(self, vm):
+    pass
 
-    @retry_authorization(max_retries=4)
-    def _Delete(self):
-        from novaclient.exceptions import NotFound
+  def Detach(self):
+    pass
 
-        if self._disk is None:
-            logging.info('Volume %s was not created. Skipping deletion.'
-                         % self.name)
-            return
 
-        sleep = 1
-        sleep_count = 0
-        try:
-            self.__nclient.volumes.delete(self._disk)
-            is_deleted = False
-            while not is_deleted:
-                volume = self.__nclient.volumes.get(self._disk.id)
-                is_deleted = volume is None
-                time.sleep(sleep)
-                sleep_count += 1
-                if sleep_count == 10:
-                    sleep = 5
-        except NotFound:
-            logging.info('Volume %s not found, might have been already deleted'
-                         % self._disk.id)
+class OpenStackBlockStorageDisk(OpenStackDisk):
+  """
+  OpenStackBlockStorageDisk is an OpenStackDisk object to represent a
+  remote block storage volume that can be attached to an instance.
+  """
 
-    def _Exists(self):
-        from novaclient.exceptions import NotFound
-        try:
-            volume = self.__nclient.volumes.get(self._disk.id)
-            return volume and volume.status in ('available', 'in-use',
-                                                'deleting',)
-        except NotFound:
-            return False
+  def __init__(self, disk_spec, name, zone):
+    super(OpenStackBlockStorageDisk, self).__init__(disk_spec, name)
+    self.__nclient = os_utils.NovaClient()
+    self.attached_vm_name = None
+    self.attached_vm_id = -1
+    self.name = 'cinder-volume-%s' % name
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -197,6 +267,43 @@ def DeleteKeyfile(self):
         except NotFound:
             logging.info("Deleting key doesn't exists")
 
+    def _GetBlockDevices(self):
```
meteorfox: This method should probably be promoted to the `BaseLinuxMixin` class

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -197,6 +267,43 @@ def DeleteKeyfile(self):
...
+        colnames = ('name', 'model', 'size_bytes',
+                    'type', 'mountpoint', 'label',)
+        blk_devices = [dict(zip(colnames, t)) for t in tuples]
+        for d in blk_devices:
+            d['model'] = d['model'].rstrip()
+            d['label'] = d['label'].rstrip()
+            d['size_bytes'] = int(d['size_bytes'])
+        return blk_devices
+
+    def _GetBootDevice(self):
```
meteorfox: This one as well. Once this PR gets merged, on a separate branch this can refactored to the `BaseLinuxMixin` class. Since it's duplicated on `rackspace_virtual_machine.py` as well.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+    pass
+
+  def Attach(self, vm):
+    pass
+
+  def Detach(self):
+    pass
+
+
+class OpenStackBlockStorageDisk(OpenStackDisk):
```
noahl: Consider naming this `OpenStackRemoteBlockStorageDisk` instead, because otherwise it sounds like it could represent a local drive as well as a remote one.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+    self.zone = zone
+    self.device = ""
+    self._disk = None
+
+  def _Create(self):
+    self._disk = self.__nclient.volumes.create(self.disk_size,
+                                               display_name=self.name,
+                                               availability_zone=self.zone,)
+    is_unavailable = True
+    while is_unavailable:
```
noahl: You should use `@Retry` from vm_util.py to handle the retries. That will automatically back off the timing, and raise an exception if it tries too many times.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+      logging.info('Volume %s was not created. Skipping deletion.'
+                   % self.name)
+      return
+
+    sleep = 1
+    sleep_count = 0
+    try:
+      self.__nclient.volumes.delete(self._disk)
+      is_deleted = False
+      while not is_deleted:
```
noahl: Again, use `@Retry` with a local function for the retries.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+                                          'deleting',)
+    except NotFound:
+      return False
+
+  def Attach(self, vm):
+    self.attached_vm_name = vm.name
+    self.attached_vm_id = vm.id
+
+    result = self.__nclient.volumes.create_server_volume(vm.id,
+                                                         self._disk.id,)
```
noahl: Nit: trailing `,` is unnecessary.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+    self.attach_id = result.id
+
+    volume = None
+    is_unattached = True
+    while is_unattached:
```
noahl: Use `@Retry` as above.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
...
+      device_path: string Linux device path of the local disk. e.g. '/dev/vdb'
+    """
+    super(OpenStackRootDisk, self).__init__(disk_spec, device_path, 'root-disk')
+    self.mount_point = disk_spec.mount_point
+
+  def GetDevicePath(self):
+    """
+    Root disk is already mounted. Formatting of of root disk is not permitted.
+    """
+    raise errors.Error('Formatting of root disk block device is not supported.')
```
noahl: I don't think we should raise an exception here, because it is possible to get the device path of a boot disk, and we might want to. For instance, a benchmark might `ls -l` the device path to figure out how big the disk is.

It might make more sense to have an `Unmount()` or `FormatDevice()` method, and either of those would raise an exception for a boot disk.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -24,96 +24,198 @@
 
 FLAGS = flags.FLAGS
 
+ROOT = 'root'
```
noahl: I think we call this disk type "boot" elsewhere in the code, so please consider that name. However, thanks to the disk refactor, OpenStack has its own disk type namespace, so if you think "root" is better, that's fine.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -58,6 +64,9 @@ def __init__(self, vm_spec):
         self.user_name = self.DEFAULT_USERNAME
         self.boot_wait_time = None
         self.image = self.image or self.DEFAULT_IMAGE
+        self.boot_device = None
+        self.root_disk_allocated = False
```
noahl: Use of "boot" and "root" seems inconsistent here. If they mean the same thing, please standardize on one of them.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -162,14 +171,75 @@ def WaitForBootCompletion(self):
...
+        self.boot_device = self._GetBootDevice()
 
-        self._CreateScratchDiskFromDisks(disk_spec, disks)
+    def CreateScratchDisk(self, disk_spec):
+        """
+        Creates instances of OpenStackDisk child classes depending on disk type.
+        """
+        if disk_spec.disk_type == os_disk.ROOT:  # Ignore num_striped_disks
+            if self.root_disk_allocated:
+                raise errors.Error('Only a Root disk can be created per VM')
```
noahl: Nit: "Only a Root disk" --> "Only one Root disk"

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -162,14 +171,75 @@ def WaitForBootCompletion(self):
...
+            device_path = '/dev/%s' % self.boot_device['name']
```
noahl: This could be `self.boot_device.GetDevicePath()` if you accept my comment above about not throwing an exception in `GetDevicePath()`.

noahl:
I left some comments, but overall, I like this PR. I'm excited for it to be merged.

meteorfox:
@noahl Thanks! I just got back from vacation, I'll resume my work on this.

hildrum:
Should this go into release 1.3?

meteorfox:
@hildrum Not for this release. I still need to update with @noahl 's feedback.

meteorfox:
Closing, need to rework this PR with the new provider implementation.

PR #760, opened 2015-12-18
Increase preload thread count for YCSB.

cmccoy:
The current default was set for HBase, which buffers mutations and sends
them in batches. 1 loading thread makes for very long
`{MongoDB,Aerospike,Cassandra}` YCSB tests. Changing to 32, and
overriding the default for HBase / Cloud Bigtable.


skschneider:
Seems strange that the default for some benchmarks is 1 thread when the flag's stated default is 32. Is there a reason not to change them all to 32? Can it work if the flag default is None instead?

cmccoy:
@skschneider  - switched to `None`.

HBase and Cloud Bigtable should keep the current value of `1` - inserts to those backends use a buffered mutator which multiplexes RPCs internally. Added a comment noting this.

skschneider:
LGTM

Issue #761, opened 2015-12-18
Problem running on Windows VMs


ehankland:
Running any benchmark with Windows VMs gives me this error:
AttributeError: 'WindowsGceVirtualMachine' object has no attribute 'PrepareBackgroundWorkload'

PR #762, opened 2015-12-18
Ensure that windows vms still work with background_cpu_threads

hildrum:
Fix bug which made windows VMs give errors even when background_cpu_threads was not set.


hildrum:
Fixes #761

ehankland:
LGTM

PR #764, opened 2015-12-20
Add os_type for Juju Mixin

AdamIsrael:
Hi Tony,

This is still a work-in-progress, with some papercuts that need to be cleaned up (including squashing my merge of upstream/master, documentation, and adding support for clouds other than AWS). I believe this branch is more inline with our email discussion, and I want to make sure that the way I implemented the cloud orchestration piece (installing and using juju) makes sense, since that was the one less well-defined. 

To run:
`./pkb.py --cloud=AWS --os_type=juju --benchmarks=cassandra_stress`


voellm:
Adam - Thanks.  I am out the rest of the year.  How about we sink up on
video chat.  Let's walk the code together.

AdamIsrael:
Hey Tony. A sync up sounds good. I'm back in the office starting today.

marcoceppi:
I confirm and have signed the CLA

voellm:
I verified it as well...
Canonical LimitedGoogle Corporate CLA

On Mon, Jan 4, 2016 at 2:31 PM, Marco Ceppi <notifications@github.com>
wrote:

> I confirm and have signed the CLA
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/764#issuecomment-168831636>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

voellm:
Adam - Since Canonical signed the CLA and you also sign it so the Googlebot is happy?

Thanks for the update changes.  I'll take a look and if would be great if ehankland can as well

voellm:
My quick pass on this is it looks much better integrated.  Nice work!

Will this have special instructions to set up Juju config files?  If so can you also update the README.md on how to use Juju?

I'll start a more detailed look soon.

AdamIsrael:
@googlebot I confirm and have signed the CLA

AdamIsrael:
I will update the README tomorrow, with instructions for exercising the juju provider, as well as how to add support to a package.

marcoceppi:
@googlebot  I have signed the CLA

voellm:
@ehankland  - Eric has kindly signed up for the indepth review.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,59 +195,34 @@ def ConstructVirtualMachines(self):
     """Constructs the BenchmarkSpec's VirtualMachine objects."""
     vm_group_specs = self.config.vm_groups
 
-    zone_index = 0
+    if FLAGS.os_type == os_types.JUJU:
```
ehankland: I noticed a few places where you check FLAGS.os_type - in order to support configs you should check vm.OS_TYPE (or the group spec's os_type) instead (since in some instances FLAGS.os_type may not be os_types.JUJU, but the VM will still have that type).

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+    environments:
+        perfkit:
+            type: manual
+            bootstrap-host: {0}
+    """
+
+    def juju_bootstrap(self):
+        resp, _ = self.RemoteHostCommand("juju bootstrap")
+        pass
```
ehankland: nit: pass is unnecessary here (and a few places below)

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+            os.unlink(f.name)
+        pass
+
+    def juju_environment(self, environment=''):
+        # TODO: set environment on demand
+        try:
+            output, _ = self.RemoteHostCommand('juju switch')
+            return output.strip()
+        except:
+            pass
```
ehankland: Why are you eating exceptions here (and below)? I see that you do return None if the commands fail, but it doesn't look like all of the return values get checked. I'd prefer to just let the exception get raised unless there's a good reason not to.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    def Wait(self, timeout=900):
+        """
+        Wait until all services are deployed, related or idle
+        """
+
+        # This should do while start+timeout < time()
+        while True:
+            ready = True
+            status = yaml.load(self.juju_status())
+            # logging.warn(status)
```
ehankland: nit: please remove commented out code

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+                for unit in status['services'][service]['units']:
+                    unit_data = status['services'][service]['units'][unit]
+                    ag = unit_data['agent-state']
+                    if ready and ag != 'started':
+                        ready = False
+
+                    ws = unit_data['workload-status']['current']
+                    if ready and ws not in ['active', 'unknown']:
+                        ready = False
+            # TODO: Exit loop if timeout exceeded
```
ehankland: Please implement this

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    environments_yaml = """
+    default: perfkit
+
+    environments:
+        perfkit:
+            type: manual
+            bootstrap-host: {0}
+    """
+
+    def juju_bootstrap(self):
```
ehankland: Nit: to keep consistent with the rest of the codebase could you please use CamelCase for methods? And if the method is private, you can prefix the method name with '_'.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+                # Install the Juju agent on the other VMs
+                for unit in self.units:
+                    unit.controller = self
+                    self.juju_addmachine(unit.internal_ip)
+
+            except errors.VirtualMachine.RemoteCommandError as e:
+                raise e
+
+    @vm_util.Retry(log_errors=False, poll_interval=1)
```
ehankland: Unless I'm missing something, it doesn't look like you need this method defined here

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -954,3 +956,191 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+        ' '.join(params)
+        try:
+            output, _ = self.RemoteHostCommand(
+                'juju set %s %s' % (service, ' '.join(params)))
+            return output.strip()
+        except:
+            pass
+
+
+    def Wait(self, timeout=900):
```
ehankland: nit: for methods that are going to be called elsewhere could you try working juju into the name so that it's obvious where the method comes from? vm.Wait() looks very generic whereas vm.JujuWait() (or similar) would make things more clear

voellm:
Ping ... looks like this pull request is out of sync and there are some comments.  Need any help?

AdamIsrael:
Hey Tony! I'll have a chance to revisit this next week. I'll address the (very useful) comments and the merge conflict.

AdamIsrael:
Hi @ehankland! I've updated this pull request based on your comments. Please let me know if there's anything else you'd like me to address.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,59 +195,35 @@ def ConstructVirtualMachines(self):
     """Constructs the BenchmarkSpec's VirtualMachine objects."""
     vm_group_specs = self.config.vm_groups
 
-    zone_index = 0
+    os_type = vm_group_specs[vm_group_specs.keys()[0]].os_type
+    if os_type == os_types.JUJU:
```
ehankland: This only tests if the first group (which isn't well defined since its not an ordereddict) has os_type == JUJU.

You probably want a controller if any group has os_type == JUJU

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,59 +195,35 @@ def ConstructVirtualMachines(self):
...
+        """
+        The Juju VM needs to be created first, so that subsequent units can
+        be properly added under its control.
+        """
+
+        juju_spec = copy.copy(vm_group_specs[vm_group_specs.keys()[0]])
```
ehankland: This probably needs to change as well

Not many benchmarks can make use of it, but different groups can span clouds. Since the controller talks over internal ips, a controller on one cloud wouldn't be able to talk to a group on another cloud. At the very least you need a controller per cloud, but a controller per group might be simpler

perfkitbenchmarker/benchmark_spec.py
AdamIsrael: Ok, I'll take a look at that. My assumption was that, when the os_type == JUJU, all of the vm_groups would share that. Is there a scenario when that is not the case?

ehankland: Users can assign arbitrary os_types to each group via config files - I haven't seen anyone using it that way, but the capability exists so I'd like to continue supporting it as much as possible

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,171 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+                self.AuthenticateVm()
+
+                self.JujuBootstrap()
+
+                # Install the Juju agent on the other VMs
+                for unit in self.units:
+                    unit.controller = self
+                    self.JujuAddMachine(unit.internal_ip)
+
+            except errors.VirtualMachine.RemoteCommandError as e:
```
ehankland: Why catch if you are just going to re-raise?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,171 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+                raise e
+
+    # @vm_util.Retry(log_errors=False, poll_interval=1)
```
ehankland: Please remove

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,171 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+            # Check if the charm is already deployed?
+            resp, _ = self.RemoteHostCommand(
+                "juju deploy %s --num-units=%d" % (charm, units))
+
+            return True
+
+    def JujuRelate(self, a, b):
+        resp, _ = self.RemoteHostCommand("juju add-relation %s %s" % (a, b))
+
+    def AuthenticateVm(self):
```
ehankland: Do you need this?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,171 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+        super(JujuMixin, self).AuthenticateVm()
+
+    def Install(self, package_name):
+        """Installs a PerfKit package on the VM."""
+        if package_name not in self._installed_packages:
+            package = linux_packages.PACKAGES[package_name]
+            package.JujuInstall(self)
```
ehankland: Maybe fall back to AptInstall if JuJuInstall hasn't been defined?

perfkitbenchmarker/linux_benchmarks/cassandra_stress_benchmark.py
```diff
@@ -355,9 +355,13 @@ def RunTestOnLoader(vm, loader_index, operations_per_vm, data_node_ips,
     population_dist: string. The population distribution.
     population_params: string. Representing additional population parameters.
   """
+
+  command = FLAGS.cassandra_stress_command
```
ehankland: Whats the reason for this? It looks like "command" gets passed in as an argument to this function

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -90,6 +95,27 @@ def AptInstall(vm):
...
+        # those in the benchmark_spec
+        vm.JujuDeploy('cs:trusty/cassandra', 3)
+        vm.JujuSet('cassandra', ['authenticator=AllowAllAuthenticator'])
+        vm.JujuDeploy('cs:~marcoceppi/trusty/cassandra-stress')
+        vm.JujuRelate('cassandra', 'cassandra-stress')
+
+        # This will wait for the cassandra cluster and cassandra-stress vms
+        # to fully deploy and be ready for benchmarking
+        if not vm.JujuWait(timeout=3600):
+            print "JujuWait failed!"
```
ehankland: Please change from print to a log statement

ehankland: Actually you may just want to log inside JuJuWait instead if it times out

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -90,6 +95,27 @@ def AptInstall(vm):
   _Install(vm)
 
 
+def JujuInstall(vm):
+    if vm.isController is True:
```
ehankland: This only gets called on the controller because the cassandra_stress benchmark calls 

vm_util.RunThreaded(lambda vm: vm.Install('cassandra'), benchmark_spec.vms)

In general you can't expect benchmarks to install all packages on all VMs. Instead, calling install on a  (non-controller) juju vm should probably trigger the controller vm installing the package on all vms in that group - it would be nice if this were generalized so that benchmarks don't have to explicitly support it

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,171 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+class JujuMixin(DebianMixin):
+    """
+    Class to allow running juju-deployed workloads
+
+    Bootstraps a Juju environment using the manual provider:
+    https://jujucharms.com/docs/stable/config-manual
+    """
+
+    OS_TYPE = 'juju'
```
ehankland: Please use the os_types.JUJU constant

AdamIsrael:
Hey @ehankland! I've finished iterating on your great feedback -- I think the CreateVirtualMachines method is much more elegantly handled now, thanks to that refactor. I look forward to any additional feedback in order to see this new functionality land.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,158 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+          unit_data = status['services'][service]['units'][unit]
+          ag = unit_data['agent-state']
+          if ready and ag != 'started':
+            ready = False
+
+          ws = unit_data['workload-status']['current']
+          if ready and ws not in ['active', 'unknown']:
+            ready = False
+
+      if start + timeout < time.time():
```
ehankland: I don't think this does what you intended it to do - it looks like it will always set ready to false after the timeout is reached instead of returning/timing out

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,158 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+        'juju set %s %s' % (service, ' '.join(params)))
+    return output.strip()
+
+  def JujuWait(self, timeout=900):
+    """
+    Wait until all services are deployed, related or idle
+    """
+
+    start = time.time()
+    while True:
```
ehankland: Consider using the vm_util.Retry function decorator instead of using your own polling logic

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,55 +195,33 @@ def ConstructVirtualMachines(self):
...
-        if disk_spec:
-          vm.disk_specs = [copy.copy(disk_spec) for _ in xrange(disk_count)]
-          # In the event that we need to create multiple disks from the same
-          # DiskSpec, we need to ensure that they have different mount points.
-          if (disk_count > 1 and disk_spec.mount_point):
-            for i, spec in enumerate(vm.disk_specs):
-              spec.mount_point += str(i)
-        vms.append(vm)
+    clouds = {}
+    for idx, (group_name, group_spec) in enumerate(vm_group_specs.iteritems()):
```
ehankland: It doesn't look like you use "idx" anywhere - maybe don't use enumerate unless you need it

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -116,6 +116,62 @@ def RedirectGlobalFlags(self):
     with flag_util.FlagDictSubstitution(FLAGS, lambda: self.config.flags):
       yield
 
+  def ConstructVirtualMachine(self, group_name, group_spec):
```
ehankland: I think this name is a bit misleading (because there is a C"onstructVirtualMachines" I would assume a method named "ConstructVirtualMachine" only operates on one VM at a time). Maybe change it so that it's clear this creates a single *group* of VMs

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -187,45 +227,51 @@ def StartCluster(seed_vm, vms):
...
-  logging.info('Starting seed VM %s', seed_vm)
-  Start(seed_vm)
-  logging.info('Waiting %ds for seed to start', NODE_START_SLEEP)
-  time.sleep(NODE_START_SLEEP)
-  for i in xrange(5):
-    if not IsRunning(seed_vm):
-      logging.warn('Seed %s: Cassandra not running yet (try %d). Waiting %ds.',
-                   seed_vm, i, NODE_START_SLEEP)
-      time.sleep(NODE_START_SLEEP)
+  if seed_vm.OS_TYPE != JUJU:
```
ehankland: Instead of 
```
if vm.OS_TYPE != JUJU:
  rest of function body
```
Maybe do
```
if vm.OS_TYPE == JUJU:
  return
rest of function body
```
This would make it easier to keep track of indent levels (this would be most helpful here since it's the longest function, but it might be nice to make the other functions match)

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,55 +195,33 @@ def ConstructVirtualMachines(self):
...
+              jujuvm.isController = True
+
+              clouds[group_spec.cloud] = jujuvm
+
+          for vm in vms:
+            vm.controller = clouds[group_spec.cloud]
+
+          jujuvm.units.extend(vms)
+          if jujuvm and jujuvm not in self.vms:
+            vms.extend([jujuvm])
```
ehankland: Adding the controller vm to vms will ultimately add the juju controller to the vm group. Since many benchmarks automatically scale to the number of vms in the group, this is probably going to break them if the controller doesn't behave like any other member of the group. 

I probably should have asked this before, but can the controller add itself as a unit? If that's the case, we could always have the first VM in a group be the controller - this would save us from ever having to create extra controller vms or keep track of the cloud

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -117,15 +142,17 @@ def Start(vm):
   Args:
     vm: The target vm. Should already be configured via 'Configure'.
   """
-  vm.RemoteCommand(
-      'nohup {0}/bin/cassandra -p "{1}" 1> {2} 2> {3} &'.format(
-          CASSANDRA_DIR, CASSANDRA_PID, CASSANDRA_OUT, CASSANDRA_ERR))
+  if vm.OS_TYPE != JUJU:
```
ehankland: please reference os_types.JUJU directly rather than redefining a JUJU constant in this module

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,158 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+  def _Bootstrap(self):
+    resp, _ = self.RemoteHostCommand("juju bootstrap")
+
+  def JujuAddMachine(self, ip):
+    resp, _ = self.RemoteHostCommand("juju add-machine ssh:%s" % ip)
+
+  def JujuConfigureEnvironment(self):
+    if self.isController:
+      resp, _ = self.RemoteHostCommand("mkdir -p ~/.juju")
+      f = tempfile.NamedTemporaryFile(delete=False)
```
ehankland: FYI we have a vm_util.NamedTemporaryFile that can be used with a "with" statement and still be closed inside the statement. I'd recommend using that

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +976,158 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    """
+    output, _ = self.RemoteHostCommand('juju status --format=json')
+    return output.strip()
+
+  def JujuVersion(self):
+    output, _ = self.RemoteHostCommand('juju version')
+    return output.strip()
+
+  def JujuSet(self, service, params=[]):
+    ' '.join(params)
```
ehankland: delete?

perfkitbenchmarker/benchmark_spec.py
marcoceppi: To address the second half of the question, the controller node /can/ have a unit of a service deployed to it, but not every unit can. For example, the controller has mongodb and a few additional packages running so a mongodb unit can't be deployed there. As a result of this we discourage units being deployed to the controller unless the user knows what they're getting into

ehankland: Okay - thanks for explaining - it would have been nice if we didn't need the extra controller VMs.

Going back to my first point, we can't have a VM in a group that doesn't behave like the other VMs  without benchmark support  (and I don't think requiring benchmark support is a good idea). What I'd suggest is to create a separate group for the controller (or one per cloud) rather than extending an existing group.

AdamIsrael: I think I need some clarification about creating a separate group for the controller. If I don't add it to vms, I presume I'd need to modify the Provision and Delete methods to run against the Juju controller(s), similar to they now run against vms.. Is that what you'd suggest?

ehankland: Ah - sorry for being unclear. I meant the local variable `vms` not `self.vms`. Controllers should definitely be part of the `self.vms` list, but not part of the same group as the units they are controlling.

So if you figure out you need a controller on a particular cloud you can copy the spec just like you are doing now, but instead of 
```
          if jujuvm and jujuvm not in self.vms:
             vms.extend([jujuvm])
```
Just create a new group  (maybe named '{cloud}_juju_controller' or similar) in `self.vm_groups` and also add the controller to `self.vms`.

Let me know if you need additional clarification.

AdamIsrael:
Thanks @ehankland, that makes perfect sense now! I think I've addressed all of the issues you raised now. Let me know if there's anything else you'd like me to address!

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -90,6 +94,25 @@ def AptInstall(vm):
   _Install(vm)
 
 
+def JujuInstall(vm):
+  if vm.isController is True:
+    if vm.vm_group_specs and 'workers' in vm.vm_group_specs:
+      vm.JujuDeploy('cs:trusty/cassandra',
+                    vm.vm_group_specs['workers'].vm_count)
```
ehankland: How are you guaranteeing that the cassandra units will be deployed to the workers group and not the loader/client?

AdamIsrael:
Good catch wrt placement within the vm_groups. I've added logic to track which group a vm belongs to and install as appropriate. Previously it was doing a round-robin install on the next available vm.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,229 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+      vm_group: The name of vm_group the unit(s) should be deployed to.
+    """
+    machines = []
+    if vm_group:
+      """Deploy (and scale) this service to the machines in its vm group."""
+
+      for unit in self.units:
+        """Find all machines in this vm_group"""
+        if unit.vm_group == vm_group:
+          machines.append(self._FindUnit(unit.internal_ip))
```
ehankland: Not a huge deal, but it might be nice to keep track of machine ids so if you ever need to deploy more than one charm to the same group you don't have to find them again. It would also help if you didn't have to make one juju status call per vm but could just make one call. Up to you if you want to fix it now.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,229 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+      # Get the name of the service
+      service = charm[charm.rindex('/') + 1:]
+
+      # Deploy to the remaining machine(s)
+      for machine in machines:
+        resp, _ = self.RemoteHostCommand(
+            "juju add-unit %s --to %s" % (service, machine))
+    else:
+      # If the vm_group isn't specified,
```
ehankland: I don't think we ever want to install a charm/package if we don't know which vms they are getting installed to. I'd feel more comfortable making the vm group required and deleting this case

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -90,6 +94,27 @@ def AptInstall(vm):
   _Install(vm)
 
 
+def JujuInstall(vm):
+  if vm.isController is True:
+    if vm.vm_group_specs and 'workers' in vm.vm_group_specs:
+      vm.JujuDeploy('cs:trusty/cassandra',
+                    vm.vm_group_specs['workers'].vm_count,
+                    'workers')
```
ehankland: I don't think we should tie the package installation to the benchmark - there could be future benchmarks that will require a cassandra cluster that don't name that group "workers" (right now there are two cassandra benchmarks, both of which happen to have named those groups workers, but this won't be true for all packages/benchmarks).

Instead I'd suggest you slightly rewrite JuJuMixin.Install() and this function:

```
  def Install(self, package_name):
     """Installs a PerfKit package on the VM."""
     package = linux_packages.PACKAGES[package_name]
     try:
       with self.controller.installation_lock: # Make sure another unit doesn't try to install the charm at the same time
         if package_name not in self.controller._installed_packages:
           package.JujuInstall(self.controller, self.group_name)  
           self.controller._installed_packages.add(package_name)
   
       except AttributeError:
         if package_name not in self._installed_packages:
           package.AptInstall(self)
           self._installed_packages.add(package_name)
```
The above assumes you also modify ConstructVirtualMachines to add "group_name" to each vm.
Let me know if you need more clarification.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -139,55 +195,36 @@ def ConstructVirtualMachines(self):
...
-          vm.disk_specs = [copy.copy(disk_spec) for _ in xrange(disk_count)]
-          # In the event that we need to create multiple disks from the same
-          # DiskSpec, we need to ensure that they have different mount points.
-          if (disk_count > 1 and disk_spec.mount_point):
-            for i, spec in enumerate(vm.disk_specs):
-              spec.mount_point += str(i)
-        vms.append(vm)
+      vms = self.ConstructVirtualMachineGroup(group_name, group_spec)
+
+      if group_spec.os_type == os_types.JUJU:
```
ehankland: It would be nice if the code under this "if" could live inside a helper method (_ConstructJuJuController?) - it would make reading the main provisioning logic easier

ehankland:
@AdamIsrael This is looking pretty good - there are only a couple of issues left.

perfkitbenchmarker/linux_packages/cassandra.py
AdamIsrael: There's a certain part of the Juju installation that has to be tied to the benchmark, i.e., the cassandra charm needs to be installed into the workers group, and the cassandra-stress charm into the clients group. I'm not sure abstracting that out and still being able to maintain that benchmark-specific installation, unless I'm totally missing your point which is entirely possible.

ehankland: It sounds like the `cassandra_stress` tool needs to be its own package if that's the case. The package itself should not need knowledge of the benchmark. This [line](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_benchmarks/cassandra_stress_benchmark.py#L320) should change to
```
vm_util.RunThreaded(lambda vm: vm.Install('cassandra'), benchmark_spec.vm_groups['workers'])
vm_util.RunThreaded(lambda vm: vm.Install('cassandra_stress'), benchmark_spec.vm_groups['client'])
```

Your `JujuInstall(controller, group_name)` function should only install charms on one group. That way if I want to reuse the `cassandra` package in another benchmark I can install the charm on a single group without having `cassandra_stress` getting installed on any of the other vms

ehankland: Here's the change I was describing: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/938
Does that make sense now?

AdamIsrael: I think I see what you're saying, about how other benchmarks may reuse the cassandra package (such as cassandra_ycsb).

So the above code would work to install both charms into the right groups. There'd need to be another logical step or two added, though, to run `vm.JujuRelate('cassandra', 'cassandra-stress')` and `vm.JujuWait()`. The relate could be called from the Install in the cassandra-stress package, since that package would require cassandra being installed, and `JujuWait()` could be installed in the package's StartCluster() method, instead of returning if os_types.JUJU.

Am I on the right track with that?

ehankland: Ah - I hadn't really noticed that you are referencing both charms on a single line... yeah that was the idea but the relate makes it more difficult. What exactly is the relate doing? The benchmark already references the ip addresses of the workers when the stress tool is run - why is the additional "add-relation" step needed?

AdamIsrael: Relations in Juju generally handles configuration details between two services. Between something like mysql and wordpress, the relation between the two would trigger the creation of the database, users, etc. needed by wordpress on mysql's end, which would then send the appropriate credentials to the wordpress service.

In the case of cassandra-stress, the relation authorizes the vm(s) running cassandra-stress to access the cassandra cluster and does any preparations required for the two services to work together.

AdamIsrael: I think the Relate line can be integrated into the cassandra-stress package; it's going to depend on cassandra in general, so it would make sense to add the relation there.

I'm working on integrating your changes from PR #938 now, along with the above revised logic.

ehankland: Ok - that sounds like it's probably the best option for now. I still wish there were a better solution - is what you are doing going to make running cassandra_ycsb work? Or will any benchmark that has a non juju component not be able to run with any juju packages (since we can't call relate in that instance)? Is there a recommended way to deploy only part of a solution using jujus (just the cassandra cluster but not the clients)?

AdamIsrael: Yes, the change I'm working on will allow partial Juju solutions -- i.e., someone could deploy the Juju-based Cassandra but run a totally different client, like cassandra_ycsb. That's the test case I'll use here.

There's potentially some issue mixing Juju and non-Juju-deployed bits, but I suspect it'll be on a case-by-case basis and we can address those if they come up, as more packages get Juju support.

AdamIsrael:
This latest set of commits addresses all of the outstanding items. Any benchmark that includes the cassandra package should just work (cassandra-stress and cassandra-ycsb being tested and verified).

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -115,6 +115,62 @@ def RedirectGlobalFlags(self):
     with self.config.RedirectFlags(FLAGS):
       yield
 
+  def ConstructVirtualMachineGroup(self, group_name, group_spec):
+    """
```
ehankland: To stay consistent, could you please make this docstring conform to this [styleguide](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments)?
If its a single line it should just be:
"""Construct the virtual machine(s) needed for a group."""

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -134,59 +190,47 @@ def _CheckBenchmarkSupport(self, cloud):
                            provider_info_class.CLOUD,
                            self.name))
 
+  def _ConstructJujuController(self, group_spec):
+    """Construct a VirtualMachine object for a Juju controller"""
```
ehankland: Please terminate the first line of a docstring with punctuation (usually a period). Details in the styleguide linked above

perfkitbenchmarker/linux_packages/cassandra_stress.py
```diff
@@ -0,0 +1,57 @@
...
+  cassandra.YumInstall(vm)
+
+
+def AptInstall(vm):
+  """Installs cassandra-stress on the VM."""
+  cassandra.AptInstall(vm)
+
+
+def JujuInstall(vm, vm_group_name):
+  if vm.isController is True:
```
ehankland: I think this is only ever called on controllers so you shouldn't need this check (correct me if I'm wrong)

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+  def SetupPackageManager(self):
+    if self.isController:
+      resp, _ = self.RemoteHostCommand(
+          "sudo add-apt-repository ppa:juju/stable"
+      )
+    super(JujuMixin, self).SetupPackageManager()
+
+  def PrepareVMEnvironment(self):
+    """
```
ehankland: nit please fix docstring to conform to style guide linked above

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    # Get the name of the service
+    service = charm[charm.rindex('/') + 1:]
+
+    # Deploy to the remaining machine(s)
+    for machine in machines:
+      resp, _ = self.RemoteHostCommand(
+          "juju add-unit %s --to %s" % (service, machine))
+
+  def JujuRelate(self, a, b):
+    """Create a relation between two services
```
ehankland: nit: please terminate the summary with a period

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+    # Get the name of the service
+    service = charm[charm.rindex('/') + 1:]
+
+    # Deploy to the remaining machine(s)
+    for machine in machines:
+      resp, _ = self.RemoteHostCommand(
+          "juju add-unit %s --to %s" % (service, machine))
+
+  def JujuRelate(self, a, b):
```
ehankland: nit: please use more descriptive argument names than a and b

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    output, _ = self.RemoteHostCommand(
+        'juju set %s %s' % (service, ' '.join(params)))
+    return output.strip()
+
+  @vm_util.Retry(poll_interval=30, timeout=3600)
+  def JujuWait(self):
+    """
+    Waits for all deployed services to be installed, configured, and idle.
+
+    Args:
```
ehankland: nit: looks like you forgot to remove the argument documentation after you removed them

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+      pattern: Optionally match machines/services with a pattern
+    """
+    output, _ = self.RemoteHostCommand('juju status %s --format=json' %
+                                       pattern)
+    return output.strip()
+
+  def JujuVersion(self):
+    """Return the Juju version"""
```
ehankland: nit: terminate summary with a period

ehankland: I see at least a few more places where this is the case - please fix all of them

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -164,6 +214,30 @@ def _StartCassandraIfNotRunning(vm):
     Start(vm)
 
 
+def GetCassandraCliPath(vm):
+  if vm.OS_TYPE == os_types.JUJU:
+    """
```
ehankland: There are a few places in this file where you use multi line strings to add comments - could you please switch to just using comments to stay consistent with the style guide?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,225 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+
+  Bootstraps a Juju environment using the manual provider:
+  https://jujucharms.com/docs/stable/config-manual
+  """
+
+  OS_TYPE = os_types.JUJU
+
+  isController = False
+
+  """
```
ehankland: Please switch to comments over multiline strings (except in the case of docstrings)

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,214 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+        if ag != 'started':
+          raise errors.Juju.TimeoutException(
+              'Service %s is not ready; agent-state is %s' % (service, ag))
+
+        ws = unit_data['workload-status']['current']
+        if ws not in ['active', 'unknown']:
+          raise errors.Juju.TimeoutException(
+              'Service %s is not ready; workload-state is %s' % (service, ws))
+
+  def JujuDeploy(self, charm, vm_group, units=1):
```
ehankland: It doesn't look like you actually use the 'units' argument anywhere

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,214 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+          type: manual
+          bootstrap-host: {0}
+  """
+
+  def _Bootstrap(self):
+    """Bootstrap a Juju environment."""
+    resp, _ = self.RemoteHostCommand("juju bootstrap")
+
+  def JujuAddMachine(self, unit):
+    """
```
ehankland: nit: can you move the summary line up to the same line as the start of the docstring?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,214 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+class JujuMixin(DebianMixin):
+  """Class to allow running juju-deployed workloads.
+
+  Bootstraps a Juju environment using the manual provider:
+  https://jujucharms.com/docs/stable/config-manual
+  """
+
+  OS_TYPE = os_types.JUJU
+
+  isController = False
```
ehankland: nit: class/instance variables shouldn't use camel case (see [here](https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming)) 

should be "is_controller"

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -196,7 +271,8 @@ def StartCluster(seed_vm, vms):
   time.sleep(NODE_START_SLEEP)
   for i in xrange(5):
     if not IsRunning(seed_vm):
-      logging.warn('Seed %s: Cassandra not running yet (try %d). Waiting %ds.',
+      logging.warn('Seed %s: Cassandra not running yet (try %d). " \
```
ehankland: What was the intent of this change?
It looks like this just changes 
```
Seed %s: Cassandra not running yet (try %d). Waiting %ds.
```
to
```
Seed %s: Cassandra not running yet (try %d). "                "Waiting %ds
```

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -974,3 +975,214 @@ def MoveFile(self, target, source_path, remote_path=''):
...
+    """Create a relation between two services.
+
+    Args:
+      service1: The first service to relate.
+      service2: The second service to relate.
+    """
+    resp, _ = self.RemoteHostCommand(
+        "juju add-relation %s %s" % (service1, service2))
+
+  def Install(self, package_name):
```
ehankland: Is there an equivalent way to uninstall jujus? When running on pre-provisioned machines, we try to clean everything up afterwards. Defining a [PackageCleanup](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_virtual_machine.py#L223) method  and an [Uninstall](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_virtual_machine.py#L783)  method + JujuUninstall functions in any packages where PackageCleanup isn't enough (similar to the AptUninstall functions [here](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/d7de2e2aa82650190b4640f8d64657846608144b/perfkitbenchmarker/linux_packages/pip.py#L57)) should accomplish this.

You could also just add a TODO and take care of this in a subsequent PR since this only affects pre-provisioned machines

perfkitbenchmarker/linux_virtual_machine.py
AdamIsrael: The docstring for the class?

ehankland: Sorry for being unclear - I just meant change it to this:
```
"""Adds a manually-created virtual machine to Juju.
....
```
rather than
```
"""
Adds a manually-created virtual machine to Juju.
...
```

perfkitbenchmarker/linux_packages/cassandra.py
AdamIsrael: I thought I hit a lint error with this line (length), but the reverted version is under 80 characters.

perfkitbenchmarker/linux_virtual_machine.py
AdamIsrael: There is. I'll add a TODO to address that functionality!

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -134,59 +188,45 @@ def _CheckBenchmarkSupport(self, cloud):
...
+    """Construct a VirtualMachine object for a Juju controller."""
+    vm_group_specs = self.config.vm_groups
+
+    juju_spec = copy.copy(group_spec)
+    juju_spec.vm_count = 1
+    jujuvms = self.ConstructVirtualMachineGroup('juju', juju_spec)
+    if len(jujuvms):
+      jujuvm = jujuvms.pop()
+      jujuvm.is_controller = True
+      jujuvm.vm_group_specs = vm_group_specs
```
ehankland: It looks like you only use this in the jujuinstall functions and I don't think you need it there (unless I'm missing something)

perfkitbenchmarker/linux_packages/cassandra_stress.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def AptInstall(vm):
+  """Installs cassandra-stress on the VM."""
+  cassandra.AptInstall(vm)
+
+
+def JujuInstall(vm, vm_group_name):
+  """Installs the cassandra-stress charm on the VM."""
+  if vm.vm_group_specs and vm_group_name in vm.vm_group_specs:
```
ehankland: I'm not sure what this line is trying to test. It looks like its just checking that the group name passed in is a vm group in the benchmarks config - I don't think you need this (see comment above in benchmark_spec.py)

AdamIsrael: Yep, you're right. Removed.

perfkitbenchmarker/linux_packages/cassandra.py
```diff
@@ -90,6 +94,40 @@ def AptInstall(vm):
...
+  vm.JujuDeploy('cs:trusty/cassandra', vm_group_name)
+
+  # The charm defaults to Cassandra 2.2.x, which has deprecated
+  # cassandra-cli. Specify the sources to downgrade to Cassandra 2.1.x
+  # to match the cassandra benchmark(s) expectations.
+  sources = ['deb http://www.apache.org/dist/cassandra/debian 21x main',
+             'ppa:openjdk-r/ppa',
+             'ppa:stub/cassandra']
+
+  keys = ["F758CE318D77295D",
```
ehankland: super nit: the style guide suggests sticking to either single quotes or double quotes and then not switching (except docstrings which are always double quotes) unless switching eliminates the need to escape quotes (e.g. even if ' is standard, you can use " ' " over  ' \' '). We've generally used ' - could you switch to using ' except when the string contains a ' ? [Here's](https://google.github.io/styleguide/pyguide.html?showone=Strings#Strings) the style guide section

ehankland:
After you fix the quotes issue I think this looks good - could you squash most of the commits together so the history is cleaner? If you aren't familiar with that, I can help.

ehankland:
@skschneider pointed out that github will let me squash the commits when I merge them into master - so don't worry about the squashing

AdamIsrael:
Feel free to squash away! I think I've caught all of the double quotes now.

ehankland:
Great - looks good to me - was there anything else you wanted to get into this PR or are you ready for me to merge?

AdamIsrael:
That's all for this PR! I'll begin work on uninstall/cleanup on pre-provisioned machines in a separate branch.

Thanks for all of your help seeing this through to completion!

marcoceppi:
Yes, thank you @ehankland your feedback has been awesome. We promise to land more bitesized changes next :)

Issue #766, opened 2015-12-22
`./pkb --helpxml` fails


jldiaz:
Aiming to create a nicely markdown formatted documentation for all `pkb` options, I thought a good idea would be to start from `--helpxml`, but it fails with error:

```txt
Traceback (most recent call last):
  File "./pkb.py", line 21, in <module>
    sys.exit(Main())
  File "[edited]/perfkitbenchmarker/pkb.py", line 588, in Main
    argv = FLAGS(argv)  # parse flags
  File "[edited]/perfkitbenchmarker/context.py", line 70, in f
    return getattr(self._thread_flag_values, f_name)(*args, **kwargs)
  File "[edited]/lib/python2.7/site-packages/gflags.py", line 1313, in __call__
    flag.Parse(arg)
  File "[edited]/lib/python2.7/site-packages/gflags.py", line 2412, in Parse
    FLAGS.WriteHelpInXMLFormat(sys.stdout)
  File "[edited]/lib/python2.7/site-packages/gflags.py", line 1743, in WriteHelpInXMLFormat
    is_key=is_key, indent=indent)
  File "[edited]/lib/python2.7/site-packages/gflags.py", line 1968, in WriteInfoInXMLFormat
    default_serialized = self.serializer.Serialize(self.default)
  File "[edited]/perfkitbenchmarker/flag_util.py", line 179, in Serialize
    for val in il.groups])
AttributeError: 'list' object has no attribute 'groups'
```

I think the problem happens when flags from module `flags_util.py` are being processed (perhaps because of the `IntegerListSerializer` serializer)

PR #767, opened 2015-12-22
Fixed --helpxml

jldiaz:
This option produced errors due to:

- Default value for fio_io_depths being a list instead of a IntegerList
- Default value for some other lists being None instead of []

Also, IntegerList lacked a proper str representation. It was added.

This solves https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/766


jldiaz:
I signed it.

PR #768, opened 2015-12-22
--helpxml fixed

jldiaz:
This option produced errors due to:

* Default value for fio_io_depths being a list instead of a IntegerList
* Default value for some other lists being None instead of []
* Also, IntegerList lacked a proper str representation. It was added.

This solves #766

### DIFF
--- a/perfkitbenchmarker/flag_util.py
+++ b/perfkitbenchmarker/flag_util.py
@@ -90,6 +90,9 @@ class IntegerList(object):
         for val in xrange(low, high + 1):
           yield val
 
+  def __str__(self):
+      return IntegerListSerializer().Serialize(self)
+
 
 class IntegerListParser(flags.ArgumentParser):
   """Parse a string containing a comma-separated list of nonnegative integers.
--- a/perfkitbenchmarker/linux_benchmarks/block_storage_workloads_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/block_storage_workloads_benchmark.py
@@ -57,7 +57,7 @@ flags.DEFINE_enum('workload_mode', LOGGING,
                   [LOGGING, DATABASE, STREAMING],
                   'Simulate a logging, database or streaming scenario.')
 
-flags.DEFINE_list('iodepth_list', None, 'A list of iodepth parameter used by '
+flags.DEFINE_list('iodepth_list', [], 'A list of iodepth parameter used by '
                   'fio command in simulated database and streaming scenarios '
                   'only.')
 
--- a/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
@@ -83,7 +83,7 @@ flags.DEFINE_string('fio_jobfile', None,
                     'Job file that fio will use. If not given, use a job file '
                     'bundled with PKB. Cannot use with '
                     '--fio_generate_scenarios.')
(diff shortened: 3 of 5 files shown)


noahl:
Hi @jldiaz , thanks for your pull request. Can you explain exactly what goes wrong to cause the error, and how changing default arguments to `None` and an explicit `IntegerList` fixes it?

jldiaz:
Hello noahl. 

All those problems are related to the way `gflags` module deals with default values for the flags and the way it stringifies them to generate the xml.

When using `--helpxml` option, `gflags.WriteInfoInXMLFormat()` is called for each flag, to generate a documentation which includes the string help, the current value, and the default value. For the current value is uses `str(self.value)`, (`self` is the flag being processed), but for the default value it uses `self.serializer.Serialize()` if provided.

In addition, the default value is parsed to produce a python object which is stored as the current `flag.value`, but not as `flag.default`, which is stored _unparsed_,

This causes two problems in the case of the flags defined via `DEFINE_integerlist()`, as in [fio_benchmark.py][2]):

```python
flag_util.DEFINE_integerlist('fio_io_depths', [1],
                             'IO queue depths to run on. Can specify a single '
                             'number, like --fio_io_depths=1, a range, like '
                             '--fio_io_depths=1-4, or a list, like '
                             '--fio_io_depths=1-4,6-8',
                             on_nonincreasing=flag_util.IntegerListParser.WARN)
```

The default value `[1]` specified in this case, is parsed and stored as a `IntegerList` object as the current `flag.value`, but stored _unparsed_ as a python list as the `flag.default`. This causes two problems:

* First, the `IntegerList` class does not provide a `__str__()` method, so when documenting the current value, it would appear as `<perfkitbenchmarker.flag_util.IntegerList object at 0x7f8f1d88e650>` instead of `[1]`, for example. This can be easily solved by providing a `__str__()` method which calls `IntegerListSerializer()` to do the stringification. I did this in the pull request.

* Second, the [IntegerListSerializer][3] expects its argument of type `IntegerList`, but when it is called with the default value as argument, it receives a plain python `list` instead, causing an exception when trying to access to its `.groups` attribute. This is solved by specifying as default value a `IntegerList` object, instead of a list.

A similar but slightly different problem happens for flags defined via `DEFINE_list()` (which expects a list of strings as `value`) when `None` is specified as default value. In this case `None` is stored both as the current `.value` of the flag, and as the `.default` value. 

For the current `.value`, the presence of `None` instead of a list poses no problems (at least for the xml help), because it uses `str(flag.value)` to produce the documentation, and this will produce the string `"None"`. But when serializing the `.default` value it causes a exception, because the serializer expects a list as argument (even if empty), and tries to iterate over it (see [here][4]). When the argument is `None` the exception `TypeError: 'NoneType' object is not iterable` is raised.

Again, this is solved by specifying as default value an object of the kind expected by the serializer, i.e. an empty list in this case instead of `None`.

[1]: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/flag_util.py#L94
[2]: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py#L108
[3]: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/flag_util.py#L175
[4]: https://github.com/gflags/python-gflags/blob/master/gflags.py#L2092

skschneider:
LGTM. Merging.

Issue #769, opened 2015-12-22
OpenStack nova error with multi region deployment. 


stoor:
Hi, 

First Thanks you very much for PrefKit! 

I am having some problem on my multi-region based cloud setup. NOVA client works fine with multiple regions:

--------- salman -----------
export OS_USERNAME=
export OS_PASSWORD=
export OS_TENANT_NAME=
export OS_PROJECT_NAME= 
export OS_REGION_NAME=
export OS_AUTH_URL=http://keystoneURL:5000/v2.0
--------------------------------
ubuntu@prefkit-test:~/perfkit$ source salman
ubuntu@prefkit-test:~/perfkit$ nova list 
+--------------------------------------+-----------+---------+------------+-------------+--------------------------------------+
| ID                                   | Name      | Status  | Task State | Power State | Networks                             |
+--------------------------------------+-----------+---------+------------+-------------+--------------------------------------+
| 9df963e8-d0ef-44e4-993c-f3e06244961f | smurf-001 | SHUTOFF | -          | Shutdown    | vmnet=192.168.10.139, 130.239.46.227 |
+--------------------------------------+-----------+---------+------------+-------------+--------------------------------------+
ubuntu@prefkit-test:~/perfkit$ vi salman 
ubuntu@prefkit-test:~/perfkit$ source salman 
< Basically I have changed the region parameter >
ubuntu@prefkit-test:~/perfkit$ nova list 
+--------------------------------------+--------------+--------+------------+-------------+------------------------------------+
| ID                                   | Name         | Status | Task State | Power State | Networks                           |
+--------------------------------------+--------------+--------+------------+-------------+------------------------------------+
| a9a6a1cd-1ffe-473a-9f27-1cc865937cbb | PerfKit-Test | ACTIVE | -          | Running     | smog-net=10.0.10.19, 130.238.29.68 |
+--------------------------------------+--------------+--------+------------+-------------+------------------------------------+

So with the Nova Client everything is working fine with multiple regions. But here what happen when I use pkb.py: 

ubuntu@prefkit-test:~/perfkit$ cd PerfKitBenchmarker/
ubuntu@prefkit-test:~/perfkit/PerfKitBenchmarker$ ./pkb.py --cloud=OpenStack --benchmarks=ping
2015-12-22 15:43:17,635 6a699ed7 MainThread INFO     Verbose logging to: /tmp/perfkitbenchmarker/run_6a699ed7/pkb.log
[...]

PR #770, opened 2015-12-23
openstack: Add mechanism to specify region name

meteorfox:
Either by using the --openstack-region flag or the environment variable
OS_REGION_NAME specifies which OpenStack region to use. If none
is specificed the 'RegionOne' is used as the default, which is OpenStack
default region name.

Use region name value for OpenStack Swift's object storage benchmark as well.

Addresses Issue #769

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -241,7 +241,8 @@ def _MakeAzureCommandSuffix(account_name, account_key, for_cli):
     return (' --azure_account=%s --azure_key=%s') % (account_name, account_key)
 
 
-def _MakeSwiftCommandPrefix(auth_url, tenant_name, username, password):
+def _MakeSwiftCommandPrefix(auth_url, tenant_name, username, password,
+                            region_name):
   """This function returns a prefix for Swift CLI command.
 
   Args:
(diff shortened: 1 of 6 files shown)


skschneider:
In light of #776, how would you like to proceed with this PR?

In its current state, I only have two comments for this code. First is that it looks like this only supports one region for all provisioned resources - is there ever a case where you would want different regions in one benchmark? The second is that there's a lot of references to the FLAGS directly. If possible, I suggest reading the FLAGS only once and propagating the region from the read value after that. As far as I can tell, each VM tries to grab a network and firewall object. Does it make sense for the network and firewall to always have the same region as the VM? Because then you can just make the region part of the VM spec, initialize it from FLAGS there, and propagate the value from the spec. This would also be a step toward enabling support for YAML config file support (but that would also require that the different regions issue be addressed).

Neither of these comments block the current PR - they are both changes that can be considered for future additions.

meteorfox:
@skschneider 

> First is that it looks like this only supports one region for all provisioned resources - is there ever a case where you would want different regions in one benchmark?

Yes, one could run a benchmark with VMs across regions.

>  As far as I can tell, each VM tries to grab a network and firewall object. Does it make sense for the network and firewall to always have the same region as the VM?

I agree it makes sense that network and firewall resource have the same region as the VM. Also for OpenStack the SSH public key is as well a per-region resource.

I think in that case we should probably close this PR, as well #734, and have just one PR that address all the issues on #776 along with your suggestions regarding the region.

voellm:
What is the status of this pull request?  Can you fix the conflicts and then we can merge.  Thanks.

meteorfox:
@voellm This has to be revisited again in light of the discussion of zones vs regions on #776.

noahl:
@meteorfox should we close this PR and maybe open a new one later, or are you planning to push more changes here?

meteorfox:
Closing, need to rework this PR after new openstack provider implementation.

PR #772, opened 2015-12-23
add device name

kivio:
add device name to volume attaching - is required by my api.

@meteorfox - can you test it on your environment? I running cassandra_ycsb benchmark, and without this command i gets error with too less arguments.

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_disk.py
+++ b/perfkitbenchmarker/providers/openstack/os_disk.py
@@ -92,7 +92,8 @@ class OpenStackDisk(disk.BaseDisk):
         self.attached_vm_id = vm.id
 
         result = self.__nclient.volumes.create_server_volume(vm.id,
-                                                             self._disk.id, self.device)
+                                                             self._disk.id, 
+                                                             self.device)
         self.attach_id = result.id
 
         volume = None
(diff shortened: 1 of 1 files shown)


meteorfox:
@kivio Sure no problem. I'll try it out later Today.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -92,7 +92,8 @@ def Attach(self, vm):
         self.attached_vm_id = vm.id
 
         result = self.__nclient.volumes.create_server_volume(vm.id,
-                                                             self._disk.id,)
+                                                             self._disk.id,
+                                                             self.device)
```
meteorfox: One question here.

If you noticed above in the `__init__` method for this class, `self.device = ""` is empty. Are you sure that's not a problem here?

According to the [docs](http://docs.openstack.org/developer/python-novaclient/ref/v2/volumes.html) `device` is optional, and at least, in KVM is not required at all, provided you have a somewhat recent kernel. [1]

[1]http://lists.openstack.org/pipermail/openstack-dev/2012-September/001117.html

kivio:
@meteorfox ok, now it's works fine for me

kivio:
@cmccoy @meteorfox @mateusz-blaszkowski @voellm - Guy's can you review that?

meteorfox:
@kivio LGTM! :+1:

kivio:
@voellm we can merge it? It's important for tests on oldest versions of OpenStack including all current available from providers like RedHat OSP or Mirantis OpenStack

PR #774, opened 2015-12-28
Add a missing "--region" flag to AWS command

cmccoy:
The current code fails when a VPC is created outside of the user's
default region.


voellm:
lgtm

PR #778, opened 2016-01-05
Azure: remove use of affinity groups.

cmccoy:
Affinity groups [are deprecated](https://azure.microsoft.com/en-us/documentation/articles/virtual-networks-migrate-to-regional-vnet/), replaced by regional virtual networks.

Also switches Endpoints to be PKB resources, which makes creating them a bit more resilient.

### DIFF
--- a/perfkitbenchmarker/providers/azure/azure_network.py
+++ b/perfkitbenchmarker/providers/azure/azure_network.py
@@ -40,6 +40,59 @@ SSH_PORT = 22
 STORAGE_ACCOUNT_PREFIX = 'portalvhds'
 
 
+class _AzureEndpoint(resource.BaseResource):
+  """An object representing an endpoint to an Azure VM.
+
+  No deletion is specified, as endpoints are deleted along with the VM.
+  """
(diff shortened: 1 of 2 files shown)


ehankland:
Could you make sure that this doesn't break remote ssd on Azure? I think we rely on affinity groups to make it work, but I could be wrong.

cmccoy:
@ehankland - will do, thanks.

ehankland:
@cmccoy thanks. Specifically, I don't think we specify which storage account to use when creating disks/vms - in the affinity groups we set up there's only one to choose from, but without it I'm not sure what will happen (if there are multiple storage accounts in the same location). I suspect everything will seem to work even if it breaks, but you'll get the wrong disk type.

hildrum:
Has the status of this changed?

cmccoy:
No, I haven't looked more.

voellm:
@ehankland - Can you work with Connor to complete this.  Thanks!

cmccoy:
#833 has a superset of these features, so I'll close this in favor of that PR.

Issue #668, opened 2015-11-24
Convert disk specs to allow decoders.


skschneider:
#664 introduces decoders/verifiers for vm specs. Disk specs should have this ability too.

Perhaps create a base spec class inherited by both.

skschneider:
The plan here is to introduce a BaseSpec object w/ #664. This also introduces features that make it easy for a spec's ConfigOptionDecoder to construct a sub-spec object from a decoded yaml object.

Next, in a separate PR, the BaseDiskSpec will be modified to also derive from BaseSpec and use ConfigOptionDecoders.

In the future, larger pieces of the yaml post-parsing verification process can be wrapped into decoders. The flow then goes yaml -> python dicts & lists -> spec objects. We can then add new functionality to the spec objects to generate yaml help showing options and usage.

skschneider:
Fixed in #779.

PR #779, opened 2016-01-05
Make BaseDiskSpec derive from BaseSpec.

skschneider:
Adds option decoder support to BaseDiskSpec.

### DIFF
--- a/tests/disk_test.py
+++ b/tests/disk_test.py
@@ -1,4 +1,4 @@
-# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


voellm:
Can you give an example of how this feature will be used and how this helps?

skschneider:
This change is to address #668.

At the moment, these only decoders used by disk specs are for type-checking, which helps prevent users from passing in bad values via a YAML config file and getting cryptic error messages later in the run when the values are used (e.g. during resource provisioning). The BaseSpec type-checking happens before provisioning begins and can catch mistakes early.

One example of where the decoders are used in BaseVmSpecs is allowing users to provide a a memory value with units ('750MiB', '1.5GiB'). In the future, providers like AWS that allow specifying disk IOPS may benefit from similar flexibility in BaseDiskSpecs.

In the future, the decoders could also be extended to provide a help string describing their accepted values, so this is a step toward auto-documenting available config options.

tests/disk_test.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
```
ehankland: nit: Change the dates to 2016 for new files

ehankland:
LGTM

tests/disk_test.py
skschneider: Done.

PR #782, opened 2016-01-06
Test that we can generate help without errors

noahl:


noahl:
@ehankland @skschneider please take a look.

tests/help_generation_test.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
```
skschneider: 2016

tests/help_generation_test.py
```diff
@@ -0,0 +1,32 @@
...
+from perfkitbenchmarker import flags
+
+
+class HelpTest(unittest.TestCase):
+  def testHelp(self):
+    # Test that help generation finishes without errors
+    flags.GLOBAL_FLAGS.GetHelp()
+
+
+class HelpXMLTest(unittest.TestCase):
```
skschneider: Maybe these could be one TestCase class with the comment above applying to the class (since it applies to both test methods).

skschneider:
Just one date issue. Otherwise LGTM.

tests/help_generation_test.py
noahl: I made them separate TestCase classes because I've been annoyed in the past that I can run a single TestCase class by itself, but not a single test method within that class. I'll change it if you still prefer having one class.

skschneider: I don't have a strong preference.

PR #784, opened 2016-01-06
kill iperf processes started by the benchmark on all machines

hildrum:
I noticed two minor problems in the iperf benchmark:
* It was only cleaning up iperf on one of the hosts
* It was was killing all iperf processes.

So I modified the benchmark to store the pid of the iperf process and only kill that iperf process.


noahl:
LGTM

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
```diff
@@ -76,6 +76,9 @@ def Prepare(benchmark_spec):
       vm.AllowPort(IPERF_PORT)
     vm.RemoteCommand('nohup iperf --server --port %s &> /dev/null &' %
```
cmccoy: In case it's useful: elsewhere we've gotten the PID directly via [`echo $!`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/traces/dstat.py#L83)

PR #785, opened 2016-01-06
Extracting PTRANS metric from hpcc benchmark run output

lkrish:
This minor change extracts an additional data point from the output file generated by the hpcc benchmark and adds it to the result set. Updated unit test.


perfkitbenchmarker/linux_benchmarks/hpcc_benchmark.py
```diff
@@ -197,6 +197,8 @@ def ParseOutput(hpcc_output, benchmark_spec):
     results.append(sample.Sample('STREAM %s Throughput' % metric, value,
                                  'GB/s'))
 
+  value = regex_util.ExtractFloat('PTRANS_GBs=([0-9]*\\.[0-9]*)', hpcc_output)
```
cmccoy: Nit: you can use a raw string here to avoid `\\`:

`r'PTRANS_GBs=([0-9]*\.[0-9]*)'`

PR #789, opened 2016-01-09
Fix some inconsistencies at the top benchmark_spec.py

skschneider:
Inconsistent import statements and non-alphabetic order. Inconsistent
spacing and quote characters in the PickleLock/UnPickleLock functions.


ehankland:
LGTM

PR #792, opened 2016-01-12
Decode spec config options in sorted order.

skschneider:
If the user runs a benchmark with an invalid VM or disk spec, the error
message will be guaranteed to be the same on subsequent runs with the
same spec.


yuyantingzero:
LGTM

PR #794, opened 2016-01-12
Fix disk integration tests.

skschneider:


skschneider:
pep8 v1.7.0 was released 6 hours ago. It's creating all kinds of problems. I'm going to kill this PR, submit another PR that specifies an older pep8 version, and then try again. I'll also create a separate issue to update our pep8 version.

PR #795, opened 2016-01-13
Manually specify pep8 version in tox.ini

skschneider:
Specify 1.5.7 for now, because that is the latest version that does
not complain about the current state of the PKB code.


cmccoy:
Change LGTM. It looks like the travis tests are still coming up with some pep8 issues though.

skschneider:
Turns out I need 1.5.7, but there were virtualenv problems that confused me into thinking 1.6.2 was okay. I'll try again.

PR #798, opened 2016-01-13
Create only one AwsRegionalNetwork per region per benchmark run.

skschneider:
Store AwsRegionalNetworks in the benchmark spec.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -250,7 +250,17 @@ class BenchmarkSpec(object):
 
   def Provision(self):
     """Prepares the VMs and networks necessary for the benchmark to run."""
-    vm_util.RunThreaded(lambda net: net.Create(), self.networks.values())
+    # Sort networks into a guaranteed order of creation based on dict key.
+    # There is a finite limit on the number of threads that are created to
+    # provision networks. Until support is added to provision resources in an
+    # order based on dependencies, this key ordering can be used to avoid
+    # deadlock by placing dependent networks later and their dependencies
+    # earlier. As an example, AWS stores both per-region and per-zone objects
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/benchmark_spec.py
```diff
@@ -250,7 +250,14 @@ def Prepare(self):
 
   def Provision(self):
     """Prepares the VMs and networks necessary for the benchmark to run."""
-    vm_util.RunThreaded(lambda net: net.Create(), self.networks.values())
+    # Sort networks into a guaranteed order of creation based on dict key.
+    # There is a finite limit on the number of threads that are created to
+    # provision networks. Until support is added to provision resources in an
+    # order based on dependencies, this key ordering can be used to avoid
+    # deadlock by placing dependent networks later and their dependencies
+    # earlier.
```
cmccoy: Thanks. Would you mind pointing out that the AWS network in particular requires this?

perfkitbenchmarker/providers/aws/aws_network.py
```diff
@@ -410,8 +416,8 @@ def _Exists(self):
 class AwsRegionalNetwork(network.BaseNetwork):
```
cmccoy: Should we make this class private, since creating one requires a lock?

cmccoy:
Thanks, @skschneider. Two small comments, but LGTM.

perfkitbenchmarker/benchmark_spec.py
skschneider: Done.

perfkitbenchmarker/providers/aws/aws_network.py
skschneider: Makes sense. It should only be created through AwsNetwork.

PR #802, opened 2016-01-14
Remove flag proxy objects.

skschneider:
Eliminates complexity in redirecting FLAG reads and writes to
benchmark-specific FlagValues.
Removes the need for separate FLAGS and GLOBAL_FLAGS that made mocking
all flags difficult.
Prevents infinite recursion and other problems that arise when trying
to work with FLAGS.
Additionally, fixes a bug in help_generation_test where all modules
defining flags were not being imported.


skschneider:
@ehankland PTAL

ehankland:
LGTM

PR #807, opened 2016-01-19
Add ListDecoder for verifying a config option that expects list values.

skschneider:
As part of this change, make the ConfigOptionDecoder's option attribute
optional.

### DIFF
--- a/perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
+++ b/perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
@@ -82,22 +82,25 @@ class MemoryDecoder(option_decoders.StringDecoder):
     match = self._CONFIG_MEMORY_PATTERN.match(string)
     if not match:
       raise errors.Config.InvalidValue(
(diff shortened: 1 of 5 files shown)


perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -128,9 +128,8 @@ def _GetOptionDecoderConstructions(cls):
 class MachineTypeDecoder(option_decoders.TypeVerifier):
   """Decodes the machine_type option of a GCE VM config."""
 
-  def __init__(self, option, **kwargs):
-    super(MachineTypeDecoder, self).__init__(option, (basestring, dict),
-                                             **kwargs)
+  def __init__(self, **kwargs):
```
noahl: Do you need to update some of the error messages in MemoryDecoder now that self.option might be None?

noahl:
Other than my one comment, LGTM

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
skschneider: Good catch. Done.

PR #808, opened 2016-01-19
Introduce Object Sizes

noahl:
Add the ObjectSize class in perfkitbenchmarker/flag_util.py to represent
object sizes, and the ParseObjectSize function to read them. Use this to
simplify the MemoryDecoder in providers/gcp/gce_virtual_machine.py.

### DIFF
--- a/perfkitbenchmarker/__init__.py
+++ b/perfkitbenchmarker/__init__.py
@@ -14,3 +14,17 @@
 
 import gflags as flags  # NOQA
 import gflags_validators as flags_validators  # NOQA
+import pint
+
+# Pint recommends one global UnitRegistry for the entire program, so
+# we create it here.
+UNIT_REGISTRY = pint.UnitRegistry()
+
+# Pint 0.6 uses 'Bo' as the abbreviation for a byte. We want to use
+# 'B', like the rest of the world.
+UNIT_REGISTRY.define('byte = 8 * bit = B')
+
+# Apparently the prefix kilo- is supposed to be abbreviated with a
+# lower-case k. However, everyone uses the upper-case K, and would be
+# very surprised to find out that 'KB' is not a valid unit.
+UNIT_REGISTRY.define('K- = 1000')
--- a/perfkitbenchmarker/flag_util.py
+++ b/perfkitbenchmarker/flag_util.py
@@ -17,6 +17,9 @@
 import logging
 import re
 
+import pint
+
+import perfkitbenchmarker
 from perfkitbenchmarker import flags
 
 
@@ -186,13 +189,13 @@ class IntegerListSerializer(flags.ArgumentSerializer):
 
 
 def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
-                       flag_values=FLAGS, **args):
+                       flag_values=FLAGS, **kwargs):
   """Register a flag whose value must be an integer list."""
 
   parser = IntegerListParser(on_nonincreasing=on_nonincreasing)
   serializer = IntegerListSerializer()
 
-  flags.DEFINE(parser, name, default, help, flag_values, serializer, **args)
+  flags.DEFINE(parser, name, default, help, flag_values, serializer, **kwargs)
 
 
 class FlagDictSubstitution(object):
@@ -219,3 +222,85 @@ class FlagDictSubstitution(object):
   def __exit__(self, *unused_args, **unused_kwargs):
     """Stops the flag substitution."""
     self._flags.__dict__['FlagDict'] = self._original_flagdict
+
+
+class UnitsParser(flags.ArgumentParser):
+  """Parse a flag containing a unit expression.
+
+  The user may require that the provided expression is convertible to
+  a particular unit. The parser will throw an error if the condition
+  is not satisfied. For instance, if a unit parser requires that its
+  arguments are convertible to bits, then KiB and GB are valid units
+  to input, but meters are not. If the user does not require this,
+  than *any* unit expression is allowed.
+  """
+
+  syntactic_help = ('A quantity with a unit. Ex: 12.3MB.')
+
+  def __init__(self, convertible_to=None):
+    """Initialize the UnitsParser.
+
+    Args:
+      convertible_to: perfkitbenchmarker.UNIT_REGISTRY.Unit or
+        None. If a unit, the input must be convertible to this unit or
+        the Parse() method will raise a ValueError.
+    """
+
+    self.convertible_to = convertible_to
+
+  def Parse(self, inp):
+    """Parse the input.
+
+    Args:
+      inp: a string or a perfkitbenchmarker.UNIT_REGISTRY.Quantity. If a string,
+        string has the format "<number><units>", as in "12KB", or "2.5GB".
+
+    Returns:
+      A perfkitbenchmarker.UNIT_REGISTRY.Quantity.
+
+    Raises:
+      ValueError if it can't parse its input.
+    """
+
+    if isinstance(inp, perfkitbenchmarker.UNIT_REGISTRY.Quantity):
+      quantity = inp
+    else:
+      try:
+        quantity = perfkitbenchmarker.UNIT_REGISTRY.parse_expression(inp)
+      except Exception as e:
+        raise ValueError("Couldn't parse unit expresion %s: %s" %
+                         (inp, e.message))
+
+    if self.convertible_to is not None:
+      try:
+        quantity.to(self.convertible_to)
+      except pint.DimensionalityError:
+        raise ValueError("Expression %s is not convertible to %s" %
+                         (inp, self.convertible_to))
+
+    return quantity
+
+
+class UnitsSerializer(flags.ArgumentSerializer):
+  def Serialize(self, units):
+    return str(units)
+
+
+def DEFINE_units(name, default, help, convertible_to=None,
+                 flag_values=flags.FLAGS, **kwargs):
+  """Register a flag whose value is a units expression.
+
+  Args:
+    name: string. The name of the flag.
+    default: perfkitbenchmarker.UNIT_REGISTRY.Quantity. The default value.
+    help: string. A help message for the user.
+    convertible_to: perfkitbenchmarker.UNIT_REGISTRY.Unit or None. If
+      a unit is provided, the input must be convertible to this unit
+      to be considered valid.
+    flag_values: the gflags.FlagValues object to define the flag in.
+  """
+
+  parser = UnitsParser(convertible_to=convertible_to)
+  serializer = UnitsSerializer()
+
+  flags.DEFINE(parser, name, default, help, flag_values, serializer, **kwargs)
--- a/tests/flag_util_test.py
+++ b/tests/flag_util_test.py
@@ -17,6 +17,7 @@
 import copy
 import unittest
 
+import perfkitbenchmarker
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import flag_util
 
@@ -130,6 +131,46 @@ class FlagDictSubstitutionTestCase(unittest.TestCase):
     flag_values.test_flag = 3
     self.assertFlagState(flag_values, 3, False)
     self.assertFlagState(flag_values_copy, 2, True)
+
+
+class TestUnitsParser(unittest.TestCase):
+  def setUp(self):
+    self.up = flag_util.UnitsParser()
+
+  def testParser(self):
+    self.assertEqual(self.up.Parse('10KiB'),
+                     10 * 1024 * perfkitbenchmarker.UNIT_REGISTRY.bytes)
+
+  def testQuantity(self):
+    quantity = 1.0 * perfkitbenchmarker.UNIT_REGISTRY.meter
+    self.assertEqual(self.up.Parse(quantity),
+                     quantity)
+
+  def testBadExpression(self):
+    with self.assertRaises(ValueError):
+      self.up.Parse('asdf')
+
+  def testBytes(self):
+    q = self.up.Parse('1B')
+    self.assertEqual(q.magnitude, 1.0)
+    self.assertEqual(q.units, {'byte': 1.0})
+
+  def testBytesWithPrefix(self):
+    q = self.up.Parse('2KB').to(perfkitbenchmarker.UNIT_REGISTRY.byte)
+    self.assertEqual(q.magnitude, 2000.0)
+    self.assertEqual(q.units, {'byte': 1.0})
+
+  def testConvertibleTo(self):
+    up = flag_util.UnitsParser(
+        convertible_to=perfkitbenchmarker.UNIT_REGISTRY.byte)
+    self.assertEqual(up.Parse('10KiB'),
+                     10 * 1024 * perfkitbenchmarker.UNIT_REGISTRY.bytes)
+
+  def testConvertibleToWrongUnit(self):
+    up = flag_util.UnitsParser(
+        convertible_to=perfkitbenchmarker.UNIT_REGISTRY.byte)
+    with self.assertRaises(ValueError):
+      up.Parse('1m')
 
 
 if __name__ == '__main__':
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,4 @@ colorlog[windows]==2.6.0
 blinker>=1.3
 futures>=3.0.3
 PyYAML==3.11
+pint>=0.6


noahl:
@skschneider @yuyantingzero @ehankland please take a look.

noahl:
In addition to the unit tests, I tested creating a GCE custom virtual machine, and verified that it generated the correct gcloud command and that gcloud accepted it without errors.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
 
   flags.DEFINE(parser, name, default, help, flag_values, serializer, **args)
+
+
+SI_SUFFIXES = ['', 'K', 'M', 'G', 'T', 'P']
+
+
+@functools.total_ordering
+class ObjectSize(object):
+  """Holds the size of data, i.e. numbers of bytes.
```
skschneider: Add "Attributes:" section for bytes.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+  """
+
+  def __init__(self, bytes=0):
+    self.bytes = bytes
+
+  def mult_for_base_and_exp(self, base, exp):
```
skschneider: Casing is odd. This is not a property. It looks like a private method.

The name is a bit confusing to me, but the docstring was enough for me to understand, so I'm not picky about it.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+    """Given base and exp, returns mult so that
+    bytes = mult * base ^ exp.
+
+    May return a float.
+    """
```
skschneider: Since this is a multi-line docstring, write it in standard style with the "Args:" and "Returns:" sections.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+  def exact_mult_for_base_and_exp(self, base, exp):
+    """Given base and exp, returns mult so that
+    bytes = mult * base ^ exp.
+
+    Raises ValueError if mult is not an integer.
+    """
+
+    base_to_exp = base ** exp
+    if self.bytes % base_to_exp == 0:
```
skschneider: Consider using the divmod function.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+    """Given base and exp, returns mult so that
+    bytes = mult * base ^ exp.
+
+    May return a float.
+    """
+
+    base_to_exp = base ** exp
+    return float(self.bytes) / base_to_exp
+
+  def exact_mult_for_base_and_exp(self, base, exp):
```
skschneider: Seems similar to the above function. Maybe combine w/ an optional boolean arg.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+  def ExactKiB(self):
+    return self.exact_mult_for_base_and_exp(1024, 1)
+
+  def ExactMiB(self):
+    return self.exact_mult_for_base_and_exp(1024, 2)
+
+  def ExactGiB(self):
+    return self.exact_mult_for_base_and_exp(1024, 3)
+
+  def AsB(self):
```
skschneider: Considering bytes is a public attribute, this method doesn't seem necessary, but the rest should be consistent. Consider making these methods properties or making bytes private.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+  def AsKiB(self):
+    return self.mult_for_base_and_exp(1024, 1)
+
+  def AsMiB(self):
+    return self.mult_for_base_and_exp(1024, 2)
+
+  def AsGiB(self):
+    return self.mult_for_base_and_exp(1024, 3)
+
+  def WriteWithBase(self, base):
```
skschneider: Name is confusing to me, since it isn't writing anything.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+      A tuple (mult, exp).
+    """
+
+    num = self.bytes
+    exp = 0
+
+    if num == 0:
+      return 0, 0
+
+    while num % base == 0 and exp < len(SI_SUFFIXES):
```
skschneider: Rather than a loop, consider using math.log10 or int.bit_length.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+      num = num // base
+      exp += 1
+
+    return num, exp
+
+  def __str__(self):
```
skschneider: Please add a docstring. I'm not sure what the output of this method is supposed to be.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+    return isinstance(other, ObjectSize) and self.bytes == other.bytes
+
+  def __lt__(self, other):
+    if isinstance(other, ObjectSize):
+      return self.bytes < other.bytes
+    else:
+      raise NotImplemented()
+
+
+OBJECT_SIZE_REGEXP = re.compile('(\d+(\.\d+)?)(K|M|G|T|P)?(i)?B')
```
skschneider: Please add a comment with a few examples. Also, is this a private variable?

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+def ParseObjectSize(inp):
+  """Parse an object size.
+
+  The user writes a decimal followed by a suffix. Allowable base
+  suffixes are K,M,G,T, and P, meaning kilo, mega, giga, tera, and
+  peta. Each suffix may have an 'i' immediately after it, which means
+  to use the base-2 interpretation instead of base-10. All suffixes
+  must end in 'B'. All sizes must be an integer number of bytes.
+  """
```
skschneider: Add Args and Returns.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+  match = OBJECT_SIZE_REGEXP.match(inp)
+  if match is None or match.end() != len(inp):
```
skschneider: You can force the pattern to match the entire string by putting ```$``` at the end. Similarly, you can put ```^``` at the beginning of the pattern to force the match to start at the beginning of the string.

perfkitbenchmarker/flag_util.py
skschneider: Some values that may be worth supporting: ".75GiB", "2.GiB".

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+    raise ValueError('Invalid object size %s. Format is <num><SI suffix>B', inp)
+
+  num = float(match.group(1))
+  suffix = match.group(3) or ''
+  base_two = bool(match.group(4))
+
+  suffix_idx = next((i for i in xrange(len(SI_SUFFIXES))
+                       if SI_SUFFIXES[i] == suffix))  # noqa
```
skschneider: Why ```# noqa``` here?

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+  if base_two:
+    multiplier = 1024 ** suffix_idx
+  else:
+    multiplier = 1000 ** suffix_idx
+
+  num_bytes = num * multiplier
+
+  if num_bytes == round(num_bytes):
```
skschneider: int is probably faster than round.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
   serializer = IntegerListSerializer()
 
   flags.DEFINE(parser, name, default, help, flag_values, serializer, **args)
+
+
+SI_SUFFIXES = ['', 'K', 'M', 'G', 'T', 'P']
```
skschneider: Considering there's apparent support for PB, you should add some test cases and verify it actually works. I imagine there will be precision issues, since Python only guarantees at least 32 bits for an int.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+  def WriteWithBase(self, base):
+    """Write bytes in scientific notation.
+
+    Find mult and exp so that self.bytes = mult * base ^ exp,
+    where exp is as large as possible but <= len(SI_SUFFIXES)
+
+    Returns:
+      A tuple (mult, exp).
+    """
```
skschneider: Add and Args section. State restrictions on the inputs - particularly regarding how self.bytes must be divisible, or explain what happens if it is not.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+      num = num // base
+      exp += 1
+
+    return num, exp
+
+  def __str__(self):
+    mult_ten, exp_ten = self.WriteWithBase(1000)
+    mult_two, exp_two = self.WriteWithBase(1024)
+
+    if exp_ten >= exp_two:
```
skschneider: Worth noting, a value like 1024000 will be "1024KB", rather than "1KiB". I think this can lead to confusion, and the preference should probably be reversed. Either way, I suggest a test case.

tests/flag_util_test.py
```diff
@@ -100,3 +100,84 @@ def testSerialize(self):
 
     self.assertEqual(ser.Serialize(il),
                      '1,2-5,9')
+
+
+class TestObjectSize(unittest.TestCase):
+  def testStrSmall(self):
+    size = flag_util.ObjectSize(bytes=3)
+
```
skschneider: I think it's cleaner to omit these blank lines.

tests/flag_util_test.py
```diff
@@ -100,3 +100,84 @@ def testSerialize(self):
...
+    self.assertEquals(str(size), '3B')
+
+  def testStrBase10(self):
+    size = flag_util.ObjectSize(bytes=3 * (1000 ** 2))
```
skschneider: Consider allowing the constructor to accept a string - it seems like it could be a common use case.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+
+SI_SUFFIXES = ['', 'K', 'M', 'G', 'T', 'P']
+
+
+@functools.total_ordering
+class ObjectSize(object):
+  """Holds the size of data, i.e. numbers of bytes.
+  """
+
+  def __init__(self, bytes=0):
```
skschneider: Consider allowing the constructor to accept a string - it seems like it could be a common use case.

tests/flag_util_test.py
```diff
@@ -100,3 +100,84 @@ def testSerialize(self):
...
+
+    self.assertEquals(str(size), '3MB')
+
+  def testStrBase2(self):
+    size = flag_util.ObjectSize(bytes=3 * (1024 ** 2))
+
+    self.assertEquals(str(size), '3MiB')
+
+  def testEq(self):
+    self.assertEquals(
```
skschneider: assertEquals is actually a deprecated alias of assertEqual.

tests/flag_util_test.py
```diff
@@ -100,3 +100,84 @@ def testSerialize(self):
...
+    self.assertEquals(flag_util.ParseObjectSize('12MiB'),
+                      flag_util.ObjectSize(bytes=12 * (1024 ** 2)))
+
+  def testBadSuffixModifier(self):
+    with self.assertRaises(ValueError):
+      flag_util.ParseObjectSize('12Mj')
+
+  def testExtraCharacters(self):
+    with self.assertRaises(ValueError):
+      flag_util.ParseObjectSize('12mikdj')
```
skschneider: This isn't actually extra characters? The first few characters are still not a valid pattern.

perfkitbenchmarker/flag_util.py
```diff
@@ -190,3 +191,149 @@ def DEFINE_integerlist(name, default, help, on_nonincreasing=None,
...
+      base = 1000
+    else:
+      mult = mult_two
+      exp = exp_two
+      base = 1024
+
+    return '%s%s%sB' % (str(mult), SI_SUFFIXES[exp],
+                        'i' if base == 1024 else '')
+
+
```
skschneider: Extra blank line here. Should be 1 between class methods.

tests/gce_virtual_machine_test.py
```diff
@@ -51,14 +51,14 @@ def testInvalidFloat(self):
...
     self.assertEqual(str(cm.exception), (
-        'Invalid test_component.memory value: "1280.9.8MiB". "1280.9.8" is not '
-        'a valid float.'))
+        'Invalid test_component.memory value: "1280.9.8MiB". Examples of valid '
+        'values: "1280MiB", "7.5GiB".'))
 
   def testNonIntegerMiB(self):
     with self.assertRaises(errors.Config.InvalidValue) as cm:
-      self.decoder.Decode('7.6GiB', _COMPONENT, _FLAGS)
+      self.decoder.Decode('7.5MiB', _COMPONENT, _FLAGS)
```
skschneider: Why was this changed? 7.6GiB is not an integer number of MiB, so as far as I can tell, gcloud does not support it (https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#create). This test case was a good contrast to the 7.5GiB one in testValidStrings, and we should verify it's still rejected.

skschneider:
Took a pass.

cmccoy:
Could we leverage [pint](https://pint.readthedocs.org/en/0.6/) for this? I understand there's an issue with floating point, but if there's a possible workaround it'd be cleaner than re-implementing unit conversions, and broadly useful.

noahl:
@cmccoy I agree that a general-purpose units library would be much better than making my own for one kind of unit. I'm writing down my evaluation of pint here.

Looking at my commit, pint would replace ObjectSize.mult_for_base_and_exp (which is just a unit conversion function), and all the As* methods. It would not replace the Exact* functions, although I could write the equivalent functions with pint.

After reading pint's included units file, I believe it could also replace ParseObjectSize (I missed this last time I evaluated pint). I don't think pint includes the functionality in __str__ that normalizes object size units, but I don't think that's clearly important anyway, and if it is, I can implement it in pint.

There's also the issue that you mentioned that the Exact* functions depend on their quantity being stored exactly, not approximately. However, I can't think of a case where storing bytes in floating-point would make a difference. What we really need is to know that if the user wrote 1024KiB, that is an exact number of MiB. We could find that by converting KiB to MiB using pint and then seeing if the floating-point number has a fractional part. I think the only way we would lose precision is if the user input some huge number of bytes that is very very close to, but not equal to, an exact multiple of 1024. But the number of bytes needed is probably much bigger than anything you can realistically provision anyway, so I'm not worried about it.

Overall, I think using pint would save us some code, and it would be great to have a tested units library in place. I will push a new version of this CL using pint soon.

cmccoy:
@noahl - thanks!

noahl:
@skschneider - I wanted to say a special thanks for reviewing so much code which I ultimately got rid of. That's always depressing, but I really appreciate you doing the review.

voellm:
Hard to see code go but less is better.

noahl:
@skschneider @cmccoy please take a look.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+def DEFINE_units(name, default, help,
+                 flag_values=flags.FLAGS, **args):
+  """Register a flag whose value is a units expression."""
+
+  parser = UnitsParser()
+
+  flags.DEFINE(parser, name, default, help, flag_values, None, **args)
+
+def DEFINE_yaml(name, default, help,
```
skschneider: flake8 is complaining that it wants another blank line above here.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+    """
+
+    if isinstance(inp, pint_wrapper.ureg.Quantity):
+      return inp
+
+    try:
+      return pint_wrapper.ureg.parse_expression(inp)
+    except Exception as e:
+      raise ValueError("Couldn't parse unit expresion %s:" % inp,
+                       e.message)
```
skschneider: spelling "expresion". Also, I'm confused why there are two arguments to ValueError instead of one. Was this meant to be passing two args to a string template? It only has one `%s`.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+class UnitsParser(flags.ArgumentParser):
+  """Parse a flag containing a unit expression."""
+
+  syntactic_help = 'A quantity with a unit. Ex: 12MB'
+
+  def Parse(self, inp):
+    """Parse the input.
+
+    Args:
+      inp: a string
```
skschneider: Please give expected format.

skschneider: Also mention pint_wrapper.ureg.Quantity is allowed.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+                       e.message)
+
+
+class UnitsSerializer(flags.ArgumentSerializer):
+  def Serialize(self, units):
+    return str(units)
+
+
+def DEFINE_units(name, default, help,
+                 flag_values=flags.FLAGS, **args):
```
skschneider: Typically we use `*args` for unnamed args and `**kwargs` for keyword args.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+      return inp
+
+    try:
+      return pint_wrapper.ureg.parse_expression(inp)
+    except Exception as e:
+      raise ValueError("Couldn't parse unit expresion %s:" % inp,
+                       e.message)
+
+
+class UnitsSerializer(flags.ArgumentSerializer):
```
skschneider: Is this class referenced somewhere?

perfkitbenchmarker/pint_wrapper.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Wrapper module for the pint library."""
+
+import pint
+
+# Pint requires one global UnitRegistry for the entire program, so we
+# create it here.
+ureg = pint.UnitRegistry()
```
skschneider: I'm not sure whether this should be its own module, or if we can just put the one line in perfkitbenchmarker/__init__.py with a more descriptive variable name. Maybe ask other people about this one.

tests/flag_util_test.py
```diff
@@ -132,5 +133,18 @@ def testReadAndWrite(self):
     self.assertFlagState(flag_values_copy, 2, True)
 
 
+class TestUnitsParser(unittest.TestCase):
+  def setUp(self):
+    self.up = flag_util.UnitsParser()
+
+  def testParser(self):
+    self.assertEqual(self.up.Parse('10KiBo'),
+                     10 * 1024 * pint_wrapper.ureg.bytes)
```
skschneider: Are KiBo and bytes built into pint? I saw a comment earlier giving "12MB" as an example, which seems to conflict with this?

skschneider:
Took a pass.

perfkitbenchmarker/pint_wrapper.py
cmccoy: Moving it to `__init__.py` SGTM. If the usage gets any more complex (e.g., loading custom units), it can be broken out into a separate file and imported in `__init__.py` to preserve compatibility. Agree on the more descriptive variable name.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+    Raises:
+      ValueError if it can't parse its input.
+    """
+
+    if isinstance(inp, pint_wrapper.ureg.Quantity):
+      return inp
+
+    try:
+      return pint_wrapper.ureg.parse_expression(inp)
+    except Exception as e:
```
cmccoy: Can this `except` be less specific (maybe `UndefinedUnitError`)?

tests/flag_util_test.py
```diff
@@ -132,5 +133,18 @@ def testReadAndWrite(self):
     self.assertFlagState(flag_values_copy, 2, True)
 
 
+class TestUnitsParser(unittest.TestCase):
+  def setUp(self):
+    self.up = flag_util.UnitsParser()
```
cmccoy: Please add more tests :)

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+                 flag_values=flags.FLAGS, **args):
+  """Register a flag whose value is a units expression."""
+
+  parser = UnitsParser()
+
+  flags.DEFINE(parser, name, default, help, flag_values, None, **args)
+
+def DEFINE_yaml(name, default, help,
+                flag_values=flags.FLAGS, **args):
+  """Register a flag whose value is a YAML expression."""
```
cmccoy: No body here?

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
...
+
+
+class UnitsSerializer(flags.ArgumentSerializer):
+  def Serialize(self, units):
+    return str(units)
+
+
+def DEFINE_units(name, default, help,
+                 flag_values=flags.FLAGS, **args):
+  """Register a flag whose value is a units expression."""
```
cmccoy: Could you document the arguments here?
Also, consider having this take `name, *args, **kwargs` and forwarding to `flags.DEFINE`. That way the default value for `flag_values` is kept in `gflags`.

perfkitbenchmarker/flag_util.py
```diff
@@ -219,3 +220,49 @@ def __enter__(self):
   def __exit__(self, *unused_args, **unused_kwargs):
     """Stops the flag substitution."""
     self._flags.__dict__['FlagDict'] = self._original_flagdict
+
+
+class UnitsParser(flags.ArgumentParser):
+  """Parse a flag containing a unit expression."""
+
+  syntactic_help = 'A quantity with a unit. Ex: 12MB'
```
stfeng: provide more help text. especially the diffferent bases one can use to specify sizes, such as MB, MiB, etc.

tests/flag_util_test.py
```diff
@@ -132,5 +133,18 @@ def testReadAndWrite(self):
...
+
+  def testParser(self):
+    self.assertEqual(self.up.Parse('10KiBo'),
```
stfeng: why is it 10KiBo with the 'o' at the end?

perfkitbenchmarker/flag_util.py
noahl: Oops, I actually meant to exclude this from this PR. It's not part of this change.

perfkitbenchmarker/flag_util.py
noahl: Yes, this should be one string with two format arguments (and an extra `%s`). I was confused about the constructor for ValueError. It looks like it accepts any number of arguments, but gets confused and doesn't let you pass a message if you pass more than one argument.

perfkitbenchmarker/flag_util.py
noahl: Done to both.

perfkitbenchmarker/flag_util.py
noahl: I was following the format of DEFINE_integerlist, which I believe was following the format of gflags. I'll update both.

perfkitbenchmarker/flag_util.py
noahl: Nope, I forgot to use it in `DEFINE_units`. Fixed, thanks.

perfkitbenchmarker/pint_wrapper.py
noahl: Moved to `perfkitbenchmarker/__init__.py`.

For the record, I don't think the name `ureg` matches our usual style, but it seems to be Pint's convention, which is why I chose it.

tests/flag_util_test.py
noahl: I'll reply to this with Stanley's comment below.

perfkitbenchmarker/flag_util.py
noahl: I know catching all exceptions is ugly. My concern here is that I don't know what exceptions `parse_expression` might throw (I'm pretty sure it can be more than `UndefinedUnitError`), but I want all errors in the flag parser to raise a `ValueError`. Is there a nicer way to do that?

perfkitbenchmarker/flag_util.py
noahl: Oops, I meant to exclude `DEFINE_yaml` from this PR. Maybe in the future. :)

perfkitbenchmarker/flag_util.py
noahl: Documented the arguments.

I'm not sure I can do the `name, *args, **kwargs` thing because I inject my own `serializer` argument in position 5 of the positional arguments. I could modify the `*args` tuple, but that seems pretty convoluted.

tests/flag_util_test.py
noahl: Added one more. This covers all code paths in the parser. Did you want me to add tests to verify pint's behavior?

tests/flag_util_test.py
noahl: In Pint 0.6 (the current release), bytes are abbreviated `Bo`. In Pint 0.7 (the development branch), bytes are abbreviated `B`. I've been trying to figure out when there will be an official 0.7 release so we don't have to worry about this. In the meantime, I will introduce my own custom bytes unit with abbreviation `B`, ignore the built-in one, and standardize on `B`. That matches the rest of the world.

perfkitbenchmarker/flag_util.py
noahl: Changed example to make it clear that values are floats, and added lots of possible bases.

noahl:
@skschneider @cmccoy I believe I've responded to all of your points. Please take a look.

perfkitbenchmarker/__init__.py
```diff
@@ -14,3 +14,12 @@
 
 import gflags as flags  # NOQA
 import gflags_validators as flags_validators  # NOQA
+import pint
+
+# Pint requires one global UnitRegistry for the entire program, so we
```
cmccoy: Nit: s/requires/recommends

tests/flag_util_test.py
```diff
@@ -132,5 +133,24 @@ def testReadAndWrite(self):
...
+    self.up = flag_util.UnitsParser()
+
+  def testParser(self):
+    self.assertEqual(self.up.Parse('10KiB'),
+                     10 * 1024 * perfkitbenchmarker.UNIT_REGISTRY.bytes)
+
+  def testQuantity(self):
+    quantity = 1.0 * perfkitbenchmarker.UNIT_REGISTRY.meter
+
+    self.assertEqual(self.up.Parse(quantity),
```
cmccoy: Nit: extraneous newline

tests/flag_util_test.py
cmccoy: No; but since you are defining the "B" type, it might be useful to add a few covering ilo and ibibytes with varying prefixes.

tests/flag_util_test.py
```diff
@@ -132,5 +133,24 @@ def testReadAndWrite(self):
...
+                     quantity)
+
+  def testBadExpression(self):
+    with self.assertRaises(ValueError):
+      self.up.Parse('asdf')
```
cmccoy: What happens with a valid input but the wrong unit? Can I pass "1 meter"?

perfkitbenchmarker/__init__.py
noahl: Changed.

tests/flag_util_test.py
noahl: Removed.

tests/flag_util_test.py
noahl: In the old version, nothing, because the UnitParser is supposed to be agnostic about what unit it's parsing. In the code I'm about to upload, I added a feature where the user can specify that the input must be convertible to some specific unit, which you can use to verify that, for instance, your input is convertible to bytes. This lets you accept MB, GB, etc., without accepting meters. (This is a hack around the fact that you can't directly ask pint whether a quantity is in a specified dimension, even though pint's concept of dimensions is exactly what we need here.)

noahl:
@cmccoy @skschneider please take a look.

tests/flag_util_test.py
```diff
@@ -132,5 +133,49 @@ def testReadAndWrite(self):
...
+
+  def testBytes(self):
+    q = self.up.Parse('1B')
+    self.assertEqual(q.magnitude,
```
cmccoy: nit: lots of unneeded newlines here.

cmccoy:
LGTM.

noahl:
I removed the extra newlines. Merging.

Issue #787, opened 2016-01-08
Config-related failures should happen before any benchmark is run


skschneider:
The current sequence of events looks like this:
1. Create the config object for every benchmark.
2. Start running each benchmark in series.
3. As part of running the benchmark, create the BenchmarkSpec and call its ConstructVirtualMachines function (which translates config sub-objects and flags into VM specs and disk specs).

A config-related error may be raised either when creating the config object or when calling ConstructVirtualMachines.

All input-checking that can happen before resources are provisioned should happen before resources are provisioned for any benchmark.

skschneider:
With #810 and #816, config options for all benchmarks are checked prior to provisioning resources for any. Closing this issue.

PR #809, opened 2016-01-20
Create BenchmarkConfigSpec to aggregate benchmark input checking.

skschneider:
Part of a solution to #787.
Creating a BenchmarkConfigSpec performs verification on inputs for a
benchmark run, including creation of sub-spec objects like the VM and
disk specs and the input-checking that they contain. In a separate PR,
the creation of the BenchmarkConfigSpec will be moved out of the
BenchmarkSpec.__init__ to before the call to RunBenchmark so that the
inputs of all configured benchmark runs can be verified before any
benchmark is started.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -24,7 +24,6 @@ import thread
 import threading
 import uuid
 
-from perfkitbenchmarker import configs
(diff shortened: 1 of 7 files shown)


skschneider:
Tox installation failed during automatic testing. Will try to relaunch.

skschneider:
test this please

PR #810, opened 2016-01-20
Create BenchmarkConfigSpec to aggregate benchmark input checking.

skschneider:
Part of a solution to #787.
Creating a BenchmarkConfigSpec performs verification on inputs for a
benchmark run, including creation of sub-spec objects like the VM and
disk specs and the input-checking that they contain. In a separate PR,
the creation of the BenchmarkConfigSpec will be moved out of the
`BenchmarkSpec.__init__` to before the call to RunBenchmark so that the
inputs of all configured benchmark runs can be verified before any
benchmark is started.

### DIFF
--- a/perfkitbenchmarker/configs/benchmark_config_spec.py
+++ b/perfkitbenchmarker/configs/benchmark_config_spec.py
@@ -1,0 +1,335 @@
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Classes that verify and transform benchmark configuration input."""
+
+import copy
+import os
+
+from perfkitbenchmarker import disk
+from perfkitbenchmarker import errors
+from perfkitbenchmarker import flag_util
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import os_types
+from perfkitbenchmarker import providers
+from perfkitbenchmarker import static_virtual_machine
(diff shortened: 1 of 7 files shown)


skschneider:
@ehankland PTAL

perfkitbenchmarker/configs/benchmark_config_spec.py
```diff
@@ -0,0 +1,335 @@
...
+          be modified by this function.
+      flag_values: flags.FlagValues. Runtime flags that may override the
+          provided config values.
+    """
+    super(_VmGroupSpec, cls)._ApplyFlags(config_values, flag_values)
+    if flag_values['cloud'].present or 'cloud' not in config_values:
+      config_values['cloud'] = flag_values.cloud
+    if flag_values['os_type'].present or 'os_type' not in config_values:
+      config_values['os_type'] = flag_values.os_type
+    if config_values.get('vm_count', 0) is None:
```
ehankland: nit: I think ('vm_count' in config_values and config_values['vm_count'] is None) better conveys the intent of this line - alternatively add a comment.

perfkitbenchmarker/configs/benchmark_config_spec.py
```diff
@@ -0,0 +1,335 @@
...
+    """Gets decoder classes and constructor args for each configurable option.
+
+    Returns:
+      dict. Maps option name string to a (ConfigOptionDecoder class, dict) pair.
+      The pair specifies a decoder class and its __init__() keyword arguments
+      to construct in order to decode the named option.
+    """
+    result = super(_VmGroupSpec, cls)._GetOptionDecoderConstructions()
+    result.update({
+        'cloud': (option_decoders.EnumDecoder, {
```
ehankland: nit: Might be nice to have module constants for strings that get used multiple times

ehankland:
LGTM

perfkitbenchmarker/configs/benchmark_config_spec.py
skschneider: Done.

perfkitbenchmarker/configs/benchmark_config_spec.py
skschneider: I'm omitting this for now because it would be non-uniform with regard to the other items in the dict, but it's a good idea for a future change. The string 'cloud' is quite common in PKB code, and there may be a better way to avoid duplication.

PR #812, opened 2016-01-21
Change MockFlags to be more like the real FlagValues.

skschneider:
Handles use cases of ```FlagValues``` that were previously unsupported, such
as deep-copying the ```FlagValues``` or substituting the inner flag dictionary.

Changes the behavior of the ```MockFlags.__setattr__``` to no longer set the
present attribute (to be more like the real ```FlagValues.__setattr__```). The new
way to set both the value and present attributes is to use
`MockFlags[<flag_name>].Parse(<value>)`.

### DIFF
--- a/tests/mock_flags.py
+++ b/tests/mock_flags.py
@@ -15,8 +15,6 @@
 
 import contextlib
 
-import mock
-
 from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import flags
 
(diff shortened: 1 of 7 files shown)


skschneider:
@cmccoy PTAL

tests/mock_flags.py
```diff
@@ -24,43 +22,67 @@
...
+    self.present = False
+    self.validators = ()
+    self.value = None
+
+  def Parse(self, argument):
+    self.present = True
+    self.value = argument
+
+
+class MockFlagDict(dict):
```
cmccoy: Nit: can this be private?

tests/mock_flags.py
```diff
@@ -24,43 +22,67 @@
...
+  string to a Flag object. This mock maps each flag name to a MockFlag object
+  that is created upon first access.
+
+  Because FlagValues and MockFlags override __getattr__ to search their flag
+  dictionaries, care is taken not to create a MockFlag when searching for
+  special attributes like the __copy__ method.
+  """
+
+  def __getitem__(self, key):
+    if not dict.__contains__(self, key) and not key.startswith('__'):
```
cmccoy: Out of curiosity, is there a reason for `super(...)` below and explicit superclass method calls here?.

tests/mock_flags.py
skschneider: Done.

tests/mock_flags.py
skschneider: For some reason, I was under the impression that super didn't work with dict subclasses. I tried it, and it looks like it does work, so I've changed them all to super.

skschneider:
PTAL

tests/mock_flags.py
```diff
@@ -24,43 +22,68 @@
 FLAGS = flags.FLAGS
 
 
+class _MockFlag(object):
+  """Mock version of a Flag object."""
+
+  def __init__(self):
+    self.present = False
+    self.validators = ()
```
cmccoy: Is this used?

cmccoy:
One more question, otherwise LGTM.

tests/mock_flags.py
skschneider: Yes, by gflags: https://github.com/gflags/python-gflags/blob/master/gflags.py#L1080
We substitute the real FlagValues instance's internal flag dict with our _MockFlagDict, so that the scattered references to FLAGS in every module will all be affected. I'll expand this class's docstring a bit to cover the attributes.

PR #813, opened 2016-01-22
Always parse flags

noahl:
Use gflags parsers for flags, even when those flags come from a YAML
config file.

### DIFF
--- a/perfkitbenchmarker/configs/__init__.py
+++ b/perfkitbenchmarker/configs/__init__.py
@@ -220,8 +220,7 @@ def GetMergedFlags(config):
       if key not in flags_values:
         raise ValueError('Flag "%s" is not defined.' % key)
       if not flags_values[key].present:
-        flags_values[key].value = value
-        flags_values[key].present += 1
+        flags_values[key].Parse(value)
 
   return flags_values
 
--- a/tests/config_test.py
+++ b/tests/config_test.py
@@ -21,7 +21,10 @@ import mock
 from perfkitbenchmarker import linux_benchmarks
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import errors
+from perfkitbenchmarker import flags
 from perfkitbenchmarker import windows_benchmarks
+
+FLAGS = flags.FLAGS
 
 CONFIG_NAME = 'a'
 INVALID_NAME = 'b'
@@ -62,6 +65,11 @@ a:
   vm_groups:
     default:
       vm_spec: *anchor_does_not_exist
+"""
+LIST_FLAG_CONFIG = """
+a:
+  flags:
+    list_flag: a,b,c
 """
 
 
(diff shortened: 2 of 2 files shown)


noahl:
A little more explanation: I plan to add a parser for distributions. I want to be able to specify distributions through configs or a flag. But that means that I need one parser that will be triggered in either case. I think always using gflags parsers is the right thing to do anyway, so I made this PR.

Sending data through YAML and then gflags means that it may be partially parsed before it gets to the gflags parser, which could cause problems. I did a check of all the parsers included in gflags, and I don't think any of them will have this problem.

noahl:
@ehankland @yuyantingzero @cmccoy  please take a look.

perfkitbenchmarker/configs/__init__.py
```diff
@@ -220,7 +220,7 @@ def GetMergedFlags(config):
       if key not in flags_values:
         raise ValueError('Flag "%s" is not defined.' % key)
       if not flags_values[key].present:
-        flags_values[key].value = value
+        flags_values[key].Parse(value)
         flags_values[key].present += 1
```
skschneider: present is set by Flag.Parse.

noahl: Good point. I'm pushing an updated version now.

tests/config_test.py
```diff
@@ -129,3 +137,13 @@ def testConfigOverrideFlag(self):
     config = configs.GetUserConfig()
     self.assertEqual(config['a']['vm_groups']['default']['vm_count'], 5)
     self.assertEqual(config['a']['flags']['flag'], 'value')
+
+  def testConfigFlagStillParsed(self):
+    flags.DEFINE_list('list_flag', ['default_val'], 'List flag.')
```
skschneider: This will probably linger beyond the lifetime of the test. I suggest creating a separate FlagValues object and passing it in as the flag_values arg of DEFINE_list.

tests/config_test.py
```diff
@@ -129,3 +137,13 @@ def testConfigOverrideFlag(self):
...
+    list_flag_config = yaml.load(LIST_FLAG_CONFIG)
+    flag_values = configs.GetMergedFlags(list_flag_config['a'])
+
+    self.assertEqual(flag_values.list_flag,
+                     ['a', 'b', 'c'])
+
+    del FLAGS.list_flag
```
skschneider: RE: previous comment, the reason is here. This cleanup statement doesn't get called if a prior step of the test (like the assert) fails.

skschneider:
Took a pass. BTW, this might be redundant w/ #810, which also calls Parse.

cmccoy:
Other than @skschneider's comments, LGTM.

noahl:
Thanks for the review. This is unnecessary given #810 , so I'm closing it.

PR #816, opened 2016-01-22
Move BenchmarkConfigSpec construction to before RunBenchmark.

skschneider:
Check config input before starting any benchmark resource provisioning.
Change `BenchmarkSpec` constructor to accept a `BenchmarkConfigSpec`.
Remove previous input checking from `configs/__init__.py`.
Update tests that were constructing a `BenchmarkSpec` the old way.
Fix tests where the thread's `BenchmarkSpec` was not being reset.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -304,9 +304,8 @@ def _GetBenchmarkSpec(benchmark_config, benchmark_name, benchmark_uid):
   created during the provision stage from a file.
 
   Args:
-    benchmark_config: A Python dictionary representation of the configuration
-        for the benchmark. For a complete explanation, see
-        perfkitbenchmarker/configs/__init__.py.
+    benchmark_config: BenchmarkConfigSpec. The benchmark configuration to use
+        while running the current stage.
     benchmark_name: string. Name of the benchmark.
     benchmark_uid: string. Identifies a specific run of a benchmark.
 
@@ -317,11 +316,9 @@ def _GetBenchmarkSpec(benchmark_config, benchmark_name, benchmark_uid):
     return benchmark_spec.BenchmarkSpec(benchmark_config, benchmark_name,
                                         benchmark_uid)
   else:
-    # TODO(skschneider): Build BenchmarkConfigSpec before RunBenchmark.
-    config = benchmark_config_spec.BenchmarkConfigSpec(
-        benchmark_name, flag_values=FLAGS, **benchmark_config)
     try:
-      return benchmark_spec.BenchmarkSpec.GetSpecFromFile(benchmark_uid, config)
+      return benchmark_spec.BenchmarkSpec.GetSpecFromFile(benchmark_uid,
+                                                          benchmark_config)
     except IOError:
       if FLAGS.run_stage == STAGE_PREPARE:
         logging.error(
@@ -342,7 +339,7 @@ def RunBenchmark(benchmark, collector, sequence_number, total_benchmarks,
     sequence_number: The sequence number of when the benchmark was started
       relative to the other benchmarks in the suite.
     total_benchmarks: The total number of benchmarks in the suite.
-    benchmark_config: The config to run the benchmark with.
+    benchmark_config: BenchmarkConfigSpec. The config to run the benchmark with.
     benchmark_uid: An identifier unique to this run of the benchmark even
       if the same benchmark is run multiple times with different configs.
   """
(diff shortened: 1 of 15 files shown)


skschneider:
@ehankland PTAL

perfkitbenchmarker/pkb.py
```diff
@@ -506,12 +503,18 @@ def RunBenchmarks(publish=True):
   for i, benchmark_tuple in enumerate(benchmark_tuple_list):
     benchmark_module, user_config = benchmark_tuple
     benchmark_name = benchmark_module.BENCHMARK_NAME
+    benchmark_config_dict = benchmark_module.GetConfig(user_config)
```
noahl: Consider making a function that turns a dictionary into a benchmark config spec, either here or (ideally) in benchmark_config_spec.py.

tests/background_cpu_test.py
```diff
@@ -44,9 +49,18 @@
 
 class TestBackgroundWorkload(unittest.TestCase):
 
-  def setupCommonFlags(self, mock_flags):
-    mock_flags.os_type = os_types.DEBIAN
-    mock_flags.cloud = providers.GCP
+  def setUp(self):
+    super(TestBackgroundWorkload, self).setUp()
```
noahl: This is not necessary because this class inherits directly from `unittest.TestCase`.

tests/background_cpu_test.py
```diff
@@ -67,85 +81,64 @@ def _CheckVMFromSpec(self, spec, num_working):
...
-      config = configs.LoadConfig(ping_benchmark.BENCHMARK_CONFIG, {}, NAME)
-      spec = benchmark_spec.BenchmarkSpec(config, NAME, UID)
-      spec.ConstructVirtualMachines()
-      with self.assertRaisesRegexp(Exception, 'NotImplementedError'):
-        spec.Prepare()
-      with self.assertRaisesRegexp(Exception, 'NotImplementedError'):
-        spec.StartBackgroundWorkload()
-      with self.assertRaisesRegexp(Exception, 'NotImplementedError'):
-        spec.StopBackgroundWorkload()
+    self._mocked_flags['background_cpu_threads'].Parse(1)
```
noahl: What does this line do? I don't see why this won't just call Parse() on a mock.MagicMock, but it must do more than that.

tests/benchmark_spec_test.py
```diff
@@ -97,16 +104,32 @@
...
 
+_DESCRIPTION = 'Test description.'
 
-class ConstructVmsTestCase(unittest.TestCase):
+
+class _BenchmarkSpecTestCase(unittest.TestCase):
 
   def setUp(self):
-    # Reset the current benchmark spec.
+    super(_BenchmarkSpecTestCase, self).setUp()
```
noahl: I don't think this is necessary, since this class inherits directly from `unittest.TestCase`.

tests/disk_metadata_test.py
```diff
@@ -21,18 +21,34 @@
...
+_BENCHMARK_NAME = 'name'
+_BENCHMARK_UID = 'uid'
 _COMPONENT = 'test_component'
 
 
-class GcpDiskMetadataTest(unittest.TestCase):
+class _DiskMetadataTestCase(unittest.TestCase):
+
+  def setUp(self):
+    super(_DiskMetadataTestCase, self).setUp()
```
noahl: Same comment as above.

noahl:
Overall, this looks good. My one general comment, in addition to the specific ones I attached to different lines, is that I don't like putting empty `'description'` flags in all of the benchmark configs in the tests. It adds noise but doesn't add anything to the code. Can we make the `description` field optional, and remove those? It's fine with me if a missing description is recorded as the empty string; I just don't want to have to write it down, or especially have to read it in code that doesn't actually use descriptions.

skschneider:
I'm not really sure why the description is in the config tree, but it should be a required part of adding a new benchmark. How about I create some test helpers for anything outside of `tests/benchmark_spec_test.py`? Like subclasses of BenchmarkConfigSpec and BenchmarkSpec that only override `__init__` to auto-fill things like the benchmark name or description.

tests/background_cpu_test.py
skschneider: Parse sets the flag object's value and marks the flag as present (i.e. the user provided a value).
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/tests/mock_flags.py#L47
MockFlags has become a minimized version of gflags.FlagValues that can behave similarly in all of PKB's uses of FLAGS after its initialization, with the difference that MockFlags does not require importing all flag definitions prior to use. This solves the problem where a test only imports a module where a flag value is read, but not the module where the flag is defined.

noahl: Oh, got it.

noahl:
I think requiring `description` in benchmark configs is a usability bug in PKB. I'd like to get rid of it, not merge code that assumes it.

I think the problem is https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/configs/benchmark_config_spec.py#L305 . You could remove that in this PR, or I can submit another one first.

tests/background_cpu_test.py
skschneider: Removed.

tests/benchmark_spec_test.py
skschneider: Removed.

tests/disk_metadata_test.py
skschneider: Removed.

perfkitbenchmarker/pkb.py
skschneider: That particular change would be creating a function for two statements (five lines of code) that's used only here. I do see that RunBenchmarks is getting a bit large, but I think a separate PR could break it up a bit better.

skschneider:
I've made the 'description' in the benchmark config object optional and removed it from the test modules.

skschneider:
PTAL

perfkitbenchmarker/pkb.py
noahl: Sounds good to me.

noahl:
LGTM

PR #824, opened 2016-02-01
Report total cluster boot time

hildrum:
Report `os_type` in cluster boot metadata and also track total boot time.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
@@ -58,7 +58,7 @@ def _GetTimeToBoot(vms, vm_index):
   """
   vm = vms[vm_index]
   metadata = {'num_cpus': vm.num_cpus, 'machine_instance': vm_index,
-              'num_vms': len(vms)}
+              'num_vms': len(vms), 'os_type': vm.OS_TYPE}
   metadata.update(vm.GetMachineTypeDict())
   assert vm.bootable_time
   assert vm.create_start_time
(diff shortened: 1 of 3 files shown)


perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
```diff
@@ -83,7 +83,26 @@ def Run(benchmark_spec):
   samples = vm_util.RunThreaded(_GetTimeToBoot, params)
   logging.info(samples)
   assert len(samples) == len(vms)
-  return samples
+  total_boot = benchmark_spec.vm_boot_endtime - benchmark_spec.vm_boot_starttime
```
skschneider: This total time includes things that are not included in the individual machine boot times. As it's currently written, this captures the entire RunThreaded(benchmark_spec.PrepareVm, ...) call. If you check the definition of [PrepareVm](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/benchmark_spec.py#L276), it calls vm.Create(), vm.AllowRemoteAccessPorts(), vm.WaitForBootCompletion(), and then a bunch of other things including vm.AddMetadata(). For individual VMs, the boot time measurement starts partway through [vm.Create()](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/resource.py#L127), within the call to vm._CreateResource(), which comes after vm.CreateDependencies(). The boot time measurement ends partway through vm.WaitForBootCompletion().
So the difference is that this total time measurement contains the time to create dependencies (e.g. Network objects) and also the time to do post-creation steps (e.g. adding metadata).

voellm: is this boot to ssh or boot to ready?

perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
```diff
@@ -83,7 +83,26 @@ def Run(benchmark_spec):
...
+      uniform_machine_type = None
+    if uniform_os_type is not None and vm.OS_TYPE != uniform_os_type:
+      uniform_os_type = None
+
+  metadata = {'num_vms': len(vms)}
+  if uniform_os_type is not None:
+    metadata['os_type'] = uniform_os_type
+  if uniform_machine_type is not None:
+    metadata['machine_type'] = uniform_machine_type
+  overall = sample.Sample('Total Boot Time', total_boot, 'seconds', metadata)
```
voellm: Boot to sshable?

Not sure what "total" is here.

perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
hildrum: Yes, it's not really a pure boot time measurement or a time-to-ssh measurement.  I should have been clearer in the original pull request.  

The problem is that the benchmark gives time-to-ssh for each individual machine.  I've been considering the maximum of those numbers as the overall cluster time-to-ssh, but that's not quite right.  This number is also not really a cluster-time-to-ready either, for the reasons you mention.  

There are some options:
* Don't try to output a number representing how long the cluster took to get ready, and just use the maximum time-to-ssh.
*  Output this number, but rename it something else ('Cluster Ready Time', 'Cluster Start Time', 'Cluster Start and Configure Time')
* Refactor the code so that it's possible to measure of how long it takes for all hosts in the cluster to be ssh-able.   

Of those options, I prefer option 2.

hildrum:
Since it looks like this might require some more discussion, I created a new pull request #826 which just adds the os_type to the metadata.

hildrum:
Offline consensus is that this isn't the right approach.  Closing.

PR #825, opened 2016-02-02
Azure: make endpoints resources

cmccoy:
This better handles command timeouts - the call to create the endpoint
may time out but eventually succeed in creating the endpoint. Using a resource
allows the test to proceed if the endpoint exists, even if the last attempt to
create the endpoint failed.


cmccoy:
@hildrum / @ehankland  - could you take a look?

hildrum:
LGTM

PR #826, opened 2016-02-02
Add os_type to cluster_boot benchmark

hildrum:
Unlike #824, makes no changes to what's measured, just adds `os_type` to the metadata for the `Boot Time` metric.

The metadata does already contain an `image` attribute, but determining the os type from the image is cloud-dependent.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
@@ -58,7 +58,7 @@ def _GetTimeToBoot(vms, vm_index):
   """
   vm = vms[vm_index]
   metadata = {'num_cpus': vm.num_cpus, 'machine_instance': vm_index,
-              'num_vms': len(vms)}
+              'num_vms': len(vms), 'os_type': vm.OS_TYPE}
   metadata.update(vm.GetMachineTypeDict())
   assert vm.bootable_time
   assert vm.create_start_time


skschneider:
LGTM. Consider treating it the same as zone and image so it applies to all benchmarks: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/publisher.py#L149

perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
```diff
@@ -58,7 +58,7 @@ def _GetTimeToBoot(vms, vm_index):
   """
   vm = vms[vm_index]
   metadata = {'num_cpus': vm.num_cpus, 'machine_instance': vm_index,
-              'num_vms': len(vms)}
+              'num_vms': len(vms), 'os_type': vm.OS_TYPE}
```
voellm: Perhaps all benchmarks should include os_type.  How about completing this change and then looking to generalize.

voellm: And this works because all VM's are homogeneous here.

voellm:
LGTM

hildrum:
Discussed with @skschneider about whether the `os_type` property should apply to just the cluster boot benchmark or be placed in all benchmark metadata.  But at the moment, the cluster_boot benchmark is the only one that runs on Windows and Linux, so it's the only benchmark where this property matters, so we'll treat it differently than zone and image.

perfkitbenchmarker/linux_benchmarks/cluster_boot_benchmark.py
hildrum: I don't think it requires that the VMs be homogenous.  A sample is added per-vm, so one cluster boot benchmark with 20 vms generates 21 samples--one for the boot time for each vm, and one for the the end-to-end runtime.     The os_type metadata is only added to the vm-specific record, and not to the end-to-end runtime record.

PR #832, opened 2016-02-04
Fix path joining so that the log location is correct on windows

hildrum:
On windows, the following message was printing out with the last slash in the wrong direction: `2016-02-03 19:40:35,526 1791fcee MainThread INFO     Complete logs can be found at: c:\users\hildrum\appdata\local\temp\2\perfkitbenchmarker\run_1791fcee/pkb.log`

This uses `os.path.join` rather than using `%s/%s` to create the full path.


skschneider:
LGTM

PR #833, opened 2016-02-05
Azure affinity group / update new features in ping test

dq922:
Azure does not allow  ping through virtual ip address interface. The VM build process now attach a public IP to each VM and expose that as an ip_address

Added ping to allow across zones pinging with public IP with trigger: --ping_net_interface_type


dq922:
I signed it!

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -20,8 +20,18 @@
 import logging
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import sample
+from perfkitbenchmarker import flags
+
 import re
 
+flags.DEFINE_enum('ping_net_interface_type', 'internal_ip',
```
cmccoy: Could you re-use the `--ip_addresses` flag defined in [`vm_util`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L95-L100)? We already have a lot of flags :)

perfkitbenchmarker/providers/azure/azure_virtual_machine.py
```diff
@@ -198,9 +198,10 @@ def _Create(self):
     create_cmd = [AZURE_PATH,
                   'vm',
                   'create',
-                  '--affinity-group=%s' % self.network.affinity_group.name,
+                  '--location=%s' % self.zone,
                   '--virtual-network-name=%s' % self.network.vnet.name,
                   '--vm-size=%s' % self.machine_type,
+                  '--public-ip=%s' % self.name,
```
cmccoy: Could you explain how this differs from the current method? Is the named public IP cleaned up when the instance is deleted?

cmccoy:
@dq922 - thanks for this! I left a few notes.

Additionally, I looked at moving to location-based vnets in #778, but couldn't figure out how to ensure that a created VM and disks were associated with a given storage account without an affinity group. Do you have any ideas there?

hildrum:
I'm creating release 1.3 soon; can this get pulled in for that?

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dq922: You're absolutely right, --ip_addresses param will cover the requirement to ping on the external interface. I'll remove the 'ping_net_interface_type' flags.

perfkitbenchmarker/providers/azure/azure_virtual_machine.py
dq922: Each VM in Azure comes automatically with a public and private IP address. The public (VIP) is not really associated with the VM but to the cloud service account that the VM lives in. By default, VIP cannot be pinged at all. 

If we do not utilize the public IP address, we would not be able to test latency across regions either on the VIP or private IP. Caveat is if we setup vnet-to-vnet VPN but it's out of scope of this test module. This contrasts with GCP where a subnet can be stretched across all of GPC's compute region. 

I'm proposing that we open up the ping test to allow cross-region ping test in addition to inter-datacenter ping.

dq922:
@cmcocy Thanks for the feedback. I saw your #778 changes but am too confuse on how to associate the VM to storage account. 

I speculate that since MSFT flattened their storage / network architecture in ASM, and now all storage accounts / network accounts are directly associated with a region. IF we create a VM without specifying the storage account, I think it would just provision to the default storage account. Have to do further testing to validate.

It might be worthwhile to invest in developing support ARM mode where storage/network/compute resources are grouped in a resource group. It should address the majority of the issues see here.

cmccoy:
> I speculate that since MSFT flattened their storage / network architecture in ASM, and now all storage accounts / network accounts are directly associated with a region. IF we create a VM without specifying the storage account, I think it would just provision to the default storage account. Have to do further testing to validate.

This is my concern - we create a standard or premium storage account for disk tests, and don't want to accidentally start always testing the user's default storage account type.

voellm:
@hildrum  @cmccoy  - Whats the next step on the pull request?

voellm:
@connormccoy - Are you able to complete this review?  It looks like you are working on it.  Thanks.

cmccoy:
I believe this is still blocked on [this concern](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/833#issuecomment-190899074), namely that applying this change will break testing premium storage. @dq922, any ideas?

noahl:
I believe this issue will be fixed by the move to Azure Resource Management in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1112

noahl:
I recently merged my ARM mode pull request, and we can now ping across zones. @dq922 , I'm going to close this, but please feel free to reopen if you want to add any more functionality on top of what is now in PKB. Contributions are always welcome! :)

Issue #819, opened 2016-01-28
[OpenStack] Errors of pkb.py when it excuted benchmarks


michael92437558:
vi test.sh
export OS_USERNAME=admin
export OS_PASSWORD=admin
export OS_TENANT_NAME=admin
export OS_PROJECT_NAME=admin
export OS_AUTH_URL=http://172.21.4.2:5000/v2.0

source test.sh
nova list -> succeed
+--------------------------------------+----------------------------------------------+---------+------------+-------------+---------------------------------------------------------------------+
| ID                                   | Name                                         | Status  | Task State | Power State | Networks                                                            |
+--------------------------------------+----------------------------------------------+---------+------------+-------------+---------------------------------------------------------------------+
| 01249794-bfa7-4b21-be99-7f38a7a59cb7 | murano-clmonijmazrklz-gateway-1-rpy22hhl3w7q | ACTIVE  | -          | Running     | kkk-network-dcd0da153ea64e10a399bb50d7f6c451=10.0.6.7, 172.21.4.232 |
| a8ecea03-0820-4651-938f-b79c582240d9 | murano-clmonijmazrklz-kube-1-jb32dpvxxpjw    | ACTIVE  | -          | Running     | kkk-network-dcd0da153ea64e10a399bb50d7f6c451=10.0.6.6, 172.21.4.233 |
| dd2da4e5-5e49-45b3-9b79-799e44635214 | murano-clmonijmazrklz-kube-2-mlkgm22dfgw6    | ACTIVE  | -          | Running     | kkk-network-dcd0da153ea64e10a399bb50d7f6c451=10.0.6.5, 172.21.4.234 |
| 02349f64-be0e-425e-a4f5-991179deaf75 | murano-clmonijmazrklz-kube-3-h23riiwgkede    | ACTIVE  | -          | Running     | kkk-network-dcd0da153ea64e10a399bb50d7f6c451=10.0.6.4, 172.21.4.230 |
| 6f73bf01-fe76-4e90-a7c7-128778e4c586 | qemu-agent-base-image                        | ACTIVE  | -          | Running     | net04=192.168.111.229, 172.21.4.212                                 |
| 8ac91c7d-8ac3-4ba0-93bf-e34869fd98d6 | test                                         | SHUTOFF | -          | Shutdown    | net04_ext=172.21.4.190; net04=192.168.111.61                        |
[...]

PR #834, opened 2016-02-05
Check requirements.txt and requirements-*.txt at runtime.

skschneider:
Previously, users have had problems because an installed Python package did not match the version required by PKB. This change checks the versions of required packages against the requirements.txt and requirements-*.txt files and raises an error if there is a mismatch.

### DIFF
--- a/perfkitbenchmarker/providers/__init__.py
+++ b/perfkitbenchmarker/providers/__init__.py
@@ -16,6 +16,7 @@ import logging
 import os
 
 from perfkitbenchmarker import import_util
+from perfkitbenchmarker import requirements
 
 # This unconditionally loads any modules in any provider
 # directory with the name 'flags'. It is expected that providers
@@ -39,18 +40,23 @@ VALID_CLOUDS = [GCP, AZURE, AWS, DIGITALOCEAN, KUBERNETES, OPENSTACK,
                 RACKSPACE, CLOUDSTACK, ALICLOUD, MESOS]
 
 
-def LoadProvider(provider_name):
+def LoadProvider(provider_name, ignore_package_requirements=True):
   """Loads the all modules in the 'provider_name' package.
 
-  This function loads all modules in the provided package. By loading these
-  modules, relevant classes (e.g. VMs) will register themselves. This should
-  be called with the exact name of the package, which is usually the name of
-  the provider in lower case (e.g. the package name for the 'GCP' provider
-  is 'gcp').
+  This function first checks the specified provider's Python package
+  requirements file, if one exists, and verifies that all requirements are met.
+  Next, it loads all modules in the specified provider's package. By loading
+  these modules, relevant classes (e.g. VMs) will register themselves.
 
   Args:
     provider_name: The name of the package whose modules should be loaded.
+        Usually the name of the provider in lower case (e.g. the package name
+        for the 'GCP' provider is 'gcp'.
+    ignore_package_requirements: boolean. If True, the provider's Python package
+        requirements file is ignored.
   """
+  if not ignore_package_requirements:
+    requirements.CheckProviderRequirements(provider_name)
   provider_package_path = os.path.join(__path__[0], provider_name)
   try:
     # Iterating through this generator will load all modules in the provider
--- a/perfkitbenchmarker/requirements.py
+++ b/perfkitbenchmarker/requirements.py
@@ -1,0 +1,79 @@
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Functions for checking that required Python packages are installed."""
+
+import os
+
+import pkg_resources
+
+from perfkitbenchmarker import errors
+
+
+# Path of the root of the current git branch.
+_BRANCH_ROOT_DIR = os.path.dirname(os.path.dirname(__file__))
+
+
+def _CheckRequirements(requirements_file_path):
+  """Checks that all package requirements specified in a file are met.
+
+  Args:
+    requirements_file_path: string. Path to a pip requirements file.
+  """
+  try:
+    with open(requirements_file_path, 'rb') as fp:
+      for line in fp:
+        pkg_resources.require(line)
+  except (pkg_resources.DistributionNotFound,
+          pkg_resources.VersionConflict) as e:
+    # In newer versions of setuptools, these exception classes have a report
+    # method that provides a readable description of the error.
+    report = getattr(e, 'report', None)
+    err_msg = report() if report else str(e)
+    raise errors.Setup.PythonPackageRequirementUnfulfilled(
+        'A Python package requirement was not met while checking "{path}": '
+        '{msg}{linesep}To install required packages, execute the following '
+        'command:{linesep}pip install -r "{path}"{linesep}To bypass package '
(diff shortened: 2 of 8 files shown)


skschneider:
@noahl PTAL
This relates to the discussion in #819.

meteorfox:
@skschneider @voellm This is somewhat unrelated to this PR but should we consider moving the requirements text files to their own directory to avoid polluting the project root?

I'm suggesting it now since the code in this PR make the assumption that requirements are at the top-level of the project and follows the `requirements-<provider>.txt` name convention, and seems it's the appropriate time to do so.

This is what I'm suggesting to try to organize the requirements files a little bit.

Create a directory named `requirements/` with a requirements file for each provider that need python package dependencies, and a base requirements file that has just the PKB dependencies and is imported by the other requirements files.

Like so:

```
...
pkb.py
requirements/
|--- base.txt    #  original requirements.txt, it could be another name like pkb-base.txt
|--- openstack.txt 
|--- aws.txt
|--- ...
setup.cfg
setup.py
tox.ini
```

Then  `requirements/openstack.txt` would look like follows:

```
-r base.txt             # Install common dependencies

python-keystoneclient==2.1.1
python-novaclient==2.35.0
```

Using `-r base.txt` within a requirements file has the same as in the command line, it will first install
the packages listed in base.txt but then continues with the packages in the file that same file. It's a way to share a set common dependencies without duplication. It works relative to the current file, so `base.txt` is expected to be in the same directory as `openstack.txt`.

Finally to install dependencies for both PKB and a particular provider it's just *one* command:

```
pip install -r requirements/openstack.txt
```

The downside is that we will need to update the docs, tox.ini and any automation that relies on the current convention. So perhaps we will need to have both approaches co-exist to give time to our users to migrate their automation (I'm thinking of my automations around PKB)

Thoughts? After writing it out, I'm now not 100% sure if it's worth the trouble, but even if it's just for awareness for those unaware that such feature exists in `pip` then it's okay that at least it was considered. :)

tests/requirements_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+  def testIncorrectVersion(self):
+    requirements_content = """
+    # The version of the installed python-gflags will be less than 42.
+    python-gflags>=42
+    """
+    with self._MockOpen(requirements_content):
+      with self.assertRaises(errors.Setup.PythonPackageRequirementUnfulfilled):
+        requirements._CheckRequirements(_PATH)
+
```
meteorfox: Can you add a test case for when a user has an old version of the package installed but the requirements specify a specific newer version, and another test case for the reverse, that is, the user has a newer version of the package but an older version is required?

**Example for test case 1:** - this one is probably covered by `testIncorrectVersion()`

User has old version of the package:
python-novaclient==2.15.0

but requirements ask for:
python-novaclient==2.35.0


**Example for test case 2:**

User has newer version of package:
python-novaclient==3.2.0

Requirements ask for:
python-novaclient==2.35.0

perfkitbenchmarker/providers/__init__.py
```diff
@@ -51,6 +52,7 @@ def LoadProvider(provider_name):
   Args:
     provider_name: The name of the package whose modules should be loaded.
   """
+  requirements.CheckProviderRequirements(provider_name)
```
noahl: Can you add a note in `LoadProvider`'s docstring that it checks requirements in addition to loading modules?

tests/providers_test.py
```diff
@@ -33,6 +34,13 @@ def testLoadInvalidProvider(self):
     with self.assertRaises(ImportError):
       providers.LoadProvider('InvalidCloud')
 
+  def testLoadProviderChecksRequirements(self):
+    p = mock.patch(requirements.__name__ + '.CheckProviderRequirements')
```
noahl: Could you simplify this function by using `with mock.patch(...)`?

noahl:
Other than my one comment and @meteorfox 's comment, LGTM. I especially like how you get the error message from newer versions of `setuptool` - that is really going above and beyond for usability.

perfkitbenchmarker/requirements.py
```diff
@@ -0,0 +1,79 @@
...
+
+import os
+import sys
+
+import pkg_resources
+
+from perfkitbenchmarker import errors
+
+
+_DEFAULT_BRANCH_ROOT_DIR = sys.path[0]
```
cmccoy: This seems a bit magical to me. I think path relative to this file would be clearer.

cmccoy:
I'm not sure about this change as is.

`requirements

cmccoy:
I view `requirements.txt` files as an easy way to obtain a consistent development environment. I'm not sure they should force the user to have *exactly* the package versions specified - if I want to update to a patch-level release of a library, I'm not sure I should have to change the PKB requirements file or wait for a maintainer to do so upstream.

Also, this will break if we make PKB installable (not a crazy idea), since the requirements files won't be in place.

skschneider:
Would it help if there were a flag to bypass requirement checks? Having the checks present does add value by preventing users from running PKB with mismatched package versions and then getting mysterious failures that none of the developers have seen.

What would PKB look like if it were installable? Would it still depend on external packages? If so, is there a place we can move the requirements files so they become a become part of the installation so that the code knows where to find them?

cmccoy:
> Having the checks present does add value by preventing users from running PKB with mismatched package versions and then getting mysterious failures that none of the developers have seen.

Certainly not disagreeing!

> What would PKB look like if it were installable? Would it still depend on external packages? If so, is there a place we can move the requirements files so they become a become part of the installation so that the code knows where to find them?

The standard way would be with setuptools / PyPI.  You can declare [installation dependencies](http://python-packaging-user-guide.readthedocs.org/en/latest/requirements/#install-requires) in a `setup.py` file, but it doesn't map well to the case where there are subpackages with varying requirements.

Ideally I think individual clouds would be plugins that could declare their own dependencies (including PKB core), but that would be a big change.

skschneider:
If I understand correctly, the need for requirements.txt and the basic requirements run-time checking would be removed if we made PKB installable. If we wanted to support running it independently and keep the run-time checking in, it looks like there's an option to have arbitrary files in the package: http://python-packaging-user-guide.readthedocs.org/en/latest/distributing/#package-data
So we could keep the requirements.txt for run-time checks and also use it to generate the `install_requires` in `setup.py`.

For individual cloud providers, if we split them into separate packages, the same would apply. Otherwise (if everything were in a single PKB package), I guess we would have to depend on the run-time checks with bundled data files, since we don't want to require dependencies for all cloud providers.

skschneider:
I like the suggestion from @meteorfox to move the requirement files to remove clutter from the root dir. In light of the discussion about potentially splitting the providers into separate installable packages, I now wonder if it makes more sense to put the requirements.txt for each provider in the corresponding directory where their code is. If it were paired with an update to make the runtime check error messages state the command for installing the missing requirements, then discoverability would not suffer, and it would have the benefit of further condensing all the changes a new provider would need to implement into a single directory (there are still a few places where this isn't true, like the default configs). But if that's the plan, I think it would make more sense to modify this PR first to get the checks in, and then move the files in the second PR. With that order, we don't step into a state where the requirement files are hard to find. So it would be something like:
1) In the first PR, add a flag to disable the run-time requirement checks, improve the error message to state some options ("run this pip command to install or use this flag to ignore"), and address other comments about tests and docstring updates.
2) In the second PR, move the requirement files and update the README and other documentation.
Thoughts?

noahl:
I also like the idea of having the requirements files in the provider directories. It's not really important to me if it's one PR or two.

perfkitbenchmarker/providers/__init__.py
skschneider: Done.

tests/requirements_test.py
skschneider: Done.

tests/providers_test.py
skschneider: Done.

perfkitbenchmarker/requirements.py
skschneider: Done.

skschneider:
Added flag to disable runtime checks, addressed other feedback. PTAL.

noahl:
LGTM. Sorry for the delay in reviewing.

PR #842, opened 2016-02-10
Upgrade Pip

noahl:
When installing pip on a VM, make sure we always have the latest
version.


noahl:
@voellm @stfeng please take a look.

stfeng:
LGTM

Issue #837, opened 2016-02-09
Remove second test (regional latency) from GCS object_storage_service


zbjornson:
The object_storage_service benchmark for GCS tests one-byte-RW (latency) for a regional bucket after testing a multi-region bucket.

My suggestion from discussion in #750 is to remove this second test and instead allow users to make use of `--object_storage_class` (proposed in #836), `--object_storage_region` and `--object_storage_scenario` to get whatever benchmark combo they want.

PR #843, opened 2016-02-10
Add GCP Region Names API

noahl:
A simple API to pull out the parts of a GCP zone, region, or
multi-region location name.

### DIFF
--- a/perfkitbenchmarker/providers/gcp/util.py
+++ b/perfkitbenchmarker/providers/gcp/util.py
@@ -113,3 +113,61 @@ class GcloudCommand(object):
     if hasattr(resource, 'zone'):
       self.flags['zone'] = resource.zone
     self.additional_flags.extend(FLAGS.additional_gcloud_flags or ())
+
+
+class GCPLocation(object):
+  """A GCP location.
+
+  This class may represent a zone, region, or continent.
+  """
+
+  MULTIREGION = 'multiregion'
+  REGION = 'region'
+  ZONE = 'zone'
+
+  def __init__(self, name):
+    parts = name.split('-')
+    if len(parts) == 0 or len(parts) > 3:
+      raise ValueError('%s is not a valid GCP Location' % name)
+
+    self._multiregion = parts[0]
+    self._region = parts[1] if len(parts) > 1 else None
+    self._zone = parts[2] if len(parts) > 2 else None
+
+  def asMultiRegion(self):
+    return self._multiregion
+
+  def asRegion(self):
+    if self._region is not None:
+      return '%s-%s' % (self._multiregion, self._region)
+    else:
+      raise ValueError('Location %s has no region' % self)
+
+  def asZone(self):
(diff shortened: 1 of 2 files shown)


noahl:
@stfeng @skschneider @cmccoy please take a look.

This is part of a plan to address #837, and also to let users select the multi-region location to test GCS in. This PR provides a general API for pulling out parts of a GCP location name. Once we have this, I can have the `object_storage_service` benchmark handle different types of location.

stfeng:
I don't think we need this class to address 837 for object storage
benchmark. All we need is a new storage_continent flag to allow user to
choose a continental location.

We should not try to infer object storage locations from *gce* zones, these
two are different concepts.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -113,3 +113,61 @@ def _AddCommonFlags(self, resource):
     if hasattr(resource, 'zone'):
       self.flags['zone'] = resource.zone
     self.additional_flags.extend(FLAGS.additional_gcloud_flags or ())
+
+
+class GCPLocation(object):
+  """A GCP location.
+
+  This class may represent a zone, region, or continent.
```
cmccoy: I don't see continent anywhere...

stfeng:
If we want to have a generic way of getting region info from a GCP "zone string" (such as "us-central1-a"), we should explore ways to use official google APIs first if possible.

Anyway, it's an interesting thought to make this a library, but I don't see an immediate use case for it (see my other comments above)

perfkitbenchmarker/providers/gcp/util.py
noahl: Oops, left over from an old version. "Multiregions" were initially "continents" until I renamed them. Looks like I missed this reference.

noahl:
@stfeng The reason I wanted an API like this is that we could make the object storage benchmark run the continental test in the continent corresponding to the region passed in by the `object_storage_region` flag, which I think would be logical (I should have done it when I added the region flag).

Also, it looks to me from https://cloud.google.com/docs/geography-and-regions like all GCP products have the same zones and regions. I think it would actually be useful to connect the two, so we could distinguish when a GCS benchmark is using VMs in a zone that is part of the bucket's region versus a zone that is not part of the bucket's region.

stfeng:
My point is we can introduce a separate flag to specify which continent to use for object storage test, e.g. "object_storage_continent", instead of overloading the object_storage_region flag, it's much simpler and more specific this way, and it's also more extensible because it allows the user to specify provider-specific continent strings, instead of implying it from regional name, which may be different. e.g., suppose in the future when a provider introduce a continental offering that is named "US_2", or ASIA_B, then it would be quite hard to jam it into the object_storage_region flag.

noahl:
@stfeng I see what you're saying, and I need to rethink the design of this. Let's talk in person.

PR #845, opened 2016-02-13
Let users choose their multiregion for GCS benchmarking

noahl:
As the title says.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -54,6 +54,9 @@ flags.DEFINE_enum('storage', providers.GCP,
 flags.DEFINE_string('object_storage_region', None,
                     'Storage region for object storage benchmark.')
 
+flags.DEFINE_string('object_storage_gcs_multiregion', None,
+                    'Storage multiregion for GCS in object storage benchmark.')
+
 flags.DEFINE_enum('object_storage_scenario', 'all',
                   ['all', 'cli', 'api_data', 'api_namespace'],
                   'select all, or one particular scenario to run: \n'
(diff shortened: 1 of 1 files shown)


noahl:
I've tested this in the US and Asia, and it worked fine. I'm running a Europe test right now, but I don't expect any surprises.

noahl:
Update: the Europe test succeeded. That covers all GCS multiregions.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -865,7 +868,11 @@ def Prepare(self, vm):
 
     vm.bucket_name = 'pkb%s' % FLAGS.run_uri
 
-    vm.RemoteCommand('%s mb gs://%s' % (vm.gsutil_path, vm.bucket_name))
+    make_bucket_command = '%s mb' % vm.gsutil_path
+    if FLAGS.object_storage_gcs_multiregion:
```
stfeng: Let's output the gcs multi-region location name, and the regional location name (of all 3 providers) as a metadata, for example - 

metadata['gcs_multiregion_location'] = "default", or whatever the user has specified via the flag.

metadata['regional_bucket_location'] = "default", or whatever the user has specified via FLAGS.object_storage_region (for all 3 providers)

This way, when pkb users collect and analyze data from each run, they can categorize based on bucket locations and compare properly.

stfeng:
Just one comment about adding metadata entries for bucket location.

Thanks for the extra effort in testing.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Done. I tested it on GCP, AWS, and Azure.

noahl:
@stfeng @yuyantingzero I added the bucket locations to the metadata. Please take a look.

stfeng:
LGTM

PR #849, opened 2016-02-22
Don't require version of gcs-oauth2-boto-plugin

noahl:
The 1.8 version we used to require is not compatible with the latest
oauth2client, and using the most recent version causes no problems.


noahl:

yuyantingzero:
LGTM

stfeng:
LGTM. Thanks for debugging and fixing this!

PR #852, opened 2016-02-25
Use boto3 for S3 benchmarking

noahl:
This forward-proofs our object storage benchmarking code, since
boto (non-3) is no longer under development. In particular, it sets us
up to benchmark different S3 storage classes.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -117,6 +117,7 @@ AZURE_CREDENTIAL_LOCATION = '.azure'
 
 DEFAULT_BOTO_LOCATION = '~/.boto'
 BOTO_LIB_VERSION = 'boto_lib_version'
+BOTO3_LIB_VERSION = 'boto3_lib_version'
 
 SWIFTCLIENT_LIB_VERSION = 'python-swiftclient_lib_version'
 
@@ -651,7 +652,7 @@ class S3StorageBenchmark(object):
       Then the final return value is the list of the above list that reflect
       the results of all scenarios run here.
     """
-    metadata[BOTO_LIB_VERSION] = _GetClientLibVersion(vm, 'boto')
+    metadata[BOTO3_LIB_VERSION] = _GetClientLibVersion(vm, 'boto3')
     results = []
     scratch_dir = vm.GetScratchDir()
 
@@ -1125,6 +1126,7 @@ def Prepare(benchmark_spec):
 
   vms[0].Install('pip')
   vms[0].RemoteCommand('sudo pip install python-gflags==2.0')
+  vms[0].RemoteCommand('sudo pip install boto3')
   azure_version_string = ''
   if FLAGS.azure_lib_version is not None:
     azure_version_string = '==%s' % FLAGS.azure_lib_version
--- a/perfkitbenchmarker/scripts/object_storage_api_tests.py
+++ b/perfkitbenchmarker/scripts/object_storage_api_tests.py
@@ -26,11 +26,13 @@ import json
 import logging
 import sys
 from threading import Thread
+import threading
 import string
 import random
 import time
 
 import boto
+import boto3
 import gflags as flags
 import gcs_oauth2_boto_plugin  # noqa
 from azure.storage.blob import BlobService
@@ -117,6 +119,10 @@ LARGE_OBJECT_FAILURE_TOLERANCE = 0.1
 # A global variable initialized once for Azure Blob Service
 _AZURE_BLOB_SERVICE = None
 
+# boto3 recommends creating per-thread sessions. We store them in
+# THREAD_LOCAL.boto_session.
+THREAD_LOCAL = threading.local()
+
 
 # When a storage provider fails more than a threshold number of requests, we
 # stop the benchmarking tests and raise a low availability error back to the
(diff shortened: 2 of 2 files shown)


noahl:
I tested this change by verifying that the api_data and api_namespace object storage scenarios still work, in addition to all of the unit tests.

noahl:

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -117,6 +117,7 @@
 
 DEFAULT_BOTO_LOCATION = '~/.boto'
 BOTO_LIB_VERSION = 'boto_lib_version'
```
stfeng: Add a comment here to indicate we use boto for GCS and boto3 for S3

<do you have any plan to update to boto3 for GCS?>

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -135,12 +141,37 @@ def _useBotoApi(storage_schema):
...
+    A boolean indicates whether or not to use boto API.
+  """
+
+  if storage_schema == 's3':
     return True
   else:
     return False
 
 
+def _getS3():
```
stfeng: Please consider a better name for this function...

e.g., _getBoto3ResourceForS3()

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -117,6 +119,10 @@
 # A global variable initialized once for Azure Blob Service
 _AZURE_BLOB_SERVICE = None
 
+# boto3 recommends creating per-thread sessions. We store them in
+# THREAD_LOCAL.boto_session.
```
stfeng: we need to clarify and confirm the life cycle of a boto3 session here, especially its relationship to the underlying http connection. (i.e., When is a session created and terminated.)

We need to make sure connections are being reused across many requests and we only pay the overhead of connection setup for the first request (per thread)

noahl:
Update: I think boto3 will figure out the right regional endpoint for us. The JSON file is https://github.com/boto/botocore/blob/develop/botocore/data/_endpoints.json#L71

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -180,21 +211,19 @@ def _ListObjects(storage_schema, bucket, prefix, host_to_connect=None):
...
-  bucket_list_result = None
   if _useBotoApi(storage_schema):
     bucket_uri = boto.storage_uri(bucket, storage_schema)
     if host_to_connect is not None:
       bucket_uri.connect(host=host_to_connect)
 
-    bucket_list_result = bucket_uri.list_bucket(prefix=prefix)
+    return [key.name for key in bucket_uri.list_bucket(prefix=prefix)]
+  elif _useBoto3Api(storage_schema):
+    s3 = _getS3()
```
stfeng: where do you implement the .connect() equivalent here in boto3?

The caller of this script passes in "host=host_to_connect" which is a fully qualified DNS name.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -180,21 +211,19 @@ def _ListObjects(storage_schema, bucket, prefix, host_to_connect=None):
...
+    bucket = s3.Bucket(bucket)
+    return [object.key for object in bucket.objects.filter(Prefix=prefix)]
```
stfeng: Should we return object.key.name?

stfeng:
The caller of this script passes in a fully qualified DNS name ("host_to_connect") to connect to a different region or a different "environment" of a particular provider, and the current code passes it to the storage_uri.connect() call, which worked very well. I don't see how this is being handled now in the new boto3 code. (Yes, I see your comment about that JSON file boto3 has which maps region to endpoints, but we still need to figure out how we are going to use that with our current code here...)

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: It took a lot of following code, but I learned that the botocore library (which boto3 is based on) uses Requests (python-requests.org) for its HTTP(S) functionality, and Requests supports connection pooling.

I know there is one pool per boto3.Endpoint object. I *think* that different threads have different Endpoints, so we have a connection pool per thread. It would be pretty difficult to override this - we would need to make our own botocore.Client objects, which is a few layers of abstraction below where this PR is. The alternative is to use the workaround in https://github.com/boto/botocore/issues/577 to share one session object across threads. Unfortunately, I noticed that the connection pool holds a maximum of 10 connections, so that might be a problem if we had lots of threads.

Do you think it would work for different threads to have their own connections? We're not doing much computation, so I think threads are spending most of their time on I/O anyway, so they're probably using their connection objects pretty well. I think that we would only experience overhead from having too many connections if we got to the point where we were saturating our network bandwidth, and in that case we would also have overhead from too many threads. So I think it might be okay to have per-thread connections. How does that sound to you?

voellm: Before you check in there final changes please add side by side comparison
numbers.  Before/after boto changes.

hildrum:
Should I wait for this one to get in for release 1.3?

stfeng:
No need for release 0.3

stfeng:
Sorry, 1.3

noahl:
I've found a way to add storage classes without using boto3, so I'm closing this PR for now.

PR #857, opened 2016-03-02
Add a YAML flag type

noahl:
This lets us read YAML data in a consistent way from command-line flags
and config files.

### DIFF
--- a/perfkitbenchmarker/flag_util.py
+++ b/perfkitbenchmarker/flag_util.py
@@ -18,6 +18,7 @@ import logging
 import re
 
 import pint
+import yaml
 
 import perfkitbenchmarker
 from perfkitbenchmarker import flags
@@ -302,5 +303,64 @@ def DEFINE_units(name, default, help, convertible_to=None,
 
   parser = UnitsParser(convertible_to=convertible_to)
   serializer = UnitsSerializer()
+
+  flags.DEFINE(parser, name, default, help, flag_values, serializer, **kwargs)
+
+
+# The YAML flag type is necessary because flags can be read either via
+# the command line or from a config file. If they come from a config
+# file, they will already be parsed as YAML, but if they come from the
+# command line, they will be raw strings. The point of this flag is to
+# guarantee a consistent representation to the rest of the program.
+class YAMLParser(flags.ArgumentParser):
+  """Parse a flag containing YAML."""
+
+  syntactic_help = 'A YAML expression.'
+
+  def Parse(self, inp):
+    """Parse the input.
+
+    Args:
+      inp. A string or the result of yaml.load. If a string, should be
+      a valid YAML document.
+    """
+
+    if isinstance(inp, basestring):
+      # This will work unless the user writes a config with a quoted
+      # string that, if unquoted, would be parsed as a non-string
+      # Python type (example: '123'). In that case, the first
+      # yaml.load() in the config system will strip away the quotation
+      # marks, and this second yaml.load() will parse it as the
+      # non-string type. However, I think this is the best we can do
+      # without significant changes to the config system, and the
+      # problem is unlikely to occur in PKB.
+      try:
+        return yaml.load(inp)
+      except yaml.YAMLError as e:
+        raise ValueError("Couldn't parse YAML string '%s': %s" %
+                         (inp, e.message))
+    else:
+      return inp
+
+
+class YAMLSerializer(flags.ArgumentSerializer):
+
+  def Serialize(self, val):
+    return yaml.dump(val)
+
+
+def DEFINE_yaml(name, default, help, flag_values=flags.FLAGS, **kwargs):
+  """Register a flag whose value is a YAML expression.
+
+  Args:
+    name: string. The name of the flag.
+    default: object. The default value of the flag.
+    help: string. A help message for the user.
+    flag_values: the gflags.FlagValues object to define the flag in.
+    kwargs: extra arguments to pass to gflags.DEFINE().
+  """
+
+  parser = YAMLParser()
+  serializer = YAMLSerializer()
 
   flags.DEFINE(parser, name, default, help, flag_values, serializer, **kwargs)
--- a/tests/flag_util_test.py
+++ b/tests/flag_util_test.py
@@ -173,5 +173,23 @@ class TestUnitsParser(unittest.TestCase):
       up.Parse('1m')
 
 
+class TestYAMLParser(unittest.TestCase):
+
+  def setUp(self):
+    self.parser = flag_util.YAMLParser()
+
+  def testValidString(self):
+    self.assertEqual(self.parser.Parse('[1, 2, 3]'),
+                     [1, 2, 3])
+
+  def testPreParsedYAML(self):
+    self.assertEqual(self.parser.Parse([1, 2, 3]),
+                     [1, 2, 3])
+
+  def testBadYAML(self):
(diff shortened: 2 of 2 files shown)


noahl:
@yuyantingzero @stfeng @skschneider please take a look.

perfkitbenchmarker/flag_util.py
```diff
@@ -304,3 +306,60 @@ def DEFINE_units(name, default, help, convertible_to=None,
...
+
+  def Parse(self, inp):
+    """Parse the input.
+
+    Args:
+      inp. A string or the result of yaml.load. If a string, should be
+      a valid YAML document.
+    """
+
+    if isinstance(inp, str):
```
skschneider: Checks like this should use basestring - it will match str and unicode strings.

perfkitbenchmarker/flag_util.py
```diff
@@ -304,3 +306,60 @@ def DEFINE_units(name, default, help, convertible_to=None,
...
+      # Python type (example: '123'). In that case, the first
+      # yaml.load() in the config system will strip away the quotation
+      # marks, and this second yaml.load() will parse it as the
+      # non-string type. However, I think this is the best we can do
+      # without significant changes to the config system, and the
+      # problem is unlikely to occur in PKB.
+      try:
+        return yaml.load(cStringIO.StringIO(inp))
+      except yaml.YAMLError:
+        raise ValueError("Couldn't parse YAML string '%s'" % inp)
```
skschneider: Consider flags.IllegalFlagValue to make it consistent with other parsers. Also, try to include the description of the YAMLError to help the user pinpoint what's wrong with the input value.

perfkitbenchmarker/flag_util.py
```diff
@@ -304,3 +306,60 @@ def DEFINE_units(name, default, help, convertible_to=None,
...
+    else:
+      return inp
+
+
+class YAMLSerializer(flags.ArgumentSerializer):
+  def Serialize(self, val):
```
skschneider: Nit: add a blank line between the class and function.

tests/flag_util_test.py
```diff
@@ -173,5 +173,18 @@ def testConvertibleToWrongUnit(self):
       up.Parse('1m')
 
 
+class TestYAMLParser(unittest.TestCase):
+  def setUp(self):
```
skschneider: nit: blank line between class and function.

perfkitbenchmarker/flag_util.py
```diff
@@ -304,3 +306,60 @@ def DEFINE_units(name, default, help, convertible_to=None,
...
+
+# The YAML flag type is necessary because flags can be read either via
+# the command line or from a config file. If they come from a config
+# file, they will already be parsed as YAML, but if they come from the
+# command line, they will be raw strings. The point of this flag is to
+# guarantee a consistent representation to the rest of the program.
+class YAMLParser(flags.ArgumentParser):
+  """Parse a flag containing YAML."""
+
+  syntactic_help = 'A YAML expression. Ex: "{x: y, foo: 2.3}".'
```
skschneider: Just something to keep in mind: if this is going to end up in --help output, it may not be useful for most cases, where a more specific format of input is expected. You may need to subclass this for specific flags to provide a more specific description, e.g. "an integer or a string", or you may want to make it an instance member and allow it to be overridden in DEFINE_yaml, but if so, none of the more specific checks would happen in the flag parser itself and would need to happen in a config option decoder later.

perfkitbenchmarker/flag_util.py
```diff
@@ -304,3 +306,60 @@ def DEFINE_units(name, default, help, convertible_to=None,
...
+      # This will work unless the user writes a config with a quoted
+      # string that, if unquoted, would be parsed as a non-string
+      # Python type (example: '123'). In that case, the first
+      # yaml.load() in the config system will strip away the quotation
+      # marks, and this second yaml.load() will parse it as the
+      # non-string type. However, I think this is the best we can do
+      # without significant changes to the config system, and the
+      # problem is unlikely to occur in PKB.
+      try:
+        return yaml.load(cStringIO.StringIO(inp))
```
skschneider: Why use a `StringIO` instead of just `yaml.load(inp)`?

skschneider:
Took a pass.

perfkitbenchmarker/flag_util.py
skschneider: Actually, I recall gflags has validators that can be attached to specific flags if you want to go the instance member route.

perfkitbenchmarker/flag_util.py
noahl: Fixed.

perfkitbenchmarker/flag_util.py
noahl: I was copying the IntegerListParser above, which also raises ValueError.

I just looked at gflags.py, and it looks like validators are expected to raise IllegalFlagValue, but parsers actually raise ValueError and the Flag object converts ValueError to IllegalFlagValue.

I'm guessing that's because the gflags authors wanted to be able to reuse builtins like `int` as parsers, although I'm not sure if that reason justifies that design.

perfkitbenchmarker/flag_util.py
noahl: Done.

tests/flag_util_test.py
noahl: Done.

perfkitbenchmarker/flag_util.py
noahl: Good point. For now, I would like to keep the generic test as "A YAML expression", but remove the example, which may be misleading. Specific flags can have examples in their help strings, like they do already for other flag types.

perfkitbenchmarker/flag_util.py
noahl: Because `yaml.load()` needs a stream, not a string. I'll add a comment about that.

noahl:
@skschneider I've updated in response to your comments. Please take another look.

perfkitbenchmarker/flag_util.py
noahl: Talked to @skschneider in person. I was wrong - `yaml.load()` does accept a string. I simplified the code.

perfkitbenchmarker/flag_util.py
skschneider: Ah, alright. Might help to have a test case where the parsing fails. I still recommend including YAMLError's more detailed information about the failure.

skschneider:
Looks like StringIO changes haven't been pushed to GH yet. Just that and a failed parsing test case, otherwise LGTM.

noahl:
Thanks.

voellm:
Is there an update to "help" on how to use this?  Is one needed?

On Wed, Mar 2, 2016 at 12:47 PM, noahl <notifications@github.com> wrote:

> Merged #857
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/857>.
>
> —
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/857#event-575277943>
> .
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382

noahl:
I don't think an update is needed for this. Individual yaml flags should have help strings describing how to use them. This is purely an internal infrastructure update. I am about to push the first commit which actually creates a YAML flag, and it will have a help string :).

PR #859, opened 2016-03-04
gce_virtual_machine: Support user provided instance metadata

wonderfly:
Two of the instance metadata keys that are particularly interesting are
'startup-script' and 'user-data', which are reserved by google-startup-script
and Cloud
Config(https://cloudinit.readthedocs.org/en/latest/topics/format.html#cloud-config-data)
to run user specified workloads at boot time. This can be a useful way to
measure VM boot time as a lot cloud workloads are started through some init
scripts, instead of SSH. Plus, sshd on GCP images is usually started at a
pretty late stage.

@voellm @skschneider @cmccoy

### DIFF
--- a/perfkitbenchmarker/providers/gcp/flags.py
+++ b/perfkitbenchmarker/providers/gcp/flags.py
@@ -35,3 +35,21 @@ flags.DEFINE_string(
     'https://cloud.google.com/sdk/gcloud/reference/compute/disks/create')
 flags.DEFINE_string('gce_network_name', None, 'The name of an already created '
                     'network to use instead of creating a new one.')
+flags.DEFINE_multistring(
+    'gcp_instance_metadata_from_file',
+    [],
+    'A colon separated key-value pair that will be added to the '
+    '"--metadata-from-file" flag of the gcloud cli (with the colon replaced by '
+    'the equal sign). Multiple key-value pairs may be specified by separating '
+    'each pair by commas. This option can be repeated multiple times. For '
+    'information about GCP instance metadata, see: --metadata-from-file from '
+    '`gcloud help compute instances create`.')
+flags.DEFINE_multistring(
+    'gcp_instance_metadata',
+    [],
+    'A colon separated key-value pair that will be added to the '
+    '"--metadata" flag of the gcloud cli (with the colon replaced by the equal '
+    'sign). Multiple key-value pairs may be specified by separating each pair '
+    'by commas. This option can be repeated multiple times. For information '
+    'about GCP instance metadata, see: --metadata from '
+    '`gcloud help compute instances create`.')
--- a/perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
+++ b/perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
@@ -25,6 +25,7 @@ operate on the VM: boot, shutdown, etc.
 """
 
 import json
+import logging
 import re
 import yaml
 
@@ -32,7 +33,9 @@ from perfkitbenchmarker import disk
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import events
 from perfkitbenchmarker import flags
+from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import linux_virtual_machine as linux_vm
+from perfkitbenchmarker import providers
 from perfkitbenchmarker import virtual_machine
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker import windows_virtual_machine
@@ -41,7 +44,6 @@ from perfkitbenchmarker.configs import spec
 from perfkitbenchmarker.providers.gcp import gce_disk
 from perfkitbenchmarker.providers.gcp import gce_network
 from perfkitbenchmarker.providers.gcp import util
-from perfkitbenchmarker import providers
 
 FLAGS = flags.FLAGS
 
@@ -283,11 +285,32 @@ class GceVirtualMachine(virtual_machine.BaseVirtualMachine):
       cmd.flags['machine-type'] = self.machine_type
     cmd.flags['tags'] = 'perfkitbenchmarker'
     cmd.flags['no-restart-on-failure'] = True
-    cmd.flags['metadata-from-file'] = 'sshKeys=%s' % ssh_keys_path
-    metadata = ['owner=%s' % FLAGS.owner]
-    for key, value in self.boot_metadata.iteritems():
-      metadata.append('%s=%s' % (key, value))
-    cmd.flags['metadata'] = ','.join(metadata)
+
+    metadata_from_file = {'sshKeys': ssh_keys_path}
+    parsed_metadata_from_file = flag_util.ParseKeyValuePairs(
+        FLAGS.gcp_instance_metadata_from_file)
+    for key, value in parsed_metadata_from_file.iteritems():
+      if key in metadata_from_file:
+        logging.warning('Metadata "%s" is set internally. Cannot be overridden '
+                        'from command line.' % key)
+        continue
+      metadata_from_file[key] = value
+    cmd.flags['metadata-from-file'] = ','.join([
+        '%s=%s' % (k, v) for k, v in metadata_from_file.iteritems()
+    ])
+
+    metadata = {'owner': FLAGS.owner} if FLAGS.owner else {}
+    metadata.update(self.boot_metadata)
+    parsed_metadata = flag_util.ParseKeyValuePairs(FLAGS.gcp_instance_metadata)
+    for key, value in parsed_metadata.iteritems():
+      if key in metadata:
+        logging.warning('Metadata "%s" is set internally. Cannot be overridden '
+                        'from command line.' % key)
+        continue
+      metadata[key] = value
+    cmd.flags['metadata'] = ','.join(
+        ['%s=%s' % (k, v) for k, v in metadata.iteritems()])
+
     if not FLAGS.gce_migrate_on_maintenance:
       cmd.flags['maintenance-policy'] = 'TERMINATE'
     ssd_interface_option = NVME if NVME in self.image else SCSI
--- a/tests/gce_virtual_machine_test.py
+++ b/tests/gce_virtual_machine_test.py
@@ -16,6 +16,7 @@
 
 import contextlib
 import mock
+import re
 import unittest
 
 from perfkitbenchmarker import benchmark_spec
@@ -227,6 +228,8 @@ class GCEVMFlagsTestCase(unittest.TestCase):
     self._mocked_flags.gcloud_path = 'test_gcloud'
     self._mocked_flags.os_type = os_types.DEBIAN
     self._mocked_flags.run_uri = 'aaaaaa'
+    self._mocked_flags.gcp_instance_metadata = []
+    self._mocked_flags.gcp_instance_metadata_from_file = []
     # Creating a VM object causes network objects to be added to the current
     # thread's benchmark spec. Create such a benchmark spec for these tests.
     self.addCleanup(context.SetThreadBenchmarkSpec, None)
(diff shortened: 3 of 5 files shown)


perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -35,3 +35,10 @@
     'https://cloud.google.com/sdk/gcloud/reference/compute/disks/create')
 flags.DEFINE_string('gce_network_name', None, 'The name of an already created '
                     'network to use instead of creating a new one.')
+flags.DEFINE_list('vm_metadata_from_string', [],
+                  'Instance metadata specified through local files. '
+                  'See: --metadata-from-file from '
+                  '`gcloud help compute instances create`.')
+flags.DEFINE_list('vm_metadata', [],
```
skschneider: I know we haven't been consistent about this with the earlier flags, but could you prefix these new ones with 'gcp_'? We've been doing that for other provider-specific flags because the names are global. Alternately, see more general comments I'll post in a bit in the Conversation tab.

perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -35,3 +35,10 @@
     'https://cloud.google.com/sdk/gcloud/reference/compute/disks/create')
 flags.DEFINE_string('gce_network_name', None, 'The name of an already created '
                     'network to use instead of creating a new one.')
+flags.DEFINE_list('vm_metadata_from_string', [],
```
skschneider: Was this meant to be vm_metadata_from_file? That's what I see in your changes to gce_virtual_machine.py and gce_virtual_machine_test.py.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +283,23 @@ def _GenerateCreateCommand(self, ssh_keys_path):
       cmd.flags['machine-type'] = self.machine_type
     cmd.flags['tags'] = 'perfkitbenchmarker'
     cmd.flags['no-restart-on-failure'] = True
-    cmd.flags['metadata-from-file'] = 'sshKeys=%s' % ssh_keys_path
-    metadata = ['owner=%s' % FLAGS.owner]
-    for key, value in self.boot_metadata.iteritems():
-      metadata.append('%s=%s' % (key, value))
-    cmd.flags['metadata'] = ','.join(metadata)
+
+    metadata_from_file = {'sshKeys': ssh_keys_path} if ssh_keys_path else {}
```
skschneider: Why the else? It looks like ssh_keys_path will always be set.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +283,23 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
+    for pair in FLAGS.vm_metadata_from_file:
+      key, value = pair.split('=')
+      metadata_from_file.setdefault(key, value)
```
skschneider: Looks like setting --vm_metadata_from_file sshKeys=value will cause the command-line value to be silently dropped. Maybe there should be an error or warning.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +283,23 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
+    cmd.flags['metadata-from-file'] = ','.join([
+        '%s=%s' % (key, value) for key, value in metadata_from_file.iteritems()
```
skschneider: flake8 complains here about key and value being redefined (it was defined once in the for loop above, and it's being redefined here in the list comprehension). Please rename these.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +283,23 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
+    ])
+
+    metadata = {'owner': FLAGS.owner} if FLAGS.owner else {}
+    metadata.update(self.boot_metadata)
+    for pair in FLAGS.vm_metadata:
+      key, value = pair.split('=')
+      metadata.setdefault(key, value)
+    cmd.flags['metadata'] = ','.join(
+        ['%s=%s' % (key, value) for key, value in metadata.iteritems()])
```
skschneider: Same flake8 issue here.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -416,6 +428,7 @@ class ContainerizedGceVirtualMachine(GceVirtualMachine,
   DEFAULT_IMAGE = UBUNTU_IMAGE
 
 
+
```
skschneider: No need for extra space here (two is enough).

skschneider:
Thanks for the PR. Two general comments:

1. Have you looked into whether other clouds support a similar metadata-on-boot that could warrant making the flag global, rather than GCP-specific? If that's the case, I recommend moving the flag out of the GCP module. If not, the flag should be prefixed, as mentioned in my earlier code comment. I don't think it would be a big deal if we started by prefixing the flag even if we were unsure, because if we later learned there was another cloud that supported the feature, we could easily make a duplicate flag without the prefix that does the same thing.
2. Would it make sense to treat these new flags more like the existing `--metadata` flag, which is a defined as a multistring and uses the format `--metadata=k0:v0,k1:v1`?  The way the `--metadata` flag is handled, you can put it twice in the same command, making it easier for scripts to add more metadata to a base command stored in a variable that already adds some metadata. On the other hand, I see that the `k=v` syntax you're using more closely matches the gcloud commands, but if we're going to be consistent with something, I'd cast my vote for the method that avoids commands like `--gcp_vm_metadata=key=value`.

perfkitbenchmarker/providers/gcp/flags.py
wonderfly: Done. Prefixed them with gcp_instance, i.e., `gcp_instance_metadata` and `gcp_instance_metadata_from_file`.

perfkitbenchmarker/providers/gcp/flags.py
wonderfly: Good catch.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done. You are right it is always set.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done. That was intended, but logging a warning seems the right call.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done. Mistake.

wonderfly:
@skschneider  Thanks for the comments. PTAL.

> Have you looked into whether other clouds support a similar metadata-on-boot that could warrant making the flag global, rather than GCP-specific?

I looked at it. EC2 seems to support similar instance metadata as GCP does, but it has a different format for configuring those metadata - may not be simple key-value pairs, or may not be applied the same way gcloud applies them. I am not familiar with other cloud providers but I assume it'll be hard to unify VM metadata across the board into one common flag, and that is clearly beyond my reach. So I'd not do it in this PR.

> Would it make sense to treat these new flags more like the existing --metadata flag, which is a defined as a multistring and uses the format --metadata=k0:v0,k1:v1?

Yes, and that seems neat.

> I'd cast my vote for the method that avoids commands like --gcp_vm_metadata=key=value

You don't necessarily have to write it that way - you can do something like `--gcp_instance_metadata k1=v1,k2=v2` when you use gflags, but looking at the full pkb command line dump, I think making all key value pairs conform with the same format makes sense, so I changed to use colon.

perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -35,3 +35,21 @@
...
 flags.DEFINE_string('gce_network_name', None, 'The name of an already created '
                     'network to use instead of creating a new one.')
+flags.DEFINE_multistring(
+    'gcp_instance_metadata_from_file',
+    [],
+    'A colon separated key-value pair that will be added to the '
+    '"--metadata-from-file" flag of the gcloud cli (with the colon replaced by '
+    'the equal sign). Multiple key-value pairs may be specified by separating '
+    'each pair by commas. This option can be repeated multiple times. For '
+    'informatin about GCP instance metadata, see: --metadata-from-file from '
```
skschneider: spelling: "informatin"

perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -35,3 +35,21 @@
...
+    '`gcloud help compute instances create`.')
+flags.DEFINE_multistring(
+    'gcp_instance_metadata',
+    [],
+    'A colon separated key-value pair that will be added to the '
+    '"--metadata" flag of the gcloud cli (with the colon replaced by the equal '
+    'sign). Multiple key-value pairs may be specified by separating each pair '
+    'by commas. This option can be repeated multiple times. For information '
+    'about GCP instance metadata, see: -metadata-from-file from '
```
skschneider: spelling: "-metadata-from-file" should have one more hyphen upfront.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +284,31 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
     cmd.flags['tags'] = 'perfkitbenchmarker'
     cmd.flags['no-restart-on-failure'] = True
-    cmd.flags['metadata-from-file'] = 'sshKeys=%s' % ssh_keys_path
-    metadata = ['owner=%s' % FLAGS.owner]
-    for key, value in self.boot_metadata.iteritems():
-      metadata.append('%s=%s' % (key, value))
-    cmd.flags['metadata'] = ','.join(metadata)
+
+    metadata_from_file = {'sshKeys': ssh_keys_path}
+    for pair in FLAGS.gcp_instance_metadata_from_file:
```
skschneider: The flag description says comma-separate key-value pairs are supported, but I think DEFINE_multistring only allows the flag to be repeated and doesn't support the comma behavior by default. The `--metadata` flag had to handle it explicitly, see here: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/publisher.py#L180

skschneider: Tangent thought, but since we have at least three flags now that will be using this behavior, it may be good to separate it out into a helper. Someone can address that in a separate PR, though.

tests/gce_virtual_machine_test.py
```diff
@@ -267,6 +270,38 @@ def testImageProjectFlag(self):
...
+      actual_metadata = re.compile('--metadata\s+(.*)(\s+--)?').search(
+          ' '.join(issue_command.call_args[0][0])).group(1)
+      self.assertIn('k1=v1', actual_metadata)
+      self.assertIn('k2=v2', actual_metadata)
+      # Assert that FLAGS.owner is honored and added to instance metadata.
+      self.assertIn('owner=test-owner', actual_metadata)
+
+  def testVmMetadataFromFileFlag(self):
+    with self._PatchCriticalObjects() as issue_command:
+      self._mocked_flags.gcp_instance_metadata_from_file = ['k1:p1', 'k2:p2']
```
skschneider: Could you add a test case or modify this one to have a value that demonstrates the comma, like `['k1:p1,k2:p2', 'k3:p3']`?

skschneider:
Took another pass - main issue at this point is handling commas.

perfkitbenchmarker/providers/gcp/flags.py
wonderfly: Done

perfkitbenchmarker/providers/gcp/flags.py
wonderfly: Done

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done.

wonderfly:
> Took another pass - main issue at this point is handling commas.

Addressed the comments and created a helper function to parse key value pairs from command line. PTAL.

perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -35,3 +35,21 @@
...
+    'information about GCP instance metadata, see: --metadata-from-file from '
+    '`gcloud help compute instances create`.')
+flags.DEFINE_multistring(
+    'gcp_instance_metadata',
+    [],
+    'A colon separated key-value pair that will be added to the '
+    '"--metadata" flag of the gcloud cli (with the colon replaced by the equal '
+    'sign). Multiple key-value pairs may be specified by separating each pair '
+    'by commas. This option can be repeated multiple times. For information '
+    'about GCP instance metadata, see: --metadata-from-file from '
```
skschneider: `--metadata-from-file` should be `--metadata`?

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +285,32 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
-    for key, value in self.boot_metadata.iteritems():
-      metadata.append('%s=%s' % (key, value))
-    cmd.flags['metadata'] = ','.join(metadata)
+
+    metadata_from_file = {'sshKeys': ssh_keys_path}
+    parsed_metadata_from_file = flag_util.ParseKeyValuePairs(
+        FLAGS.gcp_instance_metadata_from_file)
+    for key, value in parsed_metadata_from_file.iteritems():
+      if key in metadata_from_file:
+        logging.warning('Metadata "%s" is set internally. Cannot be overriden '
```
skschneider: spelling: overriden -> overridden

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -283,11 +285,32 @@ def _GenerateCreateCommand(self, ssh_keys_path):
...
+    cmd.flags['metadata-from-file'] = ','.join([
+        '%s=%s' % (k, v) for k, v in metadata_from_file.iteritems()
+    ])
+
+    metadata = {'owner': FLAGS.owner} if FLAGS.owner else {}
+    metadata.update(self.boot_metadata)
+    parsed_metadata = flag_util.ParseKeyValuePairs(FLAGS.gcp_instance_metadata)
+    for key, value in parsed_metadata.iteritems():
+      if key in metadata:
+        logging.warning('Metadata "%s" is set internally. Cannot be overriden '
```
skschneider: spelling: overriden -> overridden

perfkitbenchmarker/providers/gcp/flags.py
wonderfly: Done

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
wonderfly: Done

skschneider:
LGTM other than the three small things, which you've fixed already. Let me know when you've pushed your changes to GitHub, and I can merge it.

wonderfly:
Thanks. Just pushed a new patch with the suggested changes.

skschneider:
Looks good, thanks! Merged.

Issue #855, opened 2016-03-01
openstack provider: feature request: option to disable floating IP allocation


darrylweaver:
Currently the Openstack provider will attempt to allocate a floating IP to all Virtual Machines on creation.
Some public clouds and many private cloud deployments will either be short of real floating IPs or have no need to use floating IPs at all, if networks are routed directly from the workstation running perfkit.

Therefore, there should be an option to disable the allocation of a floating IP and rely on only the internal IP only.

A flag such as "--openstack-disable-floating-ip=true" as an option for the openstack provider, which turns off the allocation of the floating IP and the only sets the internal IP address.

All benchmarks should then use the internal address only and only run the ip_address=internal benchmarks and disable the ip_address=external benchmark runs.

voellm:
You can force internal IP allocation today with --ip_addresses=INTERNAL for
all benchmark runs.  Does that work for you?

On Tue, Mar 1, 2016 at 4:41 AM, Darryl Weaver <notifications@github.com>
wrote:

> Currently the Openstack provider will attempt to allocate a floating IP to
> all Virtual Machines on creation.
> Some public clouds and many private cloud deployments will either be short
> of real floating IPs or have no need to use floating IPs at all, if
> networks are routed directly from the workstation running perfkit.
>
> Therefore, there should be an option to disable the allocation of a
> floating IP and rely on only the internal IP only.
>
> A flag such as "--openstack-disable-floating-ip=true" as an option for the
> openstack provider, which turns off the allocation of the floating IP and
> the only sets the internal IP address.
>
> All benchmarks should then use the internal address only and only run the
> ip_address=internal benchmarks and disable the ip_address=external
> benchmark runs.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/855>.
>



-- 
Anthony F. Voellm (aka Tony)
Google Voice:  (650) 516-7382
<https://www.google.com/voice/b/0?pli=1#phones>
Blog: http://perfguy.blogspot.com/
Benchmark like a pro... PerfKitBenchmarker
<https://github.com/GoogleCloudPlatform/PerfKitBenchmarker>

meteorfox:
@darrylweaver Sorry for hijacking your post.

[...]

PR #861, opened 2016-03-08
openstack: Add support for optional floating ip pool

meteorfox:
Tested on OpenStack Kilo.

Nova API version:

Id   | Status    | Updated | Min Version | Version 
------|------------|-------------|------------------|------------
v2.0 | SUPPORTED | 2011-01-21T11:33:21Z |             |         
v2.1 | CURRENT   | 2013-07-23T11:33:21Z | 2.1         | 2.3     

Tested the following combinations:

```bash
# --openstack_network defaults 'private'
./pkb.py --cloud=OpenStack --machine_type=m1.medium --image='Ubuntu 14.04' \
         --openstack_floating_ip_pool=public --benchmarks=netperf
```

```bash
# Explicit flags
./pkb.py --cloud=OpenStack --machine_type=m1.medium --image='Ubuntu 14.04' \
         --openstack_floating_ip_pool=public --openstack_network=service \
         --benchmarks=netperf
```

```bash
# No floating IP pool
./pkb.py --cloud=OpenStack --machine_type=m1.medium --image='Ubuntu 14.04'  \
         --openstack_network=service --benchmarks=netperf
```

### DIFF
--- a/perfkitbenchmarker/providers/openstack/flags.py
+++ b/perfkitbenchmarker/providers/openstack/flags.py
@@ -52,9 +52,9 @@ flags.DEFINE_string('openstack_private_network', 'private',
                     'Name of OpenStack private network.')
 
 flags.DEFINE_string('openstack_network', 'private',
-                    'Name of OpenStack network. Network provides '
+                    'Name of OpenStack network. This network provides '
                     'automatically allocated fixed-IP addresses to attached '
-                    'instances. Typically, network is used for internal '
+                    'instances. Typically, this network is used for internal '
                     'communication between instances. '
                     'If openstack_floating_ip_pool is not '
                     'set then this network will be used to communicate with '
@@ -64,8 +64,8 @@ flags.DEFINE_string('openstack_floating_ip_pool', None,
                     'Name of OpenStack floating IP-address pool. If set, '
                     'a floating-ip address from this pool will be associated'
                     'to each instance and will be used for communicating '
-                    'with it. An internally routable network must be set as '
-                    'well using the openstack_network flag.')
+                    'with it. To use this flag, an internally routable network '
+                    'must also be specified via the openstack_network flag.')
 
 flags.DEFINE_boolean('openstack_config_drive', False,
                      'Add possibilities to get metadata from external drive')
--- a/perfkitbenchmarker/providers/openstack/flags.py
+++ b/perfkitbenchmarker/providers/openstack/flags.py
@@ -44,10 +44,28 @@ flags.DEFINE_string('openstack_nova_endpoint_type',
                     'defaults to $NOVA_ENDPOINT_TYPE.')
 
 flags.DEFINE_string('openstack_public_network', None,
-                    'Name of OpenStack public network')
+                    '(DEPRECATED: Use openstack_floating_ip_pool) '
+                    'Name of OpenStack public network.')
 
 flags.DEFINE_string('openstack_private_network', 'private',
-                    'Name of OpenStack private network')
+                    '(DEPRECATED: Use openstack_network) '
+                    'Name of OpenStack private network.')
+
+flags.DEFINE_string('openstack_network', 'private',
+                    'Name of OpenStack network. Network provides '
+                    'automatically allocated fixed-IP addresses to attached '
+                    'instances. Typically, network is used for internal '
+                    'communication between instances. '
+                    'If openstack_floating_ip_pool is not '
+                    'set then this network will be used to communicate with '
+                    'the instance.')
+
+flags.DEFINE_string('openstack_floating_ip_pool', None,
+                    'Name of OpenStack floating IP-address pool. If set, '
+                    'a floating-ip address from this pool will be associated'
+                    'to each instance and will be used for communicating '
+                    'with it. An internally routable network must be set as '
+                    'well using the openstack_network flag.')
 
 flags.DEFINE_boolean('openstack_config_drive', False,
                      'Add possibilities to get metadata from external drive')
--- a/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
@@ -17,6 +17,7 @@ import threading
 import time
 
 from perfkitbenchmarker import virtual_machine, linux_virtual_machine
+from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker import providers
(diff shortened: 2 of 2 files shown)


meteorfox:
@noahl @kivio PTAL. Make sure to skim over issue #855 for context. Thanks!

meteorfox:
@voellm Also here's another one ready for review.

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -44,10 +44,28 @@
...
+                    '(DEPRECATED: Use openstack_floating_ip_pool) '
+                    'Name of OpenStack public network.')
 
 flags.DEFINE_string('openstack_private_network', 'private',
-                    'Name of OpenStack private network')
+                    '(DEPRECATED: Use openstack_network) '
+                    'Name of OpenStack private network.')
+
+flags.DEFINE_string('openstack_network', 'private',
+                    'Name of OpenStack network. Network provides '
```
skschneider: Is this an existing network, or the name of the one to create?

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -44,10 +44,28 @@
...
+                    'automatically allocated fixed-IP addresses to attached '
+                    'instances. Typically, network is used for internal '
```
skschneider: Unless network is a variable, I think descriptions like this would read easier as "the network", "this network", or similar.

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -44,10 +44,28 @@
...
+                    'communication between instances. '
+                    'If openstack_floating_ip_pool is not '
+                    'set then this network will be used to communicate with '
+                    'the instance.')
+
+flags.DEFINE_string('openstack_floating_ip_pool', None,
```
skschneider: Similar comments for this flag's description.

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -44,10 +44,28 @@
...
+                    'Name of OpenStack floating IP-address pool. If set, '
+                    'a floating-ip address from this pool will be associated'
+                    'to each instance and will be used for communicating '
+                    'with it. An internally routable network must be set as '
+                    'well using the openstack_network flag.')
```
skschneider: I think this last sentence could be a bit clearer. For example, "To use this flag, an internally routable network must also be specified via the openstack_network flag."

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -54,21 +55,47 @@ def __init__(self, vm_spec):
                                                FLAGS.run_uri)
         self.client = os_utils.NovaClient()
         self.public_network = os_network.OpenStackPublicNetwork(
-            FLAGS.openstack_public_network
+            FLAGS.openstack_floating_ip_pool
```
skschneider: Does it make sense to instantiate `self.public_network` if openstack_floating_ip_pool is None?

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -54,21 +55,47 @@ def __init__(self, vm_spec):
...
+        self.floating_ip = None
 
     def _Create(self):
         image = self.client.images.findall(name=self.image)[0]
         flavor = self.client.flavors.findall(name=self.machine_type)[0]
 
-        network = self.client.networks.find(
-            label=FLAGS.openstack_private_network)
-        nics = [{'net-id': network.id}]
+        # FIXME(meteofox): Remove --openstack_public_network and
```
skschneider: Typo in username?

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -54,21 +55,47 @@ def __init__(self, vm_spec):
...
+                             FLAGS.openstack_private_network)
+
+        if self.network_name:
+            self.private_net = self.client.networks.find(
+                label=self.network_name)
+        if self.floating_ip_pool_name:
+            self.public_net = self.client.networks.find(
+                label=self.floating_ip_pool_name)
+
+        if self.public_net is None and self.private_net is None:
```
skschneider: Based on the later assert, it seems an important point that an error is raised here if private_net is None. Consider nested ifs:
```
if self.private_net is None:
  if self.public_net is None:
    raise ...
  else:
    raise ...
```

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -54,21 +55,47 @@ def __init__(self, vm_spec):
...
+            raise errors.Error(
+                'Cannot build instance without a network. Make sure to set '
+                'either just --openstack_network or both --openstack_network '
+                'and --openstack_private_network flags.')
```
skschneider: Why both `--openstack_network` and `--openstack_private_network`? It looks like if both are set, then only `--openstack_network` is used.

skschneider:
Took a pass.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: heh.. can't even type my own username. thanks

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: Good catch. I definitely missed this one.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: `--openstack_network` is essentially a rename of  `--openstack_private_network`.  They behave the same way  as before, but we decided to drop 'private' from the name since 'private' is relative to how OpenStack was configured and it caused confusion. For the moment I deprecated `--openstack_private_network` to give some time before I completely remove it.

So, in the case that both are set then we prefer the new one, but we could have chosen any of them, as long as they were the same.

They 'new' behavior is that setting a floating IP pool is optional, and if one is given then whatever network that floating IP pool belongs to, it **must** route to the network set with `--openstack_network` or OpenStack will not allocate a floating IP for the instance since it has no way to reach it.

The usual network configuration for OpenStack[1] is to use an internal network that provides the fixed IPs at build time to the instances and then later a user can attach a floating IP from a public network since usually there is a limited number of IP addresses available for public addressing. 

[1] http://docs.openstack.org/openstack-ops/content/network_design.html#public_addressing

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: Good point.

meteorfox:
@skschneider Thanks for your comments. I'll post another commit with the changes soon.

perfkitbenchmarker/providers/openstack/flags.py
meteorfox: Currently, it needs to be an existing network. The same for the floating IP pool.

On public clouds, this would be the equivalent of your internal network, except it can be named anything which is why we take it as a parameter.

Defaults to 'private' since a lot of people keep that the default name.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
skschneider: Sorry, I'm still confused. I think I get the purpose of each new flag, and I understand why this error message recommends setting `just --openstack_network`, but why is this error message also recommending that I set `both --openstack_network and --openstack_private_network flags`? Should the second one be `--openstack_floating_ip_pool`?

meteorfox: I think I must have read that like 10 times and didn't notice it. Yes, the second one should be `--openstack_floating_ip_pool`. Thanks for catching that one.

meteorfox:
@skschneider PTAL, hopefully, I didn't make it worse. :)

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -53,43 +54,52 @@ def __init__(self, vm_spec):
...
 
     def _Create(self):
         image = self.client.images.findall(name=self.image)[0]
         flavor = self.client.flavors.findall(name=self.machine_type)[0]
+        self.private_net = self.client.networks.find(label=self.network_name)
+        if self.floating_ip_pool_name:
+            self.public_net = self.client.networks.find(
+                label=self.floating_ip_pool_name)
+
+        if self.private_net is None:
```
skschneider: If these variables are not None, is there any other way they could evaluate False? Could this be
```
if not self.private_net:
  if self.public_net:
    ...
  else:
    ...
```

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -120,18 +143,22 @@ def _PostCreate(self):
             instance = self.client.servers.get(self.id)
             status = instance.status
 
-        with self._floating_ip_lock:
-            self.floating_ip = self.public_network.get_or_create()
-            instance.add_floating_ip(self.floating_ip)
-            logging.info('floating-ip associated: {}'.format(
-                self.floating_ip.ip))
+        assert self.private_net is not None  # Previously checked, must be true.
```
skschneider: I recommend adding a string to this assert, or at least modifying the comment to say that the previous check was in _Create.

skschneider:
Just two small comments, but LGTM.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: That's a reasonable case, a user could pass an empty string which it would evaluate to False

voellm:
This is ready for integration.... just assigning to @skschneider  because he did the final review.

@meteorfox - complete at will.

meteorfox:
@voellm sure, thanks. I'll push the changes to address the last two comments, then I'll merge.

Issue #854, opened 2016-02-29
Benchmarker releases after 1.0.1 will not reuse available OpenStack Floating IPs


cbhall:
With PerfKit Benchmarker version 1.0.1, OpenStack Floating IPs that had been created but were currently unused were being reused by Benchmarker.  In later releases (1.1.0 and 1.2.0) Benchmarker will never reuse an existing (but currently unused) Floating IP, but will instead always attempt to create a new one.  In my tests, not reusing an existing Floating IP will result in a Benchmarker failure if no more Floating IPs can be created.

I believe I know what the problem is.  In the Bencharker version 1.0.1 code in providers/openstack/os_network.py, it looks for available existing Floating IPs with:
```
            floating_ips = self.__nclient.floating_ips.findall(
                fixed_ip=None,
                pool=self.ip_pool_name
            )
```
In the later releases, this code was changed to:
```
            floating_ips = self.__nclient.floating_ips.findall(
                fixed_ip=None,
                pool=self.ip_pool_name,
                status='Down'
            )
```
The problem is that Benchmarker is using the OpenStack v2.0 API which is documented at http://docs.clouda.ca/api/api-ref-compute-v2-ext.html.  In that API a 'status' value is not returned by the 'List Floating IPs' API call.  As an example, dumping the result of that API call on my OpenStack deployment, I see a result that looks like this (with the ip obscured):
`{"floating_ips": [{"instance_id": null, "ip": "x.xx.xx.xxx", "fixed_ip": null, "id": "10efed9c-9881-47b6-aa03-bdbb39b7decc", "pool": "ext_net"}, ...`

In the OpenStack v2.1 API which is documented at http://developer.openstack.org/api-ref-compute-v2.1.html, the same API call is not an extension any more, and it is named 'List Floating IP addresses'.  In the v2.1 API there is a 'status' value returned.

meteorfox:
@cbhall Thanks for reporting this issue. I believe this should be fixed on master already. We pinned the version to v2.1 due to similar issues. See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/829

Try the master branch with the latest commit, if it's still not working I'll take a look. Thanks!

voellm:
+Carlos +Mateusz

On Mon, Feb 29, 2016 at 2:30 PM, Brian Hall <notifications@github.com>
wrote:

> With PerfKit Benchmarker version 1.0.1, OpenStack Floating IPs that had
[...]

PR #862, opened 2016-03-08
openstack: Fix floating IP allocation and deallocation

meteorfox:
Fixes issue where floating IP were not being deallocated after benchmarks
were completed. Reuses floating IPs that have already been allocated but are
not in use, otherwise allocates a new one from the pool.

Addresses Issue #854


meteorfox:
@noahl @kivio PTAL. Thanks again!

noahl:
LGTM. Sorry it took me so long to review.

PR #866, opened 2016-03-08
Install AWS requirements in the test environment.

skschneider:
AWS disk integration tests actually provision AWS VMs.

### DIFF
--- a/tests/requirements_test.py
+++ b/tests/requirements_test.py
@@ -25,6 +25,11 @@ from perfkitbenchmarker import requirements
 
 _PATH = 'path'
 
+_FAKE_PACKAGE_REQUIREMENTS = """
+# This is not a real package.
+perfkitbenchmarker-fake-package>=1.2
+"""
+
 
 class _MockOpenRequirementsFile(object):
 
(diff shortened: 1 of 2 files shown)


tests/requirements_test.py
```diff
@@ -25,6 +25,11 @@
 
 _PATH = 'path'
 
+_FAKE_PACKAGE_REQUIREMENTS = """
```
noahl: Did you mean to include this file in this PR?

noahl:
I think you might have included an extra file accidentally. I'm not sure about this change - if you need `requirements-aws.txt` to use tox, why not just include the AWS deps in `requirements-testing.txt`? Or if you don't need them for testing, then why are they in the tox testing section?

I think we might need a separate tox environment for integration tests. That way the regular testing environment won't be affected by extra requirements for integration tests.

skschneider:
Separate integration test environment: #868 and #869.
If those changes are more favorable, I will close this PR.

skschneider:
Closing in favor of #868 and #869.

PR #868, opened 2016-03-09
Create a separate tox environment for integration tests.

skschneider:
Move integration tests to a separate directory.


noahl:
The code LGTM, but I have a question about the interface: will the integration tests run automatically if I just run `tox`? I think that would be a bad behavior.

Also, sorry to make you jump through so many hoops, but I think that given that `tox -e integration` sets `PERFKIT_INTEGRATION` for you, we should just drop the environment variable requirement. I can do that in a separate PR if you want.

skschneider:
Running just `tox` will use the default envlist, which just runs flake8 and the tests in the `tests` directory (resursively). But because PERFKIT_INTEGRATION is neither forwarded nor set, the integration tests will not run.

If you want to remove the PERFKIT_INTEGRATION variable, then all unit tests will probably need to be in a separate directory, similar to how I've split out the integration tests in this PR.

noahl:
I see. I think it might be nice to have separate folders, but that can be a separate PR from this one. LGTM.

PR #869, opened 2016-03-09
Add specific version requirement for colorama.

skschneider:
awscli and colorlog both take a dependency on colorama. However,
awscli has specific version requirements that may be ignored when
installing both requirements.txt and aws-requirements.txt unless
otherwise specified.


noahl:
LGTM

Issue #856, opened 2016-03-02
SPEC CPU2006 run fails to delete VM on OpenStack


cbhall:
Running SPEC CPU2006 on OpenStack with:
`./pkb.trace.py --cloud=OpenStack --benchmarks=speccpu2006 --benchmark_subset=all --image="Ubuntu 14.04.3 LTS" --machine_type=m1.small --openstack_private_network=private_net --openstack_public_network=ext_net`
takes about 11 hours.  This run always fails to delete the VM that was created.  Dumping the API requests and responses indicates that the problem is as follows.  A delete request is made:
`DELETE /v2/4f299816e33f49c7ae3819ee2d7d19d9/servers/bbe3f0cc-04c6-46d4-87a7-5ab0f04d7e66`
and the response back is `Authentication required`.  Next, an attempt is made to authenticate using `POST /v2.0/tokens`, but the body of the request is missing a password:
`{"auth": {"passwordCredentials": {"username": "demo", "password": null}, "tenantId": "4f299816e33f49c7ae3819ee2d7d19d9"}}`
Obviously, many previous authentications have taken place with the right password.  This particular case though is the first attempt that used `passwordCredentials` with a `tenantId` (a number of previous cases used a `tenantName`).

cbhall:
I forgot to mention that there was also an exception reported:
```
Connectio2016-03-02 01:40:47,955 2b78f930 MainThread speccpu2006(1/1) ERROR    Exception occurred while calling DeleteVm(x.xx.xx.xxx):
Traceback (most recent call last):
  File ".../PerfKitBenchmarker/perfkitbenchmarker/vm_util.py", line 245, in _ExecuteThreadCall
    queue.put(ThreadCallResult(call_id, target(*args, **kwargs), None))
  File ".../PerfKitBenchmarker/perfkitbenchmarker/benchmark_spec.py", line 307, in DeleteVm
    vm.Delete()
  File ".../PerfKitBenchmarker/perfkitbenchmarker/resource.py", line 135, in Delete
    self._DeleteResource()
  File ".../PerfKitBenchmarker/perfkitbenchmarker/vm_util.py", line 481, in WrappedFunction
    return f(*args, **kwargs)
  File ".../PerfKitBenchmarker/perfkitbenchmarker/resource.py", line 117, in _DeleteResource
    self._Delete()
[...]

PR #870, opened 2016-03-09
openstack: Use Keystone session for handling authentication

meteorfox:
Using a Keystone Session allows for clients (i.e. novaclient) to authenticate
on demand only on their first request, and can also handle re-authentication
when the token expires.

Addresses Issue #856


meteorfox:
@kivio @cmccoy PTAL thanks!

meteorfox:
@voellm This one is also ready for review, which addresses issue #856

voellm:
LGTM.  Code looks much simpler.

voellm:
@kivio - Also want to take a look today?  Otherwise I think @meteorfox should complete the request.

Issue #860, opened 2016-03-07
azure vm creation errors


jdc0589:
I'm running in to errors with the VM creation portion of an fio benchmark in Azure.

Steps for (me) to reproduce:
1. confirm azure cli version: `0.9.9 (node: 5.2.0)`), and pkb version: `v1.2.0`.
2. set azure cli to asm mode
3. run `pkb --cloud Azure --benchmarks=fio --scratch_disk_type=standard --scratch_disk_size=100`
4. error: `d931267b MainThread fio(1/1) vm_util.py:292 ERROR    Exception occurred while calling PrepareVm(<BaseVirtualMachine [ip=None, internal_ip=None]>)`

Full log, with personal data scrubbed: [pkb.log.txt](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/files/161922/pkb.log.txt)

skschneider:
Looks like Popen.communicate stdout from the `azure vm image list --json` command is being truncated:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L541
Python doc: https://docs.python.org/2/library/subprocess.html#subprocess.Popen.communicate
There's a warning that "The data read is buffered in memory, so do not use this method if the data size is large or unlimited."

We should create an alternative method of getting the output. In this particular case, Python's json module could probably work directly with the stream rather than getting the entire stdout as a string, but there are plenty of other places where we're using vm_util.IssueCommand where stdout may also be too large.

Can you provide information about the environment in which you are running PKB? It could be useful for testing.

jdc0589:
The response is pretty large, I had manually been calling the same `images` method earlier in the day just to remind myself of azure's weird VM names; it is extremely slow, and large.

It would be weird if it was getting truncated, there is certainly enough free memory to handle the response, so it must be very specific issue in the subprocess  module (if thats even the issue). My python fu is weak though.

Environtment: 
- Latest OSX, recent macbook pro
- zsh 5.2

skschneider:
I'll try to reproduce this later with OS X and/or zsh. For now, you should be able to sidestep the image list command if you explicitly specify the image (e.g. `--image=...`)

The default images I'm getting for Azure are currently, by OS:
windows: `a699494373c04fc0bc8f2bb1389d6106__Windows-Server-2012-R2-20160229-en.us-127GB.vhd`
[...]

PR #878, opened 2016-03-10
Use temp files for Popen stdout and stderr in IssueCommand.

skschneider:
Avoids issue of limited pipe size.
Addresses #860.

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -525,25 +525,28 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False,
   logging.info('Running: %s', full_cmd)
 
   shell_value = RunningOnWindows()
-  process = subprocess.Popen(cmd, env=env, shell=shell_value,
-                             stdin=subprocess.PIPE, stdout=subprocess.PIPE,
-                             stderr=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/vm_util.py
```diff
@@ -525,25 +525,30 @@ def IssueCommand(cmd, force_info_log=False, suppress_warning=False,
   logging.info('Running: %s', full_cmd)
 
   shell_value = RunningOnWindows()
-  process = subprocess.Popen(cmd, env=env, shell=shell_value,
-                             stdin=subprocess.PIPE, stdout=subprocess.PIPE,
-                             stderr=subprocess.PIPE)
+  with tempfile.TemporaryFile() as tf_stderr:
+    with tempfile.TemporaryFile() as tf_stdout:
```
meteorfox: LGTM. Minor style nitpick.

Python 2.7 supports multiple context manager in the 'with' statement.
https://docs.python.org/dev/whatsnew/2.7.html#other-language-changes

So, you could simplify it something like:

```python
with tempfile.TemporaryFile() as tf_stderr, tempfile.TemporaryFile() as tf_stdout:
  ...
```

skschneider: I'll shorten variable names. I think `tf_out` and `tf_err` should still be understandable.

PR #882, opened 2016-03-14
Add unit tests for scripts

noahl:
This requires adding a new tox environment with a modified PYTHONPATH,
and making the git hook machinery use that environment.


noahl:

stfeng:
LGTM

PR #888, opened 2016-03-15
Add Object Size Distributions

noahl:
Add object size distribution support to the api_multistream scenario of
the object_storage_service benchmark. Also introduce a
SizeDistributionIterator class to provide general support for
distributions, and change the object storage upload mechanism from
strings to streams to allow use of APIs that make distribution support
easier.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -324,14 +324,20 @@ def _MakeSwiftCommandPrefix(auth_url, tenant_name, username, password):
   return ' '.join(options)
 
 
-def _ProcessMultiStreamResults(raw_result, operation, results, metadata=None):
+def _ProcessMultiStreamResults(raw_result, operation, sizes,
+                               results, metadata=None):
   """Read and process results from the api_multistream worker process.
+
+  Results will be reported per-object size and combined for all
+  objects.
 
   Args:
     raw_result: string. The stdout of the worker process.
     operation: 'upload' or 'download'. The operation the results are from.
+    sizes: the object sizes used in the benchmark, in bytes.
     results: a list to append Sample objects to.
     metadata: dict. Base sample metadata
+
   """
 
   if metadata is None:
@@ -339,27 +345,44 @@ def _ProcessMultiStreamResults(raw_result, operation, results, metadata=None):
 
   records = json.loads(raw_result)
   metric_name = 'Multi-stream %s latency' % operation
-  # TODO: when object size distributions are added, support multiple
-  # sizes.
-  object_size = records[0]['size']
 
-  logging.info('Processing %s multi-stream %s results for object size %s',
-               len(records), operation, object_size)
-
-  # Once we land size distributions, we will report different latency
-  # for each object size.
   metadata = metadata.copy()
-  metadata['object_size_B'] = object_size
   metadata['num_streams'] = FLAGS.object_storage_multistream_num_streams
   metadata['objects_per_stream'] = (
       FLAGS.object_storage_multistream_objects_per_stream)
 
+  overall_metadata = metadata.copy()
+  # I would love to publish the full distribution in the metadata, but
+  # I'm afraid that including a complex value like a dictionary in
+  # metadata that is otherwise simple Python objects would break
+  # parsers. Also, we'd need to implement a custom serializer unless
+  # we wanted regular expression parsing of metadata to depend on
+  # Python's hash function. I add 'object_size_B':'distribution' for
+  # the full results because the metadata for the per-object-size
+  # values will include the 'object_size_B' field, and searching can
+  # be easier when all the records you care about have the same
+  # metadata fields.
+  overall_metadata['object_size_B'] = 'distribution'
+  logging.info('Processing %s multi-stream %s results for the full '
+               'distribution.', len(records), operation)
   _AppendPercentilesToResults(
       results,
       (record['latency'] for record in records),
       metric_name,
       'sec',
-      metadata)
+      overall_metadata)
+
+  for size in sizes:
+    metadata = metadata.copy()
+    metadata['object_size_B'] = size
+    logging.info('Processing %s multi-stream %s results for object size %s',
+                 len(records), operation, size)
+    _AppendPercentilesToResults(
+        results,
+        (record['latency'] for record in records if record['size'] == size),
+        metric_name,
+        'sec',
+        metadata)
 
 
 def _DistributionToBackendFormat(dist):
@@ -586,7 +609,8 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
           '--scenario=MultiStreamWrite'])
       write_out, _ = vm.RobustRemoteCommand(
           multi_stream_write_cmd, should_log=True)
-      _ProcessMultiStreamResults(write_out, 'upload', results,
+      _ProcessMultiStreamResults(write_out, 'upload',
+                                 size_distribution.iterkeys(), results,
                                  metadata=metadata)
 
       logging.info('Finished multi-stream write test. Starting multi-stream '
@@ -611,7 +635,8 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
       try:
         read_out, _ = vm.RobustRemoteCommand(
             multi_stream_read_cmd, should_log=True)
-        _ProcessMultiStreamResults(read_out, 'download', results,
+        _ProcessMultiStreamResults(read_out, 'download',
+                                   size_distribution.iterkeys(), results,
                                    metadata=metadata)
       except Exception as ex:
         logging.info('MultiStreamRead test failed with exception %s. Still '
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -352,16 +352,12 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
       FLAGS.object_storage_multistream_objects_per_stream)
 
   overall_metadata = metadata.copy()
-  # I would love to publish the full distribution in the metadata, but
-  # I'm afraid that including a complex value like a dictionary in
-  # metadata that is otherwise simple Python objects would break
-  # parsers. Also, we'd need to implement a custom serializer unless
-  # we wanted regular expression parsing of metadata to depend on
-  # Python's hash function. I add 'object_size_B':'distribution' for
-  # the full results because the metadata for the per-object-size
-  # values will include the 'object_size_B' field, and searching can
-  # be easier when all the records you care about have the same
-  # metadata fields.
+  # Don't publish the full distribution in the metadata because doing
+  # so might break regexp-based parsers that assume that all metadata
+  # values are simple Python objects. However, do add an
+  # 'object_size_B' metadata field even for the full results because
+  # searching metadata is easier when all records with the same metric
+  # name have the same set of metadata fields.
   overall_metadata['object_size_B'] = 'distribution'
   logging.info('Processing %s multi-stream %s results for the full '
                'distribution.', len(records), operation)
--- a/perfkitbenchmarker/scripts/object_storage_api_tests.py
+++ b/perfkitbenchmarker/scripts/object_storage_api_tests.py
@@ -204,12 +204,10 @@ class SizeDistributionIterator(object):
       self.sizes.append(size)
       self.cutoffs.append(total_percent)
 
-    # I considered storing probabilities instead of percents, but I'm
-    # afraid that there is an edge case where the probability
-    # corresponding to a percent doesn't have a terminating binary
-    # representation and then the total percent is less than 1.0 even
-    # though the percentages add up to 100. I can't prove that this
-    # happens, but storing percentages seems safer.
+    # We store percents instead of probabilities because it avoids any
+    # possible edge cases where the percentages would add to 100.0 but
+    # their equivalent probabilities would have truncated
+    # floating-point representations and would not add to 1.0.
     assert total_percent == 100.0
 
   # this is required by the Python iterator protocol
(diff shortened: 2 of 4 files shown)


noahl:

I have verified that the `api_multistream` works on GCP, AWS and Azure (although the AWS and Azure tests were with very, very slightly different versions of the code). In addition, I checked that the `api_data` scenario works on GCP, because that depends on `WriteObjects`, which I changed.

There are also several unit tests for the distribution code itself.

stfeng:
Ack. I will take a look at this.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -339,27 +345,44 @@ def _ProcessMultiStreamResults(raw_result, operation, results, metadata=None):
...
   _AppendPercentilesToResults(
       results,
       (record['latency'] for record in records),
       metric_name,
       'sec',
-      metadata)
+      overall_metadata)
+
+  for size in sizes:
+    metadata = metadata.copy()
```
stfeng: can you make only one copy of the metadata and then overwrite the size inside the for loop?

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -187,32 +187,55 @@ class SizeDistributionIterator(object):
...
+    self.sizes = []
+    self.cutoffs = []
+
+    total_percent = 0.0
+    for size, percent in dist.iteritems():
+      total_percent += percent
+      self.sizes.append(size)
+      self.cutoffs.append(total_percent)
+
+    # I considered storing probabilities instead of percents, but I'm
```
stfeng: Consider shortening this comment and remove the usage of "I" from all comments in this file.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -187,32 +187,55 @@ class SizeDistributionIterator(object):
...
+    # afraid that there is an edge case where the probability
+    # corresponding to a percent doesn't have a terminating binary
+    # representation and then the total percent is less than 1.0 even
+    # though the percentages add up to 100. I can't prove that this
+    # happens, but storing percentages seems safer.
+    assert total_percent == 100.0
```
stfeng: if the user pass in a bad distribution, what would the error message look like? Would it be better if we do an if check, and the print out a friendly error message, and then throw exception? (depends on the answer to the first question, if the error message is obviously useful to the user with the current code, then we are fine)

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -187,32 +187,55 @@ class SizeDistributionIterator(object):
...
 
 
 def MaxSizeInDistribution(dist):
   """Find the maximum object size in a distribution."""
 
-  if len(dist) != 1:
-    raise NotImplementedError('General distributions are not implemented yet.')
-
-  return long(dist.iterkeys().next())
+  return max(dist.iterkeys())
```
stfeng: Consider saving the max in the initialization function and return the saved value directly from here?

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -354,7 +384,8 @@ def WriteObjects(storage_schema, bucket, object_prefix, count,
     host_to_connect: An optional endpoint string to connect to.
   """
 
-  payload_bytes, payload_string = GenerateWritePayload(size)
+  payload = GenerateWritePayload(size)
+  handle = cStringIO.StringIO(payload)
```
stfeng: have you confirmed whether creation of the handle creates a copy of the payload?

stfeng: Never mind, at this place we are generating the handle only once. should be fine...

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -370,14 +401,14 @@ def WriteObjects(storage_schema, bucket, object_prefix, count,
         latency_results.append(latency)
       if bandwidth_results is not None and latency > 0.0:
         bandwidth_results.append(size / latency)
+      handle.seek(0)
```
stfeng: wow.... seek is buried here. I'd strongly consider move the seek() to immediately before the place where the handle is going to be read,  that is, inside WriteObjectFromBuffer()

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -387,8 +418,8 @@ def WriteObjectFromBuffer(storage_schema, bucket_name, object_name,
     storage_schema: The address schema identifying a storage. e.g., "gs"
     bucket_name: Name of the bucket to write to.
     object_name: Name of the object.
-    payload_bytes: A bytearray to transfer.
-    payload_string: A string version of payload_bytes.
+    stream: A read()-able stream to transfer.
```
stfeng: Consider adding a condition/constraint here: 

"This function will always seek to the beginning of the stream and write the requested number of bytes from that position to the object storage"

This way, the caller does not have to shoulder the burden of doing seek(0) every time then call this function. (and currently, we don't have a use case to write from a middle of the stream).

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -405,10 +436,10 @@ def WriteObjectFromBuffer(storage_schema, bucket_name, object_name,
     if host_to_connect is not None:
```
stfeng: consider adding the seek(0) at here inside this function.

stfeng:
I've done my pass, mostly minor comments, thanks for all the unit tests!

Let me know once you are done incorporating my comments, I think we are really close to the finish line here for this one.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I don't think that will work, because each of the Sample object stores a pointer to its metadata dictionary. If I made only one copy, then the metadata for all of the samples would change when I updated inside the loop.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: If the user passes a bad distribution in, it should be caught in the PKB layer by `_DistributionToBackendFormat`, which processes all distributions before sending them to the API test script and has lots of informative error messages. It also checks that the percentages add to 100%, so this assertion should never fail in theory. It's here mostly as a sanity check.

I think it's fine to say that if you want to run the backend worker script yourself, you won't get great error messages (although the message here should still be pretty good). If you think differently, I will add nicer error reporting.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: I'm not sure which initialization function you mean. The API test script initialization? The `SizeDistributionIterator` initialization?

If you mean the `SizeDistributionIterator` initialization, that object itself won't work well because we construct the iterators in the worker threads, but `MaxSizeInDistribution` is called in the parent thread to construct the payload. I think the cleanest way to handle this would be to make an object that represents a distribution. It would have methods to get an iterator and to find the maximum object size in the distribution. I'm just not sure it's worth the extra code it would take to implement that. What would you like me to do?

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Yes, I confirmed that creating the handle does not copy the payload. I make a 2GB string in a Python process and then made 1000 handles to it. I checked top, and I couldn't see the memory use growing while I was creating the handles.

This handle is only created once, but I do create per-thread handles in `WriteWorker`.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: I see, MaxSizeInDistribution is called infrequently, by the parent thread only. (I thought it's called for each and every object, looks like not)

Then this approach of computing max over all keys every time this method is called works just fine.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: SGTM.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: SG.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
stfeng: SGTM.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Done. Moved to `WriteObjectFromBuffer()`.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Added.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: I added the seek, but moved it to before the timing starts, because I don't think it should be part of the operation latency. Other than that it's exactly like you suggested.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Done. Removed "I" from all comments, and also shortened several of them (including this one).

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -187,32 +187,53 @@ class SizeDistributionIterator(object):
...
+    # from our distribution. The data structure is two lists, sizes
+    # and cutoffs. If the random number is in the range [cutoffs[i-1],
+    # cutoffs[i]), then the random choice is sizes[i]. Cutoffs has an
+    # implied starting element 0.0, and its last element must be
+    # 100.0.
+    self.sizes = []
+    self.cutoffs = []
+
+    total_percent = 0.0
+    for size, percent in dist.iteritems():
```
meteorfox: Order matters here. You should use an `OrderedDict()`.

`dict()` does not guarantee the order of the keys. You could have the case that `next()` returns the incorrect size

Example:
```python
{1024: 10.0, 2048: 45.0, 1048576: 45.0}
```
Then but when iterating over it might give you the following sequence
```python
[(2048, 45.0), (1024, 10.0), (1048576, 45.0)]
```
and now your cutoff list is
```python
[45.0, 55.0, 100.0]
```
Now, say `random.random() * 100.0` returns `9.0`

Then your `next()` function will return `2048` but it should have been `1024`.

script-tests/object_storage_worker_tests.py
```diff
@@ -15,31 +15,61 @@
...
     iter = object_storage_api_tests.SizeDistributionIterator(dist)
 
-    lst = list(itertools.islice(iter, 5))
+    with mock.patch(random.__name__ + '.random') as rand:
+      rand.side_effect = [0.2, 0.7, 0.2]
+      values = list(itertools.islice(iter, 3))
+      # We should draw either 1, 10, 1 or 10, 1, 10, but which one
+      # depends on Python's hash function, and I don't want this test
+      # to depend on that.
+      self.assertTrue(values == [1, 10, 1] or values == [10, 1, 10])
```
meteorfox: See comment above about `OrderedDict()`

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: Such a great catch! Thank you Carlos!!!

noahl: @meteorfox thanks a lot for your review! I really appreciate you taking a look at my code.

I am happy to sort the sizes here. If this ever becomes a general-purpose distribution class, @klausw points out that sorted numbers are a much nicer interface than unsorted. Responding to your comment below, this change will also simplify the unit tests, since we will know exactly which `random()` values correspond to which sizes.

script-tests/object_storage_worker_tests.py
noahl: I responded above. :)

perfkitbenchmarker/scripts/object_storage_api_tests.py
meteorfox: @noahl No problem. Sorting the sizes in ascending order should work as well.

noahl:
@stfeng I pushed an update last night that I believe addresses all of your comments. I also verified that the `api_multistream` and `api_data` scenarios still work on GCP.

@meteorfox, @stfeng I pushed an update just now that responds to Carlos' comment. I verified that `api_multistream` still works on GCP, but I didn't check `api_data` again because the change doesn't affect that code path. Please take a look.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -187,32 +187,57 @@ class SizeDistributionIterator(object):
...
+    for size, percent in sorted_dist:
+      total_percent += percent
+      self.sizes.append(size)
+      self.cutoffs.append(total_percent)
+
+    # We store percents instead of probabilities because it avoids any
+    # possible edge cases where the percentages would add to 100.0 but
+    # their equivalent probabilities would have truncated
+    # floating-point representations and would not add to 1.0.
+    assert total_percent == 100.0
```
meteorfox: Consider adding an error message for when `total_percent != 100.0`, such as users setting percentages that do not add to exactly 100.

stfeng:
LGTM. Please try a run of api-namespace scenario with 100 or 200 iteration before merge. That test case also calls writeobject().

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: @meteorfox That error should be caught by the PKB driver script, which will print a nice error message if the percentages don't add to 100. This is just a sanity check.

noahl:
@stfeng I finished a run of 200 iterations of `api_namespace` overnight. Merging now.

PR #889, opened 2016-03-15
Add Flag for List Consistency Iterations

noahl:
Add the flag --object_storage_list_consistency_iterations, which lets
users specify the number of iterations to use for the list consistency
benchmark. (Previously the number was a constant 200.) This makes it
much easier to test changes to the object storage code.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -104,6 +104,15 @@ flags.DEFINE_integer('object_storage_multistream_num_streams', 10,
                      'Only applies to the api_multistream scenario.',
                      lower_bound=1)
 
+flags.DEFINE_integer('object_storage_list_consistency_iterations', 200,
+                     'Number of iterations to perform for the api_namespace '
+                     'list consistency benchmark. This flag is mainly for '
+                     'regression testing in the benchmarks. Reduce the number '
+                     'to shorten the execution time of the api_namespace '
+                     'scenario. However, to get useful metrics from the '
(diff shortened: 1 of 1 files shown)


noahl:
@yuyantingzero @stfeng please take a look. This is a very small PR.

I did this because when I make changes to the object storage code, I test all benchmarks that use the code. It takes a long time to wait for 200 iterations of the list consistency benchmark to complete, and all I really want is to see that the code completes without errors.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -104,6 +104,10 @@
                      'Only applies to the api_multistream scenario.',
                      lower_bound=1)
 
+flags.DEFINE_integer('object_storage_list_consistency_iterations', 200,
```
stfeng: consider add the following help text to the flag:

"this flag is introduced mainly for regression test purpose. Reduce this number to shorten execution time of the api_namespace scenario. However, to get useful metrics from the api_namespace scenario, a high number of iterations should be used (>=200)"

stfeng:
LGTM. only one minor comment about adding more helper text to the flag.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I added almost exactly your text to the flag help text.

noahl:
I added the help text you suggested and verified that the `api_namespace` scenario still works on GCP. Since you LGTM'd, I'll go ahead and merge.

PR #893, opened 2016-03-17
Add exception types to except clauses.

skschneider:


noahl:
The object storage benchmark part LGTM

perfkitbenchmarker/providers/__init__.py
```diff
@@ -67,6 +67,6 @@ def LoadProvider(provider_name, ignore_package_requirements=True):
                                               __name__ + '.' + provider_name)]
     if not modules:
       raise ImportError('No modules found for provider.')
-  except:
+  except ImportError:
```
ehankland: I think this was broad on purpose. Not all errors that you encounter while loading modules are import errors - if the python file exists, but has invalid syntax for example, I don't think that will raise an ImportError. In that case, I think it's still useful to have a message that tells you that the provider failed to load.

perfkitbenchmarker/traces/dstat.py
```diff
@@ -103,7 +103,7 @@ def _StopOnVm(self, vm):
     vm.RemoteCommand(cmd)
     try:
       vm.PullFile(self.output_directory, file_name)
-    except:
+    except Exception:
```
ehankland: I think PullFile calls RemoteCopy which raises errors.VirtualMachine.RemoteCommandError on failure.

skschneider: Changed to Exception. I'm just trying to cut down on places where things like KeyboardInterrupt or SystemExit can be involuntarily swallowed.

perfkitbenchmarker/providers/__init__.py
skschneider: Changed to Exception.

perfkitbenchmarker/traces/dstat.py
skschneider: Discussed offline. As a separate issue, we should address errors.VirtualMachine.RemoteExceptionError, which can also be raised in some situations.

ehankland:
LGTM

Issue #836, opened 2016-02-09
Add --object_storage_class flag for object_storage_service benchmark


zbjornson:
For GCS, allow users to spec standard|DRA|nearline.

(See discussion in #750.)

noahl:
This is fixed by #895.

PR #895, opened 2016-03-18
Support object storage classes

noahl:
Add --object_storage_storage_class flag, which lets users of the object
storage benchmark choose the storage class of their objects.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -59,6 +59,9 @@ flags.DEFINE_string('object_storage_region', None,
 flags.DEFINE_string('object_storage_gcs_multiregion', None,
                     'Storage multiregion for GCS in object storage benchmark.')
 
+flags.DEFINE_string('object_storage_storage_class', None,
+                    'Storage class to use in object storage benchmark.')
+
 flags.DEFINE_enum('object_storage_scenario', 'all',
                   ['all', 'cli', 'api_data', 'api_namespace',
                    'api_multistream'],
@@ -471,6 +474,8 @@ def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
         cmd_parts += [azure_command_suffix]
       if host_to_connect is not None:
         cmd_parts += ['--host', host_to_connect]
+      if FLAGS.object_storage_storage_class is not None:
+        cmd_parts += ['--storage_class', FLAGS.object_storage_storage_class]
 
       return ' '.join(cmd_parts)
 
@@ -1076,6 +1081,8 @@ class GoogleCloudStorageBenchmark(object):
     make_bucket_command = '%s mb' % vm.gsutil_path
     if FLAGS.object_storage_gcs_multiregion:
       make_bucket_command += ' -l %s' % FLAGS.object_storage_gcs_multiregion
+    if FLAGS.object_storage_storage_class:
+      make_bucket_command += ' -c %s' % FLAGS.object_storage_storage_class
     make_bucket_command += ' gs://%s' % vm.bucket_name
     vm.RemoteCommand(make_bucket_command)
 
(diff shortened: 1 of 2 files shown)


noahl:
This PR fixes #836 . @stfeng @yuyantingzero please take a look.

This PR adds an `--object_storage_storage_class` flag which applies to GCS and S3 (as far as I know, Azure and Swift don't have separate storage classes, so they are not affected). I've verified that the `api_multistream` scenario still succeeds on both GCS and S3, using NEARLINE and STANDARD_IA as the storage classes respectively. I still want to check that the other api scenarios succeed, at least on GCP, but I think code review can happen at the same time.

The patch takes two different approaches, because GCS and S3 require users to specify storage classes in different ways. In GCS, the storage class becomes an argument to the `gsutil` command that creates the bucket. In S3, the storage class becomes a property of the key that we upload with. (It doesn't affect downloads or list operations.)

I had to use a private boto method (`key._set_storage_class`) because the public method equivalent checks to make sure the passed-in storage class is valid, and unfortunately boto hasn't been updated in long enough that it would reject `STANDARD_IA` as invalid even though it is, in fact, an AWS storage class. This means that if the user passes an invalid storage class, the program will fail when the boto upload command fails, which I think is fine.

The new code path for uploading code to S3 with a storage class set is basically the implementation of `storage_uri.set_contents_from_string` (https://github.com/boto/boto/blob/develop/boto/storage_uri.py#L647), the method we were previously using to upload, but with only the parts that apply to S3. I had to use a lower-level upload method because I needed to get access to the key object so I could set its storage class.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -1049,6 +1054,8 @@ def Prepare(self, vm):
     make_bucket_command = '%s mb' % vm.gsutil_path
     if FLAGS.object_storage_gcs_multiregion:
       make_bucket_command += ' -l %s' % FLAGS.object_storage_gcs_multiregion
+    if FLAGS.object_storage_storage_class:
```
stfeng: Should this be: "if FLAGS.object_storage_storage_class is not None:"? The same as the way you do the check above in the Azure case...

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -100,6 +100,10 @@
                    'to start the operation. Only applies to the '
                    'MultiStreamRead and MultiStreamWrite scenarios.')
 
+flags.DEFINE_string('storage_class', None, 'The storage class to use for '
```
stfeng: Please rename this flag to "object_storage_class", and clarify in the text that this only applies to AWS because they support per-object storage class. For the other two providers, storage class is decided by the Bucket which is passed in as the --bucket parameter.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -405,7 +409,15 @@ def WriteObjectFromBuffer(storage_schema, bucket_name, object_name,
...
       object_uri.connect(host=host_to_connect)
 
-    object_uri.set_contents_from_string(payload_string)
+    if FLAGS.storage_class and storage_schema == 's3':
+      # We need to use a lower-level upload method when using an S3
+      # storage class because we need access to the key object so we
+      # can set its storage class.
+      key = object_uri.new_key()
+      key._set_storage_class(FLAGS.storage_class)
+      key.set_contents_from_string(payload_string)
```
stfeng: Please merge in the other PR you did that changed this function from calling set_contents_from_string() to set_contents_from_file().

stfeng:
Mostly minor comments, please merge-in the object_distribution PR which used the set_contents_from_file() function, and I will take another look at that time.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Yes, fixed.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Changed the name and the help text.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Done, and updated this section accordingly.

noahl:
@stfeng I responded to your comments and rebased to avoid merge conflicts. Please take a look.

stfeng:
LGTM

Issue #902, opened 2016-03-21
rackspace: Undefined flags AttributeError


meteorfox:
Using PKB with the Rackspace provider throws the following error:
```
 if FLAGS.use_security_group:
  File "/home/carl6796/.virtualenvs/pkb/local/lib/python2.7/site-packages/gflags.py", line 1059, in __getattr__
    raise AttributeError(name)
AttributeError: use_security_group
``

noahl:
@meteorfox is this fixed?

meteorfox:
Yes, that was the PR you just reviewed. :)

PR #844, opened 2016-02-12
rackspace: Migrate provider implementation to use official CLI.

meteorfox:
Migrate provider implementation to use official Rackspace CLI and introduces experimental support for security groups, and isolated networks.

Uses new Disk types:
  - `cbs-sata` for a HDD, or SATA, Cloud Block Storage Volume
  - `cbs-ssd` for a SSD Cloud Block Storage Volume
  - `boot` creates data disk mount point  as path within the boot disk.
  - `local` uses, if any, provided local disks that come with the machine type

Resolves Issue #533

Signed-off-by: Carlos L. Torres <carlos.torres@rackspace.com>

PR #903, opened 2016-03-21
rackspace: Fix missing flags bug

meteorfox:
After the rename for use_security_group and boot_from_cbs_volume, their
references were not updated causing fatal failure with flags not being defined.

Addresses Issue #902


meteorfox:
@noahl PTAL, This is related to commit f985590f8fbd8cac38c00ad8b028d0c3cbcbb4dd for #844. I forgot to update the references after renaming of the flags. That's embarrassing :anguished: 

I guess nobody except for me uses the rackspace implementation since nobody complained. :disappointed:

noahl:
@meteorfox LGTM. Sorry I missed that in my last code review!

And maybe it just means that no one else benchmarks with prerelease PKB code, which honestly is a pretty reasonable decision. :)

meteorfox:
Thanks!

Issue #874, opened 2016-03-10
Spurious error that deletion of OpenStack VM failed


cbhall:
In many, but not all, runs of PerfKit Benchmarker on OpenStack, the following error is reported:
`2016-03-10 03:55:46,244 73012e99 Thread-25 speccpu2006(1/1) ERROR    Got exception running _DeleteResource: Deletion of DebianBasedOpenStackVirtualMachine failed.`
I think this error is generally accompanied by a later message:
`2016-03-10 03:56:04,206 73012e99 Thread-25 speccpu2006(1/1) INFO     Instance not found, may have been already deleted`

The code that spuriously emits this error is as follows.  In perfkitbenchmarker/resource.py, `_DeleteResource` contains:
```
    self._Delete()
    try:
      if self._Exists():
        raise errors.Resource.RetryableDeletionError(
            'Deletion of %s failed.' % type(self).__name__)
```
The problem is that once `self._Delete()` has initiated the deletion of the VM, `self._Exists()` checks on the status of the VM and, depending on timing, could get a result such as this:
```
[...]

PR #904, opened 2016-03-21
openstack: Wait until VM deletion has been completed

meteorfox:
Previously, it only waited for 5 seconds after a delete call. Some occasions
deletes took longer than 5 seconds which caused a retry of the delete and
logged an error message. This patch uses vm_util.Retry to wait until VM has been
completely deleted before proceeding to the next step.

Addresses Issue: #874

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
+++ b/perfkitbenchmarker/providers/openstack/os_virtual_machine.py
@@ -164,7 +164,7 @@ class OpenStackVirtualMachine(virtual_machine.BaseVirtualMachine):
         from novaclient.exceptions import NotFound
         try:
             self.client.servers.delete(self.id)
-            time.sleep(5)
+            self._WaitForDeleteCompletion()
         except NotFound:
             logging.info('Instance not found, may have been already deleted')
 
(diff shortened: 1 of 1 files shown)


meteorfox:
@skschneider PTAL

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -190,6 +190,17 @@ def WaitForBootCompletion(self):
         if self.hostname is None:
             self.hostname = resp[:-1]
 
+    @vm_util.Retry(poll_interval=5, max_retries=-1, timeout=300,
+                   log_errors=False,
+                   retryable_exceptions=(
+                       errors.Resource.RetryableDeletionError,))
```
skschneider: A bit of a tangent, but it looks like we have a lot of places where retryable_exceptions lists only one type, and it looks icky as a tuple. As far as I can tell, Retry should be able to work if you just pass `retryable_exceptions=errors.Resource.RetryableDeletionError`, but the Retry docstring doesn't explicitly say so. Maybe we should allow this.

skschneider:
LGTM.

PR #905, opened 2016-03-21
Analysis Step for Object Storage Multistream Scenario

noahl:
Add a module 'perfkitbenchmarker.analyze' containing utility functions
for analyzing benchmark data. Use it to implement an analysis step for
the 'api_multistream' scenario of the object storage service benchmark.


noahl:

stfeng:
Ack. I will look at this.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -351,43 +354,88 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+      name_prefix=throughput_prefix,
+      metadata=distribution_metadata)
+
+  logging.info('Processing %s multi-stream %s results for net throughput',
+               len(records), operation)
+  results.append(
+      sample.Sample(
+          'Multi-stream %s net throughput' % operation,
+          records_in_interval['size'].sum() /
+          records_in_interval['latency'].sum(),
```
stfeng: I don't think this is right... the sum of all latency data points would far exceed the actual overall duration these multiple streams took to upload or download the data. 

Also, are we sure that summing up the size of each records_in_interval equals to the total number of bytes transferred? I think so, but worth confirming with you.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -351,43 +354,88 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
-        metric_name,
+    analyze.AppendStatsAsSamples(
+        this_size_records['latency'],
         'sec',
-        metadata)
+        results,
+        name_prefix=latency_prefix,
+        metadata=this_size_metadata)
+
+    analyze.AppendStatsAsSamples(
```
stfeng: The more I think about this, the more I believe it probably doesn't make sense to report throughput per size. The throughput is only interesting when reported as an aggregated number, not sure how people can make sense of a per-size throughput which is performed in the context of a multi-stream operation in concurrency with many other different sizes.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Oh, shoot! That's the wrong calculation. I think my first mistake was doing the net throughput calculation here, where it is difficult to test. So I moved it to analyze.py, added a unit test, and pushed a change that should fix the calculation.

Yes, summing the size of records in records_in_interval equals the total number of bytes transferred during the interval where all streams were active.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: That makes sense. I'll take the per-size throughput out. I still think it makes sense to report per-size latency, though, like we had been. Do you agree?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: After thinking about it more, I have a better testing strategy for this function. I'll push additional tests tomorrow.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: After thinking about it, I don't think that per-object throughput percentiles make sense either (you might have meant this in your message; I'm just confirming). If you have some large and some small object sizes, the per-object throughput percentiles would probably just tell you that startup costs are a bigger portion of total upload/download time for smaller objects than for larger ones.

noahl:
@stfeng I just pushed an update that I think addresses our conversation and your code review comments. Please take a look.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
stfeng: I meant per-size throughput reporting. As to per-object throughput percentiles, it makes sense when object size is "large enough", e.g., for the current single stream throughput test, we use 100MiB object and we report per object throughput percentiles. 

Yes, it still makes sense to report latency percentiles as per-size.

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
...
+
+def AllStreamsInterval(start_times, durations, stream_ids):
+  """Compute when all streams were active from multistream records.
+
+  Args:
+    start_times: a pd.Series of POSIX timestamps, as floats.
+    durations: a pd.Series of durations, as floats measured in seconds.
+    stream_ids: a pd.Series of any type.
+
+  All three arguments must have matching indices. Each (start_time,
```
stfeng: add some asserts at the beginning of this function to check this. 

(if this condition is not met, wrong results would be produced, I'd rather see our program crash/assert than reporting wrong results)

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
...
+
+  stream_start_times = start_times.groupby(stream_ids).min()
+  stream_end_times = (start_times + durations).groupby(stream_ids).max()
+  interval_start = stream_start_times.max()
+  interval_end = stream_end_times.min()
+
+  return interval_start, interval_end - interval_start
+
+
+def StreamStartAndEndGaps(start_times, durations,
```
stfeng: Looking at the caller of this function, it seems to be called only by one place and immediately after AllStreamsInterval(), does it make sense to merge this function into AllStreamsInterval()? it seems unnecesary here to introduce a new function here to just compute the gaps

perfkitbenchmarker/analyze.py
stfeng: consider adding such asserts to all functions in this class that depends on such assumption (of the input arrays must have matching indices).

I realized it's hard to check for "matching" in this context here, but at the minimal, we can check if these are of the same length, right?

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
...
+    - The net throughput of all streams, excluding times they were inactive.
+    - The throughput of all streams, including times they were inactive.
+    - The total time that all streams were inactive.
+    - The total time all streams were inactive, as a proportion of the
+      total benchmark time.
+  """
+
+  total_bytes_by_stream = sizes.groupby(stream_ids).sum()
+  active_time_by_stream = durations.groupby(stream_ids).sum()
+  total_overhead = interval_duration * num_streams - active_time_by_stream.sum()
```
stfeng: Is there a requirement from the caller that the "durations" array must be those durations that are covered by "interval_duration"? (i.e. when all streams are actively doing work)

If such requirement is true, then it make sense here and total_overhead must be >= 0. Please add an assertion here. (If total_overhead is negative, it doesn't make sense and something is wrong with our calculation)

Please also consider renaming total_overhead. Visually, I consider these are "gaps" in between consecutive object uploads or downloads. It's a benchmarking overhead. but in terms of analysis, these are just "gaps" in between two durations? How do you think?

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
...
+
+  Args:
+    series: a pd.Series of floats.
+    name_prefix: if given, a prefix for the statistic names.
+
+  Returns: a pd.Series with summary statistics.
+  """
+
+  # Percentiles 0, 1, 2, ..., 100
+  percentiles = range(0, 101, 1)
```
stfeng: it's an overkill to compute percentiles to the ones digit, e.g.: p33, p34, p35 and p36, etc, no one is going to use it.

Also, this function functionally duplicates with the percentile calculator that we already have, can you please use that calculator instead of introduce a new one here?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+
+  latency_prefix = 'Multi-stream %s latency ' % operation
   logging.info('Processing %s multi-stream %s results for the full '
-               'distribution.', len(records), operation)
-  _AppendPercentilesToResults(
-      results,
-      (record['latency'] for record in records),
-      metric_name,
+               'distribution.', len(records_in_interval), operation)
+  analyze.AppendStatsAsSamples(
```
stfeng: I think we should really use the existing method _AppendPercentilesToResults() for consistency and easy maintenance. (*If* we want to change that function to use pandas and such, we should do that in a separate PR and change all references to it.)

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
...
+  result[name_prefix + 'median'] = result[name_prefix + 'p50']
+  result[name_prefix + 'mean'] = sum(values) / float(count)
+
+  sum_of_squares = sum([num ** 2 for num in values])
+  result[name_prefix + 'stddev'] = (sum_of_squares / (count - 1)) ** 0.5
+
+  return pd.Series(result)
+
+
+def AppendStatsAsSamples(series, unit, samples_list,
```
stfeng: I'd like to avoid introducing another function to compute percentiles and append to samples if possible. that's a duplicate to what we already have. See my comments below in the place where this function is called.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -242,6 +245,12 @@
 # benchmark. This is the filename.
 OBJECTS_WRITTEN_FILE = 'pkb-objects-written'
 
+# If the period of time when some streams, but not all, are active is
```
stfeng: Can you simplify this comment?

From the beginning of the test to the time when all streams are started and active, we call it the "Gap", if the Gap is above a threshold of total test time, we log a warning, because --- (fill your reason here)

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
-  overall_metadata = metadata.copy()
+  records_json = json.loads(raw_result)
+  records = pd.DataFrame(records_json)
+
+  all_streams_start, all_streams_duration = analyze.AllStreamsInterval(
+      records['start_time'], records['latency'], records['stream_num'])
+  start_gap, stop_gap = analyze.StreamStartAndEndGaps(
+      records['start_time'], records['latency'],
+      all_streams_start, all_streams_duration)
+  if ((start_gap + stop_gap) / all_streams_duration <
```
stfeng: should we do:

(start_gap + stop_gap) / (all_stream_duration + start_gap + stop_gap) ?

The (all_stream_duration + start_gap + stop_gap) time is our "total benchmarking time" which you referred to in your warning message below.

That is, should we warn when the gap exceeds X% of total benchmarking time or X% of total duration when all streams are active? visually, I believe the former is easier to understand.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+      MULTISTREAM_STREAM_GAP_THRESHOLD):
+    logging.info(
+        'First stream started %s seconds before last stream started', start_gap)
+    logging.info(
+        'Last stream ended %s seconds after first stream ended', stop_gap)
+  else:
+    logging.warning(
+        'Difference between first and last stream start/end times was %s and '
+        '%s, which is more than %s of the benchmark time %s.',
```
stfeng: "the benchmark time" should be "all_streams_duration + start_gap + stop_gap", right?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
-  metadata['num_streams'] = FLAGS.object_storage_multistream_num_streams
+  metadata['num_streams'] = num_streams
   metadata['objects_per_stream'] = (
       FLAGS.object_storage_multistream_objects_per_stream)
 
-  overall_metadata = metadata.copy()
+  records_json = json.loads(raw_result)
+  records = pd.DataFrame(records_json)
+
+  all_streams_start, all_streams_duration = analyze.AllStreamsInterval(
```
stfeng: put good comments here to explain what "all_streams_start", "all_streams_duration" and start_gap, stop_gap means.

stfeng: Also, I am trying to come up with a better name than "all_streams_start", "all_streams_duration". These names actually gives people a first impression: the start time or the duration of all streams, which means earliest start time and the entire duration covers all streams, which is exactly the opposite of what they actually mean.

all_streams_start is actually the latest start time of all streams, right?
more descriptives

how about these two? - 
"latest_start_time_of_all_streams"
"duration_when_all_streams_active"

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+  latency_prefix = 'Multi-stream %s latency ' % operation
   logging.info('Processing %s multi-stream %s results for the full '
-               'distribution.', len(records), operation)
-  _AppendPercentilesToResults(
-      results,
-      (record['latency'] for record in records),
-      metric_name,
+               'distribution.', len(records_in_interval), operation)
+  analyze.AppendStatsAsSamples(
+      records_in_interval['latency'],
```
stfeng: For latency percentiles and results reporting, I think we should just keep it simple and report all latency data as is, there is no need to go through this interval selection process which is only necessary for aggregated throughput calculation.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
       'sec',
-      overall_metadata)
-
+      results,
+      name_prefix=latency_prefix,
+      metadata=distribution_metadata)
+
+  logging.info('Processing %s multi-stream %s results for net throughput',
+               len(records), operation)
+  net_thpt, thpt_with_overhead, total_overhead, overhead_prop = (
```
stfeng: use complete names "thpt" -> throughput

"overhead" means "gaps" visually.

perfkitbenchmarker/analyze.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
```
stfeng: Think about this again, the name "analyze.py" is just way too generic. This class was made for object storage benchmarks, and lots of its logic was designed for the object storage multi-stream throughput analysis, at least for now.

Please either move this to the same directory where object storage benchmark lives, or rename it to something like object_storage_multistream_analyze.py.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +355,105 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+          all_streams_duration))
+  logging.info('Benchmark overhead was %s percent of total benchmark time',
+               overhead_prop * 100.0)
+
+  results.extend([
+      sample.Sample(
+          'Multi-stream %s net throughput' % operation,
+          net_thpt, 'B/sec', metadata=distribution_metadata),
+      sample.Sample(
+          'Multi-stream %s net throughput (experimental)' % operation,
```
stfeng: I believe the "experimental" throughput reported here would be most likely lower than the other non experimental one. but the key question is which one is more useful and accurate to the user?

I have an idea that we should try to validate further: when you do your end-to-end tests on this PR, turn on dstat on a separate window that covers the entire duration of your end-to-end test. configure dstat to report network in/out throughput every second, and make sure to redirect the dstat output to a file. Use a resonable object size (distribution) to begin with, such as 10MB or 100MB objects. The numbers from dstat reflects actual user experience in terms of throughput and can be used to validate which of these two ways of calculation is more accurate.

stfeng:
I've done my first pass. Please take a look.

perfkitbenchmarker/analyze.py
noahl: @stfeng I'll respond to the rest of your comments later, but I was thinking about this one. I agree that no one wants to look at a table of single-digit percentiles, but I want them for generating a graph of percentiles. I think the best solution is to add a command-line option saying which percentiles to record, defaulting to the current list that a person would actually want to look at.

I'm about to send you a PR which adds an option to the existing percentile calculator that lets the caller specify which percentiles to compute; that would allow the functionality that I want. If you disagree, we can talk about it more, but I wanted to get that sent out tonight because if you agree, it will be a very quick review.

perfkitbenchmarker/analyze.py
noahl: I checked, and Pandas has a function making it easy to check whether indices match. I have added `assert start_times.index.equals(....)`.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Sounds good. Changed.

perfkitbenchmarker/analyze.py
noahl: Changed. I don't remember why I wrote this function in the first place.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/analyze.py
noahl: I had them together originally, and split them because the combined function had four return values and did several different things, and I thought it was hard to understand what it did or how to use it. I think it's more confusing than it's worth, but if you really want, I will merge them.

stfeng: OK. keeping them separate SGTM.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I did it this way because I thought the second one was easier to understand. :) However, I can see both sides of this question. I'll change it.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Fixed, following the change to the metric above.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I was looking for a way to make this more clear, and I ended up adding a class called Interval, which lets me condense the start time and duration of an interval into a single variable. It doesn't shorten the code much, but I think it does make it more clear what's going on. I'm still on the fence about whether it was worth doing or not - please let me know what you think.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I am worried that the latency of a single-stream transfer would be lower before the other streams start up and start stealing bandwidth. That's why I did it this way. Is that not going to happen?

stfeng: OK. SGTM.

stfeng:
There are about 3 comments not answered yet, please address those and I will take a second pass afterwards.

I am also curious to know the results of bandwidth measurements from dstat vs this script.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: After looking at it again, I think it's hard to use a function returning a 4-tuple, just because there are so many return values with no inherent names. So I made it return a dictionary that gives quantities their proper names.

I also realized that if I could bundle units with the values, then the thing that generates samples could become a simple for loop that doesn't depend on the specific metrics used. I thought about returning Sample objects, but they would make unit testing more difficult because I would have to mock the time stamp, so I used pint Quantity objects for the values, which are perfect because they bundle a magnitude and its units together with no other information. The code is much simpler, and the naming issue is gone.

perfkitbenchmarker/analyze.py
noahl: Yes, in order for the results to make sense, the durations array must cover the time interval of length `interval_duration`. I renamed `total_overhead` to `total_gap_time` and added the assert.

However, I realized that there is a problem with the metrics that use `total_gap_time`: they assume that each stream is active for the same amount of time, but if restricting to an interval cuts off a big object from the front or back of a stream, that object's transfer time will be counted as gap time. I rewrote the function to fix this issue, and in the process also made it impossible to accidentally pass in the wrong interval duration (because you don't pass in the duration any more).

perfkitbenchmarker/analyze.py
noahl: I think the cleanest solution is to make the object storage benchmark a package, with two modules inside of it: the regular benchmark module, and this analysis module. I implemented that, but the change added a bunch of boring changes to this PR, just updating names. I can merge this as-is and immediately submit that as my next PR, or I could add the extra code to this PR. The new code wouldn't be complicated to review, there would just be more of it to look at. What do you think?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Responded by personal email with a dstat chart.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: For throughput percentiles, I think one reason they can be confusing is that per-object percentiles don't represent the actual throughput you will see, because a large and a small object count the same amount in the percentiles, but you will spend much more time uploading a large object than a small object. What do you think of weighting the per-object throughput by the number of seconds it took to upload that object? (I think this would be a separate PR.)

It would also be possible to use the code in `perfkitbenchmarker/traces/dstat.py` to run dstat on the worker VM and get an empirical view of the same data.

Either of these would eliminate the issue of object sizes, but they might make the throughput data less easy to understand. What do you think?

noahl:
@stfeng I just uploaded a commit with lots of new changes. I think I addressed all of your comments, except for the two I left questions on, about changing the name of `analysis.py` and recording throughput percentiles. In addition to the unit tests, I checked that I can actually run an `api_multistream` benchmark. Please take a look.

perfkitbenchmarker/analyze.py
stfeng: I think the package idea is a good idea but it's out of scope for this PR. We have an upcoming work item to refactor the object storage benchmark and as part of that work, we can consider this option. For now, all I am asking is to rename analyze.py to something more specific to object storage benchmark, e.g.: "object_storage_multistream_analysis.py", that's it. Then we can resolve this comment and move on from this PR.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
stfeng: Let's try to keep things simple for this version of work...

Latency: we report percentiles per-size, e.g., 1KB latency percentiles, 100MB latency percentiles., etc.

Multi-stream Throughput: we report aggregated throughput as computed by your analysis script. That's it.

stfeng:
Ack. I answered two of your questions inline above. I will take a pass on the new changes tomorrow. Thanks!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -1455,7 +1450,7 @@ def Run(benchmark_spec):
 
   results = OBJECT_STORAGE_BENCHMARK_DICTIONARY[FLAGS.storage].Run(vms[0],
                                                                    metadata)
-  print results
+  # print results
```
stfeng: We have always printed out the results for ease of debugging, did you remove this (by changing this to a comment) because there are too much information here? (or is it redundant)?

Or is this an accidental change?

noahl: Accidental change. I'll remove it from this PR.

perfkitbenchmarker/analyze.py
noahl: Okay, I'll rename the module in this PR and hopefully organize better later on.

noahl:
@stfeng I just pushed an update.

I couldn't replicate the failing test on my box with `tox -r`, so I'm hoping the problem will go away this time. If not, I'll investigate more.

noahl:
Test this please

stfeng:
travis-ci still failed (see above). @ehankland investigation pointed out that the new dependency on Pandas took a very long time to install.

Can we increase the timeout of the travis build to tolerate this?

stfeng:
test this please.

stfeng:
well, the bot does not listen to me. Trying again...

stfeng:
test this please

requirements-testing.txt
```diff
@@ -30,4 +30,4 @@ nose>=1.3
 flake8>=2.1.0
 psutil==3.0.0
 gcs-oauth2-boto-plugin
-azure
+azure-storage<=0.20.3
```
stfeng: This doesn't seem necessary to fix the current build break, if so, please consider reverting it to previous text.

noahl:
test this please

noahl:
I did some more research and discovered that travis caching probably can't affect our builds, because Tox recreates the build environment from scratch on each run anyway, so I removed it. I also noticed that the build now passes. I think we were being affected by a temporary availability problem in the pip server. Unfortunately, there's not much we can do about it short of hosting our own pip packages.

perfkitbenchmarker/__init__.py
```diff
@@ -26,6 +26,10 @@
 # very surprised to find out that 'KB' is not a valid unit.
 UNIT_REGISTRY.define('K- = 1000')
 
+# Storing percentages like other quantities simplifies metric
+# processing.
+UNIT_REGISTRY.define('percent = []')
```
stfeng: do you mean "percentile"?

perfkitbenchmarker/analyze.py
```diff
@@ -19,13 +19,50 @@
...
+  def __init__(self, start, duration=None, end=None):
+    self.start = start
+
+    if duration is not None:
+      if end is not None and start + duration != end:
+        raise ValueError(
+            'Invalid arguments to interval constructor: %s + %s != %s' %
+            (start, duration, end))
+      self.duration = duration
+    elif end is not None:
```
stfeng: Should you also validate end here? (e.g. "end<start")

perfkitbenchmarker/analyze.py
```diff
@@ -19,13 +19,50 @@
...
+    return (isinstance(other, Interval) and
+            self.start == other.start and
+            self.duration == other.duration)
+
+
+  def __repr__(self):
+    return 'Interval(%s, %s, %s)' % (self.start, self.duration, self.end)
+
+
+def AnyAndAllStreamsIntervals(start_times, durations, stream_ids):
```
stfeng: The name is cryptic. I start to understand it more when reading the comments.

How about simply: GetStreamActiveIntervals()? 

And then late the comments speak more for it.

perfkitbenchmarker/analyze.py
```diff
@@ -19,13 +19,50 @@
...
+  """Compute when all streams were active and when any streams were active.
```
stfeng: Consider this:

This function computes two intervals based on the input: 1) an interval that represents the duration when ALL streams are actively doing work, 2) an interval that represents the duration when ANY ONE of the streams are doing work.

perfkitbenchmarker/analyze.py
```diff
@@ -96,107 +135,95 @@ def FullyInInterval(start_times, durations, interval_start, interval_duration):
...
 
-  Returns: a tuple of
-    - The net throughput of all streams, excluding times they were inactive.
-    - The throughput of all streams, including times they were inactive.
-    - The total time that all streams were inactive.
-    - The total time all streams were inactive, as a proportion of the
-      total benchmark time.
+  Returns: a dictionary whose keys are metric names and whose values
+  are Quantity objects holding the metric values with appropriate
+  units.
```
stfeng: Expand this comment to include specifically what's returned - 

Returns: <your existing text> + "specifically, this includes:" 
 - blah

perfkitbenchmarker/analyze.py
```diff
@@ -96,107 +135,95 @@ def FullyInInterval(start_times, durations, interval_start, interval_duration):
...
 
+  bit = perfkitbenchmarker.UNIT_REGISTRY.bit
+  sec = perfkitbenchmarker.UNIT_REGISTRY.second
 
-def SummaryStats(series, name_prefix=''):
+  end_times = start_times + durations
 
-  """Compute some summary statistics for a series.
+  stream_starts = start_times.groupby(stream_ids).min()
+  stream_ends = end_times.groupby(stream_ids).max()
```
stfeng: assert these two arrays should have the same index (size).

perfkitbenchmarker/analyze.py
```diff
@@ -96,107 +135,95 @@ def FullyInInterval(start_times, durations, interval_start, interval_duration):
...
+  total_bytes_by_stream = sizes.groupby(stream_ids).sum()
+  active_time_by_stream = durations.groupby(stream_ids).sum()
+  overall_duration_by_stream = stream_ends - stream_starts
 
-  Returns: a pd.Series with summary statistics.
-  """
+  return {
+      'net throughput':
+      (total_bytes_by_stream / active_time_by_stream).sum() * 8 * bit / sec,
+      'net throughput (experimental)':
```
stfeng: As discussed, consider renaming this to "net throughput (with gap)":

perfkitbenchmarker/__init__.py
stfeng: Never mind, it's percent! :-)

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -245,10 +246,10 @@
 # benchmark. This is the filename.
 OBJECTS_WRITTEN_FILE = 'pkb-objects-written'
 
-# If the period of time when some streams, but not all, are active is
-# more than MULTISTREAM_STREAM_GAP_THRESHOLD percent of the time when
-# all streams are active, then log a warning and put metadata in the
-# samples.
+# If the gap between different stream starts and ends is above a
```
stfeng: [ignore] A note to myself: continue review from here.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -402,42 +401,38 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
-          thpt_with_overhead, 'B/sec', metadata=distribution_metadata),
-      sample.Sample(
-          'Multi-stream %s total overhead time' % operation,
-          total_overhead, 'sec', metadata=distribution_metadata),
-      sample.Sample(
-          'Multi-stream %s overhead time proportion' % operation,
-          overhead_prop * 100.0, 'percent', metadata=distribution_metadata)])
+               gap_stats['gap time proportion'].magnitude)
 
+  for name, value in itertools.chain(throughput_stats.iteritems(),
```
stfeng: As discussed in person, we should add a very simplified version of aggregated throughput computed in following way:

(Total bytes transferred by any streams in this run) / total time taken
(my understanding is the any_streams_active array could be used for this purpose)

We can name this as "net throughput (simplified)"

We will see how these 3 numbers diverge and under which cases.

requirements-testing.txt
```diff
@@ -31,3 +31,4 @@ flake8>=2.1.0
 psutil==3.0.0
 gcs-oauth2-boto-plugin
 azure
+
```
stfeng: extra newline here.

tests/analyze_test.py
```diff
@@ -16,13 +16,13 @@
 
 import unittest
 
-import mock
 import pandas as pd
 
 from perfkitbenchmarker import analyze
-from perfkitbenchmarker import sample
+import perfkitbenchmarker
```
stfeng: As usual, thanks for the great tests here!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -346,48 +356,99 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
   logging.info('Processing %s multi-stream %s results for the full '
-               'distribution.', len(records), operation)
+               'distribution.', len(records_in_interval), operation)
   _AppendPercentilesToResults(
       results,
-      (record['latency'] for record in records),
-      metric_name,
+      records_in_interval['latency'],
+      latency_prefix,
       'sec',
```
stfeng: Please use the existing constant LATENCY_UNIT to replace all references to the constant string 'sec' here and below.

stfeng:
I took a pass, mostly minor comments. 

Note that most of my comments are made inside [this particular commit] (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/905/files/a05cf4a3528a4209ae58b9a52b80b9f1d43f786b..ca537256388b88287dd1351101eacbc7bd50b74a#r60124643)

The overview page says I made comments to an outdated diff, but it's not really "outdated".

Let me know once you addresses these comments, I will take another look.

perfkitbenchmarker/analyze.py
noahl: Done.

perfkitbenchmarker/analyze.py
noahl: SGTM. Renamed.

perfkitbenchmarker/analyze.py
noahl: Added something very similar to your text. I left the title as-is because it's supposed to be one line, then added your text as a paragraph below.

perfkitbenchmarker/analyze.py
noahl: Added.

perfkitbenchmarker/analyze.py
noahl: Done.

perfkitbenchmarker/analyze.py
noahl: Done.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Added.

requirements-testing.txt
noahl: Removed.

tests/analyze_test.py
noahl: Thanks!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/__init__.py
noahl: Acknowledged :)

noahl:
@stfeng I pushed a commit that I think addresses all of your comments. I also rebased to fix merge conflicts. PTAL

noahl:
Wow, Skyler was too fast for me. I rebased again to fix another merge conflict.

stfeng:
LGTM. 

Please feel free to merge whenever you finish a final round of testing of this PR.

noahl:
@stfeng I was testing and discovered two small bugs, which I fixed. I have verified that it works on GCP and produces results that match the `api_data` scenario. Still okay to merge?

stfeng:
LGTM. Go ahead.

Issue #899, opened 2016-03-21
Unable to use run stages with OpenStack


kamsz:
While trying to run PKB with --run_uri, pickle is stuck in never ending loop and fails with:

```
maximum recursion depth exceeded while calling a Python object
```

After some digging, I've found out that the issue is with __getattribute__ method stuck in never ending loop in perfkitbenchmarker/providers/openstack/utils.py:
```
File "perfkitbenchmarker/providers/openstack/utils.py", line 80, in __getattribute__
return self.__client.__getattribute__(item)
```

voellm:
@kamsz - Can you please add the command you ran?

@ehankland - Eric can you take a look at the pickling?

kamsz:
@voellm The command used:
```
python pkb.py --cloud=OpenStack --image=ubuntu-cloud-14.04 --openstack_private_network=net04 --openstack_public_network=net04_ext --y│
csb_operation_count=100000 --scratch_disk_size=10 --machine_type=m1.xlarge --benchmarks=cassandra_ycsb --ycsb_operation_count=5000000 --ycsb_record_count=10000 --run_stage=provision
```

After provision succeed:

```
python pkb.py --cloud=OpenStack --image=ubuntu-cloud-14.04 --openstack_private_network=net04 --openstack_public_network=net04_ext --y│
csb_operation_count=100000 --scratch_disk_size=10 --machine_type=m1.xlarge --benchmarks=cassandra_ycsb --ycsb_operation_count=5000000 --ycsb_record_count=10000 --run_stage=prepare --run_uri=26ac809c
```

meteorfox:
@voellm @kamsz I'll try to replicate on my OpenStack Kilo deployment

meteorfox:
Confirmed. I'll try to debug it.

meteorfox:
@ehankland So the issue is that `NovaClient` defers the access to dynamically fetch the attribute from its client attribute. I tried removing the `__getattribute__` approach and instead have them use the direct reference to the client object but that client object also has some dynamic attribute definition which causes the pickler to loop indefinitely again. 

I'm not too familiar with the pickler code, but it seems that since `BenchmarkSpec` has references to the Network, Firewall, and VirtualMachine resources it tries to pickle those as well which then will try to pickle  the NovaClient and its attributes.

Here are some ways I think we could fix this:
 - Somehow have the Pickler not pickle the resource implementations (not sure if viable)
[...]

PR #909, opened 2016-03-23
Fix OpenStack pickling issue

ehankland:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/899

### DIFF
--- a/perfkitbenchmarker/providers/openstack/utils.py
+++ b/perfkitbenchmarker/providers/openstack/utils.py
@@ -16,6 +16,10 @@
 import logging
 import os
 
+from keystoneclient import session as ksc_session
+from keystoneclient.auth.identity import v2
+from novaclient import client as nova
+
 from perfkitbenchmarker import flags
 
 FLAGS = flags.FLAGS
@@ -27,7 +31,7 @@ class NovaClient(object):
         try:
             return super(NovaClient, self).__getattribute__(item)
         except AttributeError:
-            return self.__client.__getattribute__(item)
+            return self._client.__getattribute__(item)
 
     def GetPassword(self):
       # For compatibility with Nova CLI, use 'OS'-prefixed environment value
@@ -51,10 +55,6 @@ class NovaClient(object):
       raise Exception(error_msg)
 
     def __init__(self):
-        from keystoneclient import session as ksc_session
-        from keystoneclient.auth.identity import v2
-        from novaclient import client as nova
-
         self.url = FLAGS.openstack_auth_url
         self.user = FLAGS.openstack_username
         self.tenant = FLAGS.openstack_tenant
@@ -62,15 +62,32 @@ class NovaClient(object):
         self.http_log_debug = FLAGS.log_level == 'debug'
         self.password = self.GetPassword()
 
-        self.__auth = v2.Password(auth_url=self.url,
-                                  username=self.user,
-                                  password=self.password,
-                                  tenant_name=self.tenant)
-        self._session = ksc_session.Session(auth=self.__auth)
-        self.__client = nova.Client(version='2', session=self._session,
-                                    http_log_debug=self.http_log_debug)
+        self._auth = v2.Password(auth_url=self.url,
+                                 username=self.user,
+                                 password=self.password,
+                                 tenant_name=self.tenant)
+        self._session = ksc_session.Session(auth=self._auth)
+        self._client = nova.Client(version='2', session=self._session,
+                                   http_log_debug=self.http_log_debug)
         # Set requests library logging level to WARNING
         # so it doesn't spam logs with unhelpful messages,
         # such as 'Starting new HTTP Connection'.
         rq_logger = logging.getLogger('requests')
         rq_logger.setLevel(logging.WARNING)
+
+    def __getstate__(self):
+      d = self.__dict__.copy()
+      del d['_client']
+      del d['_session']
+      del d['_auth']
+      return d
+
+    def __setstate__(self, state):
+      self.__dict__ = state
+      self._auth = v2.Password(auth_url=self.url,
+                               username=self.user,
+                               password=self.password,
+                               tenant_name=self.tenant)
+      self._session = ksc_session.Session(auth=self._auth)
+      self._client = nova.Client(version='2', session=self._session,
+                                 http_log_debug=self.http_log_debug)
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -23,6 +23,7 @@
 # or install command-line tools, and are not needed for unit tests.
 -rrequirements-alicloud.txt
 -rrequirements-cloudstack.txt
+-rrequirements-openstack.txt
 
 # Test requirements
 mock>=1.0.1


meteorfox:
I tried your branch with the followings commands:

```
python pkb.py --cloud=OpenStack --machine_type=m1.small --image='Ubuntu 14.04' \
         --openstack_network='private' --openstack_floating_ip_pool='public' \
         --benchmarks=ping --run_stage=provision
```

```
python pkb.py --cloud=OpenStack --machine_type=m1.small --image='Ubuntu 14.04' \
         --openstack_network='private' --openstack_floating_ip_pool='public' \
         --benchmarks=ping --run_stage=prepare --run_uri=58d202c4
```

But still the same error:

```
2016-03-23 19:16:36,690 58d202c4 MainThread pkb.py:471 INFO     PerfKitBenchmarker version: v1.3.0-46-g5d14ada
2016-03-23 19:16:36,690 58d202c4 MainThread pkb.py:426 INFO     Flag values:
--machine_type=m1.small
--openstack_network=private
--image=Ubuntu 14.04
--cloud=OpenStack
--run_stage=prepare
--run_uri=58d202c4
--benchmarks=ping
--openstack_floating_ip_pool=public
2016-03-23 19:16:36,690 58d202c4 MainThread publisher.py:557 DEBUG    Using publishers: [<LogPublisher logger=<logging.RootLogger object at 0x7f1ffe7cb9d0> level=20>, <PrettyPrintStreamPublisher stream=<open file '<stdout>', mode 'w' at 0x7f2000f84150>>, <NewlineDelimitedJSONPublisher file_path="/tmp/perfkitbenchmarker/run_58d202c4/perfkitbenchmarker_results.json" mode="wb">]
2016-03-23 19:16:36,855 58d202c4 MainThread ping(1/1) benchmark_spec.py:337 ERROR    Unable to unpickle spec file for benchmark ping0.
```

Also, fyi, when I tried the `__getstate_` and `__setstate__`, I overrode them for each resource `OpenStackVirtualMachine`, `OpenStackFirewall`, `OpenStackDisk` etc. I did not try overriding them on just `NovaClient` class.

ehankland:
@meteorfox Thanks for testing - did the exception get printed out at the end or did you kill it before that happened? When I was testing it in the python shell, pickling without my change did produce the infinite \_\_getattribute\_\_ recursion, but the change made that go away - so maybe there's something else going on - it would be useful to know what the exact error was. I can take a look at the VM/Firewall/Disk classes to see if there's anything obvious.

meteorfox:
Here's an excerpt of the traceback. The bottom part.

```
  File "/opt/PerfKitBenchmarker/pkb-pip/local/lib/python2.7/site-packages/novaclient/openstack/common/apiclient/base.py", line 490, in __getattr__
    if not self.is_loaded():
  File "/opt/PerfKitBenchmarker/pkb-pip/local/lib/python2.7/site-packages/novaclient/openstack/common/apiclient/base.py", line 527, in is_loaded
    return self._loaded
RuntimeError: maximum recursion depth exceeded
```

meteorfox:
That's the same traceback I was getting after I tried overriding `__{get,set}state__` methods. I'm thinking it could be related to the imports. I'm guessing each method gets pickled as well but since we have the imports for novaclient module in some of them then it loops indefinitely. I've seen the python-novaclient library does some dynamic definition using __init__.py when being the modules are being imported. That's my best guess anyway.

meteorfox:
Side note: does your team have a some kind of integration testing environment for testing PKB against GCP, AWS and Azure? I could help you setup a simple OpenStack All-in-One deployment, it can run on a VM if your machines support nested VMs. I know KVM does, but I'm not sure if Google enables it. :)

ehankland:
@meteorfox Ah - I think I see the issue - I thought the only problem was that we couldn't pickle the novaclient client objects, but those lines are related to novaclient resource objects. It looks like novaclient resource objects get saved as attributes of PKB objects in a few places (example here: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_disk.py#L41) which causes them to get pickled too. The problem is that the \_\_getattr\_\_ method (defined here https://github.com/openstack/python-novaclient/blob/2.35.0/novaclient/openstack/common/apiclient/base.py#L487) assumes that the objects \_\_dict\_\_ already has the "\_loaded" key which isn't the case when unpickling.

meteorfox:
Yes, that's true. That is most likely the problem, we are definitely storing a few of those novaclient resource objects around.

I think these are all of them, but I might have missed one.
- https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_virtual_machine.py#L77
- https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_virtual_machine.py#L79 
- https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_virtual_machine.py#L150
- https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_virtual_machine.py#L229
- https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_network.py#L49
-  https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/openstack/os_disk.py#L41

ehankland:
Working around all of those novaclient resources is probably at least as much work as switching to the CLI. That's seeming like the best option.

meteorfox:
Sounds good! It wasn't so bad.

It's really almost there. I've already migrated the network, ssh key pieces. Currently, the branch has a mix of the CLI and the library approach, and everything is still working.

All that is left is the VM piece which is really 3 commands boot,show,delete, and similarly the disk piece. I think probably by tomorrow I should've it done.

kimcie:
@meteorfox Have you already finished working on opencstack cli approach?

meteorfox:
@kimcie almost. For the past week and a half I was in crunch mode at work and didn't have time to work on this issue.

Anyway, today I've been working on it, I should have something up in a few hours.

kimcie:
Great, it will help to investigate why some of benchmarks doesn't work. I just want make sure, if all your changes connected with this issue are on branch openstack-cli-provider? I have just wanted to test it, at it fails at assocating floating ip.

meteorfox:
@kimcie Yeah, I'm working on the branch right now, I don't recommend trying it for now since it's not ready yet. I just finished with the floating IP part and I'm now working on the Cinder volumes code. I'll also need to test it against my OpenStack lab and update the `REAMDE.md` once it's all done.

meteorfox:
@kimcie Check out PR #942 it should be ready now. It's still under review but I would like to know if it works for you as well. Thanks!

kimcie:
It works for me as well.

meteorfox:
@kimcie nice, thanks for testing it!. @ehankland @noahl can you PTAL? Thanks!

noahl:
@meteorfox just to clarify, you want us to look at #942, right?

meteorfox:
@noahl  yes, that's correct. thanks

PR #910, opened 2016-03-25
Add percentiles option to PercentilesCalculator

noahl:
Let the user specify which percentiles to compute. Also adds a test for
the PercentilesCalculator and fixes a bug revealed by the test.

### DIFF
--- a/perfkitbenchmarker/sample.py
+++ b/perfkitbenchmarker/sample.py
@@ -25,7 +25,8 @@ def PercentileCalculator(numbers, percentiles=PERCENTILES_LIST):
 
   Args:
     numbers: The set of numbers to compute percentiles for.
-    percentiles: If given, a list of percentiles to compute.
+    percentiles: If given, a list of percentiles to compute. Can be
+      floats, ints or longs.
 
   Returns:
     A dictionary of percentiles.
@@ -35,10 +36,12 @@ def PercentileCalculator(numbers, percentiles=PERCENTILES_LIST):
   total = sum(numbers_sorted)
   result = {}
   for percentile in percentiles:
+    if percentile < 0.0 or percentile > 100.0:
+      raise ValueError('Invalid percentile %s' % percentile)
+
     percentile_string = 'p%s' % str(percentile)
     index = int(count * float(percentile) / 100.0)
-    if index == count:
-      index = count - 1
+    index = min(index, count - 1)  # Correction to handle 100th percentile.
     result[percentile_string] = numbers_sorted[index]
 
   if count > 0:
--- a/perfkitbenchmarker/sample.py
+++ b/perfkitbenchmarker/sample.py
@@ -20,11 +20,12 @@ PERCENTILES_LIST = [0.1, 1, 5, 10, 50, 90, 95, 99, 99.9]
 _SAMPLE_FIELDS = 'metric', 'value', 'unit', 'metadata', 'timestamp'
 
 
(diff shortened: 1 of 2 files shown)


noahl:
@stfeng This PR adds support for a user-specified list of percentiles to `PercentilesCalculator`, adds support for p100, and adds a test case. PTAL.

perfkitbenchmarker/sample.py
```diff
@@ -33,10 +34,12 @@ def PercentileCalculator(numbers):
   count = len(numbers_sorted)
   total = sum(numbers_sorted)
   result = {}
-  for percentile in PERCENTILES_LIST:
+  for percentile in percentiles:
```
yuyantingzero: Can you use numpy.percentile?
Any reason to write your own percentile calculator?

stfeng: thanks for the information about numpy.percentile, we will take a look at that. This percentile calculator was added long time ago, here in this PR we are just adding an option to allow it to output more/finer-grain percentiles.

The potential work to replace this percentile calculator with  numpy.percentile can be tracked separately. We need to make sure it can do what we want do here (the advantage of having our own percentile calculator is we can add any new feature/functionalities to it as necessary).

perfkitbenchmarker/sample.py
```diff
@@ -33,10 +34,12 @@ def PercentileCalculator(numbers):
...
     percentile_string = 'p%s' % str(percentile)
-    result[percentile_string] = numbers_sorted[
-        int(count * float(percentile) / 100)]
+    index = int(count * float(percentile) / 100.0)
+    if index == count:
```
stfeng: If the "percentiles" passed in is something like "101.0", then it would still overflow. (the original code works because it works on a constant list of percentiles up to 99.9), now it takes in an input list from the user, we might want to do some validation?
also, you can probably simplify the if check as one liner: index = min(index, count - 1)

noahl:
@stfeng I added input validation, simplified the code as you suggested, and added a bit more documentation. PTAL.

noahl:
@yuyantingzero please take a look.

perfkitbenchmarker/sample.py
```diff
@@ -33,10 +35,14 @@ def PercentileCalculator(numbers):
...
-  for percentile in PERCENTILES_LIST:
+  for percentile in percentiles:
+    if percentile < 0.0 or percentile > 100.0:
+      raise ValueError('Invalid percentile %s' % percentile)
+
     percentile_string = 'p%s' % str(percentile)
-    result[percentile_string] = numbers_sorted[
-        int(count * float(percentile) / 100)]
+    index = int(count * float(percentile) / 100.0)
+    index = min(index, count - 1)  # Correction to handle 100th percentile.
```
stfeng: if count is zero, this code won't work either....

stfeng: but we can probably require the caller to supply an non-empty array.

Or return an empty result if count is zero.

noahl: @stfeng Good point. I think we should raise an exception if the array is empty, because I think the caller should be able to assume that if they get a result from PercentileCalculator, then it contains the percentiles they requested.

I just pushed a commit that raises the exception, and also adds tests for this case and the case where the user requests a bad percentile. PTAL.

stfeng:
LGTM

PR #918, opened 2016-03-29
Fix bug aerospike 3.0.84 fails to compile on ppc64

mlpesant:
Updated pointer to version 4.0.4, which successfully compiles on ppc64 and corrected indentation of comment on lines 99 and 100 to pass lint test

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ PerfKit Benchmarker users must manually download SpecCPU2006 tarball from their 
 the `perfkitbenchmarker/data` folder (e.g. `~/PerfKitBenchmarker/perfkitbenchmarker/data/cpu2006v1.2.tgz`)
 
 Installing PerfKitExplorer and Prerequisites
-========================
+============================================
 Before you can run the PerfKit Benchmaker, you need account(s) on the Cloud provider(s) you want to benchmark:
 
 * Get a GCE account to run tests on GCE. Our site is https://cloud.google.com
@@ -117,14 +117,13 @@ The path to these commands should be
 
 [Download PerfKitBenchmarker](http://github.com/GoogleCloudPlatform/PerfKitBenchmarker/releases) from GitHub.
 
-## Install PerfKit Benchmakrer dependencies
+## Install PerfKit Benchmarker dependencies
 ```
 $ cd /path/to/PerfKitBenchmarker
 $ sudo pip install -r requirements.txt
 ```
 
-Cloud account setup
-===================
+##Cloud account setup
 
 This section describes the setup steps needed for each cloud system.  
 * [Google Cloud](#install-gcloud-and-setup-authentication)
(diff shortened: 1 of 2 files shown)


mlpesant:
@voellm, please help me fix the cla again...  If you can think of a way to solve this issue permanently, please let me know.  Thanks!

README.md
```diff
@@ -113,7 +113,32 @@ The path to these commands should be
 
 `C:\\Users\\\<user\>\\AppData\\Local\\GitHub\\PortableGit\_\<guid\>\\bin`
 
-## Install `gcloud` and setup authentication
+## Install PerfKit
```
voellm: Can you mark this as optional?  Its not required.

voellm:
Lorena - It looks like this issue is in the git commits (vs gitHub).  The "fix" is to follow this thread to update the user email in the commit. 

http://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

However given this change and the fact I've confirmed the CLA I'll complete it as it.  Just going to run a test or tow first.

voellm:
In a follow-up pull request I'll also update the Aerospike used by YCSB runs here - https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_packages/aerospike_server.py#L28

mlpesant:
Thanks Tony!  I changed my credentials on the system I'm sending the commits from.  Hopefully that permanently fixes the problem.

voellm:
Thanks.  I've been trying to validate the commit but am hitting an error in
the patch command.

voellm:
Looks like the aerospike.patch file does not work with this version of aerospike and is causing the benchmark to fail.  It just needs to be updated to add the code.

Would you like to update the patch file?  Its part of the data/ directory here https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/data/aerospike.patch

Just to make sure... you are trying to update the version on the aerospike benchmark vs aerospike_ycsb correct?

ehankland:
The following patch file seemed to work for me with this client version:
```
diff -rupN main/latency.c mainnew/latency.c
--- main/latency.c      2014-10-09 05:51:43.070816173 +0000
+++ mainnew/latency.c   2014-10-09 06:33:18.084487933 +0000
@@ -31,6 +31,8 @@ latency_init(latency* l, int columns, in
        l->last_bucket = columns - 1;
        l->bit_shift = shift;
        l->buckets = cf_calloc(columns, sizeof(uint32_t));
+       l->total_latency = 0;
+       l->total_samples = 0;
 }
 
 void
@@ -58,6 +60,8 @@ latency_getindex(latency* l, uint64_t el
 void
 latency_add(latency* l, uint64_t elapsed_ms)
 {
+       ck_pr_add_64(&l->total_latency, elapsed_ms);
+       ck_pr_inc_64(&l->total_samples);
        int index = latency_getindex(l, elapsed_ms);
        ck_pr_inc_32(&l->buckets[index]);
 }
@@ -137,5 +141,9 @@ latency_print_results(latency* l, const
                limit <<= shift;
                p += latency_print_column(l, limit, sumDouble, array[i], p);
        }
+
+       if (l->total_samples != 0)
+               p += sprintf(p, " Overall Average Latency (ms) %f", l->total_latency / (float) l->total_samples);
+
        *p = 0;
 }
diff -rupN main/latency.h mainnew/latency.h
--- main/latency.h      2014-10-09 05:51:12.421313546 +0000
+++ mainnew/latency.h   2014-10-09 06:33:18.084487933 +0000
@@ -27,6 +27,8 @@ typedef struct latency_t {
        uint32_t* buckets;
        int last_bucket;
        int bit_shift;
+       uint64_t total_latency;
+       uint64_t total_samples;
 } latency;
 
 void latency_init(latency* l, int columns, int shift);
```

s-deitz:

Issue #915, opened 2016-03-28
BaseOsMixin.PullFile doc confusing


skschneider:
> The argument documentation for BaseOsMixin.PullFile is confusing.
> 
>As far as I can tell
>
>source_path is a local path for the target of the copy 
>remote_path is a remote path for the source of the copy.
>
>I don't quite know what the text of the comments is saying.

PR #923, opened 2016-03-30
Rename variables and improve documentation of BaseOsMixin.PullFile.

skschneider:
Also make remote_path a required parameter. The default behavior is not
currently used, and the expected behavior is confusing.

Addresses #915.


yuyantingzero:
LGTM

Issue #916, opened 2016-03-28
Pressing Ctrl+C doesn't stop resource provisioning or wait for it to complete


skschneider:
For example, during the prepare stage, RunThreaded is used to launch parallel threads that create networks or VMs. If Ctrl+C is pressed, the main thread that is executing RunThreaded will jump out and get caught by the bare except (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/pkb.py#L403), but the background threads will continue to provision resources while the main thread goes on to try to clean them up at the same time.

skschneider:
Fixed by #926

PR #926, opened 2016-04-04
Add support for interrupting child threads created by RunThreaded.

skschneider:
Addresses #916.

Pressing CTRL+C after resource-provisioning threads have been created
will now cause a KeyboardInterrupt to be raised in each child thread.

- Move RunThreaded, RunParallelThreads, and RunParallelProcesses out of
  vm_util.py into their own file.
- Combine the bodies of RunParallelThreads and RunParallelProcesses
  into a single _RunParallelTasks function that uses separate task
  managers for each.
- Raise KeyboardInterrupt explicitly in each child thread.

### DIFF
--- a/perfkitbenchmarker/background_tasks.py
+++ b/perfkitbenchmarker/background_tasks.py
@@ -1,0 +1,636 @@
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Background tasks that propagate PKB thread context.
+
+TODO(skschneider): Many of the threading module flaws have been corrected in
+Python 3. When PKB switches to Python 3, this module can be simplified.
+
+PKB tries its best to clean up provisioned resources upon SIGINT. By default,
+Python raises a KeyboardInterrupt upon a SIGINT, but none of the built-in
+threading module classes are designed to handle a KeyboardInterrupt very well:
+
+- threading.Lock has an atomic acquire method that cannot be interrupted and
+  hangs forever if the same thread tries to acquire twice. Its release method
+  can be called by any thread but raises thread.error if an unacquired Lock is
+  released.
+
+- More complicated classes (threading.RLock, threading.Event, threading.Thread,
+  Queue.Queue) use internal Locks in such a way that a KeyboardInterrupt can
+  cause a thread that has acquired a Lock to jump out of its current action
+  without releasing the Lock. For example, in the below code, a
+  KeyboardInterrupt can be raised immediately after the acquire call but before
+  entering the try block:
+    lock.acquire()
+    try:
+      ...
+    except:
+      lock.release()
+
+Taken together, this means that there is a possibility to leave an internal Lock
+acquired, and when later cleanup steps on the same or different thread attempt
+to acquire the Lock, they will hang forever, unresponsive to even a second
+KeyboardInterrupt. A KeyboardInterrupt during Thread.start() or Thread.join()
+can even trigger an unbalanced acquire on a global lock used to keep track of
+active threads, so that later attempts to start or join any Thread will hang
+forever.
+
+While it would take a significant and impractical redesign of PKB's code to
+completely eliminate any risk of deadlock following a KeyboardInterrupt, the
+code in this module is designed to allow interrupting parallel tasks while
+keeping the risk of deadlock low.
+"""
+
+import abc
+from collections import deque
+from concurrent import futures
+import ctypes
+import functools
+import logging
+import os
+import Queue
+import threading
+import time
+import traceback
+
+from perfkitbenchmarker import context
+from perfkitbenchmarker import errors
+from perfkitbenchmarker import log_util
+
+
+# For situations where an interruptable wait is necessary, a loop of waits with
+# long timeouts is used instead. This is because some of Python's built-in wait
+# methods are non-interruptable without a timeout.
+_LONG_TIMEOUT = 1000.
+
+# Constants used for polling waits. See _WaitForCondition.
+_WAIT_MIN_RECHECK_DELAY = 0.001  # 1 ms
+_WAIT_MAX_RECHECK_DELAY = 0.050  # 50 ms
+
+# Values sent to child threads that have special meanings.
+_THREAD_STOP_PROCESSING = 0
+_THREAD_WAIT_FOR_KEYBOARD_INTERRUPT = 1
+
+
+def _GetCallString(target_arg_tuple):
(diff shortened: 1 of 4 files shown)


skschneider:
@ehankland PTAL

perfkitbenchmarker/background_tasks.py
```diff
@@ -0,0 +1,578 @@
...
+# long timeouts is used instead. This is because some of Python's built-in wait
+# methods are non-interruptable without a timeout.
+_LONG_TIMEOUT = 1000.
+
+# Constants used for polling waits. See _WaitForCondition.
+_WAIT_MIN_RECHECK_DELAY = 0.001  # 1 ms
+_WAIT_MAX_RECHECK_DELAY = 0.050  # 50 ms
+
+
+def _GetCallString(target_arg_tuple):
```
ehankland: Looks like you haven't deleted the copy of this function in vm_util - I couldn't find any references to the old function - do you still need it or can it be deleted?

skschneider: Deleted.

perfkitbenchmarker/background_tasks.py
```diff
@@ -0,0 +1,633 @@
...
+
+  def __init__(self):
+    self._deque = deque()
+    self._lock = threading.Lock()
+    self._lock.acquire()
+
+  def _WaitForItem(self):
+    self._lock.acquire()
+
+  def _SignalAvailableItem(self):
```
ehankland: Seems like [condition variables](https://docs.python.org/2/library/threading.html#condition-objects) might be appropriate here.

perfkitbenchmarker/vm_util.py
```diff
@@ -194,243 +188,9 @@ def GetSshOptions(ssh_key_filename):
...
-        msg = ('Exception occurred while calling {0}:{1}{2}'.format(
-            _GetCallString(target_arg_tuples[index]), os.linesep, stacktrace))
-        logging.error(msg)
-        error_strings.append(msg)
-  if error_strings:
-    msg = ('The following exceptions occurred during parallel execution:'
-           '{0}{1}'.format(os.linesep, os.linesep.join(error_strings)))
-    raise errors.VmUtil.CalledProcessException(msg)
-  return results
+RunParallelProcesses = background_tasks.RunParallelProcesses
```
ehankland: Doesn't look like there are any references to vm_util.RunParallelProcesses (or background_tasks.RunParallelProcesses for that matter). Who/what do we expect to use this?

perfkitbenchmarker/vm_util.py
```diff
@@ -194,243 +188,9 @@ def GetSshOptions(ssh_key_filename):
...
+RunParallelThreads = background_tasks.RunParallelThreads
```
ehankland: Looks like there are only 3 references to vm_util.RunParallelThreads (and those are all in benchmark_spec.py). Maybe just make those use background_tasks.RunParallelThreads instead and get rid of this line.

perfkitbenchmarker/vm_util.py
```diff
@@ -194,243 +188,9 @@ def GetSshOptions(ssh_key_filename):
...
+RunThreaded = background_tasks.RunThreaded
```
ehankland: Is the expectation that we are eventually going to change all RunThreaded calls to point to background_tasks.RunThreaded or are we keeping this indefinitely? Maybe add a TODO if the former.

perfkitbenchmarker/background_tasks.py
skschneider: Unfortunately not. They're not resilient to KeyboardInterrupts. Internally, they use two threading.Lock objects. One of them is the "waiter" Lock used to wait for the notification, similar to how this _lock is used to wait for items. But in order to start waiting or to send the notification, the current thread needs ownership of the other Lock (let's call it the "condition" Lock). That's where things can get tricky. Because the KeyboardInterrupt can be raised after the condition Lock is acquired, but before it's released, things can end up in a state where that condition Lock is forever tied up.

perfkitbenchmarker/vm_util.py
skschneider: Added TODO. Will update references and remove vm_util.RunParallelThreads in a separate PR.

perfkitbenchmarker/vm_util.py
skschneider: This can be a separate PR to redirect references, but we should keep the vm_util version at least until the v1.5.0 release.

perfkitbenchmarker/background_tasks.py
skschneider: Here's something you can try for yourself. Write this to a file, run it, and hit CTRL+C. Repeat a few times and see if any bad behaviors show up. Then try it using an RLock instead of a Lock.
```
import itertools
import threading

c = threading.Condition(threading.Lock())
try:
  # Infinite loop of just acquiring and releasing a lock.
  for _ in itertools.repeat([None]):
    with c:
      pass
except KeyboardInterrupt:
  print 'KeyboardInterrupt. If cleaned up properly, expect c to have been released.'
  try:
    # Release again to confirm. If error, c was already released. If no error,
    # c was not released the first time.
    c.release()
    print 'But c had not been released.'
  except threading.ThreadError:
    print 'And c had indeed been released.'
```

ehankland: Thanks for the script - I saw the bad behavior - that is definitely weird. Was there a problem using a condition variable with a normal (non-reentrant) lock?

skschneider: Yes, the Condition class has this problem whether it uses a Lock or RLock. To simplify the with statement, it appears to be something like
```
__enter__()
try:
  ...
finally:
  __exit__(...)
```
And the gap between enter and try can still be interrupted.

perfkitbenchmarker/background_tasks.py
```diff
@@ -0,0 +1,633 @@
...
+    self._task_queues[worker_id].Put((task_id, task))
+
+  def AwaitAnyTask(self):
+    worker_id, task_id = self._response_queue.Get()
+    self._available_worker_ids.append(worker_id)
+    return task_id
+
+  def HandleKeyboardInterrupt(self):
+    # Raise a KeyboardInterrupt in each child thread.
+    for thread in self._threads:
```
ehankland: Do we want all threads to log the keyboard interrupt? Depending on the number of threads this can be a lot of output

skschneider: It could be helpful to know where they were interrupted. But at the same time, threads that are not actively processing a task will likely be interrupted while within _ExecuteBackgroundThreadTasks, so that could be a lot of useless information. The only way that I can think of to differentiate helpful from unhelpful cases is to inspect the stack of the exception, which I would want to be a separate thread. How about if I switch this to debug level for now (but still include the exc_info) and make a separate PR to reduce output?

ehankland: SGTM

ehankland:
LGTM

perfkitbenchmarker/background_tasks.py
skschneider: Oops, did I say "thread"? I meant PR. Anyway, I'll go ahead with the log level change.

skschneider: Changed to debug and added a TODO.

skschneider:
GitHub has been a bit weird and unresponsive today. Merged manually (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/commit/00d68725aa94edffabe4632c9547463bb6ddef7b), so closing this PR.

PR #932, opened 2016-04-06
Fix Cloud Bigtable YCSB benchmark

cmccoy:
* Use `netty-tcnative-boringssl` in place of ALPN boot. This should give
  better performance and is simpler to configure.
* Bump client to 0.2.4.

@hildrum, could you PTAL?


hildrum:
LGTM.  I was able to run the benchmark.

Issue #900, opened 2016-03-21
tox breaks when the environment variable PYTHONPATH is not set (e.g. virtualenv)


meteorfox:
The recent `scripttests` command added to tox breaks when `PYTHONPATH`env variable is not set. Usually, by default in virtualenvs `PYTHONPATH` is not set, this causes the `setenv` command to raise an error.

```
tox.ConfigError: ConfigError: substitution env:'PYTHONPATH': unknown environment variable 'PYTHONPATH'  or recursive definition.
```

meteorfox:
Easy fix:

```diff
diff --git a/tox.ini b/tox.ini
index 9c63c5b..e477a0d 100644
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ deps =
 [testenv:scripttests]
 commands = nosetests {toxinidir}/script-tests []
 setenv = 
-  PYTHONPATH = {toxinidir}/perfkitbenchmarker/scripts:{env:PYTHONPATH}
+  PYTHONPATH = {toxinidir}/perfkitbenchmarker/scripts:{envdir}
 
 [testenv:flake8]
 skip_install = True
```

meteorfox:
@noahl PTAL, this seems to be affecting @ivonindza as well. I have been running with this change locally otherwise tox wouldn't work for me.

noahl:
Closing this because #941 fixes it for me. Please reopen if you see more problems.

PR #934, opened 2016-04-06
Cloudsuite: in-memory analytics benchmark

ivonindza:


ivonindza:
I signed it!

perfkitbenchmarker/benchmark_sets.py
```diff
@@ -138,6 +138,10 @@
     'mit_set': {
         MESSAGE: 'Massachusetts Institute of Technology benchmark set.',
         BENCHMARK_LIST: [STANDARD_SET, 'silo']
+    },
+    'cloudsuite_set': {
+        MESSAGE: 'CloudSuite benchmark set.',
+        BENCHMARK_LIST: ['cloudsuite_inmemory_analytics']
```
skschneider: Any plans to add cloudsuite_web_serving or cloudsuite_websearch?

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
```
skschneider: 2016?

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+#
```
skschneider: I suggest making the file name cloudsuite_in_memory_analytics_benchmark.py to be consistent about the 'in-memory-analytics' name. It looks like we didn't do that with 'web-search' -> 'websearch', but we did do it with 'web-serving' -> 'web_serving'.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+More info: http://cloudsuite.ch/inmemoryanalytics/
+"""
+
+import re
+
+from perfkitbenchmarker import configs
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import sample
+
+flags.DEFINE_string('ima_dataset',
```
skschneider: These are global flag names. Prefix each one with `cloudsuite_inmemory_analytics_` (or `cloudsuite_in_memory_analytics_` if the benchmark/file name changes).

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+cloudsuite_inmemory_analytics:
+  description: >
+    Run Cloudsuite in-memory analytics benchmark. Specify the number of worker
+    VMs with --num_vms.
+  vm_groups:
+    master:
+      vm_spec: *default_single_core
+    workers:
+      vm_spec: *default_single_core
+      vm_count: 1
```
skschneider: Is this supposed to be for the master group? Or both? It looks like the behavior here is that the default number of workers is 1, but it can be overridden by --num_vms below.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def GetConfig(user_config):
+  config = configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+  if FLAGS['num_vms'].present:
+    config['vm_groups']['workers']['vm_count'] = FLAGS.num_vms
+  return config
+
+
+def _HasDocker(vm):
```
skschneider: Since we have a docker package that can be used by multiple benchmarks, please move this into that package (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_packages/docker.py). It will be more helpful there.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+  resp, _ = vm.RemoteCommand('command -v docker',
+                             ignore_failure=True,
+                             suppress_warning=True)
+  if resp.rstrip() == "":
+    return False
+  else:
+    return True
```
skschneider: Looks like this can be simplified to `return bool(resp.rstrip())`

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def Prepare(benchmark_spec):
```
skschneider: Can you make a function to prepare one VM (alternately two functions - one for master, one for worker) and then run it in parallel with either RunThreaded or RunParallelThreads? I'm not sure if there's any ordering required between master_cmd and worker_cmd, but it looks like the commands above could be parallel, and at least the worker_cmd for all workers could be parallel.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+  vms = benchmark_spec.vms
+  master = benchmark_spec.vm_groups['master'][0]
+  workers = benchmark_spec.vm_groups['workers']
+
+  for vm in workers:
+    vm.RemoteCommand('sudo docker stop spark-worker')
+    vm.RemoteCommand('sudo docker rm spark-worker')
```
skschneider: Could the above commands be done in parallel?

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+
+  master.RemoteCommand('sudo docker stop spark-master')
+  master.RemoteCommand('sudo docker rm spark-master')
+
```
skschneider: Could the below commands be done in parallel?

skschneider:
I think our README.md is getting out of date. It doesn't have information about all the benchmarks. Could you add information about this one?

perfkitbenchmarker/benchmark_sets.py
ivonindza: There should be 8 benchmarks eventually: Data Analytics, Data Caching, Data Serving, Graph Analytics, In-memory Analytics, Media Streaming, Web Search, Web Serving

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
ivonindza: There is always one master (for the client and the coordinator), and there can be multiple workers (default 1) that do the actual computation. --num_vms specifies the number of worker VMs as stated in the description, there will be an additional master VM. Is this confusing?

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
ivonindza: Yes it can be parallelized per VM. Pulling different docker images within the same VM can also be parallelized, but I'm not sure how to do it in the framework. Maybe just RemoteCommand('docker pull x & docker pull y &')?

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
skschneider: Makes sense, just confirming that's what you wanted since it looks a bit weird. The master group, which should always have 1 VM, has the vm_count omitted (which functionally is the same as explicitly setting it to 1), while the worker group, which can have one or more VMs, has the vm_count explicitly set to 1 (but can be overridden elsewhere). Functionally, what you have here is fine, but it *looks* like the master group size is nebulous while the workers group size is recommended/fixed.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
skschneider: The problem with `&` is that you need to later check to make sure the pull finished, and it becomes difficult to balance the complexity of checking error status and the latency of each RemoteCommand/ssh session. I recommend either one RemoteCommand that executes both pulls back to back (`&&`) or just making more threads with RunThreaded (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/6b6eccfe318952fc75dd7f839024df817321ad0d/perfkitbenchmarker/background_tasks.py#L560) or RunParallelThreads. The current serial approach is also fine as-is. Improvements can always be added as separate pull requests later.

ivonindza: You can do wait at the end: 'docker pull x & docker pull y & wait'. Problem with multiple threads are dependencies: you can pull images in parallel, but need to have data image before starting data container and need both images and started data container before starting worker container. This is all on a single worker VM.

skschneider: `wait` doesn't propagate exit code, which is why I'm more in favor of
```
RemoteCommand('docker pull x')
RemoteCommand('docker pull y')
```
or
```
RemoteCommand('docker pull x && docker pull y')
```

skschneider: Looks like because of the dependencies, master and workers can't just spin off and do their entire initialization. But maybe everything above `master.RemoteCommand(master_cmd)` could have one thread per VM, and everything below could also have one thread per VM?

ivonindza: Yes, one thread per VM is not a problem. I was thinking about multiple threads per VM, because parts of a single VM's setup can be parallelized. But then you need the ability to wait on a thread's completion.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
ivonindza: That one is taken from https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_virtual_machine.py, where it's called `_CheckDockerExists`. The script for docker installation can cause trouble if docker is already installed (or at least that's what it prints when it detects docker is installed), so maybe it makes sense to put this check in `AptInstall` in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_packages/docker.py.

ivonindza:
I addressed everything except moving `_HasDocker` and updating README.md. For `_HasDocker` see the separate comment. There is no information about available benchmarks in the README.md file as far as I can see. Running `./pkb.py --help` lists all the benchmarks and all the benchmark sets.

tox.ini
```diff
@@ -17,7 +17,7 @@ deps =
 [testenv:scripttests]
 commands = nosetests {toxinidir}/script-tests []
 setenv = 
-  PYTHONPATH = {toxinidir}/perfkitbenchmarker/scripts:{env:PYTHONPATH}
+  PYTHONPATH = {toxinidir}/perfkitbenchmarker/scripts:{envdir}
```
skschneider: Please create a separate PR for this.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+
+def _HasDocker(vm):
+  resp, _ = vm.RemoteCommand('command -v docker',
+                             ignore_failure=True,
+                             suppress_warning=True)
+  return bool(resp.rstrip())
+
+
+def Prepare(benchmark_spec):
+  """Install docker. Pull the required images from DockerHub. Create datasets.
```
skschneider: Style: It's preferable to have a one-line summary, followed by a blank line, followed by the detailed description.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+  Start Spark master and workers.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+  master = benchmark_spec.vm_groups['master'][0]
+  workers = benchmark_spec.vm_groups['workers']
+
+  def PrepareCommon(vm):
```
skschneider: Style: It's preferable to define these at the module level, rather than within another function definition.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+    vm.RemoteCommand(start_master_cmd)
+
+  def PrepareWorker(vm):
+    PrepareCommon(vm)
+    start_worker_cmd = ('sudo docker run -d --net host --volumes-from data '
+                        '--name spark-worker cloudsuite/spark worker '
+                        'spark://%s:7077' % master.internal_ip)
+    vm.RemoteCommand(start_worker_cmd)
+
+  target_arg_tuples = ([(PrepareWorker, (vm,), {}) for vm in workers] +
```
skschneider: Style: It's preferable to use a list instead of a one-element tuple. RunParallelThreads is able to accept `[vm]` and `[master]`.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+
+  benchmark_cmd = ('sudo docker run --rm --net host --volumes-from data '
+                   'cloudsuite/in-memory-analytics %s %s '
+                   '--master spark://%s:7077' %
+                   (FLAGS.cloudsuite_in_memory_analytics_dataset,
+                    FLAGS.cloudsuite_in_memory_analytics_ratings_file,
+                    master.internal_ip))
+  stdout, _ = master.RemoteCommand(benchmark_cmd, should_log=True)
+
+  execution_time = re.findall(r'Benchmark execution time: (\d+)ms', stdout)[0]
```
skschneider: Why `re.findall(...)[0]`?

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+  results.append(sample.Sample('Benchmark execution time',
+                               float(execution_time) / 1000, 'seconds'))
+
+  return results
+
+
+def Cleanup(benchmark_spec):
```
skschneider: Similar comments as Prepare.

perfkitbenchmarker/linux_benchmarks/cloudsuite_inmemory_analytics_benchmark.py
skschneider: I don't think inside `AptInstall` would work at the moment. See where `AptInstall` is called here: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/d7de2e2aa82650190b4640f8d64657846608144b/perfkitbenchmarker/linux_virtual_machine.py#L780
Immediately after, the package gets added to the VM object's `_installed_packages` list. If the VM is a static machine (one that was set up manually by the user and used by PKB to run the benchmark), then PKB will try to clean up after itself by uninstalling the package during cleanup. Perhaps packages should have an optional check-if-exists type method, but that would be more work and require a separate PR. For now, I think conditionally installing is the way to go, but the `_HasDocker` function could be moved into docker.py to become `docker.HasDocker`.

skschneider: Or `docker.CheckIfInstalled` or some other appropriate name.

skschneider:
Took another pass. Mostly style comments at this point.

tox.ini
ivonindza: I can't commit without this one, since pre-commit tests will fail. This issue (#900) has been around for some time.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
ivonindza: I'm not really proficient at python, actually never used it before. What do you suggest here?

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
ivonindza: It makes more sense to me to define things in the narrowest scope in which they will be used. These are just helpers that are only used within the function where I defined them. That said, if you have a styleguide that recommends them to be at the module level, I'll move them.

skschneider: In python, there's a pickle module that can serialize and deserialize objects. One of its weaknesses is that it can't handle nested functions. We currently only use pickle to serialize the BenchmarkSpec to preserve state between stages of benchmark execution: provision, prepare, run, cleanup, teardown. Since the BenchmarkSpec doesn't hold any references to these functions, it's not required, just a recommendation. I'd be fine with leaving them like this.

perfkitbenchmarker/linux_benchmarks/cloudsuite_in_memory_analytics_benchmark.py
skschneider: Using `re.findall` could imply it's expected to find more than one instance of the pattern (of which only the first matched result is later selected by `[0]`). If only one instance is ever expected, `re.search(pattern, stdout).group(1)` would be equivalent. Either way, the resulting error message if the pattern was not found will be uninformative (IndexError or AttributeError) unless this is broken into two steps with a check in between.
```
matches = re.findall(...)
if len(matches) != 1:
  raise errors.Benchmarks.RunError('Expected ... but ... happened instead')
execution_time = matches[0]
```
This is probably the most resilient approach, since it covers both the cases of 0 and >1 matches found. I don't know much about cloudsuite, but if a successful command implies only zero or one match will be found, then what you already have is probably good enough, since the stdout is also added to the PKB log file.

tox.ini
skschneider: Thanks. Looks like #941 addresses this.

skschneider:
The tox.ini merge conflict is the major blocker at this point. If that's resolved, this LGTM. Beyond that, the next major thing is having the docker helper function moved into the docker.py module. The rest is less important. When you're done making changes, let me know, and I can merge this into master.

ivonindza:
OK, I'm done

skschneider:
Merged. Thanks for contributing!

PR #941, opened 2016-04-08
Handle unset PYTHONPATH in tox.ini

noahl:
Provide default value so Tox won't error when PYTHONPATH is unset.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ deps =
 [testenv:scripttests]
 commands = nosetests {toxinidir}/script-tests []
 setenv = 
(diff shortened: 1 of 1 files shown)


noahl:
This addresses #900 .

@meteorfox please take a look. This isn't quite the patch you suggested because `{envdir}` would put the top-level PKB directory in the PYTHONPATH. What this does is keep using the PYTHONPATH from the environment, but default to an empty string if it's unset so Tox won't error in that case.

I still don't love the PYTHONPATH handling, because I don't know how it interacts with the way virtualenv sets PYTHONPATH, but I can't find any documentation on the Tox documentation site, and this is definitely an improvement over the current situation.

meteorfox:
:+1: Works for me

PR #942, opened 2016-04-08
OpenStack CLI provider implementation

meteorfox:
OpenStack provider implementation that similar to the other providers uses the CLI utilities instead of Python libraries.

Using the CLI utilities has several benefits:
- Interacts with cloud provider the same way a user would.
- Lists the commands used to interact with OpenStack while previously only HTTP requests were logged.
- Pushes backward-compatibility support to the CLI instead of PKB managing it itself.
- OpenStack CLI utilities are commited to backward-compatiblity of supported releases.
- Allows `Specs` implementation to be easily pickled.
- Configuration is handled by the CLI, simplifies options and configuration.
- Removes PKB's direct dependency to OpenStack python libraries.
- Simplifies error handling (internally)

Addresses Issues #909 #899

Full log showing run across all run stages:
https://gist.github.com/meteorfox/e14434b6f88a18f287c3c8d918c78b84


perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -12,107 +12,154 @@
...
+
+def DeleteVolume(resource, volume_id):
+  """Deletes a remote (Cinder) block volume."""
+  vol_cmd = os_utils.OpenStackCLICommand(resource, 'volume', 'delete',
+                                         volume_id)
+  del vol_cmd.flags['format']  # volume delete does not support json output
+  vol_cmd.Issue()
+
+
+@vm_util.Retry(poll_interval=1, max_retries=-1, timeout=300, log_errors=False,
```
noahl: The `poll_interval` seems a little low to me, but I guess it's your backend that's going to be hit by this, so up to you. I just want to make sure you didn't set this low for development and forget to turn it back up.

perfkitbenchmarker/providers/openstack/os_network.py
```diff
@@ -27,134 +29,143 @@
...
+  """
+
+  CLOUD = providers.OPENSTACK
+
+  def __init__(self):
+    self._lock = threading.Lock()  # Guards security-group rule set
+    self.sec_group_rules_set = set()
+
+    with self._lock:
+      cmd = utils.OpenStackCLICommand(self, 'security group', 'show',
```
noahl: Please move hard-coded constants (`'security group'`, `'perfkit_sc_group'`) to module-level constants at the top of the file.

perfkitbenchmarker/providers/openstack/os_disk.py
meteorfox: For the most part, this operation is really quick (< 5 seconds for empty volumes) which is why I have the small `poll_interval`, but thinking about it, when we create a boot volume with an image it might take longer, but it will depend on the backend implementation they configured to use (LVM, Ceph, etc). I can bump it up to 5 seconds, that should be good enough for most cases.

perfkitbenchmarker/providers/openstack/os_network.py
meteorfox: :+1:

perfkitbenchmarker/providers/openstack/os_network.py
```diff
@@ -27,134 +29,143 @@
...
+    if to_port is None:
+      to_port = port
+
+    sec_group_rule = (port, to_port, vm.group_id)
+
+    with self._lock:
+      cmd = utils.OpenStackCLICommand(vm, 'security group rule', 'create',
+                                      vm.group_id)
+      cmd.flags['dst-port'] = '%d:%d' % (port, to_port)
+      if sec_group_rule in self.sec_group_rules_set:
```
noahl: Should this check be before you create the command?

perfkitbenchmarker/providers/openstack/os_network.py
```diff
@@ -27,134 +29,143 @@
...
+      _, stderr, _ = cmd.Issue()
+      if stderr:
+        raise errors.Error(stderr)
+      return floating_ip
+
+  def _get_or_create(self, vm):
+    list_cmd = [FLAGS.openstack_nova_path, 'floating-ip-list']
+    stdout, stderr, _ = vm_util.IssueCommand(list_cmd)
+    floating_ip_dict_list = utils.ParseFloatingIPTable(stdout)
+    for flip_dict in floating_ip_dict_list:
```
noahl: Please rename `flip_dict` to something more clear, like `floating_ip`. `ip_dict` would also work. The first time I read the code I wondered what flipping had to do with IPs.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -33,214 +43,328 @@
...
+      return False
+
+    show_cmd = os_utils.OpenStackCLICommand(self, 'server', 'show', self.id)
+    stdout, _, _ = show_cmd.Issue(suppress_warning=True)
+    try:
+      resp = json.loads(stdout)
+    except ValueError:
+      return False
+
+    return resp is not None
```
noahl: Can `resp` ever be `None` here? I don't understand under what circumstances `json.loads` would return `None`. If I don't understand `json.loads`, please correct me, but otherwise I think you can just return from inside the `try` block.

perfkitbenchmarker/providers/openstack/os_network.py
meteorfox: haha.. good point. We called them "flips" at Rackspace. It's a terrible name, I'll fix it.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -33,214 +43,328 @@
...
+      cmd = os_utils.OpenStackCLICommand(self, 'keypair', 'create',
+                                         self.key_name)
+      cmd.flags['public-key'] = self.ssh_public_key
+      cmd.IssueRetryable()
+      self.uploaded_keypair_set.add(self.zone)
+      if self.zone in self.deleted_keypair_set:
+        self.deleted_keypair_set.remove(self.zone)
+
+  def _DeleteSSHPublicKey(self):
+    """Deletes SSH publick key used for the VM."""
```
noahl: Nit: spelling of `publick`.

noahl:
@meteorfox a few small comments, but mostly this looks great. I'm excited to get it merged.

meteorfox:
@noahl thanks for the review. I'll work on your suggestions.

After this gets merged I'll resume my work on the disk types for OpenStack  #759

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -12,107 +12,154 @@
...
+  """Creates a remote (Cinder) block volume."""
+  vol_cmd = os_utils.OpenStackCLICommand(resource, 'volume', 'create', name)
+  vol_cmd.flags['availability-zone'] = resource.zone
+  if image:
+    vol_cmd.flags['image'] = image
+    vol_cmd.flags['size'] = GetImageMinDiskSize(resource, image)
+  if FLAGS.openstack_volume_size:
+    vol_cmd.flags['size'] = FLAGS.openstack_volume_size
+  else:
+    vol_cmd.flags['size'] = REMOTE_BOOT_DISK_SIZE_GB
```
kimcie: Maybe you should change it to use disk size spec from default_config_constrants.yaml file?

meteorfox:
@noahl I've made the changes based on your feedback. can you PTAL? thanks!

meteorfox:
@noahl @cmccoy Please, let me know if I need to improve something here. thanks!

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -12,36 +12,24 @@
...
+flags.DEFINE_string('openstack_cli_path',
+                    default='openstack',
+                    help='The path to the OpenStack CLI binary.')
 
-flags.DEFINE_string('openstack_tenant',
-                    os.getenv('OS_TENANT_NAME', 'admin'),
-                    'OpenStack tenant name, defaults to $OS_TENANT_NAME.')
+flags.DEFINE_string('openstack_nova_path',
+                    default='nova',
+                    help='the path to the Nova CLI binary.')
```
noahl: Nit: capital T in first "the"

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -12,107 +12,155 @@
...
 
 
+def CreateVolume(resource, name, image=None):
+  """Creates a remote (Cinder) block volume."""
+  vol_cmd = os_utils.OpenStackCLICommand(resource, 'volume', 'create', name)
+  vol_cmd.flags['availability-zone'] = resource.zone
+  if image:
+    vol_cmd.flags['image'] = image
+    vol_cmd.flags['size'] = GetImageMinDiskSize(resource, image)
+  if FLAGS.openstack_volume_size:
```
noahl: It looks like this `if` will always override the value of `vol_cmd.flags['size']` set in the line above. Did you mean to replace the `else` part with `GetImageMinDiskSize`?

perfkitbenchmarker/providers/openstack/flags.py
meteorfox: done

perfkitbenchmarker/providers/openstack/os_disk.py
meteorfox: Good catch. Actually, so in this case, we want to set it to the image min_size required if the volume is being used as a bootable volume. Otherwise, it should just default to either the flag or the default volume size.

meteorfox:
If there are no other comments and it LGTY I would like to merge this PR.

Note that this will require the user to install and configure OpenStack CLI, which a link to the instructions are included in the README, but the change should be highlighted in the CHANGELOG as a breaking change and a feature.

noahl:
@meteorfox I'm so sorry, I still need to review the rest of this. Completely my fault. I will do it today, and let you know when I'm done.

meteorfox:
@noahl No worries. I know this is a big PR. Thanks for reviewing it.

perfkitbenchmarker/providers/openstack/os_network.py
```diff
@@ -11,150 +11,181 @@
...
+    if vm.is_static:
+      return
+
+    sec_group_rule = (ICMP, icmp_type, icmp_code, vm.group_id)
+    with self._lock:
+      cmd = utils.OpenStackCLICommand(vm, OSC_SEC_GROUP_RULE_CMD, 'create',
+                                      vm.group_id)
+      cmd.flags['dst-port'] = str(icmp_type)
+      cmd.flags['proto'] = ICMP
+      if sec_group_rule in self.sec_group_rules_set:
```
noahl: Maybe do this check right after creating `sec_group_rule`?

meteorfox: The idea is to avoid creating the `sec_group_rule` if it's already in the set. which it would immediately return from the function.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -33,214 +43,327 @@
...
+                      'For valid floating IP pools run',
+                      '"openstack ip floating pool list".',))
+      raise errors.Config.InvalidValue(msg)
+
+  def _CheckNetworkExists(self):
+    cmd = os_utils.OpenStackCLICommand(self, 'network', 'show',
+                                       self.network_name)
+    stdout, stderr, _ = cmd.Issue()
+    if stderr:
+      msg = ' '.join(('Network %s could not be found.' % self.network_name,
```
noahl: You should be able to write this message with strings on different lines that are automatically appended, the same way you did in `_CheckImage` and `_CheckFlavor`.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -33,214 +43,327 @@
...
+      if flip_pool['Name'] == self.floating_ip_pool_name:
+        break
+    else:
+      msg = ' '.join(('Floating IP pool %s could not be found.'
+                      % self.floating_ip_pool_name,
+                      'For valid floating IP pools run',
+                      '"openstack ip floating pool list".',))
+      raise errors.Config.InvalidValue(msg)
+
+  def _CheckNetworkExists(self):
```
noahl: Suggestion: make one method that runs an OpenStack command and raises an error with a custom message if that command has non-empty stderr. Then use that function to implement `_CheckImage`, `_CheckFlavor`, and `_CheckNetworkExists`.

perfkitbenchmarker/providers/openstack/os_network.py
noahl: I don't understand. I see why you want to avoid creating the `sec_group_rule`, but the way this function is written, it looks like you create an OpenStackCLICommand object and set some properties on it before you check whether `sec_group_rule` already exists and isn't needed. I'm suggesting that you check whether `sec_group_rule` exists before creating the OpenStackCLICommand, so you can save a bit more work. Will that not work? Am I missing something?

noahl:
Other than my few very minor comments, LGTM.

perfkitbenchmarker/providers/openstack/os_network.py
meteorfox: @noahl You are absolutely right. When I read your first comment I understood it as actually executing the `sec_group_rule` command and then checking for it to be in the set.

Yes, I'll make the change, that should save us a little bit of work and memory. We do that already on the function for TCP, UDP ports, I don't know why I missed this one. Thanks

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: Good call. I'll give that a try.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
meteorfox: `_CheckImage` and `_CheckFlavor` also use `join()`. The tricky bit is replacing the value, I guess another way to do it is using `format()`. Create the automatically appended string using `{n}` placeholders and assign it to `msg`, then call `msg.format(...)` with the values.

Unless you are suggesting to do everything inline like I did for `_CheckImage` and `_CheckFlavor` instead of assingning it to `msg`.

cbhall:
I picked up this pull request and tried it out.  I see some anomalous elements, and one bug, related to network handling.  The anomalous elements are that the new code uses a strange mixture of openstack, nova-networking, and neutron, as shown in these extracts from the stderr output of a successful run:
```
2016-06-23 15:02:45,778 4d2d632e Thread-1 coremark(1/1) INFO     Running: openstack ip floating pool list --format json
2016-06-23 15:03:05,013 4d2d632e Thread-1 coremark(1/1) INFO     Running: nova floating-ip-list
2016-06-23 15:03:05,718 4d2d632e Thread-1 coremark(1/1) INFO     Running: openstack ip floating create ext_net --format json
2016-06-23 15:04:20,650 4d2d632e Thread-33 coremark(1/1) INFO     Running: neutron floatingip-show c77b9d1b-aa29-43ea-946a-b70595a2ca03 --format json
2016-06-23 15:04:21,109 4d2d632e Thread-33 coremark(1/1) INFO     Running: neutron floatingip-delete c77b9d1b-aa29-43ea-946a-b70595a2ca03 --format json
```
I think the `openstack` command should be used exclusively, eliminating the requirement to have python-novaclient and python-neutronclient installed.

The bug I saw was in an unsuccessful run and was related to the `nova floating-ip-list` command that was used.  That command returned one floating IP that was allocated but currently unused.  However, `openstack ip floating create` was nonetheless called and failed because the OpenStack deployment had only one floating IP available to the tenant.  I suspect part of the problem here was that `--format json` was not used on the `nova` command.

One other benign issue I noted was that even though `openstack server create` was invoked with the `--wait` option, there were two (apparently) unnecessary `openstack server show` invocations right after that which returned essentially the same information as the original `create` command.

One last comment I have is unrelated to the new CLI code.  I'm not sure why all PerfKit runs use `perfkit_sc_group`, rather than creating a new unique security group rule.  The security group rule is the only artifact left behind from a PerfKit Benchmarker run.  I think that any artifact left behind is undesirable.  Also, the use of a common security group rule for all workloads seems inflexible, should particular workloads have specific requirements that other workloads don't have.

meteorfox:
@cbhall Great feedback! Thanks, I appreciate you taking the time to test it and even analyzing the run.

- Strange mixture of OpenStack commands (e.g. OpenStack, nova, neutron):
  - I had to rely on nova and neutron due to certain functionality missing in the OpenStack CLI.  Work is on-going upstream to add these to OpenStack CLI. Nova outputs to which pool each floating IP belongs to while OpenStack CLI does not, if I had used just the OpenStack CLI, I would have to execute about 3 commands to figure out the same information.
  - I opened an issue upstream for this (https://bugs.launchpad.net/python-openstackclient/+bug/1566090)

- Floating IP bug:
  - This sounds definitely like a bug, I will try to recreate it on my deployment.

- `openstack server create` and invocations of  `openstack server show`:
  - Actually, this one is by design. The `server create` does the creation of the instance,  then the first `server show` is to make sure it does exists and the second `server show` is to grab the IP addresses.
  - I think I can add a fast path to avoid the 2 extra calls if the create call returns with all the data needed (e.g status=ACTIVE, ip addresses, etc).

- `perfkit_sc_group`:
  - That was the behavior the previous provider implementation had and I just ported it. However, you made an excellent point about not necessarily using the same security group for the workloads. Can you please open an issue against it?

cbhall:
I looked into the floating IP bug I saw and determined the reason for the failure.  The problem is that the regular expression used to match the lines of interest in the table output is expecting only one string/word in each column.  The header of the `nova floating-ip-list` command violates that assumption because there are two words in a couple of cases:
```
+----+----+-----------+----------+------+
| Id | IP | Server Id | Fixed IP | Pool |
+----+----+-----------+----------+------+
```
The code in `ParseNovaTable` skips the first match using `next(tuples)  # Skip table header`, but in the problematic case the match that is skipped is actually the data for the available floating IP.

I can see two possible fixes for the problem.  The first is to change the regular expression such that one or two words can be matched, such as in:
```
FIVE_COLUMNS_ROW_REGEX = (r'\|\s+(\S+|\S+ \S+)\s+\|\s+(\S+|\S+ \S+)\s+\|\s+(\S+|\S+ \S+)\s+'
                          r'\|\s+(\S+|\S+ \S+)\s+\|\s+(\S+|\S+ \S+)\s+\|')
```
The second, and better fix is to look for matches only in the data section of the table, using:
```
groups = (regex_pattern.match(line) for line in stdout_lines[3:-1])
```
and eliminate the later line that skips the header (`next(tuples)  # Skip table header`).

It is easy to demonstrate the bug on any OpenStack deployment.  Just create one floating IP using `nova floating-ip-create`, and then run PerfKit.  The run will never use that available floating IP, but will instead always create another.

meteorfox:
@cbhall Great catch! You are absolutely right, the regex will not match properly for lines that contain more than 1 word per column. I'll add a unit test for this, and then code a solution for it.

I think we will have to go with a modified version of your first proposed solution that works for 1 or more space-separated words per column.  `ParseNovaTable` is used to parsed all outputs from the Nova commands, so we can't just grab the same number of columns since the number of columns changes depending on the command output. Eventually, this problem will go away once we get this functionality under OpenStack CLI since we get JSON formatted output.

Thanks for catching this one!

meteorfox:
@vhe182 Go here: https://cla.developers.google.com/ and sign the CLA agreement as "Only yourself". You'll need a gmail account. Then make sure that email is on your Github's profile email addresses.

cbhall:
I upgraded my python-openstackclient to (the latest) version 2.6.0, and was able to trivially replace all usage of `neutron` with `openstack`.  Please consider incorporating the following.  (There is some lingering problem with our CLA, so I can only suggest changes and not yet contribute any.)
In os_network.py, `release`, replace
```
    # TODO(meteorfox): Change floating IP commands to OpenStack CLI once
    # support for them is added.
    show_cmd = [FLAGS.openstack_neutron_path,
                NEUTRON_FLOAT_IP_SHOW_CMD,
                floating_ip_dict['id'], '--format', 'json']
    stdout, stderr, _ = vm_util.IssueCommand(show_cmd, suppress_warning=True)
```
with
```
    cmd = utils.OpenStackCLICommand(vm, OSC_IP_CMD, OSC_FLOATING_SUBCMD, 'show',
                             floating_ip_dict['id'])
    stdout, stderr, _ = cmd.Issue(suppress_warning=True)
```
and replace:
```
      delete_cmd = [FLAGS.openstack_neutron_path,
                    NEUTRON_FLOAT_IP_DELETE_CMD, updated_floating_ip_dict['id'],
                    '--format', 'json']
      stdout, stderr, _ = vm_util.IssueCommand(delete_cmd,
                                               suppress_warning=True)
```
with
```
      cmd = utils.OpenStackCLICommand(vm, OSC_IP_CMD, OSC_FLOATING_SUBCMD, 'delete',
                               floating_ip_dict['id'])
      del cmd.flags['format']  # Command does not support json output format
      stdout, stderr, _ = cmd.Issue(suppress_warning=True)
```
As well, these two lines:
```
NEUTRON_FLOAT_IP_SHOW_CMD = 'floatingip-show'
NEUTRON_FLOAT_IP_DELETE_CMD = 'floatingip-delete'
```
and `is_attached` (which I believe is unused) can be removed from os_network.py.

cbhall:
Further to my previous comment, I have prototyped some code to remove all use of the `nova` command.  In two cases, it was trivial to change from `nova` to `openstack`.
1) In os_virtual_machine.py, `_CreateServerGroup`, replace
```
    nova_cmd = [FLAGS.openstack_nova_path, 'server-group-create',
                group_name, FLAGS.openstack_scheduler_policy]
    stdout, _, _ = vm_util.IssueCommand(nova_cmd, suppress_warning=True)
    server_group = os_utils.ParseServerGroupTable(stdout)
```
with
```
    cmd = os_utils.OpenStackCLICommand(self, 'server', 'group', 'create',
              group_name)
    cmd.flags['policy'] = FLAGS.openstack_scheduler_policy
    stdout, _, _ = cmd.Issue(suppress_warning=True)
    server_group = json.loads(stdout)
```
2) In os_virtual_machine.py, `_DeleteServerGroup`, replace
```
      nova_cmd = [FLAGS.openstack_nova_path, 'server-group-delete',
                  self.server_group_id]
      vm_util.IssueCommand(nova_cmd, suppress_warning=True)
```
with
```
      cmd = os_utils.OpenStackCLICommand(self, 'server', 'group', 'delete',
                self.server_group_id)
      del cmd.flags['format']  # server group delete does not support json output
      cmd.Issue(suppress_warning=True)
```

The third and final case where the `nova` command is used was problematic as @meteorfox had suggested in another comment.  A minor change to the `openstack` command should make this easier as documented below.  However, I also came upon another issue that further complicated things and that appears to be an OpenStack bug, as I will explain in a follow up comment.

3a) In os_virtual_machine.py, `_CheckFloatingIPNetworkExists`, append
```
    cmd = os_utils.OpenStackCLICommand(self, 'network', 'show',
                    self.floating_ip_pool_name)
    stdout, _, _ = cmd.Issue()
    network_show = json.loads(stdout)
    self.floating_ip_pool_id = network_show['id']
```
3b) In os_virtual_machine.py, `_CreateDependencies`, replace
```
    self.public_network = os_network.OpenStackFloatingIPPool(
        self.floating_ip_pool_name)
```
with
```
    self.public_network = os_network.OpenStackFloatingIPPool(
        self.floating_ip_pool_name, self.floating_ip_pool_id)
```
3c) In os_network.py, `OpenStackFloatingIPPool`, replace
```
  def __init__(self, pool_name):
    self.ip_pool_name = pool_name
```
with:
```
  def __init__(self, pool_name, pool_id):
    self.ip_pool_name = pool_name
    self.ip_pool_id = pool_id
```
3d) In os_networks.py, `_get_or_create`, replace
```
    list_cmd = [FLAGS.openstack_nova_path, NOVA_FLOAT_IP_LIST_CMD]
    stdout, stderr, _ = vm_util.IssueCommand(list_cmd)
    floating_ip_dict_list = utils.ParseFloatingIPTable(stdout)
    for floating_ip_dict in floating_ip_dict_list:
      if (floating_ip_dict['pool'] == self.ip_pool_name and
              floating_ip_dict['instance_id'] is None):
        return floating_ip_dict
```
with
```
    cmd = utils.OpenStackCLICommand(vm, OSC_IP_CMD, OSC_FLOATING_SUBCMD, 'list')
    stdout, stderr, _ = cmd.Issue()
    flip_list = json.loads(stdout)
    for flip in flip_list:
      if flip['Fixed IP Address'] is None:
        cmd = utils.OpenStackCLICommand(vm, OSC_IP_CMD, OSC_FLOATING_SUBCMD, 'show',
                                 flip['Floating IP Address'])
        stdout, stderr, _ = cmd.Issue()
        flip_show = json.loads(stdout)
        if flip_show['floating_network_id'] == self.ip_pool_id:
          return flip_show
```

The extra `openstack ip floating show` command could be eliminated if the `openstack ip floating list` command would include the `floating_network_id` in the response.  In fact, the underlying API request used by the `openstack` command, the neutron API `GET /v2.0/floatingips`, does include that information.

cbhall:
I found a particular case where the code I presented in my last comment does not work properly.  @meteorfox, I'm hoping you can help me determine if this is an OpenStack bug, and if so, to get it addressed.

The problem I found is related to an OpenStack user being granted administrative privileges.  When a user **does not** have an admin role, these three commands produce the same list of available floating IPs, all allocated to the project/tenant specified by the OS_TENANT_NAME environment variable:
```
nova floating-ip-list
neutron floatingip-list
openstack ip floating list
```
On the other hand, when a user has multiple roles that **does** include the `admin` role, then:
- `nova floating-ip-list` will show only the floating IPs allocated to the project/tenant specified by the OS_TENANT_NAME environment variable
- `neutron floatingip-list` or `openstack ip floating list` will show **all** of the floating IPs that have been allocated, regardless of which project/tenant allocated them.  It would appear that the OS_TENANT_NAME environment variable was simply ignored in that case.

The code I posted above which scans the list floating IPs looking for an unused one can then fail because it could choose a floating IP allocated by a different project/tenant.  When there is a subsequent attempt to run `genesis ip floating add <floatingIP> <serverID>`, it can fail with:
```
Unable to associate floating ip <floatingIP> to fixed ip <fixedIP> for instance <serverID>. Error: Bad floatingip request: Port 25473942-4e42-4d95-b0ed-d6f025e10ea8 is associated with a different tenant than Floating IP c24f0ed0-da68-4495-8be9-a79c201eab9e and therefore cannot be bound. (HTTP 400) (Request-ID: req-1ac53847-2da2-4d73-96b9-b15605def13a)
```

Tracing the actual API request that the `openstack ip floating list` command is using, it is `GET /v2.0/floatingips`, documented at http://developer.openstack.org/api-ref-networking-v2-ext.html#listFloatingIps.  The documentation claims that the API request "Lists floating IPs that are accessible to the tenant who submits the request."

So, is this API request broken when it returns floating IP that were not allocated to the tenant specified by OS_TENANT_NAME?

meteorfox:
@cbhall This is awesome. I'll try the changes you suggested, hopefully, we can get everything working with just 1 CLI.

I will take a look at the tenant problem with floating IPs as well. My guess is there way to use the CLI to pass the `tenant_id` HTTP URL query parameter, with that parameter you can filter the floating IPs for the given `tenant_id`, if you are admin you can ask for anyone's floating IPs. By default, it should just default to that user's floating IPs.

meteorfox:
@noahl Can I signoff Git commits for a colleague's commits, instead of having him go through the CLA process?

cbhall:
@meteorfox, I'm seeing another strange problem I'd like your help with.  This problem occurs with the current code in this pull request (that is, without the code changes I suggested in previous comments).

The problem is related to security groups.  After I run the 1.5.0 version of PerfKit, this is what I see for security groups:
```
$ nova secgroup-list
+-----+------------------+------------------------------------------------+
| Id  | Name             | Description                                    |
+-----+------------------+------------------------------------------------+
| 2   | default          | default                                        |
| 206 | perfkit_sc_group | Firewall configuration for Perfkit Benchmarker |
+-----+------------------+------------------------------------------------+
$ openstack security group list
+--------------------------------------+---------+------------------------+----------------------------------+
| ID                                   | Name    | Description            | Project                          |
+--------------------------------------+---------+------------------------+----------------------------------+
| b3b6ff53-ad79-4ac7-b084-efcd66a7d8a5 | default | Default security group | 4f299816e33f49c7ae3819ee2d7d19d9 |
+--------------------------------------+---------+------------------------+----------------------------------+
```
The `perfkit_sc_group` shown by the `nova` command is obviously the security group created by the version 1.5.0 code.

If I then run the modified PerfKit that includes #942, the run succeeds and afterwards I see:
```
$ nova secgroup-list
+-----+------------------+------------------------------------------------+
| Id  | Name             | Description                                    |
+-----+------------------+------------------------------------------------+
| 2   | default          | default                                        |
| 206 | perfkit_sc_group | Firewall configuration for Perfkit Benchmarker |
+-----+------------------+------------------------------------------------+
$ openstack security group list
+--------------------------------------+------------------+------------------------+----------------------------------+
| ID                                   | Name             | Description            | Project                          |
+--------------------------------------+------------------+------------------------+----------------------------------+
| b3b6ff53-ad79-4ac7-b084-efcd66a7d8a5 | default          | Default security group | 4f299816e33f49c7ae3819ee2d7d19d9 |
| c0572462-9593-4b6e-8039-b3ec2cbd59d9 | perfkit_sc_group | perfkit_sc_group       | 4f299816e33f49c7ae3819ee2d7d19d9 |
+--------------------------------------+------------------+------------------------+----------------------------------+
```

The problem I've run into is that if I then run the commands:
```
$ nova secgroup-delete perfkit_sc_group
$ openstack security group delete perfkit_sc_group
```
and then again try to run the modified PerfKit that includes #942, the run fails.  The output on stderr shows the failure as:
```
2016-07-06 17:02:23,523 3f07315a Thread-1 coremark(1/1) INFO     Ran openstack server create pkb-3f07315a-0 --format json --flavor m1.small --security-group perfkit_sc_group --key-name perfkit_key_3f07315a --availability-zone nova --nic net-id=private --wait --image Ubuntu 14.04.3 LTS. Got return code (1).
STDOUT:
STDERR: Security group perfkit_sc_group not found for project 4f299816e33f49c7ae3819ee2d7d19d9. (HTTP 400) (Request-ID: req-20dbddbf-51b3-455a-a26f-c7dfe897bc21)
```
Once again checking the security groups after the failing run, it shows:
```
$ nova secgroup-list
+----+---------+-------------+
| Id | Name    | Description |
+----+---------+-------------+
| 2  | default | default     |
+----+---------+-------------+
$ openstack security group list
+--------------------------------------+------------------+------------------------+----------------------------------+
| ID                                   | Name             | Description            | Project                          |
+--------------------------------------+------------------+------------------------+----------------------------------+
| 1da216bb-3215-4e46-9bed-76374378e317 | perfkit_sc_group | perfkit_sc_group       | 4f299816e33f49c7ae3819ee2d7d19d9 |
| b3b6ff53-ad79-4ac7-b084-efcd66a7d8a5 | default          | Default security group | 4f299816e33f49c7ae3819ee2d7d19d9 |
+--------------------------------------+------------------+------------------------+----------------------------------+
```

meteorfox:
@cbhall I can't recreate it. I deleted the security group with both commands and it still created it and worked. 

Does your user has an admin role? I just want to know if that somehow changes the behavior. Also, the other strange part of your output is that `nova secgroup-list` shows different output than `openstack security group list`. They should be the same since they talk to the same services. Unless they are being run with different projects/tenants, or something else is going on.

cbhall:
@meteorfox, thanks for checking on the security group problem I was seeing.  With more investigation, I found that 3 OpenStack deployments I have access to are all broken as I indicated, presumably because they are older releases.  I did identify a fourth OpenStack deployment I have access to and that works as you indicated, and thus must be a newer release.  I will focus my efforts on that fourth deployment.

cbhall:
For a variety of reasons, I've been looking at the OpenStack API requests that the `openstack` command makes while running a simple benchmark (Coremark).  This exercise has revealed some interesting observations on how to efficiently use the CLI.  As an example, I will look at how the PerfKit code handles the checks for `--openstack_network` and `--openstack_floating_ip_pool` and propose some code changes there.  I'll try to summarize by suggesting some general approaches for improved efficiency in using the CLI.

To validate the private network to be used in a PerfKit run, the command `openstack network show <name>` is used.  The `openstack` command then first tries the API `GET /v2.0/networks/<name>`, based on the assumption that `<name>` is a network ID rather than a name.  When this API fails, the `openstack` command instead uses the API `GET /v2.0/networks` and then searches the result to extract the information for `<name>`.

This indicates that more efficient code to check the private and public networks in PerfKit would be something like:
```
  def _CheckNetworksExist(self):
    cmd = os_utils.OpenStackCLICommand(self, 'network', 'list')
    stdout, _, _ = cmd.Issue()
    resp = json.loads(stdout)
    # check the private network
    for network in resp:
      if network['Name'] == self.network_name:
        break
    else:
      raise errors.Config.InvalidValue(' '.join(
          ('Network %s could not be found.'
           % self.network_name,
           'For valid networks run '
           '"openstack network list".',)))
    if self.floating_ip_pool_name:
      # check the public network
      for network in resp:
        if network['Name'] == self.floating_ip_pool_name:
          self.floating_ip_pool_id = network['ID']
          break
      else:
        raise errors.Config.InvalidValue(' '.join(
            ('Network %s could not be found.'
             % self.floating_ip_pool_name,
             'For valid networks run '
             '"openstack network list".',)))
      # make sure the public network is a floating IP pool
      cmd = os_utils.OpenStackCLICommand(self, 'ip', 'floating', 'pool', 'list')
      stdout, _, _ = cmd.Issue()
      resp = json.loads(stdout)
      for flip_pool in resp:
        if flip_pool['Name'] == self.floating_ip_pool_name:
          break
      else:
        raise errors.Config.InvalidValue(' '.join(
            ('Floating IP pool %s could not be found.'
             % self.floating_ip_pool_name,
             'For valid floating IP pools run '
             '"openstack ip floating pool list".',)))
```

Some general suggestions:
- The checks for image and flavor show the same behavior inside the `openstack` command (attempting to use the name as if it was an ID, then falling back on asking for and searching a list).
- In general, PerfKit would be better to convert names into IDs as soon as possible, and then use the IDs in subsequent commands.  As a case in point, the command to create a VM will follow the same general behavior if presented with names for the network, etc, having to use a list command and doing a search to find the ID.

meteorfox:
@cbhall Great find. While I see the benefit of reducing the number of calls when a name instead of ID is given, I'm not completely sure if we should immediately convert names into IDs. For the following reasons:

 -  IMHO, I think it will be confusing for the user if they pass in a name to PKB and the commands logged don't ever include that name, so the user will have to determine what name correspond to the ID. But, I think there are commands that require the ID, I need to verify this.
 -  The user can already use IDs and PKB w/ OpenStack CLI will do the right thing, that is it won't have to fall back to the list search.

Of course, to address the first point we could always output in the log that such 'translation' happened so the reference of the `name -> ID` is there in the same log. Then from there forward use IDs to reduce the number of HTTP requests as you suggested.

cbhall:
@meteorfox I agree with your concerns about the readability of the log.

Also, I should point out that the `openstack` command is generally not that efficient internally, in ways unrelated to PerfKit usage.  With python-openstackclient==2.6.0 for example, I see that almost every invocation of the `openstack` command does two identical invocations of the `POST /v2.0/tokens` API.  The token returned by the first invocation is never used.

meteorfox:
@noahl My colleague @vhe182 already signed the CLA, is there something else we need to do? Thanks man!

noahl:
@meteorfox It looks like all we need is for @vhe182 to reply here and confirm that he is okay contributing his commits to Google. I know that Rackspace has a corporate CLA signed, so since you both work there you should be covered by that (and Googlebot found the CLA, as you can see in its comment above).

hildrum:
The problem is that `victor.estrada@rackspace.com` is not listed as being associated with the rackspace CLA according to Google.   A corporate CLA has a google group associated with it, and people in that google group are the ones that are allowed to contribute under that CLA.

What makes this more interesting is that `vhe182@gmail.com` is listed as having a CLA.  But since the commits are labeled with `victor.estrada@rackspace.com` and not  `vhe182@gmail.com`, the googlebot is labelling them as "cla: no"

meteorfox:
@hildrum Thanks I see, @vhe182 signed the CLA for contributions made by himself as well that's why his personal email shows up under the CLA. I already requested the owner of the rackspace-google contributors group to add him but I'm still waiting on that.

vhe182:
I, Victor Hugo Estrada, am totally fine with contributing my commits to Google projects and understand what contributing entails.  As it was in the beginning and ever shall be, so sayeth Frank.

noahl:
My mistake - @hildrum found the problem. We're waiting on the Rackspace-google contributors group.

noahl:
@meteorfox quick ping - have you fixed the Google group problem yet? I'd love to get this merged.

meteorfox:
@noahl Actually perfect timing. Finally, I've been told @vhe182 was added to the Google group. Can we do a recheck for the CLA, see if it's working?

noahl:
@meteorfox @vhe182 I *think* that all you have to do is post a comment saying "I signed it!" and CLAbot will recheck. If that doesn't work, I'll investigate more. Thanks for your patience!

hildrum:
I think vhe182 might need to reply here with I signed it!

vhe182:
I signed it!

PR #944, opened 2016-04-12
Update speccpu2006 benchmark to support iso file.

skschneider:
Additionally, update flags and log messages to match the new behavior
and to standardize to "SPEC CPU2006" spelling.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
@@ -12,120 +12,301 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Runs SpecCPU2006.
+"""Runs SPEC CPU2006.
 
-From SpecCPU2006's documentation:
-The SPEC CPU2006 benchmark is SPEC's industry-standardized, CPU-intensive
-benchmark suite, stressing a system's processor, memory subsystem and compiler.
+From the SPEC CPU2006 documentation:
+"The SPEC CPU 2006 benchmark is SPEC's next-generation, industry-standardized,
+CPU-intensive benchmark suite, stressing a system's processor, memory subsystem
+and compiler."
 
-SpecCPU2006 homepage: http://www.spec.org/cpu2006/
+SPEC CPU2006 homepage: http://www.spec.org/cpu2006/
 """
 
+import itertools
 import logging
+import os
 import posixpath
 import re
+import tarfile
 
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import data
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import sample
+from perfkitbenchmarker import stages
+from perfkitbenchmarker.configs import benchmark_config_spec
+
 
 FLAGS = flags.FLAGS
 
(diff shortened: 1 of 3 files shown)


skschneider:
Not sure why the travis tests failed. Local tests are currently working for me, and the error message looks unrelated to speccpu2006 changes. I'll look into it before I merge anything.

skschneider:
@ehankland PTAL

perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
```diff
@@ -12,120 +12,301 @@
...
+_TAR_REQUIRED_MEMBERS = 'cpu2006', 'cpu2006/bin/runspec'
 
 
 def GetConfig(user_config):
   return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
 
 
-def CheckPrerequisites():
-  """Verifies that the required resources are present.
+class SpecCpu2006ConfigSpec(benchmark_config_spec.BenchmarkConfigSpec):
```
ehankland: Can you explain why creating a BenchmarkConfigSpec subclass is preferable over just using CheckPrerequisites? CheckPrerequisites seems more appropriate to me, but I might be missing something

skschneider: The config is constructed before RunBenchmarks starts calling RunBenchmark, which actually creates the BenchmarkSpec and starts provisioning resources. So in the case of a ten-benchmark run, this will catch the missing resource up front rather than in the middle of the run after five benchmarks have already run. I've considered moving CheckPrerequisites, but there are a few benchmarks that use it in peculiar ways, so I'd prefer to hold off on that for now.

ehankland: Other benchmarks that use CheckPrerequisites will also suffer from this late check problem too though right? It sounds like the correct thing to do is to move the call to check prerequisites if we want it to catch problems early. That's probably outside the scope out of this PR, but could you please note or TODO that mentions that this is a temporary workaround until that gets cleaned up?

perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
```diff
@@ -12,120 +12,301 @@
...
+  if not examine_members:
+    return
+
+  with tarfile.open(tar_file_path, 'r') as tf:
+    members = tf.getnames()
+  cfg_member = 'cpu2006/config/{0}'.format(runspec_config)
+  required_members = itertools.chain(_TAR_REQUIRED_MEMBERS, [cfg_member])
+  for member in required_members:
+    if member not in members:
+      raise errors.Benchmarks.PrepareException(
```
ehankland: nit: you'll only catch one missing member per failed run

perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
skschneider: Added a TODO.

perfkitbenchmarker/linux_benchmarks/speccpu2006_benchmark.py
skschneider: Changed.

ehankland:
LGTM

Issue #945, opened 2016-04-12
Can't run with a static vm with a disk unless you specify disk_size


ehankland:
I got this error:
```
2016-04-12 10:07:11,731 ee17f302 MainThread fio(1/1) ERROR    Error during benchmark fio
Traceback (most recent call last):
  File "/usr/local/google/home/ehankland/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 381, in RunBenchmark
    detailed_timer)
  File "/usr/local/google/home/ehankland/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 269, in DoRunPhase
    collector.AddSamples(samples, name, spec)
  File "/usr/local/google/home/ehankland/PerfKitBenchmarker/perfkitbenchmarker/publisher.py", line 598, in AddSamples
    sample['metadata'], benchmark_spec)
  File "/usr/local/google/home/ehankland/PerfKitBenchmarker/perfkitbenchmarker/publisher.py", line 168, in AddMetadata
    data_disk.disk_size * data_disk.num_striped_disks)
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```
Looks like the problem is this [line](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/publisher.py#L168)

ehankland:
@noahl This looks related to your disk refactor - mind taking a look?

noahl:
@ehankland yep, I've got it. Looks like with the old metadata keys, we just ignore or use None as the size if we don't know it, but the computation assumes that we actually have that information.

noahl:
This should be fixed with the merge of #946 . Closing now, but please reopen if you see more problems.

PR #946, opened 2016-04-13
Fix publisher bug when disk_size is None

noahl:
This situation can happen when using static VMs. Also fix an unrelated
bug in the SampleCollectorTestCase that was causing our tests to fail.


noahl:
@ehankland please take a look. This addresses #945 by adding a unit test to catch the error in the future and a fix.

I also discovered an unrelated unit test failure, which this PR also fixes. I don't understand why it wasn't caught by Travis earlier.

ehankland:
LGTM

Issue #907, opened 2016-03-22
CheckPrerequisites does not have access to the BenchmarkSpec or benchmark-specific flag values


skschneider:
CheckPrerequisites was originally designed when we just had Prepare, Run, and Cleanup stages. At that time, the BenchmarkSpec constructor provisioned resources, so it wasn't created until the Prepare stage.

If there's no problem with it, I think the construction of BenchmarkSpec and its corresponding redirection of the global flags to benchmark-specific flags can be moved above the call to CheckPrerequisites, and the call to CheckPrerequisites can take the spec as an arg.

Either existing implementations of CheckPrerequisites will have to be modified to accept the spec, or the caller will need to inspect the function definition or try/catch the call with arguments and fall back to a call without arguments.

skschneider:
#933 moved CheckPrerequisites after the benchmark-specific flag overrides, so all that's missing is that CheckPrerequisites doesn't have access to the BenchmarkSpec.

skschneider:
I've checked all uses of CheckPrerequisites. The two main uses are:
A) Checking that a file exists using os.path.exists or data.ResourcePath.
B) Verifying or modifying FLAGS values (this actually wasn't even working properly until #933).

Considering these uses, it makes more sense to me at this time to drop the idea of providing access to the BenchmarkSpec. Instead, it seems like all the checks can be done after the config object is constructed, but before RunBenchmark is called for any benchmark. This will allow missing file errors to be caught earlier in multi-benchmark runs, before any resources are provisioned. The purpose of CheckPrerequisites then becomes an input validation step, so if anything, access to the config object is more in alignment with purpose B than providing access to the BenchmarkSpec.

I'm in favor of the following:
1) Creating a PR to move CheckPrerequisites out of RunBenchmark but after the config object has been constructed. The call in the new location would be wrapped by the flag redirection helper introduced in #936, so purpose B would still be possible.
2) Considering the possibility of providing CheckPrerequisites with access to the config object. If this is feasible, creating a separate PR to add that functionality.

skschneider:
#947 is merged and addresses step 1.

PR #947, opened 2016-04-13
Call CheckPrerequisites earlier.

skschneider:
Call CheckPrerequisites before RunBenchmark so that any input errors or
missing files will cause a failure before any resources are provisioned.

Remove the code specific to the speccpu2006 benchmark that was doing the
same thing.

Make some minor changes to the RunBenchmarks and RunBenchmark args.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -563,7 +563,7 @@ def RunBenchmarks():
                        gsutil_path=FLAGS.gsutil_path,
                        prefix=FLAGS.run_uri + '_')
   all_benchmarks_succeeded = all(r[2] == benchmark_status.SUCCEEDED
-                                 for r in run_status_lists)
+                                 for _, r in benchmark_run_list)
   return 0 if all_benchmarks_succeeded else 1
 
 
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -195,6 +195,51 @@ MAX_RUN_URI_LENGTH = 8
 events.initialization_complete.connect(traces.RegisterAll)
 
 
+def _CreateBenchmarkRunList():
+  """Create a list of configs and states for each benchmark run to be scheduled.
+
+  Returns:
+    list of (args, run_status_list) pairs. Contains one pair per benchmark run
+    to be scheduled, including multiple pairs for benchmarks that will be run
+    multiple times. args is a tuple of the first five arguments to pass to
+    RunBenchmark, and run_status_list is a list of strings in the order of
+    [benchmark_name, benchmark_uid, benchmark_status].
+  """
+  result = []
+  benchmark_tuple_list = benchmark_sets.GetBenchmarksFromFlags()
+  total_benchmarks = len(benchmark_tuple_list)
+  benchmark_counts = collections.defaultdict(itertools.count)
+  for i, benchmark_tuple in enumerate(benchmark_tuple_list):
+    # Construct benchmark config object.
+    benchmark_module, user_config = benchmark_tuple
+    name = benchmark_module.BENCHMARK_NAME
+    config_dict = benchmark_module.GetConfig(user_config)
+    config_spec_class = getattr(
+        benchmark_module, 'BENCHMARK_CONFIG_SPEC_CLASS',
+        benchmark_config_spec.BenchmarkConfigSpec)
+    config = config_spec_class(name, flag_values=FLAGS, **config_dict)
+
+    # Assign a unique ID to each benchmark run. This differs even between two
+    # runs of the same benchmark within a single PKB run.
+    uid = name + str(benchmark_counts[name].next())
+
+    # Optional step to check flag values and verify files exist.
+    check_prereqs = getattr(benchmark_module, 'CheckPrerequisites', None)
+    if check_prereqs:
+      try:
+        with config.RedirectFlags(FLAGS):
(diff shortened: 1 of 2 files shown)


skschneider:
Related to #907.
@ehankland PTAL

skschneider:
The Travis test failure looks unrelated to this PR. I have opened #948 for it.

perfkitbenchmarker/pkb.py
```diff
@@ -195,6 +195,51 @@
 events.initialization_complete.connect(traces.RegisterAll)
 
 
+def _CreateBenchmarkRunList():
+  """Create a list of configs and states for each benchmark run to be scheduled.
+
+  Returns:
+    list of (args, run_status_list) pairs. Contains one pair per benchmark run
```
ehankland: Passing around tuples of arguments and lists of run statuses is getting a bit hard to keep track of. It might make it a bit more readable if we bound the arguments directly to the RunBenchmark function and just passed around a bunch of partials that we called later. Making the benchmark status an attribute of the BenchmarkSpec (or if that's not appropriate, defining a Benchmark class that can keep track of status and stage etc) would probably also help. This PR doesn't make things any worse than they were, so I'm fine skipping this for now, but it would be nice if this could eventually get cleaned up.

perfkitbenchmarker/pkb.py
```diff
@@ -195,6 +195,51 @@
...
+    # Assign a unique ID to each benchmark run. This differs even between two
+    # runs of the same benchmark within a single PKB run.
+    uid = name + str(benchmark_counts[name].next())
+
+    # Optional step to check flag values and verify files exist.
+    check_prereqs = getattr(benchmark_module, 'CheckPrerequisites', None)
+    if check_prereqs:
+      try:
+        with config.RedirectFlags(FLAGS):
+          check_prereqs()
```
ehankland: This isn't a change from the way it worked before, but do we want to re-check prerequisites every time we run a stage or just on the "provision" stage?

perfkitbenchmarker/pkb.py
skschneider: I agree that things are a bit complicated at the moment. I see some duplication in fields of the args and the run_status_list, and they could be worked out.

In the case of staged runs, the status actually represents whether the desired stage was run and whether it was successful for each benchmark. So making it part of the BenchmarkSpec may not make sense, since it's a state that has more to do with the current PKB process, and not the overall benchmark run (provision-prepare-run-cleanup-teardown).

The only thing missing from a partial is the collector argument, which currently is only created once and shared for all benchmarks. It would need to be passed to this function, which somehow seems unclean to me.

Perhaps a stronger type definition or two might help. What we essentially have at this point is:
1) Information needed by RunBenchmark.
2) Information needed by benchmark_status.CreateSummary.

The history behind building the lists ahead of time is also something that is tied to our now-removed --parallelism flag, so there may not even be a need for it.

I guess what I'm leaning toward at this time is something like a map of uid to (module, config, status), and adapting the later loop in RunBenchmarks and CreateSummary to use the map. But it's a somewhat sizable change from the current behavior, so I'd prefer to think a bit more and then make a separate PR.

ehankland: > In the case of staged runs, the status actually represents whether the desired stage was run and whether it was successful for each benchmark. So making it part of the BenchmarkSpec may not make sense, since it's a state that has more to do with the current PKB process, and not the overall benchmark run (provision-prepare-run-cleanup-teardown).

Yeah - that's why I mentioned having a Benchmark class may be more appropriate. It could keep track of things like the current run stage and the status for that stage and have methods related to running itself.

> The only thing missing from a partial is the collector argument, which currently is only created once and shared for all benchmarks. It would need to be passed to this function, which somehow seems unclean to me.

You could pass it to the partial later - it doesn't need to be bound right away.

> But it's a somewhat sizable change from the current behavior, so I'd prefer to think a bit more and then make a separate PR.

That's what I was figuring - SGTM

perfkitbenchmarker/pkb.py
skschneider: I think that's generally reasonable, but there may be a case where it's less desirable. Currently, we allow re-running stages with mixed levels of support depending on the benchmark (e.g. provision-prepare-run-run-run-cleanup-teardown, or provision-prepare-run-cleanup-prepare-run-cleanup-teardown). This allows changing a few benchmark parameters and rerunning the experiment on the same resources. The problem is that there are some inputs (FLAGS) that control the properties of the provisioned resources, which don't make sense to change between running stages, and there are some inputs that control other benchmark behavior. To my knowledge, the latter only applies to the following benchmarks:
- cassandra_stress
- cloudsuite_websearch
- speccpu2006
- specsfs2014
- ycsb benchmarks
which all involve copying config files to the VMs (prepare stage). However, if we're running in stages and repeating the prepare stage, there's less concern about early failure (since the resources have already been provisioned), but it means a failure may arise in the middle of Prepare when trying to copy the files rather than before Prepare, in CheckPrerequisites.

So I think what you're suggesting is reasonable, but it may merit more thought as a separate issue along with the whole repeating stages discussion. Currently, aside from speccpu2006 peeking into a tar file, there aren't any checks that take a lot of time, so there wouldn't be a big performance drawback to letting the call happen in all stages by default.

ehankland: SGTM

ehankland:
LGTM

Issue #745, opened 2015-12-16
Clean up os_type use.


skschneider:
There are a few places where we check the --os_type flag that aren't compatible with the benchmark config "os_type" option. Considering OS type is a property of a VM, we should also look into allowing this to be set separately per VM group.

os_type constant definitions are also repeated in scattered locations, and these should be cleaned up.

skschneider:
After #950 and #952 are addressed, there will only be four ways in which FLAGS.os_type is used:

1) _GetValidBenchmarks in benchmark_sets.py. This restricts the set of benchmarks in a PKB run to those that match the Linux/Windows classification of FLAGS.os_type.
2) RunBenchmarks in pkb.py. This restricts Windows benchmarks so that they can only be run if PKB is also executing on a Windows machine.
3) _CreateBenchmarkRunList in pkb.py. This verifies that if a Linux benchmark is being run, all of its VM groups are configured for Linux VMs, and if a Windows benchmark is being run, all of its VM groups are configured for Windows VMs.
4) ReadStaticVirtualMachineFile in static_virtual_machine.py. This method reads each VM's properties from a config file. It expects a certain set of required properties based on FLAGS.os_type, but os_type is actually one of the properties that can be configured. There is a check to ensure that the os_type of each read configuration matches the Linux/Windows classification of FLAGS.os_type, but the VM os_type need not match FLAGS.os_type if they are both different flavors of Linux. For now, because of the restriction from (1) above of only running either Linux or Windows benchmarks in one PKB run based on FLAGS.os_type, it is also guaranteed that all static VM configurations will match the requirements of the benchmarks that are run. That is, when the BenchmarkSpec calls GetStaticVirtualMachine, it will always get a VM that matches the Linux/Windows state of the current benchmark and VM group.

In summary, the current state of the code is such that one run of PKB can include
A) Linux benchmarks with Linux VMs, or
B) Windows benchmarks with Windows VMs (this is only allowed when running PKB on a Windows machine)
No mixing is allowed. The only way to run Windows benchmarks is via --os_type=windows. If --os_type is anything else, PKB will run Linux benchmarks and expect Linux VMs. Until we decide to change this behavior, I think that's all the work for this issue.

Issue #741, opened 2015-12-16
disk.GetDevicePath() doesn't work on Windows


noahl:

noahl:
This has been resolved by #743 .

It doesn't make GetDevicePath() work on Windows, because that doesn't seem to make sense, but it adds a GetDeviceId() method that does make sense on Windows and implements it.

PR #950, opened 2016-04-14
Remove use of FLAGS.os_type from BaseDisk.

skschneider:
Initialize BaseDisk.os_type attribute from BaseDiskSpec.os_type.
Initialize BaseDiskSpec.os_type to match parent _VmGroupSpec.os_type.
Related to #745.

### DIFF
--- a/perfkitbenchmarker/disk.py
+++ b/perfkitbenchmarker/disk.py
@@ -21,7 +21,6 @@ import abc
 import logging
 
 from perfkitbenchmarker import flags
-from perfkitbenchmarker import os_types
 from perfkitbenchmarker import resource
 from perfkitbenchmarker.configs import option_decoders
 from perfkitbenchmarker.configs import spec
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/disk.py
```diff
@@ -172,6 +172,8 @@ class BaseDiskSpec(spec.BaseSpec):
     mount_point: None or string. Directory of mount point.
     num_striped_disks: int. The number of disks to stripe together. If this is
         1, it means no striping will occur. This must be >= 1.
+    os_type: None or string. OS type of the VM to which the disk will be
```
ehankland: Giving a data disk an os_type feels a bit wrong to me - could we give it a reference to the VM it's attached to (when Attach() is called) and use that VM's os_type instead?

skschneider: Attach gets complicated with all the per-provider overriding. It may make sense to just remove the FLAGS.os_type checks entirely. It seems like the use cases are pretty well partitioned by how we've split the code into linux_benchmarks, linux_packages, and linux_virtual_machine vs windows_benchmarks, windows_packages, and windows_virtual_machine. Also, at this point most providers override GetDevicePath and ignore this check anyway. I can't even find the use case for GetDeviceId. @noahl opened #741, so perhaps he can comment on this.

noahl: @skschneider DeviceId is a Windows-specific concept. It looks like a path, but I don't think it really is a path. I agree with @ehankland that disks don't really have an inherent OS. Maybe it would make more sense to make GetDeviceId be a method of windows_virtual_machine that takes an attached disk as an argument.

noahl: For now I'm okay just getting rid of the `os_type` stuff. You can make `GetDeviceId` check if `self.disk_number` is defined and return the DeviceId if it is.

skschneider:
Pushed a new version that simply removes the FLAGS.os_type check in favor of direct attribute check. PTAL.

Issue #948, opened 2016-04-13
background_cpu_test.py is flaky


skschneider:
I've hit this twice so far in the automated tests that run for each pull request.

Example failure:
```
======================================================================
FAIL: Check that the background_cpu_threads flags overrides the config
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/GoogleCloudPlatform/PerfKitBenchmarker/tests/background_cpu_test.py", line 129, in testBackgroundWorkloadVanillaConfigFlag
    self._CheckVMFromSpec(spec, 2)
  File "/home/travis/build/GoogleCloudPlatform/PerfKitBenchmarker/tests/background_cpu_test.py", line 81, in _CheckVMFromSpec
    expected_remote_post_stop)
AssertionError: 3 != 4
    '3 != 4' = '%s != %s' % (safe_repr(3), safe_repr(4))
    '3 != 4' = self._formatMessage('3 != 4', '3 != 4')
>>  raise self.failureException('3 != 4')
```

background_cpu_test patches out commands like vm.Install or vm.RemoteCommand, then calls BenchmarkSpec.Prepare, which is expected to call those VM functions, and finally checks how many times the patched functions were called.

The problem is that the BenchmarkSpec commands call the patched functions in a different thread per VM: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/c5e2c395f89cf4e54993b82fa8674048c3772d2e/perfkitbenchmarker/benchmark_spec.py#L233

The default MagicMocks used to patch the functions do not appear to be threadsafe. Example:
```
import sys
import threading

import mock


_CALLS_PER_THREAD = 100
_THREAD_COUNT = 1000
_EXPECTED_CALL_COUNT = _CALLS_PER_THREAD * _THREAD_COUNT


def Hello():
  pass


def CallHello():
  for _ in range(_CALLS_PER_THREAD):
    Hello()


module = sys.modules[__name__]
with mock.patch.object(module, 'Hello'):
  threads = []
  for _ in range(_THREAD_COUNT):
    t = threading.Thread(target=CallHello)
    threads.append(t)
  for t in threads:
    t.start()
  for t in threads:
    t.join()
  print 'call_count is {0} (expected {1})'.format(
      module.Hello.call_count, _EXPECTED_CALL_COUNT)
```

I suspect background_network_workload_test.py may also be vulnerable as well, but I haven't encountered a failure on it yet.

PR #954, opened 2016-04-15
Change bg tests to patch and verify call counts for each VM separately.

skschneider:
MagicMock.call_count is unreliable in multithreaded environments.
Rather than patching the class's methods, patch each VM's methods.
Addresses #948.

### DIFF
--- a/tests/background_cpu_test.py
+++ b/tests/background_cpu_test.py
@@ -14,8 +14,10 @@
 
 """Tests for background cpu workload """
 
+import itertools
 import unittest
 
+import contextlib2
 from mock import patch
 
 from perfkitbenchmarker import benchmark_spec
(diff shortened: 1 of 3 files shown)


skschneider:
@hildrum PTAL

tests/background_cpu_test.py
```diff
@@ -61,24 +67,46 @@ def _CreateBenchmarkSpec(self, benchmark_config_yaml):
...
+                                                     vm_expected_call_counts)
+      for function_name, expected_call_count in iter_mocked_functions:
+        call_count = getattr(vm, function_name).call_count
+        self.assertEqual(call_count, expected_call_count, msg=(
+            'Expected {0} from VM group "{1}" to be called {2} times, but it '
+            'was called {3} times.'.format(function_name, group_name,
+                                           expected_call_count, call_count)))
+
+  def _CheckVMFromSpec(self, spec, working_groups=(), non_working_groups=()):
+    with contextlib2.ExitStack() as stack:
```
hildrum: Maybe a comment here?

hildrum:
LGTM.

tests/background_cpu_test.py
skschneider: ExitStack is a standard python 3 function: https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack
contextlib2 is a python package that backports it to python 2.

Issue #521, opened 2015-10-09
Adding better error handling for mismatched CLI versions with PKB


voellm:
A couple of users have been hit by unmatched PKB and CLI.

Here is what I propose we do:

have PKB check the version of the CLI and error+exit if it has not been verified to work.
have a flag to turn the error+exit into a warning
The error+exit message should include the following:

A clear message on why the benchmarker exited. No exception stacks as the message.
A message on how to get the CLI version (eg: /npm install azure-cli#0.9.3 -g)
A message on how to override the error+exit using the flag "--ignore_cli_version" for example.
This should be a quick feature... and marking it P0 because it has been a pain point more than once.

(related issue - https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/469)

skschneider:
#958 and #967 address this.

PR #958, opened 2016-04-18
Fix Azure CLI version check.

skschneider:
Related to #521

Previous behavior:
- Upon PKB initialization, providers packages were imported recursively
  (including `providers/azure/__init__.py`, its dependency `util.py`, and
  potentially other subpackages like `providers/azure/subpkg/__init__.py`)
- `CheckAzureVersion` was registered as a handler for the event
  `initialization_complete`, which was sent prior to dynamically loading
  providers, so `CheckAzureVersion` was never called.
- Missing Azure CLI error was suppressed.

New behavior:
- Upon PKB initialization, only `providers/azure/__init__.py` and
  `providers/azure/flags.py` are imported.
- `CheckAzureVersion` is registered as a handler for the new event,
  `provider_imported`, which is called at the end of `LoadProvider`.
- Missing Azure CLI error triggers a graceful exit.

### DIFF
--- a/perfkitbenchmarker/events.py
+++ b/perfkitbenchmarker/events.py
@@ -33,6 +33,11 @@ parsed, temporary directory initialized, run_uri set).
 Sender: None
 Payload: parsed_flags, the parsed FLAGS object.""")
 
+provider_imported = _events.signal('provider-imported', doc="""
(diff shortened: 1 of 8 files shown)


skschneider:
@ehankland PTAL

ehankland:
LGTM

PR #959, opened 2016-04-18
Only check python package requirements if requirements.txt exists.

skschneider:
Most users download PKB in its entirety, including requirements.txt, so
the default behavior will still be to verify package versions.


hildrum:
LGTM

PR #960, opened 2016-04-19
Move pint unit registry code into a separate module.

skschneider:


skschneider:
@noahl PTAL

noahl:
The code LGTM, but I'm worried that the renaming makes it harder to use this module. How about putting `units = pint_registry.UnitRegistry()` into `__init__.py`, and then other modules can do `perfkitbenchmarker.units.parse_expression` (or whatever method)? Or alternatively, directly export the registry methods that we use from the `pint_registry` module, so other modules can do `pint_registry.parse_expression()`? This could also be combined with renaming `pint_registry` to `units`, since its function would be expanded from just holding the registry.

I think I slightly prefer the second of my two options above, but I'm curious what you think of all of them.

skschneider:
The second one sounds preferable to me.

skschneider:
PTAL

noahl:
LGTM. This makes using units much easier - thank you.

PR #962, opened 2016-04-20
Don't request SSH verbose output when --log_level=debug

noahl:
PKB's current behavior is to request verbose SSH output (via "ssh -v") whenever --log_level=debug. However, verbose SSH output is *very* verbose. And the problem is frequently not SSH, meaning you have to dig through lots and lots of SSH logs looking for the rare bit of actual PKB debug output to investigate a problem.

This PR adds a --verbose_ssh flag that enables verbose SSH output, and disables
verbose SSH output when --log_level=debug but the verbose_ssh flag is
not set.

I tested by running the fio benchmark with --log_level=debug with and without --verbose_ssh. The output was as expected.

### DIFF
--- a/perfkitbenchmarker/vm_util.py
+++ b/perfkitbenchmarker/vm_util.py
@@ -75,8 +75,6 @@ flags.DEFINE_integer('background_network_mbits_per_sec', None,
                      'Number of megabits per second of background '
                      'network traffic to generate during the run phase '
                      'of the benchmark')
-flags.DEFINE_boolean('verbose_ssh', False,
-                     'Request verbose output from SSH commands run by PKB.')
 
 
 class IpAddressSubset(object):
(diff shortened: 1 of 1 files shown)


noahl:
@skschneider @ehankland @yuyantingzero please take a look.

skschneider:
It sounds reasonable to decouple the ssh output (which is present both in stderr and the log file) from the `--log_level` flag, which is tied with values from the python logging system and controls only stderr output. But I see that the `--ssh_options` flag could also be used to enable verbose ssh output, so I suggest omitting the new flag.

noahl:
@skschneider good point. I pushed a new commit without the `--verbose_ssh` flag. PTAL.

skschneider:
LGTM. Make sure to modify the title when merging.

noahl:
@skschneider I changed the title and I'm merging. Thanks.

PR #963, opened 2016-04-20
Add a helpful error message if a UnitsParser parses a unitless value.

skschneider:


skschneider:
@noahl PTAL

Previous behavior was an AttributeError and a stacktrace to `quantity.to(self.convertible_to)`

noahl:
@skschneider I've been going back and forth on whether it should be valid to parse a dimensionless quantity if `convertible_to` is not specified. I think the right answer is that it doesn't make any sense to have a units flag without a convertible_to unit (or possibly a list of them). LGTM to your code, and I will submit a PR soon to require convertible_to.

PR #964, opened 2016-04-20
Add percent (%) as a recognized pint unit.

skschneider:


skschneider:
@noahl PTAL

noahl:
LGTM. This looks great. One question - can I define a units flag with `convertible_to=units.percent`? That would be very useful.

skschneider:
Good question. The answer is yes, and I've added a test case for it.

noahl:
LGTM

PR #965, opened 2016-04-20
Reorder some PKB initialization steps.

skschneider:
Cleans up some of the extraneous information that is printed when it is not needed.
For example, there is no reason to print warnings about other flags if --version is set.


hildrum:
LGTM

PR #967, opened 2016-04-20
Improve Azure CLI version check.

skschneider:
Add or expand error messages for when `azure -v` fails or returns an
incorrect version. Exit PKB unless the version is correct or the
user specifies --azure_ignore_cli_version.

Addresses #521.


ehankland:
LGTM

PR #968, opened 2016-04-20
Modify semantics of UnitsParser's convertible_to parameter.

skschneider:
Make it required.
Allow strings.
Allow multiple unit types.


skschneider:
@noahl PTAL

noahl:
LGTM. Thank you!

PR #969, opened 2016-04-21
Move to gflags version 3.0.4

skschneider:


requirements.txt
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-python-gflags==2.0
+python-gflags==3.0.4
```
noahl: Can we make it >= 3.0.4 instead of ==?

noahl:
One minor comment, otherwise LGTM

requirements.txt
skschneider: Discussed offline. We've had a previous issue where an updated version of gflags broke PKB: https://github.com/google/python-gflags/issues/23

Issue #797, opened 2016-01-13
Move to pep8 1.7


skschneider:
The current version of pep8 is 1.7.0. By default, it seems to be more picky about style than 1.5.7 (the version we use in tox when running our hooks). The changelog can be found at https://pypi.python.org/pypi/pep8

PR #970, opened 2016-04-22
Add CloudSuite Data-caching benchmark.

neo-apz:


perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+
+from perfkitbenchmarker import configs
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import sample
+from perfkitbenchmarker.linux_packages import docker
+
+flags.DEFINE_string('cloudsuite_data_caching_memcached_flags',
+                    '-t 1 -m 2048 -n 550',
+                    'Flags to be given to memcached.')
+flags.DEFINE_string('cloudsuite_data_caching_rps',
```
skschneider: I see this is cast to an integer below. How about using DEFINE_integer?

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+  vms = benchmark_spec.vms
+  server_vm = benchmark_spec.vm_groups['server'][0]
+  client_vm = benchmark_spec.vm_groups['client'][0]
+
+# Make sure docker is installed on all VMs.
+  for vm in vms:
```
skschneider: How about `for vm in (server_vm, client_vm)`? Then `vms` above isn't needed.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+
+  Returns:
+    A list of sample.Sample objects.
+  """
+
+  client_vm = benchmark_spec.vm_groups['client'][0]
+
+  benchmark_cmd = ('sudo docker run --rm --name dc-client --net host'
+                   ' cloudsuite/data-caching:client -rps %d' %
+                   int(FLAGS.cloudsuite_data_caching_rps))
```
skschneider: If the flag is defined as an integer, this cast shouldn't be necessary.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+                          'cloudsuite/data-caching:server %s' %
+                          FLAGS.cloudsuite_data_caching_memcached_flags)
+
+# Prepare the client.
+  client_vm.RemoteCommand('sudo docker pull cloudsuite/data-caching:client')
+  client_vm.RemoteCommand("echo '%s    dc-server' | sudo tee -a /etc/hosts >"
+                          " /dev/null" % server_vm.internal_ip)
+
+
+def _ParseOutput(output_str):
```
skschneider: Could this function be redone to provide more insight into what's going on? I'm worried it might be a bit flaky and difficult to debug. From the stdout I got, it looks like dc-server periodically takes measurements, and then the `numbers` assignment throws away everything except the last four lines. For reference, here's an excerpt from my output:
```
Configuration:

nProcessors on system: 8
nWorkers: 4
runtime: 123
Get fraction: 0.800000
Naggle's algorithm: False


host: dc-server
address: 10.0.0.2
Loading key value file...created uniform distribution 1000
rps 18000 cpus 4
meanInterarrival 222
num_worker_connections 50
num_worker_connections 50
Creating worker on tid 3852093184
starting receive base loop
num_worker_connections 50
Creating worker on tid 3843700480
starting receive base loop
num_worker_connections 50
Creating worker on tid 3835307776
starting receive base loop
Created 200 connections total
Creating worker on tid 3623876352
starting receive base loop
Stats:
-------------------------
  timeDiff,     rps,        requests,     gets,       sets,      hits,       misses,   avg_lat,      90th,      95th,        99th,       std,       min,        max,    avgGetSize
1461365108.363647,       0.0,       36429,      29235,       7194,      29235,          0,   0.218154,   0.264100,   0.296100,   0.389100,   0.044939,   0.000000,   1.110000, 840.211869
Outstanding requests per worker:
2 1 0 2 
  timeDiff,     rps,        requests,     gets,       sets,      hits,       misses,   avg_lat,      90th,      95th,        99th,       std,       min,        max,    avgGetSize
120.000120,   18074.9,     2168993,    1734696,     434297,    1734696,          0,   0.237822,   0.295100,   0.335100,   0.497100,   0.096340,   0.111000,  15.449000, 843.741086
Outstanding requests per worker:
1 2 1 0 
  timeDiff,     rps,        requests,     gets,       sets,      hits,       misses,   avg_lat,      90th,      95th,        99th,       std,       min,        max,    avgGetSize
120.000120,   18085.9,     2170316,    1736253,     434063,    1736253,          0,   0.240551,   0.297100,   0.339100,   0.516000,   0.149249,   0.111000,  35.241000, 843.748133
Outstanding requests per worker:
2 0 0 1 
Ran for 242.000242, exiting
```
That makes my `numbers` equal `[120.00012, 18085.9, 2170316.0, 1736253.0, 434063.0, 1736253.0, 0.0, 0.240551, 0.2971, 0.3391, 0.516, 0.149249, 0.111, 35.241, 843.748133, 2.0, 0.0, 0.0, 1.0, 242.000242]`.

Rather than using indexes and a flat array, could you parse this table to associate each value with the name of the column, and then extract the value by column name? I think that would be clearer. With the current single-regex methods, I'm worried about nan, since I see that in a previous table in my log output. I also think the outstanding requests per worker could be done as a separate step, since it doesn't fit the format of the rest of the chart.

Adding a parsing test case under tests/linux_benchmarks/cloudsuite_data_caching_benchmark_test.py would also be helpful, since it would give an example of what the output should look like.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+                                 numbers[8], "ms"))
+    results.append(sample.Sample("95th percentile latency",
+                                 numbers[9], "ms"))
+    results.append(sample.Sample("99th percentile latency",
+                                 numbers[10], "ms"))
+
+    req_rems = numbers[15:-1]
+
+    results.append(sample.Sample("Average outstanding requests per requester",
+                                 float(sum(req_rems)) / len(req_rems), "reqs"))
```
skschneider: The sum of floats should already be a float.

perfkitbenchmarker/benchmark_sets.py
```diff
@@ -190,7 +191,7 @@ def GetBenchmarksFromFlags():
     else:
       benchmark_names.add(benchmark)
 
-  # Expand recursive sets
+# Expand recursive sets
```
skschneider: Please align comments to match the rest of the code (applies to the other file as well).
```
# Comment
x = 1
if condition:
  # Comment
  x = 2
```

skschneider:
Thanks for the pull request. I've made a pass and added comments. Let us know if you have any questions.

perfkitbenchmarker/benchmark_sets.py
neo-apz: It was like that, but `flake8` gave me an error that `indentation is not a multiple of four (comment)`. So I changed it to this. What should I do here?

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
neo-apz: I fixed it.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
neo-apz: I fixed it.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
neo-apz: I fixed it.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
neo-apz: Yes, it's not really clean now, but it works without any problem. The program works for 4 minutes and every two minutes it's going to log those numbers. I only report the latest values because it needs a little bit of time to warm up. So used that regular expression to get the latest numbers and then I add them to the result set. For now it's fine and works without any problem. In future I plan to modify the code and give the user the ability to set the running time and other parameters. At that time I should and will do something like you said to parse the whole table and based on the parameters figure out what to put in the result set.

perfkitbenchmarker/benchmark_sets.py
skschneider: Are you manually running flake8? If you've installed the hooks in this repository, they should be running it through tox, which is configured to use a specific version of flake8: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/c4289c6035f95f6e3964273ebf43dd1788a25a9a/tox.ini#L27
We have an open issue to make changes to support newer versions of pep8 (#797), but for now `tox -r -e flake8` should re-download required packages at the specified versions and execute flake8 with them.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
skschneider: My biggest concern here is that if anything changes in the output format, this method won't complain. It will just silently use the wrong numbers, and the users may not notice. If you want to check this method in as-is, please take other precautions to make the benchmark as stable as possible, for example by pulling a specific tag/digest in the Prepare docker commands.

perfkitbenchmarker/benchmark_sets.py
neo-apz: As far as I understand it doesn't use the `flake8` version installed within the `tox` folder. The pre-commit hook executes `check` which then calls `check-lint.sh`. In that script, at first it complains if you don't have `flake8` installed, then it uses the `flake8` version installed on the system (`/usr/bin`), not the one within the `tox` folder. Am I missing something here?

skschneider: Ah, you're right. That's weird, and I think we should change that to tox - it doesn't look like the current implementation actually checks the installed version. To get around it, you can temporarily disable that hook, or you can install the flake8 and pep8 versions specified in the tox.ini file:
```
pip install pep8==1.5.7
pip install flake8==2.4.1
```
If you're not working in a virtualenv, you might need sudo for those commands.

neo-apz: OK, fixed.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
neo-apz: Yeah I see. But for now there is no concern because I'm developing the benchmark itself. I created the docker images myself, and I'm pulling those. So no worries. :)

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,143 @@
...
+    client:
+      vm_spec: *default_single_core
+"""
+
+
+def GetConfig(user_config):
+  return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+
+
+def Prepare(benchmark_spec):
```
skschneider: The three line comments in this function still have odd spacing.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,143 @@
...
+  """Run the data-caching benchmark.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+
+  Returns:
+    A list of sample.Sample objects.
+  """
+
```
skschneider: Nit: blank line unnecessary

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_caching_benchmark.py
```diff
@@ -0,0 +1,143 @@
...
+
+def Cleanup(benchmark_spec):
+  """Stop and remove docker containers. Remove images.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+
```
skschneider: Nit: blank line unnecessary

skschneider:
Took another pass. Just small style comments now (spacing and blank lines). Let me know when you have fixed them and are ready for us to merge this in.

neo-apz:
OK. I have fixed them all.

skschneider:
Merged.

PR #973, opened 2016-04-26
Add back support for python-gflags version 2.

skschneider:
Keep support for the older version until we have a full release that
supports the new one.


ehankland:
LGTM

PR #975, opened 2016-04-26
Remove duplicate requirements files

noahl:
Provider-specific requirements files were moved to provider-specific
directories, but there were still copies in the top-level
directory. Remove those copies.


noahl:
@skschneider please take a look.

noahl:
@hildrum PTAL.

skschneider:
Conceptually LGTM, but I suggest not merging this until after the v1.5 release, and then have it go into effect in v1.6.

noahl:
@skschneider sounds good to me. I'll wait on this one.

hildrum:
LGTM.  I agree with waiting until after v1.5, so I set it to a v1.6.0 milestone so we don't forget it.

hildrum:
Should we pull this in now?

noahl:
Yes, I think we should go ahead and pull it in. Thanks for remembering this.

PR #984, opened 2016-05-03
Support S3 Signature Version 4

noahl:
Supply a '--region' flag to 'aws s3 rm' and 'aws s3 rb', which make it
use signature version 4, which is required for certain S3 regions.


noahl:
@yuyantingzero @stfeng @skschneider @ehankland please take a look.

skschneider:
LGTM

PR #987, opened 2016-05-04
Factor object_storage_api_tests.py with separate backend objects

noahl:
Move the code for each object storage provider in
object_storage_api_tests.py to a separate per-provider "backend" object,
and update the benchmarks to work with this system. Add a new
TestBackend scenario for testing these backends (clearly marked as for
debugging only in the docstring). Also take the opportunity to improve
testing of the benchmarks, by implementing a mock backend and adding
some integration tests that use it.

### DIFF
--- a/perfkitbenchmarker/scripts/object_storage_api_tests.py
+++ b/perfkitbenchmarker/scripts/object_storage_api_tests.py
@@ -279,6 +279,8 @@ class ObjectStorageBackend(object):
     # an exception if the flags are invalid for its particular
     # provider.
 
+    pass
+
 
   @abc.abstractmethod
   def ListObjects(self, bucket, prefix):
--- a/perfkitbenchmarker/scripts/object_storage_api_tests.py
+++ b/perfkitbenchmarker/scripts/object_storage_api_tests.py
@@ -22,6 +22,7 @@
    run this script.
 """
 
+import abc
 import cStringIO
 import json
 import logging
@@ -64,13 +65,15 @@ flags.DEFINE_string('azure_key', None,
 flags.DEFINE_enum(
     'scenario', 'OneByteRW', ['OneByteRW', 'ListConsistency',
                               'SingleStreamThroughput', 'CleanupBucket',
-                              'MultiStreamWrite', 'MultiStreamRead'],
+                              'MultiStreamWrite', 'MultiStreamRead',
+                              'TestBackend'],
     'The various scenarios to run. OneByteRW: read and write of single byte. '
     'ListConsistency: List-after-write and list-after-update consistency. '
     'SingleStreamThroughput: Throughput of single stream large object RW. '
     'CleanupBucket: Cleans up everything in a given bucket.'
     'MultiStreamWrite: Write objects with many streams at once.'
-    'MultiStreamRead: Read objects with many streams at once.')
+    'MultiStreamRead: Read objects with many streams at once.'
+    'TestBackend: Check that a backend works. For debugging only.')
 
 flags.DEFINE_integer('iterations', 1, 'The number of iterations to run for the '
                      'particular test scenario. Currently only applicable to '
@@ -169,21 +172,7 @@ class LowAvailabilityError(Exception):
     pass
 
 
-def _useBotoApi(storage_schema):
-  """Decides if the caller should use the boto api given a storage schema.
-
-  Args:
-    storage_schema: The schema represents the storage provider.
-
-  Returns:
-    A boolean indicates whether or not to use boto API.
-  """
-
-  if storage_schema == 'gs' or storage_schema == 's3':
-    return True
-  else:
-    return False
-
+# ### Utilities for workload generation ###
 
 class SizeDistributionIterator(object):
   """Draw object sizes from a distribution.
@@ -245,6 +234,8 @@ def MaxSizeInDistribution(dist):
   return max(dist.iterkeys())
 
 
+# ### Utilities for data analysis ###
+
 def PercentileCalculator(numbers):
   numbers_sorted = sorted(numbers)
   count = len(numbers_sorted)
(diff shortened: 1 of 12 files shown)


noahl:
I have verified that all three backends (GCS, S3, and Azure) pass the TestBackend test, and that the GCS backend works with all of our benchmark scenarios (OneByteRW, SingleStreamThroughput, ListConsistency, MultiStreamWrite, MultiStreamRead, and CleanupBucket). I have also verified that all scenarios except MultiStreamRead pass the new integration tests with the mock backend that I added. (There's a comment explaining why I couldn't test MultiStreamRead that way.)

I also wanted to mention that the new tests don't run as part of normal unit testing because they take around 6 minutes to complete.

I tried to keep this PR as small as possible, but it's still pretty large because I had to update almost all of the API test script.

I think everyone agrees that we would like to do more refactoring on this benchmark eventually, but I hope we also agree that any refactor would include something like this splitting into backends. I think even in the context of wanting to do a bigger refactor, it makes sense to go piece by piece, which is why I made this PR with exactly one refactoring change.

stfeng:
Please test all 3 providers with real environment. 

Re multistream read scenario, you can still mock it. Just consider and read object request valid in your mock.

noahl:
@stfeng I've started tests running and I'll update when they're complete.

The problem with mocking the MultiStreamRead benchmark is that it unconditionally tries to read from the filesystem, so I can't do a completely in-Python mock. However, I think I have a workaround.

noahl:
@stfeng I have verified that the `api_data`, `api_namespace`, and `api_multistream` scenarios run successfully on AWS, Azure, and GCP. I also pushed a small commit that adds MultiStreamReads to the integration tests, which means that all benchmark scenarios are covered. Please take a look.

stfeng:
Ack. I will look at this in the next few days.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -271,81 +262,238 @@ def PercentileCalculator(numbers):
...
+
+class AzureBackend(ObjectStorageBackend):
+  def __init__(self):
+    if FLAGS.azure_key is None or FLAGS.azure_account is None:
+      raise ValueError('Must specify azure account and key.')
+    self.blobService = BlobService(FLAGS.azure_account, FLAGS.azure_key)
+
+  def ListObjects(self, bucket, prefix):
+    return [obj.name
+            for obj in self.blobService.list_blobs(bucket, prefix=prefix)]
```
stfeng: Is this indentation right?

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -1200,13 +1237,49 @@ def ListConsistencyBenchmark(storage_schema, host_to_connect=None):
...
   # Final clean up: delete the objects still remaining.
-  DeleteObjectsConcurrently(storage_schema,
+  DeleteObjectsConcurrently(backend,
                             per_thread_objects_written,
-                            host_to_connect,
                             per_thread_objects_deleted)
   return final_result
 
 
+def TestBackend(backend):
```
stfeng: Move the test backend out to a separate file.

perfkitbenchmarker/scripts/object_storage_api_tests.py
```diff
@@ -271,81 +262,238 @@ def PercentileCalculator(numbers):
...
-def _ListObjects(storage_schema, bucket, prefix, host_to_connect=None):
-  """List objects under a bucket given a prefix.
+# ### Backends ###
 
-  Args:
-    storage_schema: The address schema identifying a storage. e.g., "gs"
-    bucket: Name of the bucket.
-    prefix: A prefix to list from.
-    host_to_connect: An optional endpoint string to connect to.
+class ObjectStorageBackend(object):
```
stfeng: Should we move this class and the backend implementation classes to their own files and probably under the one directory?

If we do that, we do need to change the high level benchmarking script in PKB to properly provision the test script on a VM (right now it just copies one file over).

stfeng:
One high level comment - 

Should we rename "backend" into "provider", or "service"? e.g., ObjectStorageService() rather than ObjectStorageBackend(). And we can have "S3ObjectStorageService", "GCSObjectStorageService", etc. 

I know it's going to be a bit painful to rename all these, but getting the naming right is important. I think "Backend" is usually used in the context of a two/multi-tier service to refer to the second layer. In this case, what we have is a test client, and that client talks to a service.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: I believe it matches the style guide, although I don't know if it's "right" :) What indentation would you expect here?

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: This is actually a poorly-named benchmark, not a backend implementation. I moved it to the script integration tests directory, and hopefully also improved the name.

noahl: I've been working on it more, and I'm not sure that moving it is the right choice. It is definitely a test, but it's a test that needs the flags `azure_account`, `azure_key`, `bucket`, and `storage_provider` that are normally passed to `object_storage_api_tests.py`. I can make it into three separate unit tests, one for each provider, but the user will still have to provide the name of a test bucket they created for each service and their Azure account and key. I think it might be simpler to leave it in `object_storage_api_tests.py` for now. If we can automate the bucket setup, then I think it would make more sense to move it to the tests directory. What do you think?

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: Eventually, I'd like to move the backend implementation classes to the provider directories, and this one to its own separate file somewhere. How does that sound?

The reason I didn't do it in this script is exactly what you said, that it would require changing how the object storage benchmarking script provisions scripts, but I'm happy to make that part of this PR.

noahl:
@stfeng You're right, "Backend" is a bad name. I like "Service" the best out of your two names. The only thing I'm unsure about is that I also want to factor the provider-specific parts of the high-level benchmarking script into their own objects. Maybe these could be called "DataService" and the high-level ones could be called "ControlService", since this is roughly the data plane and that would be the control plane? Or it might make sense to make them the same and import the service objects from two places. I won't know for sure until I write the code.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: It seems to me that for a test method like this, it shouldn't depend on actual bucket or actual provider. It should be runnable in a tox environment as a unit-test. 

Can you make a fake backend, and call this test using a fake backend? Then you can move this method out of this file to the test directory.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: I'd put this script and the provider specific scripts into one single directory here, then the high level driver script only needs to copy everything under that directory to the target VM, that's it. 

I'm fine with doing it in a separate CR, but let's make sure we put a TODO here, and also file an issue to track this.

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: SG.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: This code is not for testing the benchmark, though. It's how I tested that I implemented the service provider backends correctly. So it does have to touch a real bucket, because I want to make sure that each backend can write to a bucket and then read its objects back.

I used it for manually testing all of the providers. I created sample buckets in the provider UIs, then did something like `object_storage_api_tests.py --storage_provider=GCS --bucket=... --scenario=TestBackend`, and if it printed the right thing, I knew the backend code was okay.

stfeng:
Another idea: ObjectStorageApiProvider

perfkitbenchmarker/scripts/object_storage_api_tests.py
stfeng: In this case, should it be part of the backend base class? e.g., a set of basic test that any backend providers must pass, so it naturally belongs to the base class. Also, all parameters/flags that are provider specific should be passed in as parameters to this method so it can be moved out of this file. Basically, the point I want to touch on is: this method is specific to "backend" implementation, it should not be part of the benchmarking logic. It should be encapsulated to the "backend" class hierarchy.

Also, consider rename it to ValidateBackend() when you do that.

<"backend" really needs to have a new name! ;)>

noahl:
I renamed `ObjectStorageBackend` to `ObjectStorageApiProvider`. The provider-specific classes are `AzureProvider`, `S3Provider`, and `GCSProvider`. Is that better? :)

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: I see why putting them under one directory makes it easier, but my ideal vision is that the benchmark doesn't have to copy over all of the provider scripts, only the one that's actually going to be used. Here's my proposed final organization, after this and 1-2 more PRs:

  - Each provider directory has a `scripts/` subdirectory, containing their `ObjectStorageApiProvider`.
  - Each provider directory also has a new resource called `ObjectStorageService`, which encapsulates the CLI commands that live in the overall benchmarking script right now. It also does two more things: provides a list of packages that need to be installed on a VM to use that provider's object storage API (right now I think we spend a lot of our overall provisioning time installing Azure's client library on VMs where we're not going to use it), and provides a list of files that need to be copied to a VM to provide the `ObjectStorageApiProvider` object.

Then the benchmarking script can install only the packages that are actually needed, and copy over only the provider object that is actually needed. How does this end-state sound to you?

noahl: However, I am happy to move everything to one directory right now. Until we have the `ObjectStorageService` objects, it would be much easier that way.

perfkitbenchmarker/scripts/object_storage_api_tests.py
noahl: I changed to `ValidateProvider`, since we are now calling backends providers. I don't think it should be part of the provider class hierarchy, because it's not needed to use a provider. But it's also not part of the benchmarker, so I moved it to its own script inside the `script-tests/integration` directory. How does that sound?

noahl:
@stfeng I made the renames we talked about in person. All tests still pass, I can run a benchmark, and I am able to validate the GCS provider. Please take a look.

stfeng:
LGTM. Please make sure you test all providers and all major api test scenarios *end-to-end* before merge. 

The Travis CI build starts to have issues again, obviously that needs to be fixed before merge.

noahl:
I'll test everything before merging, and make sure Travis works again.

It looks like the same package issue we've seen before.

noahl:
I tested all scenarios on all providers, and everything passed.

I also discovered that I can make Travis retry by pushing and then reverting a dummy commit. The issue was the same thing we've seen before - temporary availability problem of package downloads, which Travis considers a failure.

I'll merge this now.

PR #999, opened 2016-05-24
Use posixpath in MongoDB YCSB benchmark

noahl:
The benchmark incorrectly uses os.path when it should use posixpath. Fix
this bug.

In addition to the unit tests, I have verified that I can actually run a MongoDB benchmark with this patch.


noahl:
@ehankland @hildrum PTAL

hildrum:
LGTM.

PR #1006, opened 2016-06-07
Fix path to `cassandra-cli` when running benchmark from upstream repo

AdamIsrael:
This fixes a bug I introduced in a previous PR, which would cause the cassandra benchmarks to fail when run from non-Juju os_types.


hildrum:
LGTM.

hildrum:
Should I go ahead and merge?

AdamIsrael:
Maybe @ehankland can take a look. He helped with the initial pull request where I introduced the bug.

ehankland:
LGTM thanks for fixing this

PR #1014, opened 2016-06-10
Add support to AWS and GCP to get standard out

hildrum:
Also modified spark benchmark to demonstrate the features.  

@ehankland, PTAL.  @pmkc, you may also be interested.


hildrum:
Currently adding the system-reported time the operations take, so you might want to wait on your views until that change is in.

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
```diff
@@ -65,13 +67,37 @@ def _Exists(self):
...
     stdout, stderr, retcode = cmd.Issue()
-    return retcode == 0
+    logging.debug('STDOUT: {0}\nSTDERR: {1}\n'.format(stdout, stderr))
+    if retcode != 0:
+      return False
+    if job_stdout_file:
+      with open(job_stdout_file, 'w') as f:
+        # dataproc prints progress lines that end with a cr and so
+        # overwrite themselves.  We need to move past that.
+        last_cr_index = stderr.rfind('\r')
```
pmkc: I assume PKB doesn't have to run on windows without cygwin?

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
```diff
@@ -65,13 +67,37 @@ def _Exists(self):
     stdout, stderr, retcode = cmd.Issue()
     return retcode == 0
 
-  def SubmitJob(self, jarfile, classname, job_poll_interval=None):
+  def SubmitJob(self, jarfile, classname, job_poll_interval=None,
+                job_arguments=None, job_stdout_file=None):
     cmd = util.GcloudCommand(self, 'dataproc', 'jobs', 'submit', 'spark')
     cmd.flags['cluster'] = self.cluster_id
     cmd.flags['jar'] = jarfile
```
pmkc: Technically we suggest using --jars with --class, because --jar implies you are using the main method. This works fine mod an angry stderr message.

pmkc:
SGTM

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
hildrum: The spark benchmark as currently written can't be run on Windows at all.  But I could imagine someone writing a windows-based benchmark, and it'd be nice if I could make this not break.

Do you think it'd be more stable to eliminate lines based on a regular expression of what dataproc currently outputs?

pmkc: regex would probably be better. You could also split on os.linesep or something. I'm not asking for Windows client support, I'm just pointing out that this probably won't work if you want it.

hildrum:
Output of the benchmark now includes the wall_time, the run time, and the pending time, where the runtime and pending time as reported by the platform.
```
SPARK:
  class="org.apache.spark.examples.SparkPi" jarfile="file:///usr/lib/spark/lib/spark-examples.jar" job_arguments="[]" machine_type="None" num_workers="4" print_stdout="True" spark_cluster_id="foo" spark_service="dataproc"
  wall_time                            29.074358 seconds                       
  runtime                              16.781000 seconds                       
  pending_time                          9.554000 seconds                       
  End to End Runtime                   29.079056 seconds
```

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
hildrum: I'm going to leave it as-is, unless there is a strong feeling it should be changed now.

perfkitbenchmarker/linux_benchmarks/spark_benchmark.py
```diff
@@ -74,23 +91,51 @@ def Run(benchmark_spec):
...
     raise Exception('Class {0} from jar {1} did not run'.format(
         FLAGS.spark_classname, FLAGS.spark_jarfile))
   jar_end = datetime.datetime.now()
-
-  metadata = spark_cluster.GetMetadata().update(
-      {'jarfile': FLAGS.spark_jarfile, 'class': FLAGS.spark_classname})
+  if stdout_path:
+    with open(stdout_path, 'r') as f:
+      logging.info('The output of the job is ' + f.read())
+    os.remove(stdout_path)
```
ehankland: Looks like there are cases where this file can be leaked - for instance if the job didn't complete successfully you raise an exception and skip executing this line.

hildrum:
@ehankland, PTAL.

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -61,19 +62,35 @@ def __init__(self, spark_service_spec):
       self.machine_type = DEFAULT_MACHINE_TYPE
     self.cmd_prefix = util.AWS_PREFIX + ['emr']
     if FLAGS.zones:
-      region = util.GetRegionFromZone(FLAGS.zones[0])
-      self.cmd_prefix += ['--region', region]
+      self.region = util.GetRegionFromZone(FLAGS.zones[0])
+    else:
+      cmd = ['aws', 'configure', 'get', 'region']
```
ehankland: If the default region is already set then you shouldn't need to explicitly set it. And if it isn't set, then "aws configure get region" doesn't return anything.

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -61,19 +62,35 @@ def __init__(self, spark_service_spec):
...
+      stdout, _, _ = vm_util.IssueCommand(cmd)
+      self.region = stdout.strip()
+    self.cmd_prefix += ['--region', self.region]
+    self.bucket_to_delete = None
+
+  def _CreateLogBucket(self):
+    bucket_name = 's3://pkb-' + FLAGS.run_uri + '-emr-' + self.region
```
ehankland: I think the style guide prefers a format or % in cases like these https://google.github.io/styleguide/pyguide.html?showone=Strings#Strings

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
ehankland: How likely is it that a spark job will contain '\r' in its stdout? It looks like this will break if it does.

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -112,8 +132,40 @@ def _WaitUntilReady(self):
...
+
+    tries = 0
+    while tries < 30:
+      _, _, rc = vm_util.IssueCommand(cmd)
+      if rc == 0:
+        return True
+      else:
+        time.sleep(poll_interval)
+        tries += 1
+    return False
```
ehankland: It doesn't look like you check whether this method returns True or False (in submit job). If you are fine raising an exception instead of returning True/False, then you could use the vm_util.Retry decorator instead of having your own polling logic

perfkitbenchmarker/providers/aws/aws_emr.py
hildrum: I use the region as part of the logs bucket name when the logs bucket is created, which is why I wanted to get the explicit region string.

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
hildrum: Okay, changed.  Now I parse the lines carefully to separate the dataproc lines from the stdout lines.

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -84,6 +99,10 @@ def _Delete(self):
     cmd = self.cmd_prefix + ['terminate-clusters', '--cluster-ids',
                              self.cluster_id]
     vm_util.IssueCommand(cmd)
+    if self.bucket_to_delete is not None:
```
ehankland: I don't think you need to explicitly check to make sure that bucket_to_delete is not none - I think you can just do
```
if self.bucket_to_delete:
```

perfkitbenchmarker/resource.py
```diff
@@ -144,13 +145,18 @@ def _DeleteResource(self):
 
   def Create(self):
     """Creates a resource and its dependencies."""
+    @vm_util.Retry()
```
ehankland: You may want to specify some retry parameters like poll time and fuzz

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -98,22 +117,45 @@ def _Exists(self):
...
+    if 'LogUri' in result['Cluster']:
+      self.logging_enabled = True
+      log_uri = result['Cluster']['LogUri']
+      if log_uri.startswith('s3n'):
+        log_uri = 's3' + log_uri[3:]
+      return log_uri
+    else:
+      return None
+
+  @vm_util.Retry()
```
ehankland: If you aren't going to specify parameters to Retry you might as well just use vm_util.IssueRetryableCommand

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -98,22 +117,45 @@ def _Exists(self):
...
+  def _WaitForFile(self, filename, poll_interval=10):
```
ehankland: You no longer use poll_interval here

perfkitbenchmarker/resource.py
```diff
@@ -142,15 +143,39 @@ def _DeleteResource(self):
...
+      ready_max_retires: Maximum number of _IsReady retries.
+      ready_timeout: Maximum time to wait for the object to be
+          ready.
+      ready_fuzz: The fuzz on the _IsReady checks.  Only useful
+          if the subclass implements _IsReady.  See vm_util.Retry
+          for details.
+      ready_log_errors: A boolean describing whether errors should
+          be logged for the _IsReady check.
+    """
+    @vm_util.Retry(poll_interval=ready_poll_interval,
```
ehankland: I think accepting optional retry parameters to Create is probably a good thing, but only applying them to the "waituntilready" and not "createresource" itself seems like it might create confusion to me. Also if we are going to allow this, we may also want these options available to Delete as well

ehankland: Ah I guess I didn't look closely enough at the argument names + docstring - that does make it pretty clear which retry args they are, but it still seems a bit weird to me

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -98,22 +116,94 @@ def _Exists(self):
...
+      self.logging_enabled = True
+      log_uri = result['Cluster']['LogUri']
+      if log_uri.startswith('s3n'):
+        log_uri = 's3' + log_uri[3:]
+      return log_uri
+    else:
+      return None
+
+
+  @vm_util.Retry()
```
ehankland: I'm not sure I understand why you need a retry here

hildrum: I don't, thanks.

perfkitbenchmarker/providers/aws/aws_emr.py
```diff
@@ -98,22 +116,94 @@ def _Exists(self):
...
-      result = json.loads(stdout)
-      if result['Cluster']['Status']['State'] == READY_STATE:
-        return True
-      time.sleep(READY_CHECK_SLEEP)
-    return False
-
-  def SubmitJob(self, jarfile, classname, job_poll_interval=JOB_WAIT_SLEEP):
+    stdout, _, rc = vm_util.IssueCommand(cmd)
+    result = json.loads(stdout)
+    if result['Cluster']['Status']['State'] == READY_STATE:
```
ehankland:
```
return result['Cluster']['Status']['State'] == READY_STATE
```

perfkitbenchmarker/resource.py
```diff
@@ -142,15 +142,30 @@ def _DeleteResource(self):
...
+      kwargs: Arguments to pass to retry for Create and
+          _IsReady checks.
+    """
+    @vm_util.Retry(**kwargs)
+    def WaitUntilReady():
+      if not self._IsReady():
+        raise Exception('Not yet ready')
+
+    @vm_util.Retry(retryable_exceptions=(
+        errors.Resource.RetryableDeletionError,), **kwargs)
```
ehankland: Should this be RetryableCreationError?

perfkitbenchmarker/resource.py
```diff
@@ -109,7 +110,7 @@ def _DeleteDependencies(self):
     """
     pass
 
-  @vm_util.Retry(retryable_exceptions=(errors.Resource.RetryableCreationError,))
+  @vm_util.Retry(retryable_exceptions=(errors.Resource.RetryableDeletionError,))
```
ehankland: Should be RetryableCreationError

perfkitbenchmarker/providers/gcp/gcp_dataproc.py
```diff
@@ -65,13 +95,49 @@ def _Exists(self):
...
+    # Its stderr contains a mix of the stderr of the job, and the
+    # stdout of the job.  We set the driver log level to FATAL
+    # to supress those messages, and we can then separate, hopefully
+    # the job standard out from the log messages.
+    cmd.flags['driver-log-levels'] = 'root=FATAL'
+    if job_arguments:
+      cmd.additional_flags = job_arguments
     stdout, stderr, retcode = cmd.Issue()
-    return retcode == 0
+    logging.debug('STDOUT: {0}\nSTDERR: {1}\n'.format(stdout, stderr))
```
ehankland: Does this not already get logged at the debug level automatically?

hildrum:
@ehankland, Added support for subnets on AWS, which are needed for m4 machine types.

ehankland:
LGTM

PR #1023, opened 2016-06-22
Object storage refactor

noahl:
This is a major refactor with a few different components:
  - Run object storage benchmarks from provider-independent driver
  functions which call into provider-specific callbacks, instead of
  having per-provider driver functions.

  - Move provider-specific object storage functionality into provider
    directories. This refactor changes the object storage functionality
    that runs in the PKB controller. A separate refactor already
    addressed the API test script.

  - Modify the API test script slightly to support loading only the
    requested provider instead of all of them. The changes to the
    controller created provider-specific Prepare functions, which let us
    prepare the worker VM for only the provider that we are actually
    going to use. This change to the API test script is to enable that
    simpler Prepare step, which is why it is part of this refactor and
    not the previous one of the API test script.

  - Move the different object storage scenarios into their own functions
    with a uniform interface, instead of grouping them into two
    functions with lots of if statements inside. This change is in this
    refactor because it was enabled by the provider-independent driver
    function, and because the change to provider-specific callbacks
    required substantial changes to the benchmarks anyway, which made
    this seem more practical.

I have tested that AWS, Azure and GCP can all run the CLI benchmark and one of the API benchmarks, and that all of the benchmarks will run on at least one provider. I can complete the test matrix, it will just take some time, and I wanted to start getting comments on this because I'm sure there will be lots of them.

@hildrum @yuyantingzero @ehankland @meteorfox please take a look. Carlos, please especially look at the Swift implementation because I don't have one to test on.


perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -242,11 +194,13 @@
 DEFAULT = 'default'
 
 # This accounts for the overhead of running RemoteCommand() on a VM.
-MULTISTREAM_DELAY_PER_VM = 5.0
+MULTISTREAM_DELAY_PER_VM = 10.0
```
yuyantingzero: Consider adding units in naming: MULTISTREAM_DELAY_PER_VM_IN_SEC

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+        self.test_script_path,
+        '--storage_provider=%s' % self.storage
+    ] + args + self.service.APIScriptArgs()
+    if FLAGS.object_storage_storage_class is not None:
+      cmd_parts += ['--object_storage_class',
+                    FLAGS.object_storage_storage_class]
+
+    return ' '.join(cmd_parts)
+
+# The benchmark functions below all have the same arguments:
```
yuyantingzero: It looks cleaner to have documents inline IMO. Jumping back and forth harms readability and pydoc probably won't work.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+  logging.info('Starting multi-stream write test.')
+
+  objects_written_file = posixpath.join(vm_util.VM_TMP_DIR,
+                                        OBJECTS_WRITTEN_FILE)
+
+  size_distribution = _DistributionToBackendFormat(
+      FLAGS.object_storage_object_sizes)
+
+  def StartMultiStreamProcess(cmd_args, proc_idx, out_array):
+    # vm_idx = proc_idx // PROCESSES_PER_VM
```
yuyantingzero: Remove commented code.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+    cmd = command_builder.BuildCommand(
+        cmd_args + ['--stream_num_start=%s' % proc_idx])
+    # out, _ = vms[vm_idx].RobustRemoteCommand(cmd, should_log=True)
```
yuyantingzero: Remove commented code.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -890,485 +792,37 @@ def _CliBasedTests(output_results, metadata, vm, iteration_count,
...
-    if FLAGS.object_storage_scenario in ('all', 'api_data',):
-      logging.warn('Skipping API-based scenario for object storage service. '
-                   'Reason: Not yet implemented for OpenStack Swift')
-
-    return results
+  # Prepare data on vm, create a run directory on scratch drive, and add
+  # permission.
+  scratch_dir = vm.GetScratchDir()
+  vm.RemoteCommand('sudo mkdir -p %s/run/' % scratch_dir)
+  vm.RemoteCommand('sudo chmod 777 %s/run/' % scratch_dir)
```
yuyantingzero: +rw probably is enough

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -890,485 +792,37 @@ def _CliBasedTests(output_results, metadata, vm, iteration_count,
...
 
-  def Cleanup(self, vm):
-    """Clean up OpenStack Swift container and uninstall packages on vm.
+  vm.RemoteCommand('sudo mkdir -p %s/run/temp/' % scratch_dir)
+  vm.RemoteCommand('sudo chmod 777 %s/run/temp/' % scratch_dir)
```
yuyantingzero: same here

perfkitbenchmarker/object_storage_service.py
```diff
@@ -0,0 +1,250 @@
...
+    """Get ready to use object storage.
+
+    This method should be called before any other method of this
+    class. Once it is called, all of this class' methods should work
+    with data in the given location.
+
+    Args:
+      location: where to place our data.
+    """
+
```
yuyantingzero: Remove empty lines between docstring and pass.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+        ['azure', 'storage', 'container',
+         'delete', '-q', bucket,
+         '-a', self.storage_account,
+         '-k', self.azure_key])
+
+  def EmptyBucket(self, bucket):
+    # Emptying buckets on Azure is hard. We pass for now - this will
+    # increase our use of storage space, but should not affect the
+    # benchmark results.
+    pass
```
yuyantingzero: could you delete bucket with blobs inside? if not, I prefer cleanup the bucket at the end of benchmark.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+
+  _, raw_result = vm.RemoteCommand(single_stream_throughput_cmd)
+  logging.info('SingleStreamThroughput raw result is %s', raw_result)
+
+  for up_and_down in ['upload', 'download']:
+    search_string = 'Single stream %s throughput in Bps: (.*)' % up_and_down
+    result_string = re.findall(search_string, raw_result)
+    sample_name = SINGLE_STREAM_THROUGHPUT % up_and_down
+
+    if len(result_string) > 0:
```
meteorfox: style nitpick: why not check first for the error condition  and raise the error immediately? That way you don't have to indent the for-loop into the if block.

I believe `re.findall()` will either return an empty list or non-empty list, you can use Python's coercion to booleans False and True, respectively. (e.g. `if not result_string:`)

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+      '--scenario=ListConsistency'])
+
+  _, raw_result = vm.RemoteCommand(list_consistency_cmd)
+  logging.info('ListConsistency raw result is %s', raw_result)
+
+  for scenario in LIST_CONSISTENCY_SCENARIOS:
+    metric_name = '%s %s' % (scenario, LIST_CONSISTENCY_PERCENTAGE)
+    search_string = '%s: (.*)' % metric_name
+    result_string = re.findall(search_string, raw_result)
+    if len(result_string) > 0:
```
meteorfox: same as my comment above.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+                                   metadata))
+    else:
+      raise ValueError(
+          'Cannot get percentage from ListConsistency test.')
+
+    # Parse the list inconsistency window if there is any.
+    metric_name = '%s %s' % (scenario, LIST_INCONSISTENCY_WINDOW)
+    search_string = '%s: (.*)' % metric_name
+    result_string = re.findall(search_string, raw_result)
+    if len(result_string) > 0:
```
meteorfox: Just  `if result_string:` is ok.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +433,245 @@ def _DistributionToBackendFormat(dist):
...
+                                     metric_name,
+                                     LATENCY_UNIT,
+                                     metadata)
+
+    # Also report the list latency. These latencies are from the lists
+    # that were consistent.
+    metric_name = '%s %s' % (scenario, LIST_LATENCY)
+    search_string = '%s: (.*)' % metric_name
+    result_string = re.findall(search_string, raw_result)
+    if len(result_string) > 0:
```
meteorfox: Same as above.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Since we already have the units module, I made these into `units.second(...)` instead of plain floats. That documents what the units are, and it also means that PKB will automatically check that there are no unit conversion errors!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: You're right, thanks. Fixed.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Sorry, I didn't mean to put that code into this PR. Removed.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Fixed, thanks. I think we need `+rwx` just to be safe, to make sure we can access the directory, but you're completely right that `chomd 777` is too much.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Fixed, thanks.

perfkitbenchmarker/object_storage_service.py
noahl: I thought that looked weird, because we have that blank line when there's a real function body, but what you said seems to match the style guide. Done.

perfkitbenchmarker/providers/openstack/swift.py
```diff
@@ -0,0 +1,97 @@
...
+    vm.InstallPackages('python-urllib3')
+    vm.RemoteCommand('sudo pip install python-keystoneclient')
+    vm.RemoteCommand('sudo pip install python-swiftclient')
+
+  def CleanupVM(self, vm):
+    vm.RemoteCommand('/usr/bin/yes | sudo pip uninstall python-swiftclient')
+    vm.RemoteCommand('/usr/bin/yes | sudo pip uninstall python-keystoneclient')
+    vm.RemoteCommand('/usr/bin/yes | sudo pip uninstall python-gflags')
+
+  def CLIUpload(self, vm, file, bucket):
```
meteorfox: This should be `CLIUploadDirectory`

perfkitbenchmarker/providers/openstack/swift.py
```diff
@@ -0,0 +1,97 @@
...
+    return vm.RemoteCommand(
+        'time swift %s upload %s %s'
+        % (self.swift_command_prefix, bucket, file))
+
+  def CLIDownload(self, vm, bucket, object, dest):
```
meteorfox: This should be `CLIDownloadBucket`

meteorfox:
@noahl I tried your PR. Currently is throwing an error, it seems you forgot to rename `CLIUpload` and `CLIDownload` to `CLIUploadDirectory` and `CLIDownloadBucket` respectively.

Error:
```
2016-06-24 22:38:42,083 7ddd3fa7 MainThread object_storage_service(1/1) ERROR    Error during benchmark object_storage_service
Traceback (most recent call last):
  File "/opt/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 473, in RunBenchmark
    DoPreparePhase(benchmark, benchmark_name, spec, detailed_timer)
  File "/opt/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 351, in DoPreparePhase
    benchmark.Prepare(spec)
  File "/opt/PerfKitBenchmarker/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py", line 838, in Prepare
    service = object_storage_service.GetObjectStorageClass(FLAGS.storage)()
TypeError: Can't instantiate abstract class SwiftStorageService with abstract methods CLIDownloadBucket, CLIUploadDirectory
```

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: I'm not sure I understand the question. The function is mostly a space-saving optimization (and therefore money-saving optimization) for long-running benchmarks. In the CLI-based benchmark, where we upload files and then download them many times, this is the function that empties the bucket after each download. The cleanup at the end of the benchmark is the `DeleteBucket` function above, which works even with objects in the bucket.

On GCS and S3, you can issue a single command to delete all objects in a bucket. On Azure, it appears that you have to list the contents of the bucket and then issue delete commands for every object individually, which is one reason I didn't want to implement this. In the current benchmark, we empty buckets on Azure using the API test script, but on GCS and S3 using the CLI tool, which feels very inconsistent to me.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Fixed, thanks! That's much cleaner.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Done, thanks!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I think this and the following `if`statement must always be taken, because if `len(result_string) == 0` then we already raised an exception above. So I removed them.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Removed, thanks.

perfkitbenchmarker/providers/openstack/swift.py
noahl: Oops, I didn't update the Swift provider for one of the refactorings I did. Fixed, thanks.

perfkitbenchmarker/providers/openstack/swift.py
noahl: Same as above, and fixed. Thank you!

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: The changes failed in my testing, and I realized why we need `777`. Because we're using `sudo`, the directories will be created with owner `root` and group whatever group `root` is in. But later on, we want to run commands as PKB's user. Since we can't guarantee that PKB's user is in the same group as `root`, we have to allow `rwx` access for non-group members, which winds up being permissions `777`.

I think ideally we would let the PKB user own the directories, but that is more complexity. I'd like to leave this as `777` for now.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Same as above.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
yuyantingzero: Thanks for the explanation. My concern is we left blobs and buckets behind, but since we can still delete the bucket with blobs inside, it is fine.

Correct me if I am wrong, current behavior is during CLI tests, GCS and S3 remove all objects after each round of uploads. But on Azure we overwrite the objects? Which still seems a bit inconsistent to me.

Consider adding a TODO, we should support this method eventually (when Azure supports recursive/wildcard deletes) or maybe consider invoke api delete script if possible.

noahl: The current behavior is actually to use the API test script on a remote VM to clean up the objects, as you suggested. I didn't want to do that because running on a remote VM is unnecessary and very different than how we would handle cleanup for other providers. However, it does make the behavior less consistent. I have been thinking about whether I could run the API script on the controller machine to do the delete, but I am also okay letting Azure benchmarks be a little weird. Azure CLI uploads and downloads have always been different than GCS and S3 because the CLI can't upload/download multiple files at once, and this seems like a pretty small issue compared to that. (And also, bad interface design *should* get you lower performance numbers.)

noahl:
@meteorfox @yuyantingzero I pushed an update in response to your comments. Please take a look.

meteorfox:
@noahl PR is working now. One problem that is probably unrelated is a pip dependency fails to install due to missing the Python.h header file. That's probably that I need to install the 'python-dev' package in Debian or 'python-devel' on RHEL-based distro.

noahl:
@meteorfox great! About the Python issue, is the thing that's failing to install Swift-specific, or used by all providers? I'd like to add it to a `PrepareVM` method somewhere either way; I just need to know which one.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -501,278 +454,289 @@ def _DistributionToBackendFormat(dist):
   return val
 
 
-def ApiBasedBenchmarks(results, metadata, vm, storage, test_script_path,
-                       bucket_name, regional_bucket_name=None,
-                       azure_command_suffix=None, host_to_connect=None):
+class APIScriptCommandBuilder(object):
```
yuyantingzero: Add comments for APIScriptCommandBuilder attributes.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -310,55 +270,37 @@ def _GetClientLibVersion(vm, library_name):
...
+
+  return (
+      MULTISTREAM_DELAY_CONSTANT +
+      MULTISTREAM_DELAY_PER_VM * num_vms +
+      MULTISTREAM_DELAY_PER_STREAM * (num_vms * threads_per_vm))
 
 
 def _ProcessMultiStreamResults(raw_result, operation, sizes,
                                results, metadata=None):
+
```
yuyantingzero: Remove this empty line?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -1465,22 +960,40 @@ def Run(benchmark_spec):
...
+    metadata[GCS_MULTIREGION_LOCATION] = FLAGS.object_storage_gcs_multiregion
+  else:
+    metadata[GCS_MULTIREGION_LOCATION] = DEFAULT
+
+  metadata.update(service.Metadata(vms[0]))
+
+  results = []
+  test_script_path = '%s/run/%s' % (vms[0].GetScratchDir(), API_TEST_SCRIPT)
+  command_builder = APIScriptCommandBuilder(
+      test_script_path, STORAGE_TO_API_SCRIPT_DICT[FLAGS.storage], service)
```
meteorfox: `STORAGE_TO_API_SCRIPT_DICT` does not include a key for `providers.OPENSTACK`. This raises `KeyError: 'OpenStack'`

meteorfox:
This is the patch I added to get it going.

```diff
diff --git a/perfkitbenchmarker/providers/openstack/swift.py b/perfkitbenchmarker/providers/openstack/swift.py
index d7a3b76..86e0f08 100644
--- a/perfkitbenchmarker/providers/openstack/swift.py
+++ b/perfkitbenchmarker/providers/openstack/swift.py
@@ -73,7 +73,7 @@ class SwiftStorageService(object_storage_service.ObjectStorageService):
 
   def PrepareVM(self, vm):
     # Upgrade to latest urllib3 version fixes SSL SNMI error
-    vm.InstallPackages('python-urllib3')
+    vm.InstallPackages('python-dev python-urllib3')
     vm.RemoteCommand('sudo pip install python-keystoneclient')
     vm.RemoteCommand('sudo pip install python-swiftclient')
 
```

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Fixed. I also made sure that PKB would print a useful error message if the user tries to run API benchmarks on Swift.

noahl:
@meteorfox I incorporated that patch into this PR, but if you'd rather submit it yourself to keep the history accurate, I would be happy to remove.

noahl:
@yuyantingzero @meteorfox I updated in response to your comments; please take a look.

meteorfox:
@noahl Nah, go ahead leave that patch in.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -242,11 +194,12 @@
...
 # This accounts for the overhead of running RemoteCommand() on a VM.
-MULTISTREAM_DELAY_PER_VM = 5.0
-# We wait this many seconds for each stream. Note that this is
-# multiplied by the number of streams per VM, not the total number of
-# streams.
-MULTISTREAM_DELAY_PER_STREAM = 0.1
+MULTISTREAM_DELAY_PER_VM = 10.0 * units.second
+# We wait this long for each stream. Note that this is multiplied by
+# the number of streams per VM, not the total number of streams.
+MULTISTREAM_DELAY_PER_STREAM = 15.0 * units.second
```
yuyantingzero: We probably need a better way to handle this. Say we have 200 streams and 10 vms, we will be waiting 8,9 hours before this test is run.

noahl: Good point. I'd like to set these back to the old values for now - I didn't mean to put the new values in this PR. If we want to do this right in the long run, we'll have to benchmark how fast we can start processes on VMs and set these constants based on that, but I think that's outside the scope of this change.

noahl:
@yuyantingzero I set the multistream delays back to their old values. PTAL.

noahl:
@yuyantingzero @meteorfox I forgot to update a test in my previous commit; I just pushed a fix for that. PTAL.

meteorfox:
Works for me now! :+1:

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -310,44 +270,25 @@ def _GetClientLibVersion(vm, library_name):
...
-             '--os-tenant-name', tenant_name,
-             '--os-username', username,
-             '--os-password', password,
-             '--insecure' if FLAGS.openstack_swift_insecure else '',)
-  return ' '.join(options)
+
+  return (
+      MULTISTREAM_DELAY_CONSTANT +
+      MULTISTREAM_DELAY_PER_VM * num_vms +
+      MULTISTREAM_DELAY_PER_STREAM * (num_vms * threads_per_vm))
```
hildrum: I am confused with this and the comment on line 198-199

noahl: Oops, clear mistake on my part. Updated this formula.

perfkitbenchmarker/object_storage_service.py
```diff
@@ -0,0 +1,241 @@
...
+    No other method of this class should be called after
+    CleanupProvider.
+    """
+    pass
+
+  # Bucket management
+
+  @abc.abstractmethod
+  def MakeBucket(self, bucket):
+
```
yuyantingzero: More consistent on the empty line.

perfkitbenchmarker/object_storage_service.py
```diff
@@ -0,0 +1,241 @@
...
+    pass
+
+  @abc.abstractmethod
+  def DeleteBucket(self, bucket):
+    """Delete an object storage bucket.
+
+    Args:
+      bucket: the name of the bucket to delete.
+
+    This method should succeed even if bucket contains objects.
```
yuyantingzero: Move documentation in front of Args section.

perfkitbenchmarker/object_storage_service.py
```diff
@@ -0,0 +1,241 @@
...
+
+def FindBotoFile():
+  """Return the path to the boto file."""
+
+  boto_file = FLAGS.boto_file_location or DEFAULT_BOTO_LOCATION
+  boto_file = os.path.expanduser(boto_file)
+
+  if not os.path.isfile(boto_file):
+    raise errors.Benchmarks.MissingObjectCredentialException(
+        'Boto file cannot be found in %s.' % FLAGS.boto_file_location)
```
yuyantingzero: Replace FLAGS.boto_file_location with boto_file?

perfkitbenchmarker/providers/aws/s3.py
```diff
@@ -0,0 +1,101 @@
...
+
+  def EmptyBucket(self, bucket):
+    vm_util.IssueCommand(
+        ['aws', 's3', 'rm',
+         's3://%s' % bucket,
+         '--region', self.region,
+         '--recursive'])
+
+  def PrepareVM(self, vm):
+    vm.RemoteCommand('sudo pip install awscli boto')
```
yuyantingzero: Consider add a TODO to refactor out s3 installation as a package?

perfkitbenchmarker/providers/aws/s3.py
```diff
@@ -0,0 +1,101 @@
...
+        'time aws s3 sync s3://%s/ %s' % (bucket, dest))
+
+  def Metadata(self, vm):
+    return {object_storage_service.BOTO_LIB_VERSION:
+            linux_packages.GetPipPackageVersion(vm, 'boto')}
+
+  def APIScriptArgs(self):
+    hostname = AWS_S3_REGION_TO_ENDPOINT_TABLE[self.region]
+
+    return ['--host=%s' % hostname + AWS_S3_ENDPOINT_SUFFIX]
```
yuyantingzero: This looks a bit inconsistent, please either use + or %s

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+
+  def PrepareVM(self, vm):
+    # Documentation:
```
yuyantingzero: Please add comments into doc string

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+    # http://azure.microsoft.com/en-us/documentation/articles/xplat-cli/
+
+    vm.Install('node_js')
```
yuyantingzero: Consider refactor out azure cli installation into a package or add a TODO

perfkitbenchmarker/object_storage_service.py
```diff
@@ -0,0 +1,241 @@
...
+    """Empty an object storage bucket.
+
+    Args:
+      bucket: the name of the bucket to empty.
+    """
+    pass
+
+  # Working with a VM
+
+  def PrepareVM(self, vm):
```
yuyantingzero: Any reason why PrepareVM and CleanupVM method is not abstract?

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+        ['azure', 'storage', 'account', 'create',
+         '--type', 'ZRS',
+         '-l', location or DEFAULT_AZURE_REGION,
+         self.storage_account])
+
+    output, _, _ = vm_util.IssueCommand(
+        ['azure', 'storage', 'account',
+         'keys', 'list', self.storage_account])
+
+    key = re.findall(r'Primary:* (.+)', output)
```
yuyantingzero: please reuse perfkitbenchmarker/regex_util.py

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+                                                  AZURE_CREDENTIAL_LOCATION),
+        AZURE_CREDENTIAL_LOCATION)
+
+
+  def CleanupVM(self, vm):
+    pass
+
+  def CLIUploadDirectory(self, vm, directory, file_names, bucket):
+    return vm.RemoteCommand(
+        'time for file in %s; '
```
yuyantingzero: In case there are lots of %s, it is more readable using ".format" and give each %s a name.
e.g. 'time for file in {files} ... '.format(files=' '.join(file_names), ...)

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,119 @@
...
+        'do azure storage blob upload -q %s/$file %s -a %s -k %s; '
+        'done' % (' '.join(file_names), directory, bucket,
+                  self.storage_account, self.azure_key))
+
+  def CLIDownloadBucket(self, vm, bucket, objects, dest):
+    return vm.RemoteCommand(
+        'time for object in %s; '
```
yuyantingzero: same here

perfkitbenchmarker/providers/gcp/gcs.py
```diff
@@ -0,0 +1,152 @@
...
+        ['gsutil', 'rb',
+         'gs://%s' % bucket])
+
+  def EmptyBucket(self, bucket):
+    vm_util.IssueCommand(
+        ['gsutil', '-m', 'rm', '-r',
+         'gs://%s/*' % bucket])
+
+  def PrepareVM(self, vm):
+    vm.Install('wget')
```
yuyantingzero: Consider refactor gcloud cli installation to package

perfkitbenchmarker/providers/openstack/swift.py
```diff
@@ -0,0 +1,97 @@
...
+
+    vm_util.IssueCommand(
+        ['swift'] + self.swift_command_parts + ['delete', bucket])
+
+  def EmptyBucket(self, bucket):
+    vm_util.IssueCommand(
+        ['swift'] + self.swift_command_parts + ['delete', bucket])
+
+  def PrepareVM(self, vm):
+    # Upgrade to latest urllib3 version fixes SSL SNMI error
```
yuyantingzero: Consider refactoring swift

tests/linux_benchmarks/object_storage_service_benchmark_test.py
```diff
@@ -14,40 +14,89 @@
 
 """Tests for ping_benchmark."""
```
yuyantingzero: ping_benchmark? :)

perfkitbenchmarker/object_storage_multistream_analysis.py
```diff
@@ -98,8 +100,15 @@ def GetStreamActiveIntervals(start_times, durations, stream_ids):
   interval_start = stream_start_times.max()
   interval_end = stream_end_times.min()
 
-  return (Interval(first_start, end=last_end),
-          Interval(interval_start, end=interval_end))
+  logging.info('Stream ends after stream starts: %s',
+               (stream_end_times > stream_start_times).all())
+  if interval_end <= interval_start:
+    raise Exception('No interval when all streams were active.')
```
yuyantingzero: Be more specific on Exception raised. e.g. define something like ObjectStorageServiceException

perfkitbenchmarker/object_storage_service.py
noahl: Fixed, thanks.

perfkitbenchmarker/object_storage_service.py
noahl: Done.

perfkitbenchmarker/object_storage_service.py
noahl: Done.

perfkitbenchmarker/providers/aws/s3.py
noahl: Done.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: Done.

perfkitbenchmarker/object_storage_service.py
noahl: If they are abstract, then any subclass has to implement them. I wanted to make it possible for subclasses to not implement them if they have no preparation or cleanup to do.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: Done. I added a new function in regex_util.py to get exactly the semantics I wanted and used that.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: Done, thanks.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: Done.

tests/linux_benchmarks/object_storage_service_benchmark_test.py
noahl: Fixed :)

perfkitbenchmarker/object_storage_multistream_analysis.py
noahl: Done.

perfkitbenchmarker/providers/aws/s3.py
noahl: Actually, I went ahead and just factored it out.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
noahl: Factored out.

perfkitbenchmarker/providers/gcp/gcs.py
noahl: I would like to, but I don't think our current `vm.Install` interface can handle installing specific package versions. I'd like to keep that change separate from this PR.

perfkitbenchmarker/providers/openstack/swift.py
noahl: Moved to its own package.

perfkitbenchmarker/object_storage_multistream_analysis.py
noahl: Done.

noahl:
@yuyantingzero I updated in response to your comments. Please take a look.

yuyantingzero:
LGTM

PR #1034, opened 2016-06-30
side-by-side improvements

hildrum:
* When the base and the head are both zero, report the relative difference as 1 and the percent difference as 0.
Old:
image
New:
image

* Force metrics to be the same rather than just being in a similar place in the sequence.  
Before:
image

After:
image


ehankland:
LGTM

PR #1042, opened 2016-07-06
[datastore_ycsb] Support overriding vm_count in benchmark specs

k1xme:
The previous implementation cannot override `vm_count` field in `BENCHMARK_CONFIG`. Not sure why.
Took this snippet from other benchmarks, explicitly overriding `vm_count`.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cloud_datastore_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_datastore_ycsb_benchmark.py
@@ -68,7 +68,10 @@ flags.DEFINE_string('google_datastore_debug',
 
 
 def GetConfig(user_config):
-    return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+    config = configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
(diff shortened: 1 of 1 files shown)


yuyantingzero:
I believe for all ycsb benchmarks, the client vm numbers are set by ycsb_client_vms flags. Number of server vms are decided by benchmarks. 
(For datastore benchmark, there is no server vm.)

k1xme:
@yuyantingzero I'ved tried that flag too, but it didn't work neither. Gonna double check this.

k1xme:
@yuyantingzero Double checked on this. `ycsb_client_vms' doesn't work.

yuyantingzero:
Sorry. My bad. By looking at cassandra_ycsb and other ycsb benchmarks, they check ycsb_client_vms flags in GetConfig, if the flag is present, use that flag to decide the number of client vms: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_benchmarks/cassandra_ycsb_benchmark.py#L65

k1xme:
@yuyantingzero No prob. Done changing the flag from `num_vms` to `ycsb_client_vms`.

yuyantingzero:
LGTM

k1xme:
Ready to merge

PR #1043, opened 2016-07-06
[ycsb] Upgrade YCSB to 0.9.0 and support latest logging format.

k1xme:
YCSB 0.9.0 output results and log to `stdout` and `stderr`.

Previously we only parse stdout log to get the info we care about, but with YCSB 0.9.0, that is not the case anymore. We need to combine both log files to get all the info.

### DIFF
--- a/perfkitbenchmarker/linux_packages/ycsb.py
+++ b/perfkitbenchmarker/linux_packages/ycsb.py
@@ -55,7 +55,7 @@ from perfkitbenchmarker import vm_util
 
 FLAGS = flags.FLAGS
 
-YCSB_VERSION = '0.7.0'
+YCSB_VERSION = '0.9.0'
 YCSB_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases/'
                 'download/{0}/ycsb-{0}.tar.gz').format(YCSB_VERSION)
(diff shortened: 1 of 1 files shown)


hildrum:
Which results are we missing?

k1xme:
@hildrum The client version and command line info are now output to `stderr`. Previously, we only get info from `stdout`. So if we use any YCSB version > 0.7.0, we cannot get these 2 lines info.

perfkitbenchmarker/linux_packages/ycsb.py
```diff
@@ -627,8 +627,11 @@ def _Run(self, vm, **kwargs):
       param, value = pv.split('=', 1)
       kwargs[param] = value
     command = self._BuildCommand('run', **kwargs)
-    stdout, _ = vm.RobustRemoteCommand(command)
-    return ParseResults(str(stdout))
+    # YCSB version greater than 0.7.0 output some of the
```
yuyantingzero: Do we still need the workaround on line 221?

k1xme: Yes, we still need that, since stderr contains not only the messages we need but also a lot of debugging, error messages.

yuyantingzero:
LGTM

k1xme:
Ready to merge

PR #1050, opened 2016-07-13
[openjdk7] Support Ubuntu 16.04

k1xme:
Quick fix for #1038 .

Solution based on [Install openjdk-7-jdk on Ubuntu16.04](http://askubuntu.com/questions/761127/ubuntu-16-04-and-openjdk-7).

### DIFF
--- a/perfkitbenchmarker/linux_packages/openjdk7.py
+++ b/perfkitbenchmarker/linux_packages/openjdk7.py
@@ -14,6 +14,11 @@
 
 
 """Module containing OpenJDK7 installation and cleanup functions."""
+import logging
+
+from perfkitbenchmarker import errors
+from perfkitbenchmarker import vm_util
+
 
 JAVA_HOME = '/usr'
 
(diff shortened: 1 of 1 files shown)


hildrum:
I'm not sure this is the right solution.  Maybe we shouldn't be trying to install openjdk-7-jdk on Ubuntu 16.04, but rather we should be installing a newer version of java.  @yuyantingzero ? 
This is related to issue #1050.

k1xme:
Upgrading to newer openjdk version might incur compatibility issues that
breaks other benchmars. If we decide to do so, it's better to mske sure
other packages are compatible with it.

On Wed, Jul 13, 2016, 11:32 PM Kris Hildrum <notifications@github.com>
wrote:

> I'm not sure this is the right solution. Maybe we shouldn't be trying to
> install openjdk-7-jdk on Ubuntu 16.04, but rather we should be installing a
> newer version of java. @yuyantingzero <https://github.com/yuyantingzero>
> ?
> This is related to issue #1050
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1050>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1050#issuecomment-232569774>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AGxoeokRiKZbjIMe_t5ONBMRnX4qMSJ1ks5qVdffgaJpZM4JL2AT>
> .
>
-- 

Kexi Long

hildrum:
Yes, it might not be compatible with the older benchmarks.  As far as I can tell, though, ycsb is compatible with java8.  But basic problem with that we want to use the new version of Ubuntu because that's the logical thing to use for comparisons.  But why would we use a new OS and an outdated java?

yuyantingzero:
ubuntu 16.04 LTS looks like a replacement for 14.04LTS. We definitely should upgrade the default image.  But as k1xme mentioned, a handful benchmarks might be affected. I did a search in repo:
1. all ycsb benchmarks
2. scimark2
3. cassandra_stress
4. hadoop_terasort,
5. tomcat_wrk_benchmark
6. cloudsuite_web_serving

k1xme:
Yeah, that's what I'm talking about. I agree we should upgrade to newer packages and OS's, but it's gonna take more work to test.

hildrum:
Since we have a different solution, I'm closing this.

PR #1052, opened 2016-07-14
Minor fixes in flag descriptions

hildrum:
Was reading flags, founds some stuff.  Flag types that are multistream automatically get a `repeat this option to specify a list of values`, so there's no need for the description itself to say how to repeat the flag.


yuyantingzero:
LGTM

Daareyreg:
Thanks

PR #1065, opened 2016-07-27
Use the BigQuery Streaming API

noahl:
When using a BigQuery Table publisher, do 'bq insert' rather than 'bq
load', which will use the BigQuery streaming API rather than the bulk
load API.


noahl:
@yuyantingzero @hildrum @ehankland please take a look.

hildrum:
Using the streaming API costs money, no?  Does that make a difference?
https://cloud.google.com/bigquery/pricing

PR #1067, opened 2016-08-03
Add support for st1 and sc1 disks on EBS

noahl:
Support two new EBS disk types, st1 and sc1.


noahl:
I verified that I can run `fio` on both of the new disk types.

yuyantingzero:
LGTM

PR #1074, opened 2016-08-19
Multi-VM Support for the Object Storage Benchmark

noahl:
Let the object storage benchmark run from multiple VMs. Results are
aggregated and reported the same way as for a single VM, except that
throughput is no longer bounded by a single VM's limits.

Also add a new object naming scheme to the benchmark, and report QPS in
addition to latency and throughput.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -110,6 +110,16 @@ flags.DEFINE_integer('object_storage_list_consistency_iterations', 200,
                      'scenario. However, to get useful metrics from the '
                      'api_namespace scenario, a high number of iterations '
                      'should be used (>=200).')
+flags.DEFINE_enum('object_storage_object_naming_scheme', 'sequential_by_stream',
+                  ['sequential_by_stream',
+                   'approximately_sequential'],
+                  'How objects will be named. Only applies to the '
+                  'api_multistream benchmark. '
+                  'sequential_by_stream: object names from each stream '
+                  'will be sequential, but different streams will have '
+                  'different name prefixes. '
+                  'approximately_sequential: object names from all '
+                  'streams will roughly increase together.')
 
 
 FLAGS = flags.FLAGS
@@ -135,6 +145,7 @@ object_storage_service:
     default:
       vm_spec: *default_single_core
       disk_spec: *default_500_gb
+      vm_count: null
 """
 
 DATA_FILE = 'cloud-storage-workload.sh'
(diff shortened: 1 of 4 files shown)


noahl:

One design point I'm not sure about  - should the benchmark still have a `num_streams=...` option, or should it be `streams_per_vm=...`, and then you can multiply by `num_vms` to get the total stream count?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -404,6 +429,16 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
         'Multi-stream ' + operation + ' ' + name,
         value.magnitude, str(value.units), metadata=distribution_metadata))
 
+  # QPS metrics
+  results.append(sample.Sample(
+      'Multi-stream ' + operation + ' QPS',
```
hildrum: Do you want to append a 'any streams active' here?  Or are you trying to keep it backward compatible?

hildrum:
I think it'd be cleaner to do `streams_per_vm` and `num_vms`.  You'd be able to avoid the `NaturalDivisionRoundingUp` function entirely.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Yes, that would make more sense. Done.

noahl:
@hildrum I think I've addressed your comments. Please take a look.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -97,9 +97,9 @@
                       'Size of objects to send and/or receive. Only applies to '
                       'the api_multistream scenario. Examples: 1KB, '
                       '{1KB: 50%, 10KB: 50%}')
-flags.DEFINE_integer('object_storage_multistream_num_streams', 10,
-                     'Number of independent streams to send and/or receive on. '
-                     'Only applies to the api_multistream scenario.',
+flags.DEFINE_integer('object_storage_multistream_streams_per_vm', 10,
```
hildrum: Did you really want multistream in this name?  It's okay if you do.

hildrum:
LGTM.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: I don't really. I'll remove it.

PR #1075, opened 2016-08-21
Documentation Website using MkDocs

meteorfox:
As an attempt to improve adoption and lowering the entry bar for others to use and contribute to PerfKit Benchmarker. I proposed this PR which replaces the current documentation site with MkDocs.

Why MkDocs? MkDocs was designed exactly for this purpose, for software documentation, it easy to get started and to customize. Supports search and as well as other features. See here for more information:
http://www.mkdocs.org/

Here's a staging version of the PerfKit Benchmarker Docs website: 
http://meteorfox.github.io/PerfKitBenchmarker/

Along the way, I reorganized the documentation and broke it down into self-contained sections. Of course, we can always change it to better fits our needs. I also tried to style it with the look-and-feel of the PerfKit Explorer dashboard.

The way this is deploy is that the docs themselves are stored under a directory `docs/` in master, then they are build using `mkdocs build --clean` which produces a `site/` directories with all the static assets needed for the page. The `site/` contents is what is pushed to the `gh-pages` branch.

Thoughts?


meteorfox:
@voellm PTAL, I want to know your thoughts as well.

meteorfox:
@noahl I would also appreciate your feedback. :smile:

docs/documentation/benchmarks/benchmark-sets.md
```diff
@@ -0,0 +1,26 @@
...
+## Standard Benchmark Set
+---
+
+### Running the Standard Benchmark Set
+Run without the `--benchmarks` parameter and every benchmark in the standard set
+will run serially which can take a couple of hours (alternatively, run with
+`--benchmarks="standard_set"`).  Additionally, if you don't specify
+`--cloud=...`, all benchmarks will run on the Google Cloud Platform.
+
+## Named Benchmark Set
```
noahl: Nit: should be "Named Benchmark Sets", with a final 's'

hildrum:
I think this new website looks good.  The Getting Started page doesn't include the provider-specific installation instructions; was that intentional or just for this demo?

docs/documentation/configurations.md
```diff
@@ -0,0 +1,193 @@
+# Configuring PerfKit Benchmarker
+
+## Configurations and Configuration Overrides
+---
+Each benchmark now has an independent configuration which is written in YAML.
```
noahl: Nit: remove "now", since new users don't know about PKB's history.

docs/documentation/configurations.md
```diff
@@ -0,0 +1,193 @@
...
+Users may override this default configuration by providing a configuration.
+This allows for much more complex setups than previously possible,
+including running benchmarks across clouds.
+
+A benchmark configuration has a somewhat simple structure. It is essentially
+just a series of nested dictionaries. At the top level, it contains VM groups.
+VM groups are logical groups of homogenous machines. The VM groups hold both a
```
noahl: Nit: "homogenous" -> "homogeneous".

noahl:
@meteorfox the new website looks good to me. I left a couple of very minor nitpicky comments, but other than that it looks great. One general point is that I see a couple empty files in the checkin. Are those required by MkDocs? If not, can you remove them?

A broader concern is that I don't feel qualified to review all of these docs for content, but I don't think you should wait for a knowledgeable reviewer to appear before submitting. Instead, I have two questions for you: 1) did you get all of the information that was in our old site into the new one? and 2) what information did you add?

meteorfox:
I was actually looking for validation regards this approach to documentation. Perhaps I should have opened an issue instead of the pull request. I took the existing documentation and broke it down into several modules as a proof of concept. My intention was not to merge, at least not yet.

@hildrum Thanks for reviewing. Good catch, it was intentional as I was trying to keep the 'Getting Started' very minimal. But I should have added at least the instructions for the Google provider and link to the other providers. I'll make those changes.

@noahl Thanks for reviewing. I wasn't looking for a review of the docs, but more of a validation of using MkDocs. The docs are essentially the same, just broken down into pages. The empty files correspond to the new sections I want to add, they are needed otherwise it complains that they are missing. I can remove them for now and only add them when actually have content for them. 1) Yes, all the information is from the original website and the README.md. 2) Only the 'Downloading Perfkit Benchmarker' in the Getting Started is new. Everything else is the same.

It seems, there's agreement to use MkDocs for documentation. Here's what I'll do next then:

1. Transfer all the contents from the README.md and www.perfkitbenchmarker.org  (done).
2. Make a change to the 'Getting Started' to include a link to instructions on how to setup a provider.
3. Remove sections that currently have no contents or were not included in either the README or the original website.
4. Review and merge just that. No new docs will be added. 
5. (Optionally) Go live with the new docs site.
6. Make separate pull requests for the other docs (.e.g CONTRIBUTING.md, CHANGES.md, FAQs,  and the docs in the Wiki, etc)
7. Review, and merge those.
8. Finally, then we can make separate pull requests for new sections, like how to contribute to the Docs, Dev environment setup, dev workflow, etc.

Sounds good?

noahl:
@meteorfox sounds good to me!

hildrum:
Were you thinking this would replace http://googlecloudplatform.github.io/PerfKitBenchmarker/ or be linked to from it?

And what would the development process look like?  Where would we check in the md files used to generate the web pages?

meteorfox:
@hildrum The idea is to replace it but we could also have a landing page for it, something like [gRPC](http://grpc.io) and [Kubernetes](http://kubernetes.io) have and then link the docs from there. We would just have a simple static HTML page (index.html) that has the same look-and-feel as the docs and simply link there.

The md files would go under a directory `docs/` in the master branch. The `docs/` directory has the same structure as how the docs would be organized.

e.g.
```
carl6796@t450s:~/programming/github.com/GoogleCloudPlatform/PerfKitBenchmarker$ tree docs/
docs/
├── community.md
├── contributing
│   ├── dev-env.md
│   ├── dev-workflow.md
│   └── license.md
├── documentation
│   ├── alicloud.md
│   ├── aws.md
│   ├── azure.md
│   ├── benchmarks
│   │   ├── benchmark-sets.md
│   │   ├── linux-benchmarks.md
│   │   ├── object-storage.md
│   │   └── windows-benchmarks.md
│   ├── cloudstack.md
│   ├── configurations.md
│   ├── digitalocean.md
│   ├── gcp.md
│   ├── getting-started.md
│   ├── image-prereqs-docker-based-clouds.md
│   ├── kubernetes.md
│   ├── mesos.md
│   ├── openstack.md
│   └── rackspace.md
├── extra.css
├── img
│   ├── .... (*.png, *.svg, *.png)
├── index.md
└── media.md
```

At the root level of the project, there would be a file `mkdocs.yml` that is used to build the static website. The `mkdocs.yml` is very simple, it just describes the different "menus" in a hierarchical fashion, and includes the path to which md file it maps to. There is also some extra information about theme, and extensions applied.

To build the website, first `mkdocs` Python package needs to be installed. You can use `pip` as follows:

```
pip install mkdocs
```

Once you have the `mkdocs` dependency, you can preview it locally with:
```
mkdocs serve
```
and to build it:
```
mkdocs build --clean
```

After the website gets built, it will place a directory `site/` with the static assets in it. The contents of this  directory are what it is pushed to the `gh-pages` branch. See my [`gh-pages` branch](https://github.com/meteorfox/PerfKitBenchmarker/tree/gh-pages) that currently contains the staging version of the docs.

If we want to change the landing page, then I would have an `index.html` page to serve as the landing page at the root level and then the `site/` directory next to it. That way from the landing page we just link to `site/index.html`.

hildrum:
Thanks for the details on the process; it makes sense to me.  I definitely like the looks of the mkdocs pages for documentation, and that seems to be a sensible thing to do.

We may want to keep the landing page with its graphic, though.

meteorfox:
The latest commit (sorry, I had to amend it I wanted to keep it clean) is up. The version for that commit is live on my gh-pages fork here: http://meteorfox.github.io/PerfKitBenchmarker/

Basically, I removed the empty pages. Moved everything from the REAMDE.md and http://googlecloudplatform.github.io/PerfKitBenchmarker/ to the docs.

No new content was added, I only re-organized the docs into different sections. 

I suggest reviewing it from the rendered version.

tedsta:
Hey @meteorfox sorry this sat for so long. The site looks good! Is this still ok to merge?

s-deitz:

PR #1076, opened 2016-08-23
Isolate Pandas Dependency

noahl:
Move all code that depends on pandas to a script, where it can be
invoked when needed by the object storage benchmark but not affect any
of the other benchmarks or our CI infrastructure.


noahl:

hildrum:
What's the reason you are using pandas to do the data analysis?  I'm not seeing anything in the output that seems hard to compute without a library.  Is the size of the input data the problem?

noahl:
@hildrum the first-level reason is that the old analysis step used Pandas, and I was hoping to make a minimal PR to remove the Pandas dependency so I could get it in the next release without delaying it.

The second-level reason is that when I wrote the old analysis script, I thought that using a general-purpose data analysis library would make it easy to implement my analysis, and I wanted the speed. In retrospect I'm not sure the library makes the analysis any easier, but I do still want the speed because this needs to scale to analyzing input from a lot of VMs. I could move to Numpy (Pandas is built on Numpy anyway), but I am hesitant to use raw Python code. I do feel like the analysis step is a mess right now, though, and I would love to fix that.

noahl:
After talking with @hildrum, @ehankland and @yuyantingzero offline, we agreed to remove Pandas and go to Numpy right now, and maybe move to a separate script later. I will close this PR and open a new one for the new direction.

PR #1080, opened 2016-08-24
Allow user-defined metadata for the vms

hildrum:
Allow users to specify vm metadata with a flag.  @ehankland PTAL.

### DIFF
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -341,12 +341,11 @@ class BenchmarkSpec(object):
     vm_metadata = {'benchmark': self.name,
                    'perfkit_uuid': self.uuid,
                    'benchmark_uid': self.uid}
-    if FLAGS.vm_metadata:
-      for item in FLAGS.vm_metadata:
-        if ':' not in item:
-          raise Exception('\'%s\' not in expected key=value format' % item)
-        key, value = item.split(':', 1)
-        vm_metadata[key] = value
+    for item in FLAGS.vm_metadata:
+      if ':' not in item:
+        raise Exception('"%s" not in expected key=value format' % item)
+      key, value = item.split(':', 1)
+      vm_metadata[key] = value
 
     vm.Create()
 
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -343,7 +343,7 @@ class BenchmarkSpec(object):
                    'benchmark_uid': self.uid}
     for item in FLAGS.vm_metadata:
       if ':' not in item:
-        raise Exception('"%s" not in expected key=value format' % item)
+        raise Exception('"%s" not in expected key:value format' % item)
       key, value = item.split(':', 1)
       vm_metadata[key] = value
 
--- a/perfkitbenchmarker/benchmark_spec.py
+++ b/perfkitbenchmarker/benchmark_spec.py
@@ -344,8 +344,15 @@ class BenchmarkSpec(object):
     logging.info('Waiting for boot completion.')
     vm.AllowRemoteAccessPorts()
     vm.WaitForBootCompletion()
-    vm.AddMetadata(benchmark=self.name, perfkit_uuid=self.uuid,
-                   benchmark_uid=self.uid)
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    if FLAGS.vm_metadata:
+      for item in FLAGS.vm_metadata:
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/benchmark_spec.py
```diff
@@ -344,8 +344,15 @@ def PrepareVm(self, vm):
     logging.info('Waiting for boot completion.')
     vm.AllowRemoteAccessPorts()
     vm.WaitForBootCompletion()
-    vm.AddMetadata(benchmark=self.name, perfkit_uuid=self.uuid,
-                   benchmark_uid=self.uid)
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    if FLAGS.vm_metadata:
```
ehankland: Since the default is an empty list (and not none), I don't think you need this guard

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -344,8 +344,15 @@ def PrepareVm(self, vm):
...
+      for item in FLAGS.vm_metadata:
+        logging.info('pair is ' + item)
+        key, value = item.split('=', 1)
```
ehankland: Can we fail before creating VMs if the flag doesn't follow the expected format?

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -344,8 +344,15 @@ def PrepareVm(self, vm):
...
     vm.AllowRemoteAccessPorts()
     vm.WaitForBootCompletion()
-    vm.AddMetadata(benchmark=self.name, perfkit_uuid=self.uuid,
-                   benchmark_uid=self.uid)
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    if FLAGS.vm_metadata:
+      for item in FLAGS.vm_metadata:
+        logging.info('pair is ' + item)
```
ehankland: This feels a little verbose to me (especially for info), but it's up to you

perfkitbenchmarker/virtual_machine.py
```diff
@@ -41,6 +41,10 @@
 _VM_REGISTRY = {}
 
 
+flags.DEFINE_list('vm_metadata', [], 'Metadata to add to the vm '
```
ehankland: Please describe the expected format (key=value)

perfkitbenchmarker/benchmark_spec.py
hildrum: Good catch; I'd meant to take that out.

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -338,14 +338,23 @@ def PrepareVm(self, vm):
     Args:
         vm: The BaseVirtualMachine object representing the VM.
     """
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    if FLAGS.vm_metadata:
+      for item in FLAGS.vm_metadata:
+        if ':' not in item:
+          raise Exception('\'%s\' not in expected key=value format' % item)
```
ehankland: in order to avoid escaping quotes you can either use double quotes on the string or use double quotes on "%s" -  this isn't that bad though so only fix if you feel like it

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -338,14 +338,23 @@ def PrepareVm(self, vm):
     Args:
         vm: The BaseVirtualMachine object representing the VM.
     """
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    if FLAGS.vm_metadata:
```
ehankland: Why do you need this line?

hildrum: Checking to see if you read code carefully?

perfkitbenchmarker/benchmark_spec.py
```diff
@@ -338,14 +338,22 @@ def PrepareVm(self, vm):
     Args:
         vm: The BaseVirtualMachine object representing the VM.
     """
+    vm_metadata = {'benchmark': self.name,
+                   'perfkit_uuid': self.uuid,
+                   'benchmark_uid': self.uid}
+    for item in FLAGS.vm_metadata:
+      if ':' not in item:
+        raise Exception('"%s" not in expected key=value format' % item)
```
ehankland: not in expected key:value format

ehankland:
LGTM

PR #1081, opened 2016-08-25
Replace Pandas with Numpy

noahl:
Remove the Pandas dependency from the object storage benchmarker, and
replace it with a Numpy-based analysis. Take the opportunity to simplify
the analysis code and move it all into
perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py.

I also checked that the results from this benchmark match the results
from the previous version, at least for my test configuration.


noahl:

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -379,68 +402,73 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
+                      start_times[i][active_start_indexes[i]]
+                      for i in xrange(num_streams)]
+  active_sizes = [np.sum(size) for size in active_sizes]
+  results.append(sample.Sample(
+      'Multi-stream ' + operation + ' net throughput',
+      np.sum((size / active_time * 8
+              for size, active_time in zip(active_sizes, active_times))),
+      'bit / second', metadata=distribution_metadata))
+  results.append(sample.Sample(
+      'Multi-stream ' + operation + ' net throughput (with gap)',
```
hildrum: is `with gap` equivalent to `any stream active` and `simplified` equivalent to `all streams active`?  If so, can you be consistent?  If you can't, for backwards compatibility reasons, can you add a comment?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -695,8 +723,64 @@ def ListConsistencyBenchmark(results, metadata, vm, command_builder,
...
+            json_out[0]['stream_num'] == prev_stream_num + 1)
+    prev_stream_num = prev_stream_num + 1 if prev_stream_num else 0
+
+    start_time = np.zeros([num_records], dtype=np.float64)
+    latency = np.zeros([num_records], dtype=np.float64)
+    size = np.zeros([num_records], dtype=np.int64)
+
+    prev_start = None
+    prev_latency = None
+    for i in xrange(num_records):
```
hildrum: No need to change, but any reason you're not doing:
`start_time = [r['start_time'] for r in json_out]`?

I'd guess you'd still need a for loop to check your assertion.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Short answer: this is all about backwards compatibility, and I would love to get rid of it. Long answer: the "without gap" computation (just labelled as "net throughput") is supposed to compute what the throughput would be if there were no benchmarking overhead. It takes the series of events from each worker process, drops the gap time so the uploads or downloads start one after another, and then computes the throughput of that alternate history. I would *love* to drop it if you agree.

In practice, however, the gaps are less than 0.1% of the total benchmarking time for every benchmark I've seen, so it's not a huge deal.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: @hildrum No reason, your version of the loop is better. I'd like to push a new commit anyway to fix some confusing variable names, so I will also make the change then.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
hildrum: I think it makes sense to report the throughput as seen once all workers are going;  if the time before that happens is so small as to be negligible (and you check that), then you can ignore the fact they don't all start at once as far as I'm concerned.  

But my bigger question here--even if you don't change the names, since you're using two names for the same thing, could you make sure to add a comment saying that?

hildrum:
LGTM.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -308,48 +306,52 @@ def MultiThreadStartDelay(num_vms, threads_per_vm):
...
   Args:
-    raw_result: list of strings. The stdouts of the worker processes.
+    start_times: a list of numpy arrays. Operation start times, as
+      POSIX timestamps.
+    latencies: a list of numpy arrays. Operation durations, in seconds.
+    sizes: a list of numpy arrays. Object sizes used in each
+      operation, in bytes.
     operation: 'upload' or 'download'. The operation the results are from.
-    sizes: the object sizes used in the benchmark, in bytes.
+    all_sizes: all object sizes in the distribution used, in bytes.
```
yuyantingzero: what is the type of all_sizes? list of numpy arrays?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Never mind, I realized I needed the index to set the ndarrays `start_time`, `latency` and `size`.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: A sequence of integers. Added the type to the docstring.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -695,8 +723,64 @@ def ListConsistencyBenchmark(results, metadata, vm, command_builder,
                                    metadata)
 
 
+def LoadWorkerOutput(output):
+  """Load output from worker processes to our internal format.
+
+  Args:
+    output: list of strings. The stdouts of all worker processes.
```
yuyantingzero: maybe have a sample output here for reference

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -695,8 +723,64 @@ def ListConsistencyBenchmark(results, metadata, vm, command_builder,
...
+    seconds. size stores int64s, holding object size in
+    bytes.
+
+  Raises:
+    AssertionError, if an individual worker's input includes
+    overlapping operations, or operations that don't move forward in
+    time, or if the input list isn't in stream number order.
+  """
+
+  start_times = []
```
yuyantingzero: is this necessary? given line 760 - 762

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -360,13 +362,34 @@ def _ProcessMultiStreamResults(raw_result, operation, sizes,
...
     metadata['stream_gap_above_threshold'] = True
 
-  records_in_interval = records[
-      analysis.FullyInInterval(records['start_time'],
-                               records['latency'],
-                               all_streams_active)]
+  # Find the indexes in each stream where all streams are active,
+  # following Python's [inclusive, exclusive) index convention.
+  active_start_indexes = []
+  for start_time in start_times:
```
yuyantingzero: you probably can use np.where function here. http://docs.scipy.org/doc/numpy/reference/generated/numpy.where.html

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: Added. Also rewrote the description to hopefully be more clear.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
noahl: @yuyantingzero I looked at the docs, but I'm not sure it applies in this case. It looks like `where` gives you all of the indices where some condition is true, but I only want the first index. I could use `where` and then only take the first index, but that seems like extra work.

noahl:
Travis seems to be stalled waiting for resources. I've verified on my own machine that the tests pass. Merging now.

Issue #1082, opened 2016-08-26
Failed to build collectd


shalq:
When run benchmark with collectd trace enabled,  the job failed due to failing to build collectd, the error is as below:

WARNING: `aclocal' not found!
    Please make sure that `aclocal' is installed and is in one of the
    directories listed in the PATH environment variable.
WARNING: `automake' not found!
    Please make sure that `automake' is installed and is in one of the
    directories listed in the PATH environment variable.

From the log, we need to install some dependencies for building collectd. The test image is ubuntu 14.04.

shalq:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1083

shalq:
A fix for this issue:　https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1083

voellm:
Please see the comment on the pull request.

PR #1083, opened 2016-08-26
Fix error about failing to build collectd

shalq:
Install dependencies pkg for building collectd for ubuntu image. The patch is to fix issue https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/1082


shalq:
I signed it !

voellm:
@shalq It looks like you have not signed the Google CLA.  Can you please sign that and then this code and be reviewed.   Instructions are in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/CONTRIBUTING.md

s-deitz:

PR #1089, opened 2016-09-01
Made changes so that kubectl reads from file instead of piping

nehayward:


nehayward:
I signed it!

hildrum:
I see you have signed it; I'm not sure why the CLAbot didn't see that.  You might try a "I signed it!" again, but it looks like you also have some flake errors that need to be fixed.

tedsta:
There are some linter errors. Could you run `hooks/check-everything` and fix the errors you get?

nehayward:
Yes I'll make those changes

PR #1090, opened 2016-09-02
Fix wait_for_command.py: wait for status file to avoid race condition.

yuyantingzero:


yuyantingzero:
@ehankland , ptal

ehankland:
LGTM

PR #1093, opened 2016-09-08
Fix flags in static vm files.

yuyantingzero:
Fix flag override in static vms.


perfkitbenchmarker/static_virtual_machine.py
```diff
@@ -252,7 +256,8 @@ def VerifyItemFormat(item):
           'static_vm_file', ip_address=ip_address, user_name=user_name,
           ssh_port=ssh_port, install_packages=install_packages,
           ssh_private_key=keyfile_path, internal_ip=internal_ip, zone=zone,
-          disk_specs=disk_kwargs_list, password=password)
+          disk_specs=disk_kwargs_list, password=password,
```
ehankland: I think you only need this change since it should already accept flag_values without modifying the \_\_init\_\_ method

ehankland:
LGTM

PR #1096, opened 2016-09-09
Add --sysctl and --set_files

noahl:
These new flags let users configure guest Linux kernels, via setting
options in procfs and sysfs. Also record the configuration in the output
and update some tests.

### DIFF
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -232,9 +232,9 @@ class BaseLinuxMixin(virtual_machine.BaseOsMixin):
     def ConfigureSubtree(subtree):
       for name, value in subtree.iteritems():
         path_components.append(name)
-        if isinstance(value, basestring):
+        if not isinstance(value, dict):
           self.RemoteCommand('echo "%s" | sudo tee %s' %
-                             (value, posixpath.join(*path_components)))
+                             (str(value), posixpath.join(*path_components)))
         else:
           ConfigureSubtree(value)
         path_components.pop()
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -39,6 +39,7 @@ import yaml
 from perfkitbenchmarker import disk
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
+from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import linux_packages
 from perfkitbenchmarker import os_types
 from perfkitbenchmarker import virtual_machine
@@ -72,6 +73,28 @@ WAIT_FOR_COMMAND = 'wait_for_command.py'
 flags.DEFINE_bool('setup_remote_firewall', False,
                   'Whether PKB should configure the firewall of each remote'
                   'VM to make sure it accepts all internal connections.')
+
+flag_util.DEFINE_yaml('procfs_config', {},
+                      'Values to write in procfs. This flag should be a tree '
+                      'of dictionaries and strings corresponding to the folder '
+                      'structure of procfs. Strings will be written to the '
+                      'corresponding file. For example, if you pass '
+                      '--procfs_config=\'{"sys": {"vm": '
+                      '{"dirty_background_ratio": "10"}}}\','
+                      'then PKB will write "10" to '
+                      '/proc/sys/vm/dirty_background_ratio before starting the '
+                      'benchmark.')
+
+flag_util.DEFINE_yaml('sysfs_config', {},
+                      'Values to write in sysfs. This flag should be a tree of '
+                      'dictionaries and strings corresponding to the folder '
+                      'structure of sysfs. Strings will be written to the '
+                      'corresponding file. For example, if you pass '
+                      '--sysfs_config=\'{"kernel": {"mm": '
+                      '{"transparent_hugepage": {"enabled": "always"}}}}\','
+                      'then PKB will write "always" to '
+                      '/sys/kernel/mm/transparent_hugepage/enabled before '
+                      'starting the benchmark.')
 
 
 class BaseLinuxMixin(virtual_machine.BaseOsMixin):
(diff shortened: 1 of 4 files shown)


noahl:
@yuyantingzero @ehankland @hildrum @cwilkes please take a look.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -198,8 +221,29 @@ def PrepareVMEnvironment(self):
...
 
+  def ConfigureVMKernel(self):
+    """Apply {procfs,sysfs}_config to VM."""
+
+    path_components = None
+
+    def ConfigureSubtree(subtree):
+      for name, value in subtree.iteritems():
+        path_components.append(name)
+        if isinstance(value, basestring):
```
cwilkes: PyYaml will automagically cast string numbers into numbers for you:

```
$ cat /tmp/sample.yaml 
a: {b: c}
d: {e: 4}

In [6]: y=yaml.load(open('/tmp/sample.yaml'))

In [7]: y
Out[7]: {'a': {'b': 'c'}, 'd': {'e': 4}}
```

Note that the value for "e" is the number 4.  maybe check that the value is *not* another dict?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -198,8 +221,29 @@ def PrepareVMEnvironment(self):
...
+          self.RemoteCommand('echo "%s" | sudo tee %s' %
```
cwilkes: Do we have a way to mock the RemoteCommand so that we can get back what will be run?  That might be too much testing ... but it shouldn't be that brittle.

perfkitbenchmarker/linux_virtual_machine.py
noahl: My original plan was that users would explicitly enter a string value, like {"e": "4"} in your example. That would make the value they enter fit with what we're actually going to do, which is write out the value as a string.

However, putting quotes around integers is annoying. I will implement the non-dict check you suggested and `str`-convert the output to strings. Thanks!

perfkitbenchmarker/linux_virtual_machine.py
noahl: We do! I added a few tests.

noahl:
@cwilkes I just pushed a new commit which I think addresses both of your comments. Please take a look.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -198,8 +221,29 @@ def PrepareVMEnvironment(self):
...
+    """Apply {procfs,sysfs}_config to VM."""
+
+    path_components = None
+
+    def ConfigureSubtree(subtree):
+      for name, value in subtree.iteritems():
+        path_components.append(name)
+        if not isinstance(value, dict):
+          self.RemoteCommand('echo "%s" | sudo tee %s' %
+                             (str(value), posixpath.join(*path_components)))
```
cwilkes: nit: you shouldn't have to do str(x) as the %s will do that for you.  But no need to change just for that.

perfkitbenchmarker/publisher.py
```diff
@@ -185,6 +185,9 @@ def AddMetadata(self, metadata, benchmark_spec):
           for key, value in data_disk.metadata.iteritems():
             metadata[name_prefix + 'data_disk_0_' + key] = value
```
cwilkes: Not related to this CR but the "key" in here could be an integer and so this should really be done with in the %s formatting way otherwise an error will be thrown about adding a string and an integer.

tests/linux_virtual_machine_test.py
```diff
@@ -0,0 +1,96 @@
...
+    vm = LinuxVM()
+
+    self.mocked_flags.procfs_config = {'sys': {'vm':
+                                               {'dirty_background_ratio': 10,
+                                                'dirty_ratio': 50}}}
+    self.mocked_flags.sysfs_config = {}
+    with mock.patch.object(vm, 'RemoteCommand') as remote_command:
+      vm.ConfigureVMKernel()
+
+    self.assertEqual(remote_command.call_args_list,
```
cwilkes: Awesome, thanks for adding.  Hopefully the order of these mocks is not important but we can fix that if it comes up.

perfkitbenchmarker/linux_virtual_machine.py
cwilkes: What happens if that path does not exist?  Ie "echo foo > /proc/foo/bar.txt" where /proc/foo/ does not exist as a directory?  I'm not sure if this is even a case -- if the user passes in a value to set that the OS does not know about (the "foo" in this case) should we error out?

noahl: Thanks for pointing out the `%s` conversion.

If the path does not exist, I expect the remote command will return a nonzero status code, which will make `RemoteCommand` raise an exception.

perfkitbenchmarker/publisher.py
noahl: Thanks for pointing that out! I wrote that code. Right now `key` will always be a string, but I see your point that it is more robust to use `%s`-conversion. However, that's a separate PR. If you write it, I will LGTM :)

tests/linux_virtual_machine_test.py
noahl: Ah, good point. The order of the args actually is important, and right now it depends on the order of entries in a dictionary, which is very brittle. I wrote it thinking that using an OrderedDict would be more trouble than it was worth. However, your comment made me realize there is a better approach of converting the call_args_list to a set. I will do that. Thanks!

noahl:
@cwilkes I pushed another new commit. This reflects my responses to you above, and should make the tests much cleaner and easier to read. Thanks for the reviews! Please take a look.

cwilkes:
The changes look good to me, but someone else should lgtm it

noahl:
I changed to a new and better interface thanks to comments from @hildrum and @ehankland . Please take a look.

perfkitbenchmarker/publisher.py
```diff
@@ -185,6 +185,9 @@ def AddMetadata(self, metadata, benchmark_spec):
...
 
+    metadata['set_files'] = ','.join(FLAGS.set_files)
```
ehankland: I think it might be nicer not to add a metadata key unless we actually modified something (set a file or sysctl) since in the vast majority of cases things wont be modified

noahl: Changed.

noahl:
@ehankland I pushed a new version that doesn't add the metadata keys if the values would be empty. Please take a look.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -198,8 +215,23 @@ def PrepareVMEnvironment(self):
...
+
+    for pair in FLAGS.set_files:
+      path, value = pair.split('=')
+      self.RemoteCommand('echo "%s" | sudo tee %s' %
+                         (value, path))
+
+  def DoSysctls(self):
+    """Apply --sysctl to the VM."""
+
+    self.RemoteCommand('sudo sysctl -w %s' % (' '.join(FLAGS.sysctl),))
```
ehankland: Does this break with no sysctls specified? Even if it doesn't probably don't want to execute an unnecessary command

noahl: Good point. Fixed.

noahl:
@ehankland I made it not run sysctl with no variables. Please take a look.

ehankland:
LGTM

Issue #1094, opened 2016-09-08
fio fails to build when using Ubuntu 16.04


meteorfox:
I'm using a recent version of the Ubuntu 16.04 LTS image on OpenStack, and FIO is failing to build with the following error:

```
Makefile:287: recipe for target 'lib/libmtd_legacy.o' failed
STDERR: Warning: Permanently added '10.50.50.55' (ECDSA) to the list of known hosts.
FIO_VERSION = fio-2.2.10
In file included from lib/libmtd_legacy.c:38:0:
lib/libmtd.h:288:8: error: unknown type name 'uint8_t'
        uint8_t mode);
        ^
lib/libmtd.h:305:4: error: unknown type name 'uint64_t'
    uint64_t start, uint64_t length, void *data);
    ^
lib/libmtd.h:305:20: error: unknown type name 'uint64_t'
    uint64_t start, uint64_t length, void *data);
                    ^
lib/libmtd.h:322:5: error: unknown type name 'uint64_t'
     uint64_t start, uint64_t length, void *data);
     ^
lib/libmtd.h:322:21: error: unknown type name 'uint64_t'
     uint64_t start, uint64_t length, void *data);
                     ^
make: *** [lib/libmtd_legacy.o] Error 1
```
After I quick Google of the error, this came up
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=815735

Apparently, the fix is already in fio (`#include <stdint.h>`)

## Steps to recreate

1. Add official Ubuntu Xenial 16.04 image (http://docs.openstack.org/image-guide/obtain-images.html) as `ubuntu-1604-lts`
2. Run `fio` benchmark against, in this case, OpenStack provider
```bash
python pkb.py --cloud=OpenStack --machine_type=m1.medium --image='ubuntu-1604-lts'\
              --openstack_network=private --openstack_floating_ip_pool=public \
              --benchmarks=fio
```

meteorfox:
According to the bug report linked above, this issue was fixed iin version fio/2.6-1

meteorfox:
The tag `fio-2.7` is the one that contains the fix.

See https://github.com/axboe/fio/commit/d7bb6180f831091c468e5aa749b142efd5eddda4 for more info

PR #1097, opened 2016-09-09
fio: Bump Git tag to fio-2.7

meteorfox:
On modern Linux kernels  fio-2.2.10 fails to build from
source due to missing a header file stdint.h that used to be included in
mtd-user.h. fio-2.7 tag includes the fix for this issue.

For more info on the bug see:
https://github.com/axboe/fio/commit/d7bb6180f831091c468e5aa749b142efd5eddda4

Addresses issue #1094

### DIFF
--- a/perfkitbenchmarker/linux_packages/fio.py
+++ b/perfkitbenchmarker/linux_packages/fio.py
@@ -23,7 +23,7 @@ from perfkitbenchmarker import vm_util
 
 FIO_DIR = '%s/fio' % vm_util.VM_TMP_DIR
 GIT_REPO = 'http://git.kernel.dk/fio.git'
-GIT_TAG = 'fio-2.2.10'
+GIT_TAG = 'fio-2.7'
 FIO_PATH = FIO_DIR + '/fio'
 FIO_CMD_PREFIX = '%s --output-format=json' % FIO_PATH
 SECTION_REGEX = r'\[(\w+)\]\n([\w\d\n=*$/]+)'


meteorfox:
Here's a [side-by-side comparison of fio-2.2.10 vs 2.7.](http://0754139ce78ddd85f423-69c99af980606451da8f30552ebc67a9.r69.cf2.rackcdn.com/fio-2.2-10_vs_fio-2.7.html)

noahl:
LGTM.

I'm wondering if we should pin a specific fio version or just always use the latest, but that's a separate question than this PR. Thanks for fixing this so quickly!

meteorfox:
There are later versions of fio available, I choose the earliest version that included the fix, in this case, fio-2.7. Not sure if we should just update it to latest or pick the earliest that solves our issue.

As for your question, I think we should pin to a specific version of a workload for a given PKB version, that way you know you will be able to reproduce the benchmark(s) if we leave it floating it might introduce regressions. PKB is supposed to be a "living benchmark", so we should also not refrain from making updates when necessary especially for bug fixes and features that we might want to include but my suggestion is to bump the PKB version as well if we do so.

For every time we might want to update a workload, IMHO, we should analyze side-by-side comparisons of the workload to make sure it doesn't break on an older, but still supported, OS and that there aren't obvious regressions or significant changes in performance between the versions.

noahl:
@meteorfox leaving the version fixed for a specific PKB version makes sense to me. That just means continuing with our current approach, so I am happy. Feel free to merge this PR any time!

meteorfox:
Thanks!

PR #1101, opened 2016-09-13
Use python string formatting 

cwilkes:
When I first came across this code I thought "key" was going to be an integer based on the "_0_" that was going to be prepended to it when making a key for the hashmap.  It probably is a string normally as this hasn't caused a problem but it is better to be safe than sorry.

I decided to go with the minimal change to cause less distractions.  There's a lot of `metadata[name_prefix + somekey] = somevalue` in this method which could be simplified but later on.


noahl:
Thanks for doing this! The code LGTM, but please fix the CLA issue. It's probably related to either what email address is in the author field of the commits or what email addresses are associated with your GitHub account.

PR #1103, opened 2016-09-13
Use string formatting instead of addition to concatenate

cwilkes:
First pass through I thought "key" would be an integer which would throw an exception if added to a string.  Moved to use python string formatting to safely prevent future problems.


noahl:
LGTM

PR #1109, opened 2016-09-19
Add support for AWS dedicated hosts

ehankland:


perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -97,6 +97,100 @@ def IsPlacementGroupCompatible(machine_type):
...
+
+  def _Delete(self):
+    delete_cmd = util.AWS_PREFIX + [
+        'ec2',
+        'release-hosts',
+        '--region=%s' % self.region,
+        '--host-ids=%s' % self.id]
+    vm_util.IssueCommand(delete_cmd)
+
+  @vm_util.Retry()
```
hildrum: Why are you retrying exists?

ehankland: Because if stdout isn't valid json I want to retry instead of simply blowing up

hildrum:
LGTM.

PR #1112, opened 2016-09-20
Use Azure Resource Management Mode

noahl:
Use Azure CLI in the (newer) resource management mode, rather than
the (older) service management mode. Besides changing all of our Azure
commands slightly, also
  - get rid of the Azure CLI version check, since we can now use the
  latest version like other providers do
  - simplify Azure's allocation of disk LUNs
  - support Blob-specific storage accounts, which are necessary for
  using Hot and Cool tier blobs
  - make OS disk handling more robust
  - improve lifecycle management for subsidiary resources like storage
    accounts

And very importantly, THIS IS A BREAKING CHANGE. Azure resource
management mode uses different names for regions, so this PR will
require changing all Azure commands. This should go in CHANGES.md for
the next release.

I've verified that the following benchmarks work:
  - iperf in-region
  - fio in-region
  - object storage in-region
  - iperf cross-region


noahl:
@ehankland @yuyantingzero @hildrum @cwilkes please take a look.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -104,80 +109,79 @@ def __init__(self, disk_spec, vm_name, machine_type):
...
     # compute the right offset for the alphabetical ordering, but it's
     # simpler and more resilient to just search the array, and the
     # size is limited to 15 so it will never take too long.
-    data_disk = (disk for disk in response['DataDisks']
-                 if int(disk.get('logicalUnitNumber', 0)) == self.lun).next()
-    assert ((self.lun == 0 and 'logicalUnitNumber' not in data_disk)
-            or (self.lun == int(data_disk['logicalUnitNumber'])))
-    self.name = data_disk['name']
+    data_disk = (disk for disk in response
+                 if int(disk.get('lun', 0)) == self.lun).next()
```
cwilkes: Is this generator always going to have a response?  The case I'm thinking about is if nothing passes the filter.  Looks like a StopIteration will be thrown:
``` r = (_ for _ in range(10) if _ > 11).next()```
which might be what you want as the `assert` will throw an exception if data_disk isn't what  you want.

File: perfkitbenchmarker/providers/azure/azure_network.py [hunk 2]
```diff
@@ -21,264 +21,284 @@
...
-    _AzureEndpoint(vm.name, port, 'udp').Create()
-
-  def DisallowAllPorts(self):
-    """Closes all ports on the firewall."""
-    pass
+def GetResourceGroup():
+  """Get the resource group for the current benchmark."""
+  spec = context.GetThreadBenchmarkSpec()
+  if hasattr(spec, 'azure_resource_group'):
+    return spec.azure_resource_group
```
cwilkes: I've seen people say that try/except is more pythonic, but I could go either way.  It stops you from having to repeat yourself:
```
try:
  return spec.azure_resource_group
except:
  group = ....
```
I personally do that now as I've had to change code with the attribute name and didn't like having to do it in two places

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
+
+  def AllowPort(self, vm, port):
+    """Opens a port on the firewall.
+
+    Args:
+      vm: The BaseVirtualMachine object to open the port for.
+      port: The local port to open.
+    """
+    if vm.is_static or port == SSH_PORT:
+      return
```
cwilkes: Is this supposed to do something?

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -104,80 +109,79 @@ def __init__(self, disk_spec, vm_name, machine_type):
...
+        raise errors.Resource.RetryableCreationError(
+            'Error creating Azure disk.')
+
       self.created = True
 
+    @vm_util.Retry()
+    def GetName():
+      disk = self._GetDiskJSON()
+      self.name = disk['name']
+    GetName()
```
cwilkes: Did you mean to have `return self.name` here?  Maybe I'm confused at what is going on.

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
+               kind=None, access_tier=None):
     super(AzureStorageAccount, self).__init__()
-    self.name = name
     self.storage_type = storage_type
-    self.affinity_group_name = affinity_group_name
+    self.name = name
+    self.resource_group = GetResourceGroup()
+    self.location = location
+    self.kind = kind or 'Storage'
+    self.access_tier = (access_tier or 'Hot') if kind == 'BlobStorage' else None
```
cwilkes: Hrm, this looks somewhat complicated.  If someone only passes in an `access_tier` they have to pass in `kind='BlobStorage'` in order to get `self.access_tier` to be set to a non-null?  I suppose that's okay but maybe we should throw an exception if we're setting something to None when they might think it will be set?

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: The `GetName` function is trying to set `self.name` to the correct value. The reason it's a function is so that I can wrap it in `vm_util.Retry()`. There's no need to return it - once the variable is set, that's it.

However, your question made me realize that I can get the same retry behavior in a much cleaner way by changing `_GetDiskJSON`. Thank you! I will do that.

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: If the generator does not find a value, then a) Azure has changed its behavior or output format and b) we have no sensible way to continue, so raising an exception is the best we can do. Looking at this again, the `assert` is completely redundant with the generator, so I will remove it.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Changed, thanks. It does look nicer this way. But I still have the attribute name in two places - am I doing something wrong?

perfkitbenchmarker/providers/azure/azure_network.py
noahl: You're right, thanks! Added an assert and hopefully made the code a little more clear.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: I don't know. It looks like diff didn't pick it up, but this is actually old code. I haven't touched it.

perfkitbenchmarker/providers/azure/azure_blob_storage.py
```diff
@@ -38,38 +39,34 @@ class AzureBlobStorageService(object_storage_service.ObjectStorageService):
...
-         '-l', location or DEFAULT_AZURE_REGION,
-         self.storage_account])
-
-    output, _, _ = vm_util.IssueCommand(
-        ['azure', 'storage', 'account',
-         'keys', 'list', self.storage_account])
-
-    self.azure_key = regex_util.ExtractExactlyOneMatch(
-        r'Primary:* (.+)', output)
+    prefix = 'pkb%sabs' % FLAGS.run_uri
```
noahl: abs is my abbreviation for "Azure Blob Storage". I've been trying to give my accounts semi-semantic names. (Semi because the names are very limited.)

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
 
 class AzureVirtualNetwork(resource.BaseResource):
   """Object representing an Azure Virtual Network."""
 
-  def __init__(self, name):
+  def __init__(self, location, name):
     super(AzureVirtualNetwork, self).__init__()
     self.name = name
+    self.resource_group = GetResourceGroup()
+    self.address_space = '10.0.0.0/8'
```
noahl: There will be one virtual network per zone, and it will include all of the resources for that zone. `10.0.0.0/8` is the largest possible address space for a virtual network, so it's sort of logical to just take it all. Would you consider that a justification? (And virtual networks in different zones have distinct address spaces, so there shouldn't be an overlap issue.)

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
-  def DisallowAllPorts(self):
-    """Closes all ports on the firewall."""
-    pass
+def GetResourceGroup():
+  """Get the resource group for the current benchmark."""
+  spec = context.GetThreadBenchmarkSpec()
+  if hasattr(spec, 'azure_resource_group'):
+    return spec.azure_resource_group
+  else:
+    group = AzureResourceGroup('pkb%s%s' % (FLAGS.run_uri, spec.uid))
```
noahl: I did it this way because the name will be a prefix of the name of the storage account, and storage account names can't have any characters except for lower-case letters and numbers. But no, I really want a hyphen. I'll go back and fix that up.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Oh, my mistake. This function originally contained two more lines, which constructed `AzureEndpoint` objects. I took them out because endpoints no longer exist, but I didn't clean up the function. Let me take a look at this.

hildrum:
Are you going to support [AddMetadata](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/virtual_machine.py#L298)?

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
-    except ValueError:
-      return False
-    return True
+    _, _, retcode = vm_util.IssueCommand(
+        [azure.AZURE_PATH, 'resource', 'list', self.name])
+
+    return retcode == 0
+
+  def _Delete(self):
+    vm_util.IssueCommand(
```
voellm: When you delete an Azure resource group it will delete all resource in the group.  This should make cleanup a lot easier and more reliable.

perfkitbenchmarker/providers/azure/azure_network.py
voellm: /8 seems limiting.  You might want to make it a /16.  There are situations where more than 255 VM's are run.

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,284 @@
...
+    Args:
+      vm: The BaseVirtualMachine object to open the port for.
+      port: The local port to open.
+    """
+    if vm.is_static or port == SSH_PORT:
+      return
+
+  def DisallowAllPorts(self):
+    """Closes all ports on the firewall."""
+    pass
```
voellm: You might want to add a log message here staying this is not implemented.  Seems kind of dangerous to ignore a security request.

voellm:
Thanks for taking this on.  ARM is a lot more reliable than ASM and it should help about the issues of left over resources and locked storage devices.

README.md
```diff
@@ -363,30 +363,32 @@ $ aws configure
...
+$ azure vm list
 ```
 
-Test that `azure` is installed correctly:
+Finally, make sure Azure is in Resource Management mode and that your account is
+authorized to allocate VMs and networks from Azure:
 
 ```bash
-$ azure vm list
+$ azure config mode arm
```
voellm: I did not see a check in the code that verifies that azure is in arm mode.  You might want to add a check and notify the user to make the switch.  Running without the check will lead to lots of confusing error messages.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: @voellm I don't understand. A /8 is more address space than a /16. Can I only connect 255 VMs to a subnet regardless of address space?

perfkitbenchmarker/providers/azure/azure_network.py
noahl: @voellm I think the desired end state is that PKB makes a graph of all resources it wants to allocate, then constructs a resource template for them and Azure allocates them all at once. I was trying to stick to PKB's normal deployment model until we could move to this option. Are you suggesting that I just not implement `_Delete` for Azure resources since the resource group will get them?

perfkitbenchmarker/providers/azure/azure_network.py
noahl: @voellm I'm planning to implement NSGs instead, so I won't have to ignore it. It looks to me from the docs that the typical architecture is an NSG per subnet, so that's my plan. Does that sound about right to you?

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Update: I implemented firewalls with Azure Network Security Groups. It took a while, but I think it's the right solution to this comment.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Update: I moved to /16s so each subnet would have non-overlapping internal address spaces. @ehankland pointed out that that would eliminate a possible class of errors. That might have been what you were talking about too, but I didn't understand.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Update: NSGs are implemented.

README.md
noahl: Done.

noahl:
@cwilkes @hildrum @voellm I pushed some new commits which I think address all of your comments. Please take a look.

noahl:
Oops, I had a whitespace fix that I never checked in. I pushed that, and it should now pass tests on Travis. Please take a look.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -104,80 +109,73 @@ def __init__(self, disk_spec, vm_name, machine_type):
...
+          [azure.AZURE_PATH, 'vm', 'disk', 'attach-new',
+           '--host-caching', self.host_caching,
+           self.vm_name, str(self.disk_size)] +
+          self.resource_group.args)
+
+      if retcode:
+        raise errors.Resource.RetryableCreationError(
+            'Error creating Azure disk.')
+
       self.created = True
```
tedsta: Why do some Resource._Create implementations do `self.created = True` while others don't (like `AzureResourceGroup`)? Is it because `AzureResourceGroup` doesn't support creation retries?

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,428 @@
...
+           '--quiet',
+           rule] +
+          self.resource_group.args +
+          self.args)
+
+  def AllowPort(self, vm, port):
+    with self.rules_lock:
+      if port in self.rules:
+        return
+      if self.deny_all_mode:
```
tedsta: Having `AllowPort` silently override a previous `DisallowAllPorts` makes me feel a little uneasy. My intuition would make me expect that all of the ports except this port I'm allowing should still be closed. But meh, I doubt this will cause any issues.

perfkitbenchmarker/providers/azure/azure_network.py
```diff
@@ -21,264 +21,428 @@
...
+
+  def DeleteRule(self, rule):
+      vm_util.IssueRetryableCommand(
+          [azure.AZURE_PATH, 'network', 'nsg', 'rule', 'delete',
+           '--quiet',
+           rule] +
+          self.resource_group.args +
+          self.args)
+
+  def AllowPort(self, vm, port):
```
tedsta: I changed the API of AllowPort in both `BaseVirtualMachine` and `BaseFirewall` to to have `(..., start_port, end_port=None,...)` to support optional port ranges in #1141. Could you add port range support here and in `AzureFirewall.AllowPort`?

noahl: Done.

perfkitbenchmarker/providers/azure/azure_disk.py
noahl: Hmm. That's carried over from old code, but I don't think we need it any more. Originally there was some complicated logic where we used `self.created` and maybe another variable to figure out whether a disk was an Azure image, in which case we can't actually create or delete it. I replaced that with `self.is_image`. I might have left this in by accident.

perfkitbenchmarker/providers/azure/azure_network.py
noahl: Your interpretation of the methods makes sense to me, too. Changed. (I'm guessing that when the interface was made, we assumed that firewalls defaulted to deny-all. That's not true right now, but maybe PKB should enforce it. That's a separate change, anyway.)

noahl:
@tedsta I pushed a new commit that I think addresses all of your comments. Please take a look.

perfkitbenchmarker/providers/azure/azure_disk.py
```diff
@@ -127,7 +127,7 @@ def _Create(self):
         raise errors.Resource.RetryableCreationError(
             'Error creating Azure disk.')
 
-      self.created = True
+      # self.created = True
```
tedsta: Forgot to remove this after testing that it worked without it? :)

noahl: Yes, exactly :). Removed, thanks for catching it!

tedsta:
LGTM

PR #1123, opened 2016-09-27
Object storage multistream write

tedsta:
Added a api_multistream_writes scenario to object_storage_service benchmark where only write latency and throughput is measured (no reads).

Summary:
- Copy and paste MultiStreamRWBenchmark into MultiStreamWriteBenchmark. Remove the bottom section that did the reads.
- Refactor out the now-redundant nested _RunMultiStreamProcesses and _StartMultiStreamProcess closures.
- Nest _StartMultiStreamProcess under _RunMultiStreamProcess so that fewer parameters need to be passed to it.
- Refactor out the now-redundant redundant multistream writes code into a separate _MultiStreamWrites function.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -78,9 +78,9 @@ flags.DEFINE_enum('object_storage_scenario', 'all',
                   'api_namespace: runs API based benchmarking for namespace '
                   'operations. \n'
                   'api_multistream: runs API-based benchmarking with multiple '
-                  'upload/download streams.'
+                  'upload/download streams.\n'
                   'api_multistream_writes: runs API-based benchmarking with '
-                  'multiple upload streams')
+                  'multiple upload streams.')
 
 flags.DEFINE_enum('cli_test_size', 'normal',
                   ['normal', 'large'],
@@ -802,6 +802,7 @@ def _RunMultiStreamProcesses(vms, command_builder, cmd_args,
      until they complete.
 
   Args:
+    vms: the VMs to run the benchmark on.
     command_builder: an APIScriptCommandBuilder.
     cmd_args: arguments for the command_builder.
     streams_per_vm: number of threads per vm.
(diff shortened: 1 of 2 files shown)


tedsta:
@noahl @ehankland @hildrum Please take a look

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -70,15 +70,17 @@
...
+                   'api_multistream', 'api_multistream_writes'],
                   'select all, or one particular scenario to run: \n'
                   'ALL: runs all scenarios. This is the default. \n'
                   'cli: runs the command line only scenario. \n'
                   'api_data: runs API based benchmarking for data paths. \n'
                   'api_namespace: runs API based benchmarking for namespace '
                   'operations. \n'
                   'api_multistream: runs API-based benchmarking with multiple '
-                  'upload/download streams.')
+                  'upload/download streams.'
```
noahl: Looks like you need an explicit '\n' at the end of this line.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -901,6 +926,46 @@ def RunMultiStreamProcesses(command):
...
+    regional_bucket_name: the secondary bucket to benchmark.
+
+  Raises:
+    ValueError if an unexpected test outcome is found from the API
+    test script.
+  """
+
+  logging.info('Starting multi-stream write test on %s VMs.', len(vms))
+
+  objects_written_file = posixpath.join(vm_util.VM_TMP_DIR,
```
noahl: Do you need to pass this argument to the API test script when you're not planning on using the output? I can't remember how it works, but I don't think it *should* be necessary.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -793,68 +795,52 @@ def LoadWorkerOutput(output):
...
+  for thread in threads:
+    thread.start()
+  logging.info('Started %s processes.', num_streams)
+  for thread in threads:
+    thread.join()
+  logging.info('All processes complete.')
+  return output
+
+
+def _MultiStreamWrites(results, metadata, vms, command_builder,
```
noahl: Possible suggestion: merge the code in this function into `_RunMultiStreamProcesses`, adding a few more configuration parameters to that function (you may need to make an object holding the configuration just to make it all readable). That should simplify the read part of `MultiStreamRWBenchmark` as well as the write part. What do you think?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
tedsta: Fixed

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
tedsta: Ended up going with a single function that handles both reads and writes

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
tedsta: You're right, don't need it. But it'd add complexity to remove it for only the api_multistream_writes scenario. I guess there's a tiny bit of overhead writing all those names to the file, but it should be negligible, right?

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -793,27 +795,61 @@ def LoadWorkerOutput(output):
...
+    vms: the VMs to run the benchmark on.
+    command_builder: an APIScriptCommandBuilder.
+    cmd_args: arguments for the command_builder.
+    streams_per_vm: number of threads per vm.
+    num_streams: total number of threads to launch.
+  """
+
+  output = [None] * num_streams
+
+  def _StartMultiStreamProcess(proc_idx):
```
noahl: Suggestion: rename this to `RunOneProcess`. (I know I chose the old name, my fault.)

PR #1124, opened 2016-09-29
Automatically use amazon linux if os_type is rhel

ehankland:
Also use client token when creating VM for idempotency


yuyantingzero:
LGTM

PR #1126, opened 2016-09-30
Enable multiple configs for the same benchmark in the same file

ehankland:


hildrum:
LGTM, but do you need a test?

hildrum:
Also, an example config would be super-nice.

ehankland:
Test added

PR #1129, opened 2016-09-30
Add a flag that allows run retries

ehankland:


perfkitbenchmarker/pkb.py
```diff
@@ -376,12 +382,20 @@ def DoRunPhase(benchmark, name, spec, collector, timer):
   """
   deadline = time.time() + FLAGS.run_stage_time
   run_number = 0
+  consecutive_failures = 0
   while True:
+    samples = []
     logging.info('Running benchmark %s', name)
```
yuyantingzero: Add logging about current retry count?

perfkitbenchmarker/pkb.py
```diff
@@ -376,12 +382,20 @@ def DoRunPhase(benchmark, name, spec, collector, timer):
...
     events.before_phase.send(events.RUN_PHASE, benchmark_spec=spec)
     try:
       with timer.Measure('Benchmark Run'):
         samples = benchmark.Run(spec)
+    except Exception:
+      consecutive_failures += 1
+      if consecutive_failures > FLAGS.run_stage_retries:
+        raise
+    else:
+      consecutive_failures = 0
```
yuyantingzero: why do you need line 397,398?

ehankland: Because I only want to fail on consecutive failures not total failures

cwilkes: the if/else whitespacing looks off.

Maybe put in the exception the reason why it failed, ie
```
raise Exception('Failed more then %d times' % (FLAGS.run_stage_retries, ))
```

ehankland: The else matches the except (not the if). Sorry - hadn't pushed latest changes - the latest logs exceptions on every retry

perfkitbenchmarker/pkb.py
ehankland: Added logging on every failure below

PR #1132, opened 2016-10-05
Include project in metadata for GCP

hildrum:


hildrum:
@ehankland, PTAL.  Note that this removes the 'preemptible' key as a top level metadata key, but I think this is okay, because it's still part of the machine type information.

perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -264,12 +263,14 @@ def __init__(self, vm_spec):
     self.max_local_disks = vm_spec.num_local_ssds
     self.memory_mib = vm_spec.memory
     self.preemptible = vm_spec.preemptible
-    self.project = vm_spec.project
+    if vm_spec.project:
```
ehankland: self.project = vm_spec.project or util.GetDefaultProject()

PR #1134, opened 2016-10-06
Add config matrices

ehankland:


tests/benchmark_sets_test.py
```diff
@@ -29,6 +29,20 @@
   flags:
     ip_addresses: INTERNAL
 """
+MATRIX_CONFIG = """
+netperf:
+  test_matrix: GCP
```
hildrum: I think better names might be possible here.  Maybe `flag_set` and `flag_sets`?  Or `flag_set` and `flag_set_definitions`?

hildrum:
LGTM.  I have a minor quibble with the naming--your choice on changing.

PR #1136, opened 2016-10-07
Terasort bugs

hildrum:
Allow dataproc submit job to go as long as it needs to.
Also, fix `_Exists` method for AWS security groups.


hildrum:
@noahl Even if this has been merged, please check to make sure the disk spec handling is okay.

hildrum:
@ehankland -- I know you already approved, but since then I pushed a couple more changes.  One is that I added handling for AWS disk specs.  Second, instead of a flag each for a generate dir, sort dir, and validate dir, I have one flag for the base path of those three. 

The only good reason to have three flags is if you are only going to run some stages, and that isn't supported.  When it is, someone can add flags for the generated dir, sorted dir, and validate dir.

ehankland:
LGTM

PR #1137, opened 2016-10-07
Add another flag for zones

ehankland:
This allows flag matrices to have two zone axes


hildrum:
I'm not seeing how this would be used.

hildrum:
LGTM, but please add a test, comment, or some sort of example in the [configuration wiki page]( https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/wiki/PerfKitBenchmarker-Configurations)

ehankland:
Updated the config wiki

PR #1139, opened 2016-10-08
Firewall port range

tedsta:
Azure doesn't seem to provide a straightforward way to do this. I'll look into it on Monday. I need this to open a big range of ports for the netperf benchmark.

@ehankland @cwilkes  @yuyantingzero PTAL

### DIFF
--- a/perfkitbenchmarker/providers/aws/aws_network.py
+++ b/perfkitbenchmarker/providers/aws/aws_network.py
@@ -51,7 +51,7 @@ class AwsFirewall(network.BaseFirewall):
     self.firewall_set = set()
     self._lock = threading.Lock()
 
-  def AllowPort(self, vm, port):
+  def AllowPort(self, vm, port, to_port=None):
     """Opens a port on the firewall.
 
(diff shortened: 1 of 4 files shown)


perfkitbenchmarker/providers/aws/aws_network.py
```diff
@@ -51,7 +51,7 @@ def __init__(self):
     self.firewall_set = set()
     self._lock = threading.Lock()
 
-  def AllowPort(self, vm, port):
+  def AllowPort(self, vm, port, to_port=None):
```
noahl: When I first saw the name `to_port`, I thought this was a firewall rule targeting connections with different source and dest ports. Maybe rename `port` to `start_port` and then `to_port` to `end_port`, or something like that?

tedsta: Yeah, I agree. Some of the other smaller providers already implemented this so I just kept their naming scheme, but I'll go ahead and change all of them.

PR #1140, opened 2016-10-08
Add Elasticsearch publisher

tedhtchang:
The Elasticsearch publisher automatically uploads test result to
the specified Elasticsearch server at the end of test.

### DIFF
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -15,11 +15,13 @@
 """Classes to collect and publish performance samples to various sinks."""
 
 import abc
+import copy
 import csv
 import io
 import itertools
 import json
 import logging
+import math
 import operator
 import pprint
 import sys
@@ -87,6 +89,15 @@ flags.DEFINE_string(
     'cloud_storage_bucket',
     None,
     'GCS bucket to upload records to. Bucket must exist.')
+
+flags.DEFINE_string(
+    'es_uri', None,
+    'The Elasticsearch address and port. e.g. http://localhost:9200')
+
+flags.DEFINE_string(
+    'es_index', 'perfkit', 'Elasticsearch index name to store documents')
+
+flags.DEFINE_string('es_type', 'result', 'Elasticsearch document type')
 
 flags.DEFINE_multistring(
     'metadata',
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/publisher.py
```diff
@@ -538,6 +548,87 @@ def PublishSamples(self, samples):
...
+    """Convert the floating epoch timestamp in micro seconds epoch_us to
+    yyyy-MM-dd HH:mm:ss.SSSSSS in string
+    """
+    ts = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(epoch_us))
+    num_dec = ("%.6f" % (epoch_us - int(epoch_us))).split('.')[1]
+    new_ts = '%s.%s' % (ts, num_dec)
+    return new_ts
+
+  def _deDotKeys(self, res):
+    """Recursively replace dot with underscore in all keys in a """
```
tedsta: Looks like an incomplete doc string. Assuming you meant "all keys in a dictionary"?

perfkitbenchmarker/publisher.py
```diff
@@ -538,6 +548,87 @@ def PublishSamples(self, samples):
...
+    for key, value in res.items():
+        if isinstance(value, dict):
+            self._deDotKeys(value)
+        if key.replace(".", "_") != key:
```
tedsta: Consider:

```
new_key = key.replace('.', '_')
if new_key != key:
  res[new_key] = res.pop(key)
```

perfkitbenchmarker/publisher.py
```diff
@@ -538,6 +548,87 @@ def PublishSamples(self, samples):
...
+      # as each ES's document's unique _id
+      es.create(index=self.es_index, doc_type=self.es_type,
+                id=sample['sample_uri'], body=json.dumps(sample))
+
+  def _HumanReadableTimestamp(self, epoch_us):
+    """Convert the floating epoch timestamp in micro seconds epoch_us to
+    yyyy-MM-dd HH:mm:ss.SSSSSS in string
+    """
+    ts = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(epoch_us))
+    num_dec = ("%.6f" % (epoch_us - int(epoch_us))).split('.')[1]
```
tedsta: Nitpick, but maybe using `floor(epoch_us)` to round down instead of `int(epoch_us)` would make this code slightly clearer

perfkitbenchmarker/publisher.py
```diff
@@ -31,6 +31,7 @@
 from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import version
 from perfkitbenchmarker import vm_util
+from elasticsearch import Elasticsearch
```
noahl: This will cause an import error on machines without the elasticsearch module installed. Please move this import statement inside of `ElasticsearchPublisher` instead.

perfkitbenchmarker/publisher.py
```diff
@@ -538,6 +548,87 @@ def PublishSamples(self, samples):
...
+    es = Elasticsearch([self.es_uri])
+    if not es.indices.exists(index=self.es_index):
+      es.indices.create(index=self.es_index, body=self.mapping)
+      logging.info('Create index %s and default mappings', self.es_index)
+    for sample in samples:
+      sample = sample.copy()
+      # Make timestamp understandable by ES and human.
+      sample['timestamp'] = self._HumanReadableTimestamp(sample['timestamp'])
+      # Keys cannot have dots for ES
+      sample = self._deDotKeys(sample)
```
tedsta: `self._deDotKeys(self, res)` modifies the dictionary you pass in rather than making a copy. Are we sure stuff happening after `PublishSamples` isn't going to rely on those original key names?

perfkitbenchmarker/publisher.py
noahl: Please complete the sentence :)

requirements.txt
```diff
@@ -21,3 +21,4 @@ PyYAML==3.11
 pint>=0.7
 numpy
 functools32
+elasticsearch>=2.0.0,<3.0.0
```
noahl: Please don't add `elasticsearch` to the requirements.txt file, because not all users need to install the module. Instead, add a note to `README.txt` that the `elasticsearch` module is required to use the elasticsearch publisher. Also, when you move the `elasticsearch` import inside of the elasticsearch publisher, make sure there's an understandable error message when the module isn't found.

perfkitbenchmarker/publisher.py
```diff
@@ -538,6 +548,87 @@ def PublishSamples(self, samples):
...
+      # Add sample to the "perfkit index" of "result type" and using sample_uri
+      # as each ES's document's unique _id
+      es.create(index=self.es_index, doc_type=self.es_type,
+                id=sample['sample_uri'], body=json.dumps(sample))
+
+  def _HumanReadableTimestamp(self, epoch_us):
```
noahl: Since it sounds like the conversion is required for Elasticsearch, maybe rename this to `_FormatTimestampForElasticsearch`? That makes it easier to understand the method's purpose at a glance.

requirements.txt
tedhtchang: Thanks! One question for the comment. 
Would it be better to also add a separate requirement file, i.e. elasticsearch-requirement.txt ?

noahl: @tedhtchang I don't think it's worth adding the file for a single module. If elasticsearch eventually requires several new modules to be installed, then we can add a requirements file.

besanradwan:
Hello,

I noticed that you stated on line 557 of publisher.py that the default es_uri parameter is set to "http://localhost:9200". However, on line 93 of the same file you do not do this, and es_uri is whatever it has been set to seeing as it is required to be set up in order to use the publisher (line 696 of publisher.py)

tedhtchang:
Thanks! I fixed it.

noahl:
Looks good to me. I will defer to @tedsta on merging.

PR #1145, opened 2016-10-12
Add blazemark.

yuyantingzero:


perfkitbenchmarker/linux_benchmarks/blazemark_benchmark.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2014 PerfKitBenchmarker Authors. All rights reserved.
```
ehankland: 2016

perfkitbenchmarker/linux_benchmarks/blazemark_benchmark.py
```diff
@@ -0,0 +1,95 @@
...
+    A list of sample.Sample objects with individual machine boot times.
+  """
+  vm = benchmark_spec.vms[0]
+  all_tests = blazemark.GetBinaries(vm)
+  requested = set(FLAGS.blazemark_set)
+  run = set(all_tests)
+  if 'all' not in FLAGS.blazemark_set:
+    if requested - run:
+      logging.warning(
+          'Binaries %s not available, skipping. Options are %s',
```
ehankland: It would be nice if we could give this error in checkprerequisites rather than here - how many tests are there in "all_tests" - would it be easy to keep a list of them somewhere so you wouldn't need to find all blazemark binaries at runtime?

perfkitbenchmarker/linux_packages/blaze.py
```diff
@@ -0,0 +1,67 @@
...
+  """
+  vm.PushDataFile(CONFIG, BLAZE_DIR)
+  # Adjust cache size
+  cache_in_KB, _ = vm.RemoteCommand(
+      'cat /proc/cpuinfo | grep "cache size" | awk \'{print $4}\'')
+  cache_in_B = int(1024 * sum([float(c) for c in cache_in_KB.split()]))
+  vm.RemoteCommand(
+      'sed -i \'s/constexpr size_t cacheSize = 3145728UL;/constexpr '
+      'size_t cacheSize = %sUL;/g\' %s' % (
+          max(cache_in_B, MAX_BLAZE_CACHE_SIZE_IN_B - 1), os.path.join(
```
ehankland: why "MAX_BLAZE_CACHE_SIZE_IN_B - 1" instead of just MAX_BLAZE_CACHE_SIZE_IN_B?

perfkitbenchmarker/linux_packages/blaze.py
```diff
@@ -0,0 +1,67 @@
...
+              BLAZE_DIR, 'blaze', 'config', 'CacheSize.h')))
+  vm.RemoteCommand('cd %s; ./configure %s; make' % (BLAZE_DIR, CONFIG))
```
ehankland: nit: I think chaining commands with && leads to easier to read error messages because it won't continue executing if one of the earlier steps fails

perfkitbenchmarker/linux_packages/blazemark.py
```diff
@@ -0,0 +1,206 @@
...
+
+def _ParseThroughput(out, test):
+  """Parse throughput section of blazemark results.
+
+  Args:
+    out: string. Blazemark output in raw string format.
+    test: string. Name of the test ran.
+
+  Returns:
+    A list of samples. Each sample if a 4-tuple of (benchmark_name, value, unit,
```
ehankland: why not return sample.Sample objects directly so you don't need to do the conversion later?

perfkitbenchmarker/linux_packages/blazemark.py
```diff
@@ -0,0 +1,206 @@
...
+      results.append((
+          '_'.join([test, lib, 'Throughput']),  # Metric name
+          float(v[1]),  # Value
+          unit,  # Unit
+          copy.deepcopy(metadata)))  # Metadata
+  logging.info('Results for %s:\n %s', test, results)
+  return results
+
+
+def _ParseResult(out, test):
```
ehankland: Why do you have this function in addition to _ParseThroughput?

perfkitbenchmarker/linux_packages/g++5.py
```diff
@@ -0,0 +1,34 @@
...
+
+def _Install(vm):
+  """Installs the g++-5 package on the VM."""
+  pass
+
+
+def YumInstall(vm):
+  """Installs the g++-5 package on the VM."""
+  # TODO: Figure out how to install gcc/g++5 with yum
+  _Install(vm)
```
ehankland: raise notimplemented instead of doing nothing?

perfkitbenchmarker/linux_benchmarks/blazemark_benchmark.py
ehankland: ditto for the other files

tests/data/blazemark-output.txt
```diff
@@ -0,0 +1,48 @@
+ Dense Vector/Dense Vector Addition:
```
ehankland: Did you forget to git add a test that uses this? I don't see this getting used anywhere

perfkitbenchmarker/data/Configfile_blaze
```diff
@@ -0,0 +1,92 @@
+#==================================================================================================
```
ehankland: Could you document somewhere the changes you made to these config files and why you made them? Put it where ever you think is appropriate - not necessarily in the files themselves

yuyantingzero:
@ehankland PTAL, addressed all comments

PR #1147, opened 2016-10-13
Remove scratch disk from netperf benchmark

tedsta:


tedsta:
Alright I've added that to my TODO list

PR #1151, opened 2016-10-14
Add unzip to linux_packages

peterfeiner:
Now we install unzip in the standard way via vm.Install('unzip').


peterfeiner:
Rolled into https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1152.

PR #1152, opened 2016-10-15
Fix packages

peterfeiner:
I'm working toward making a perfkitbenchmarker standalone installation script so the benchmarks can be run on nodes that don't have internet access. I've been slogging through making vm.Install(package) work for every package in linux_packages. Along with my other two pull requests, everything builds!


peterfeiner:
My pleasure.

Do I need to do something to submit the change?

yuyantingzero:
@peterfeiner , my bad. forget to merge your pr.

PR #1159, opened 2016-10-19
Make OpenStack provider work with Neutron floating IPs

shakhat:
OpenStack network management has switched from nova-network to Neutron
and all modern OpenStack setups are based on Neutron. This patch
switches floating IPs management to Neutron.

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_network.py
+++ b/perfkitbenchmarker/providers/openstack/os_network.py
@@ -173,3 +173,68 @@ class OpenStackFloatingIPPool(object):
                                              updated_floating_ip_dict['id'])
       del delete_cmd.flags['format']  # Command not support json output format
       stdout, stderr, _ = delete_cmd.Issue(suppress_warning=True)
+
+
+class OpenStackNeutronFloatingIPPool(object):
+
+  _floating_ip_lock = threading.Lock()  # Guards floating IP allocation/release
+
+  def __init__(self, pool_name):
+    self.ip_pool_name = pool_name
+
+  def associate(self, vm):
+    with self._floating_ip_lock:
+      floating_ip = self._get_or_create(vm)
+      cmd = utils.OpenStackCLICommand(vm, 'server', 'add', 'floating', 'ip',
+                                      vm.id, floating_ip['ip'])
+      del cmd.flags['format']  # Command does not support json output format
+      _, stderr, _ = cmd.Issue()
+      if stderr:
+        raise errors.Error(stderr)
+      return floating_ip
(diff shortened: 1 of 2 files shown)


cbhall:
My understanding is that the underlying change here is just to the `openstack` command.  For example, python-openstackclient version 2.6.0 accepts the `openstack ip floating` list/create/show/delete commands.  On the other hand, python-openstackclient 3.3.0 indicates those commands are deprecated and are replaced with `openstack floating ip` list/create/show/delete commands.  I have verified that on my OpenStack test system (older I'm afraid) the underlying OpenStack APIs those older and newer commands invoke is not changed.

As such, and given that the version of python-openstackclient used by PerfKit Benchmarker is carefully specified and controlled, why should there be a dynamic run-time check on whether the old deprecated commands can be used, or whether the new commands must be used?  Instead of supporting both, it appears you could just support the new commands and mandate that a newer version of python-openstackclient be installed.

As well, the proposed code here replaces the deprecated `openstack ip floating pool list` with `openstack network list`.  While it is true that a network list will make a check that the specified external network exists, it will not actually prove that the network is external.  Is there a way to use `openstack network show` and determine that the network is external (such as with the `router:external` field)?

An alternative is to use `openstack floating ip pool list` (which the `openstack` command indicates is the replacement for the deprecated `openstack ip floating pool list`.  However, I cannot get `openstack floating ip pool list` to work, so I can see why you may have rejected using that.  Should `openstack floating ip pool list` be working?

meteorfox:
@cbhall You are right. The problem really stems from the deprecated command `openstack ip floating pool list`, the newer versions of python-openstackclient (I believe starting from > 2.6) prints a warning on `stderr` that is being deprecated, the current code simply exits thinking is an error.

A quick fix is to simply check for the 'deprecated' warning, and if it is, then move on since the command succeeded in that case. Of course, that will only work until the OpenStack Ocata release.

Another solution would be, as you suggested, run `openstack network show <network_name_or_id>` and check if `router:external` property is `"External"` or `"Internal"` and drop Compute v2 support since it has been deprecated and its[ set to be removed in the 2.35 microversion update](http://developer.openstack.org/api-ref/compute/#floating-ip-pools-os-floating-ip-pools-deprecated).

s-deitz:

shakhat:
@s-deitz Most probably the original issue is covered by https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1265, but I do not have OpenStack now to verify this. So, since this PR hangs for too long, I'd suggest to close it.

PR #1160, opened 2016-10-19
Fix issue with GetConfig using flags

ehankland:
Flags set in user configs were not respected by GetConfig


yuyantingzero:
Consider adding test for GetBenchmarksFromFlags and _CreateBenchmarkRunList

PR #1171, opened 2016-10-27
ContainerizedDebianMixin: Ping to a fixed docker image and remove the sudo hack

wonderfly:
The fix in #1130 was not complete: if `sudo` existed in a piped command, e.g.,
`sudo command1 && sudo command2`, only the first `sudo` would be stripped.

We could try searching and removing every occurrence of `sudo`, but the root
cause of the breakage was that the base image we used, `ubuntu:latest`, kept
changing. Instead, we should ping to a fixed version, and only upgrade when
needed.

I verified that `ubuntu:trusty-20161006` had `sudo` installed and basic
benchmarks passed for `os_type=ubuntu_container`.

@ehankland Can you review?


wonderfly:
Thanks Eric!

PR #1173, opened 2016-10-28
Netperf thinktime flag

tedsta:
Patched netperf to include a "think time" flag, which, if specified, makes netserver traverse an array of configurable size in a random order. You can also specify what fraction of the array to shuffle.

One issue with this patch is that the system I've devised to ensure that all elements in the array get traversed no matter how we shuffle it is pretty slow. Specifically, it's O(n*m), where n is the size of the array and m is the number of swaps to perform. This puts a practical limit on how large our arrays can be. Shuffling a 1GB array takes an obscene amount of time. I'll look into finding a faster way to shuffle properly in the future.


tedsta:
@ehankland @yuyantingzero Please take a look

tedsta:
Dang that is one messy rebase for such a tiny change. Weird that it's showing commits that have already been merged into master.... Did I do it wrong?

tedsta:
This is weird. I'm going to close this and reopen.

Issue #1122, opened 2016-09-27
OpenStack "openstack_volume_size" not included in json results output


akrzos:
OpenStack provider specific flags such as "openstack_volume_size" are not making their way into the result json which could be misleading, since defaults make their way in.

Flag Values:
```
2016-09-27 13:46:35,551 browbeat MainThread INFO     Flag values:
--openstack_volume_size=5
--machine_type=m1.small
--os_type=rhel
--benchmarks=copy_throughput
--openstack_network=browbeat_private
--image=centos7
--cloud=OpenStack
--timing_measurements=runtimes
--copy_benchmark_mode=cp
--run_uri=browbeat
--openstack_floating_ip_pool=browbeat_public
--openstack_image_username=centos
```

And in the newline delimited results json:

```
{"labels": "|data_disk_0_type:standard|,|zone:nova|,|scratch_disk_type:standard|,|perfkitbenchmarker_version:v1.7.0|,|num_striped_disks:1|,|machine_type:m1.small|,|scratch_disk_size:500|,|data_disk_0_size:500|,|data_disk_0_num_stripes:1|,|image:centos7|,|cloud:OpenStack|,|vm_count:1|", "timestamp": 1474998635.938512, "metric": "cp throughput", "official": false, "value": 197.3035439137134, "test": "copy_throughput", "run_uri": "browbeat-eed78df5-5389-4b37-b717-ed062cea8d99", "owner": "stack", "product_name": "PerfKitBenchmarker", "unit": "MB/sec", "sample_uri": "3f3a66a2-7a1c-4be7-9378-1a9533518d92"}
```

Looking through the results, you might think there is a 500G disk vs just a 5G cinder volume.  Shouldn't "openstack_volume_size" be included under "labels"?

meteorfox:
Yes,  good catch.  

So, we could remove deprecate `openstack_volume_size` and use `data_disk_size`, and if `openstack_volume_size` is set then use it as the `data_disk_size`.

What do you think?

akrzos:
> Yes, good catch.
> 
> So, we could remove deprecate openstack_volume_size and use data_disk_size, and if openstack_volume_size is set then use it as the data_disk_size.
> 
> What do you think?

Thanks, For consistency with the other providers it sounds like `data_disk_size` would be the way to go. 


On another note, should `--data_disk_type=local` allow the use of my flavor's ephemeral disk?

meteorfox:
[...]

PR #1174, opened 2016-10-31
OpenStack: Deprecate openstack_volume_size

vhe182:
So as to standardize flags as with other providers, 'data_disk_size'
flag should be used instead of openstack_volume_size.  So as not to
break users' scenarios, openstack_volume_size will still be included
and used but if it is not set, data_disk_size will be used before
using the default.


vhe182:
Addresses Issue #1122

PR #1177, opened 2016-11-01
Make a few netperf changes

ehankland:
- Make netperf_num_streams a list so we can run with several different stream counts in a single run
- Don't report latency info for TCP_STREAM
- Add num_streams metadata to samples

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
@@ -32,6 +32,7 @@ from collections import Counter
 
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import data
+from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import sample
 from perfkitbenchmarker import vm_util
@@ -49,8 +50,9 @@ flags.DEFINE_integer('netperf_test_length', 60,
 flags.DEFINE_bool('netperf_enable_histograms', True,
                   'Determines whether latency histograms are '
                   'collected/reported. Only for *RR benchmarks')
-flags.DEFINE_integer('netperf_num_streams', 1,
-                     'Number of netperf processes to run.')
+flag_util.DEFINE_integerlist('netperf_num_streams', flag_util.IntegerList([1]),
+                             'Number of netperf processes to run. Netperf '
+                             'will run once for each value in the list.')
 
 ALL_BENCHMARKS = ['TCP_RR', 'TCP_CRR', 'TCP_STREAM', 'UDP_RR']
 flags.DEFINE_list('netperf_benchmarks', ALL_BENCHMARKS,
@@ -104,7 +106,7 @@ def Prepare(benchmark_spec):
   vms = vms[:2]
   vm_util.RunThreaded(PrepareNetperf, vms)
 
-  num_streams = FLAGS.netperf_num_streams
+  num_streams = max(FLAGS.netperf_num_streams)
 
   # Start the netserver processes
   if vm_util.ShouldRunOnExternalIpAddress():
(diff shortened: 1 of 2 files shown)


ehankland:
@tedsta would you mind taking a look?

tests/linux_benchmarks/netperf_benchmark_test.py
```diff
@@ -105,19 +105,7 @@ def testExternalAndInternal(self):
          ('TCP_CRR_Latency_max', 2500.0, 'us'),
          ('TCP_CRR_Latency_stddev', 551.07, 'us'),
          ('TCP_STREAM_Throughput', 1187.94, mbps),
-         ('TCP_STREAM_Latency_p50', 2.0, 'us'),
```
tedsta: Ah - I see why the `if unit != MBPS` was included above now. I was confused why TCP_STREAM had latency stats, but if you look in the output file that we read this data from, the stats are there. Do we know whether a human or a machine made that file?

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -240,18 +242,18 @@ def _ParseNetperfOutput(stdout, metadata, benchmark_name,
...
-      ('90th Percentile Latency Microseconds', 'p90'),
-      ('99th Percentile Latency Microseconds', 'p99'),
-      ('Minimum Latency Microseconds', 'min'),
-      ('Maximum Latency Microseconds', 'max'),
-      ('Stddev Latency Microseconds', 'stddev')]:
-    if metric_key in results:
-      latency_samples.append(
-          sample.Sample('%s_Latency_%s' % (benchmark_name, metric_name),
-                        float(results[metric_key]), 'us', metadata))
+  if unit != MBPS:
```
tedsta: I think this is redundant - the outer `if enable_latency_histograms` should prevent throughput benchmarks from making it here.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -388,25 +391,24 @@ def Run(benchmark_spec):
...
-      internal_ip_results = RunNetperf(client_vm, netperf_benchmark,
-                                       server_vm.internal_ip, num_streams)
-      for internal_ip_result in internal_ip_results:
-        internal_ip_result.metadata.update(metadata)
-        internal_ip_result.metadata['ip_type'] = 'internal'
-      results.extend(internal_ip_results)
+  for num_streams in FLAGS.netperf_num_streams:
+    assert(num_streams >= 1)
+
+    for netperf_benchmark in FLAGS.netperf_benchmarks:
```
tedsta: So the desired behavior here is to run each benchmark multiple times - each with a different `num_streams` value? Is this better than having each `num_streams` value associated with a single benchmark and then run netperf_benchmark multiple times?

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
ehankland: Nope - the for loop isn't inside the `if enable_latency_histograms` block - that's why the latency samples for TCP_STREAM started showing up after your change

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
ehankland: Yes - this makes it a lot easier to run with several different stream counts - otherwise we'd have to provision and tear down vms for each of the netperf_benchmark tests.

ehankland:
@tedsta PTAL?

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
tedsta: Oh you are right

tedsta:
Thanks for clarifying. LGTM

PR #1178, opened 2016-11-02
Fix bug during cleanup phase of object_storage_service benchmark

tedsta:
`/tmp/run` folder is created with `sudo`, and we were trying to delete it without sudo. I'm not sure whether root permissions are actually needed to create/work within `/tmp`.


tedsta:
@ehankland Please take a look

Issue #696, opened 2015-12-07
Failed cloudsuite_web_serving


ehankland:
One run of cloudsuite_web_serving ultimately failed with this:
```
RemoteCommandError: Got non-zero return code (1) executing cd /tmp/pkb/out1 && cd $(ls -Art | tail -n 1) && cat summary.xml
Full command: ssh -A -p 22 perfkit@130.211.138.42 -2 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o IdentitiesOnly=yes -o PreferredAuthentications=publickey -o PasswordAuthentication=no -o ConnectTimeout=5 -o GSSAPIAuthentication=no -o ServerAliveInterval=30 -o ServerAliveCountMax=10 -i /tmp/perfkitbenchmarker/run_21a0e34a/perfkitbenchmarker_keyfile cd /tmp/pkb/out1 && cd $(ls -Art | tail -n 1) && cat summary.xml
STDOUT: STDERR: Warning: Permanently added '130.211.138.42' (ECDSA) to the list of known hosts.^M
cat: summary.xml: No such file or directory
```
Before that, there were many messages that looked like this:
```
Dec 03, 2015 2:38:32 AM sun.rmi.transport.tcp.TCPTransport$AcceptLoop executeAcceptLoop
WARNING: RMI TCP Accept-0: accept loop for ServerSocket[addr=localhost/127.0.0.1,localport=41444] throws
java.net.SocketException: Too many open files
        at java.net.PlainSocketImpl.socketAccept(Native Method)
        at java.net.AbstractPlainSocketImpl.accept(AbstractPlainSocketImpl.java:398)
        at java.net.ServerSocket.implAccept(ServerSocket.java:530)
        at java.net.ServerSocket.accept(ServerSocket.java:498)
        at sun.rmi.transport.tcp.TCPTransport$AcceptLoop.executeAcceptLoop(TCPTransport.java:399)
        at sun.rmi.transport.tcp.TCPTransport$AcceptLoop.run(TCPTransport.java:371)
        at java.lang.Thread.run(Thread.java:745)
```
And before that many of these:
```
Dec 03, 2015 2:20:36 AM com.sun.faban.driver.engine.AgentThread logError
WARNING: UIDriverAgent[1].291.doAddPerson: org/json/JSONException
Note: Error not counted in result.
Either transaction start or end time is not within steady state.
[...]

PR #1166, opened 2016-10-24
Update the cloudsuite web-serving benchmark

nooshin-mirzadeh:
We restructured the cloudsuite web-serving benchmark. This commit is an update for it.

PR #1180, opened 2016-11-03
Fix cloudsuite_web_serving benchmark

peterfeiner:
The cloudsuite commit ("Fix web-serving login problem (#70)"), added a
mandatory $WEB_SERVER_IP parameter to the
cloudsuite/web-serving:db_server image:

https://github.com/ParsaLab/cloudsuite/commit/41d417234e719083ae5edcb657f7cde7f3c8fbc5

Without the parameter, the db_server container fails at startup
with the error message "Web server IP is a mandatory parameter."

This commit also explicitly sets the two optional parameters to
cloudsuite/web-serving:web_server image as a defence against another
backwards-incompatible change like this.

Note that this commit doesn't fix the intermittent failure reported in
issue #696.


peterfeiner:
@tedsta Please take a look.

peterfeiner:
Looks like somebody already got to this in pull request #1166.

PR #1183, opened 2016-11-03
Fix sysbench05plus installation on ubuntu16.04

peterfeiner:
Percona's trusty repo doesn't work on ubuntu-16. Use the proper
release repo instead.


peterfeiner:
@tedsta Please take a look.

PR #1184, opened 2016-11-03
Install docker images as packages

peterfeiner:
Install docker images with vm.Install(...) so they work with an
upcoming package-installation benchmark.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cloudsuite_data_serving_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloudsuite_data_serving_benchmark.py
@@ -83,13 +83,13 @@ def Prepare(benchmark_spec):
 
   def PrepareServerSeed(vm):
     PrepareCommon(vm)
-    vm.RemoteCommand('sudo docker pull cloudsuite/data-serving:server')
+    vm.Install('cloudsuite/data-serving:server')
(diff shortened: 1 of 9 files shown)


peterfeiner:
@tedsta Please take a look.

perfkitbenchmarker/linux_benchmarks/cloudsuite_data_serving_benchmark.py
```diff
@@ -212,10 +210,6 @@ def CleanupServer(vm):
     vm.RemoteCommand('sudo docker stop cassandra-server')
     CleanupServerCommon(vm)
 
-  def CleanupClient(vm):
-    vm.RemoteCommand('sudo docker rmi cloudsuite/data-serving:client')
```
tedsta: Is this not needed because we assume the images will be removed when docker is uninstalled?

peterfeiner: Not when docker is removed, but when the package installed with vm.Install('cloudsuite/data-serving:client') is automatically uninstalled by PKB. That is, during teardown, PKB uninstalls packages that were installed with vm.Install(X) via vm.Uninstall(x). Caveat: PKB only bothers with the uninstall on static VMs; other VMs are deleted, so there's no point in uninstalling.

PR #1189, opened 2016-11-07
Added fully sequential object naming scheme to object_storage_service benchmark

tedsta:
Use an atomic counter shared by all the object writer threads.


tedsta:
@noahl @ehankland @yuyantingzero PTAL

tedsta:
Just realized this doesn't handle multiple VMs properly

PR #1190, opened 2016-11-07
Race condition in Aerospike benchmark.

loosecannon93:
Fixes race condition in Aerospike benchmark.

The fucntion ConfigureAndStart(..) installs and starts the Aerospike server
onto the target VM. Once this finishes the framework moves on and tries to
start the benchmarking, and enters a race between the clients attempting to
connect to the server and the server finishing its initialization.

Previously this was solved by a sleep(5) at the end of this function, however
this was insufficient for most of the runs I attempted. Increasing this to
sleep(15) did work for me, but an actual check seemed superior.

This change adds a proactive check that the server is up and ready for
connections before returning from the "ConfigureAndStart" function. We
accomplish this by sshing into the server vm, and connecting to  Aerospike's
out of band telnet management interface on the internal IP address and issueing
 the `status` command. Using the @Retry decorator from vm_util.py we
continuously check until a 5 minute timeout.

See the aerospike documentation for more info on the status check.
  http://www.aerospike.com/docs/reference/info#status
  http://www.aerospike.com/docs/tools/asinfo#aerospike-s-telnet-port

### DIFF
--- a/perfkitbenchmarker/linux_packages/aerospike_server.py
+++ b/perfkitbenchmarker/linux_packages/aerospike_server.py
@@ -15,9 +15,10 @@
 
 """Module containing aerospike server installation and cleanup functions."""
 
-import time
+import logging
 
 from perfkitbenchmarker import data
+from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import vm_util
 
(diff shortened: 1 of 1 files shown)


peterfeiner:
Please squash the two commits into one.

peterfeiner:
@tedsta Please take a look.

PR #1192, opened 2016-11-10
Add a way to run benchmarks in parallel

ehankland:


perfkitbenchmarker/background_tasks.py
```diff
@@ -409,6 +410,10 @@ def _ExecuteProcessTask(task):
     element is the exception traceback string, or None if the function
     succeeded.
   """
+  def handle_sigint(signum, frame):
```
yuyantingzero: add documents of interrupt behaviors.

perfkitbenchmarker/pkb.py
```diff
@@ -509,6 +518,42 @@ def RunBenchmark(spec, collector):
...
+    collector: SamplesCollector. The collector to call RunBenchmark with.
+
+  Returns:
+    A tuple of BenchmarkSpec, list of samples.
+  """
+  if _TEARDOWN_EVENT.is_set():
+    return spec, []
+  # Many providers name resources using run_uris. When running multiple
+  # benchmarks in parallel, this causes name collisions on resources.
+  # By modifying the run_uri, we avoid the collisions.
```
ehankland: Not any more often than they could before - this guarantees that each benchmark in a run will have a different run uri. The chance of runs having the same base run uri is the same as before this change.

PR #1197, opened 2016-11-15
Enbale OpenStack Cinder Volume Type Option

rossmartyn04:
This small change enables the user to include a Cinder volume type when
running benchmarks such as FIO. This is useful for users of cloud
providers who offer multiple volume types.

### DIFF
--- a/perfkitbenchmarker/providers/openstack/flags.py
+++ b/perfkitbenchmarker/providers/openstack/flags.py
@@ -64,6 +64,10 @@ flags.DEFINE_boolean('openstack_boot_from_volume', False,
 flags.DEFINE_integer('openstack_volume_size', None,
                      'Size of the volume (GB)')
 
+flags.DEFINE_string('openstack_volume_type',
+                    default=None,
+                    help='Optional Cinder volume type to use.')
+
 flags.DEFINE_string('openstack_image_username', 'ubuntu',
                     'Ssh username for cloud image')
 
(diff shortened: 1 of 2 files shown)


rossmartyn04:
Will resubmit following #1196 being merged.

rossmartyn04:
Reopening

perfkitbenchmarker/providers/openstack/flags.py
```diff
@@ -64,6 +64,10 @@
 flags.DEFINE_integer('openstack_volume_size', None,
                      'Size of the volume (GB)')
 
+flags.DEFINE_string('openstack_volume_type',
```
tedsta: It looks like this could probably be an enum?

rossmartyn04: Hi,

Thank you for the response. Please tell me if I have misunderstood, but it sounds like you mean that I should define some options for the volume type.

Due to the flexibility in OpenStack (Cinder) it is possible to name a 'volume type' as the cloud administrator sees fit, therefore I believe the user should have full control of what they input after this flag.

Thanks.

meteorfox:
@rossmartyn04 Thank you for your contribution. 

I recommend that you extend the BaseDiskSpec implementation to add the configuration option decoder to support `volume_type` as well. This way, we can pass the option through the YAML config.

Here's an example of the `AWS` provider, which add supports for the flag `aws_provisioned_iops` and the YAML config option called `iops`.
- [`AWSDiskSpec`](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/aws/aws_disk.py#L102-L139)

Then you can simply use the value directly from `disk_spec` object.

perfkitbenchmarker/providers/openstack/flags.py
tedsta: Yes that is what I meant. Ah alright, thanks for clarifying.

PR #1201, opened 2016-11-15
Add the minimum tox version to the *tox is not installed* message

gareth-ferneyhough:
Tox less than version 2.0 does not work with PKB. A message was added to clarify this point when the user is informed that he/she does not have tox installed.


gareth-ferneyhough:
I used my personal email address in my gitconfig, so the cla check failed. I will fix it next time.

PR #1202, opened 2016-11-16
Fix failing boilerplate checks

gareth-ferneyhough:
Fix two files that were failing the boilerplate check (if a .py file is to have a shebang on the first line, it must be executable)


gareth-ferneyhough:
I will recreate with the correct commits. I included an extra one.

PR #1205, opened 2016-11-17
fio: Add flags to capture latency, bandwidth and IOPS logs

meteorfox:
Add the ability to optionally collect the latency, bandwidth and IOPS logs
for a fio job. This also includes the flag fio_log_avg_msec that averages
the log entries over the given interval.

These log files produced by fio consist of the current format:

`time, rate/latency, data direction, block size`

where:

Time: in milliseconds.
Rate/latency: For bandwidth, this is in KB/sec. For latency, it's microseconds.
Data direction: 0 is read, 1 is write. 
Block size: block size in bytes.

Since PKB needs to pull the logs from the VMs, we could not simply use the `--fio_parameters` flag. This PR add the ability to fetch those logs.

The generated logs are copied from the VMs into the local machine under the generated temporary directory `/tmp/pkb/runs/<run_uri>`.

Using these logs more detail graphs such as histograms, or time series can be produced. Example
image

Note: The graph above was generated using R and ggplot2 as follows:
```r
pkb_iops <- read.csv(file = 'Downloads/pkb_fio_avg_iops.log',header = FALSE)
pkb_iops <- data.frame(time_ms = pkb_iops$V1, iops = pkb_iops$V2, dir = factor(pkb_iops$V3, levels=c(0, 1), labels=c("read", "write")), blocksize = pkb_iops$V4)
p <- ggplot(data=pkb_iops, aes(x=time_ms, y=iops, group=dir, color=dir)) + geom_line() + geom_point() + ylim(0, 500) + ggtitle("Random Read 4K")
```

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
@@ -34,6 +34,7 @@ from perfkitbenchmarker import units
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.linux_packages import fio
 
+PKB_FIO_LOG_FILE_NAME = 'pkb_fio_avg'
 LOCAL_JOB_FILE_NAME = 'fio.job'  # used with vm_util.PrependTempDir()
 REMOTE_JOB_FILE_PATH = posixpath.join(vm_util.VM_TMP_DIR, 'fio.job')
 DEFAULT_TEMP_FILE_NAME = 'fio-temp-file'
@@ -135,6 +136,19 @@ flags.DEFINE_integer('fio_runtime', 600,
 flags.DEFINE_list('fio_parameters', [],
                   'Parameters to apply to all PKB generated fio jobs. Each '
                   'member of the list should be of the form "param=value".')
+flags.DEFINE_boolean('fio_lat_log', False,
+                     'Whether to collect a latency log of the fio jobs.')
+flags.DEFINE_boolean('fio_bw_log', False,
+                     'Whether to collect a bandwidth log of the fio jobs.')
+flags.DEFINE_boolean('fio_iops_log', False,
+                     'Whether to collect an IOPS log of the fio jobs.')
+flags.DEFINE_integer('fio_log_avg_msec', 1000,
+                     'By default, this will average each log entry in the '
+                     'fio latency, bandwidth, and iops logs over the specified '
+                     'period of time in milliseconds. If set to 0, fio will '
+                     'log an entry for every IO that completes, this can grow '
+                     'very quickly in size and can cause performance overhead.',
+                     lower_bound=0)
 
 
 FLAGS_IGNORED_FOR_CUSTOM_JOBFILE = {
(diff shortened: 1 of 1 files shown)


meteorfox:
FYI... This has not been tested on Windows. Can someone verify if it works on Windows?

meteorfox:
@noahl PTAL, or please recommend someone that might be able to review this. Thanks!

noahl:
@meteorfox great patch! Unfortunately I am moving on from spending most of my time on PKB. I think @tedsta is the right person to review this now.

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
```diff
@@ -460,6 +489,9 @@ def Run(benchmark_spec):
   stdout, stderr = vm.RobustRemoteCommand(fio_command, should_log=True)
   samples = fio.ParseResults(job_file_string, json.loads(stdout))
 
+  if collect_logs:
+    vm.PullFile(vm_util.GetTempDir(), '%s_*.log' % PKB_FIO_LOG_FILE_NAME)
```
tedsta: It's a little weird to be getting the tmp dir from two different sources. Suggest using `vm_util.VM_TMP_DIR` as above?

meteorfox: True. I'll make the change.

meteorfox: While going over your suggestion, I remembered now why I didn't use `vm_util.VM_TMP_DIR`. I want to store the collected logs under `/tmp/pkb/runs/<run_uri>/` so they are grouped with the results for that run, `vm_util.VM_TMP_DIR` is the path to just `/tmp/pkb/` directory.

tedsta: I don't see anything passed to fio telling to put the logs there?

tedsta: Ahhh I see.

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
```diff
@@ -135,6 +136,19 @@
 flags.DEFINE_list('fio_parameters', [],
                   'Parameters to apply to all PKB generated fio jobs. Each '
                   'member of the list should be of the form "param=value".')
+flags.DEFINE_boolean('fio_lat_log', False,
+                     'Whether to collect a latency log of the fio jobs.')
+flags.DEFINE_boolean('fio_bw_log', False,
+                     'Whether to collect a bandwidth log of the fio jobs.')
+flags.DEFINE_boolean('fio_iops_log', False,
+                     'Whether to collect an IOPS log of the fio jobs.')
+flags.DEFINE_integer('fio_log_avg_msec', 1000,
```
tedsta: fio's `--log_avg_msec` defaults to 0. Why do we default to 1000?

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
```diff
@@ -373,6 +387,17 @@ def GetConfig(user_config):
...
 
 
+def GetLogFlags():
+  collect_logs = FLAGS.fio_lat_log or FLAGS.fio_bw_log or FLAGS.fio_iops_log
+  fio_log_flags = [(FLAGS.fio_lat_log, '--write_lat_log=%(filename)s',),
+                   (FLAGS.fio_bw_log, '--write_bw_log=%(filename)s',),
+                   (FLAGS.fio_iops_log, '--write_iops_log=%(filename)s',),
+                   (collect_logs, '--log_avg_msec=%(interval)d',)]
+  fio_command_flags = ' '.join([flag for given, flag in fio_log_flags if given])
+  return fio_command_flags % {'filename': PKB_FIO_LOG_FILE_NAME,
```
tedsta: We want all these logs written to the same file? FIO can have concurrent jobs, so does FIO ensure that writes to any one of these logs is the only write to these logs occurring at any given time? (atomic)

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
meteorfox: By default, it will log data for every IO completion which can create a huge file. I thought 1 second, was a sensible default, so it aggregates the metrics per 1 second interval.

If you feel, we should use the same as fio as to not "surprise" the user, I can make the change.

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
tedsta: Ah ok thanks for clarifying

PR #1206, opened 2016-11-17
openstack: Deprecate flag

vhe182:
This PR addresses #1122.  The flag 'openstack_volume_size' has been deprecated and the expected flag to pass is now 'data_disk_size'.  The following changes were made:

* Added a deprecated flag message to inform user
* Added abstract class that overrides BaseDiskSpec class from disk.py
* Added check for 'openstack_volume_size' in _ApplyFlags() method of BaseDiskSpec class

### DIFF
--- a/perfkitbenchmarker/providers/openstack/os_disk.py
+++ b/perfkitbenchmarker/providers/openstack/os_disk.py
@@ -18,6 +18,7 @@ import logging
 from perfkitbenchmarker import disk
 from perfkitbenchmarker import errors
 from perfkitbenchmarker import flags
+from perfkitbenchmarker import providers
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.providers.openstack import utils as os_utils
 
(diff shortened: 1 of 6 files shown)


vhe182:
If the minimum disk size is not met by the disk size value given by the user, the min disk for the image is used (command: openstack image show...).  A warning could be added to inform the user that the user's requested disk size was overridden.  Not an error per se, but a heads-up on the correction.  Although, maybe it is preferred that PKB immediately exits with the error.

Thoughts?

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -81,6 +88,38 @@ def WaitForVolumeCreation(resource, volume_id):
     raise errors.Resource.RetryableCreationError(msg)
 
 
+disk.RegisterDiskTypeMap(providers.OPENSTACK, DISK_TYPE)
```
meteorfox: I don't think is necessary to register a disk type map in this case.

perfkitbenchmarker/providers/openstack/os_disk.py
```diff
@@ -81,6 +88,38 @@ def WaitForVolumeCreation(resource, volume_id):
...
+    Can be overridden by derived classes to add support for specific flags.
+
+    Args:
+      config_values: dict mapping config option names to provided values. May
+          be modified by this function.
+      flag_values: flags.FlagValues. Runtime flags that may override the
+          provided config values.
+    """
+    super(OpenStackDiskSpec, cls)._ApplyFlags(config_values, flag_values)
+    if flag_values['openstack_volume_size'].present \
```
meteorfox: Instead of using `\` to split the line, surround the boolean expression with parenthesis `( )`.  Break the line before the `and` binary operator, add an indentation level.[1]

[1] https://www.python.org/dev/peps/pep-0008/#indentation

vhe182: Made the change as you suggested.

PR #1214, opened 2016-11-22
Add version option to spark_service.

jasonkuster:
Signed-off-by: Jason Kuster <jasonkuster@google.com>

Allows specifying version of Spark to use in managed Spark benchmarks.

### DIFF
--- a/perfkitbenchmarker/configs/benchmark_config_spec.py
+++ b/perfkitbenchmarker/configs/benchmark_config_spec.py
@@ -203,6 +203,9 @@ class _SparkServiceSpec(spec.BaseSpec):
             'default': spark_service.PROVIDER_MANAGED,
             'valid_values': [spark_service.PROVIDER_MANAGED,
                              spark_service.PKB_MANAGED]}),
+        'version': (option_decoders.EnumDecoder, {
+            'default': spark_service.SPARK_1_6_2,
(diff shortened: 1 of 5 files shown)


asaksena:
2 comments
- How are we planning to use this version information for a pkb managed cluster ? I guess this will be ignored in the interim ?
- If you have not tested it yet, I would like to make the change to the default config of other benchmarks that rely on spark service and make sure they continue to work

perfkitbenchmarker/configs/benchmark_config_spec.py
```diff
@@ -203,6 +203,9 @@ def _GetOptionDecoderConstructions(cls):
             'default': spark_service.PROVIDER_MANAGED,
             'valid_values': [spark_service.PROVIDER_MANAGED,
                              spark_service.PKB_MANAGED]}),
+        'version': (option_decoders.EnumDecoder, {
+            'default': spark_service.SPARK_1_6_2,
```
amitsela: Spark runner upgraded to 1.6.3 - though it really shouldn't matter much, and would be compatible.

s-deitz:

jasonkuster:
I think we can close this. I can't remember why this was necessary and I've been working on other things for a while now.

PR #1216, opened 2016-11-23
Allow override hpcc binary.

yuyantingzero:


perfkitbenchmarker/linux_packages/hpcc.py
```diff
@@ -22,21 +22,48 @@
...
+
+def CheckPrerequisites():
+  """Verifies that required resources are present.
+
+  Raises:
+    perfkitbenchmarker.data.ResourceNotFound: On missing resource.
+  """
+  if __name__.split('.')[-1] in FLAGS.override_packages:
+    for k, _ in OVERRIDABLE.items():
+      data.ResourcePath(k)
```
tedsta: If more packages decide to be override-able, this could be reusable code. But I guess we can deal with that when the need arises.

perfkitbenchmarker/linux_packages/hpcc.py
```diff
@@ -22,21 +22,48 @@
...
+
+
+def _Override(vm):
+  """Override hpcc binaries."""
+  for k, v in OVERRIDABLE.items():
+    vm.PushDataFile(k, v)
```
tedsta: Same as above - probably reusable later on.

perfkitbenchmarker/linux_packages/hpcc.py
```diff
@@ -50,6 +77,16 @@ def _Install(vm):
...
+  vm.RemoteCommand('cp %s/hpcc hpcc' % HPCC_DIR)
+
+
+def _Install(vm):
+  """Installs the HPCC package on the VM."""
+  vm.Install('openmpi')
+  if __name__.split('.')[-1] in FLAGS.override_packages:
+    _Override(vm)
+  else:
+    _Build(vm)
```
tedsta: I bet this logic could be moved to `LinuxVirtualMachine.Install` so that it could be shared among override-able packages.

yuyantingzero:
Thanks @tedsta for the review, ehankland gave exact same comment offline :) . My concern is the override behavior is specific to each packages. Some packages have dependencies. But most packages should be able to override by scp necessary binaries. I will try to come up with a more general solution.

Issue #1217, opened 2016-11-24
Cannot run bonnie++ on Azure


manavgarg:
Cannot create the resource group on Azure because of a regexp mismatch.
I believe this is because Azure doesn't allow "++" in the resource group name.
Should be an easy fix though. 

This is the full error trace:

2016-11-24 01:37:29,329 d9c0f2e7 Thread-2 bonnie++(1/1) INFO     Running: azure group create pkbd9c0f2e7-bonnie++0 westus2
2016-11-24 01:37:31,056 d9c0f2e7 Thread-2 bonnie++(1/1) INFO     Ran azure group create pkbd9c0f2e7-bonnie++0 westus2. Got return code (1).
STDOUT: info:    Executing command group create
info:    Getting resource group pkbd9c0f2e7-bonnie++0

STDERR: error:   "resourceGroupName" should satisfy the constraint - "Pattern": /^[-w._]+$/
error:   Error information has been recorded to /home/godha/.azure/azure.err
error:   group create command failed

PR #1218, opened 2016-11-24
Fix for "Cannot run bonnie++ on Azure #1217"

rgodha:
Windows Azure cannot create resource group with '+' symbol in the name. We have to match the regex as  /^[-w._]+$/ as pointed out in issue #1217.

Solution:
The benchmark name is changed from bonnie++ to bonnieplusplus.
Change has been made in two places:
(i) benchmark_name changed to bonnieplusplus, and
(ii) benchmark config yaml 'key' value changed to bonnieplusplus to match the benchmark name.

### DIFF
--- a/tests/benchmark_sets_test.py
+++ b/tests/benchmark_sets_test.py
@@ -171,7 +171,8 @@ class BenchmarkSetsTestCase(unittest.TestCase):
       self.assertIn(benchmark_tuple[0].BENCHMARK_NAME,
                     self.valid_benchmark_names)
     # make sure bonnie++ is a listed benchmark
-    self.assertTrue(self._ContainsModule('bonnieplusplus', benchmark_tuple_list))
+    self.assertTrue(self._ContainsModule('bonnieplusplus', 
+                                         benchmark_tuple_list))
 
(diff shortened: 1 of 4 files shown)


tests/benchmark_sets_test.py
```diff
@@ -163,15 +163,16 @@ def _ContainsModule(module_name, module_list):
...
     # to a valid set of benchmarks
-    self.mock_flags.benchmarks = ['standard_set', 'bonnie++']
+    self.mock_flags.benchmarks = ['standard_set', 'bonnieplusplus']
     benchmark_tuple_list = benchmark_sets.GetBenchmarksFromFlags()
     self.assertIsNotNone(benchmark_tuple_list)
     self.assertGreater(len(benchmark_tuple_list), 0)
     for benchmark_tuple in benchmark_tuple_list:
       self.assertIn(benchmark_tuple[0].BENCHMARK_NAME,
                     self.valid_benchmark_names)
     # make sure bonnie++ is a listed benchmark
```
yuyantingzero: Update the reference in comments as well

PR #1219, opened 2016-11-24
 Fix for "Cannot run bonnie++ on Azure #1217"

rgodha:
Test case failed as bonnie++ was modified to bonnieplusplus.


rgodha:
Fix for "Cannot run bonnie++ on Azure #1217"

rgodha:
Fix for "Cannot run bonnie++ on Azure #1217"

PR #1221, opened 2016-11-28
Fix gcs credential conflict.

yuyantingzero:


noahl:
Quick question - what happens if you use this to access GCS from a non-GCE host? I think the gsutil credentials were mostly important for that. There might also be an issue if your GCE account didn't have access to the GCS bucket you wanted to use but you yourself did have access, although that is a weird scenario that maybe is not important to support.

PR #1227, opened 2016-11-30
Always set an internal ip address to OpenStack VMs

yremmet:
in order for all benchmarks to function.
(e.g. Cloudsuite Data Caching)


yremmet:
Signed CLA.

PR #1228, opened 2016-11-30
Always set internal IP-Address on OpenStack VMs

yremmet:
in order for all benchmarks to function. (e.g. cloud suite_data_serving)


perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -340,6 +340,7 @@ def _SetIPAddresses(self):
     stdout, _, _ = show_cmd.Issue()
     server_dict = json.loads(stdout)
     self.ip_address = self._GetNetworkIPAddress(server_dict, self.network_name)
+    self.internal_ip = self.ip_address
     if self.floating_ip_pool_name:
       self.floating_ip = self._AllocateFloatingIP()
       self.internal_ip = self.ip_address
```
yuyantingzero: still need this line?

tedsta: I guess it's too late, but we should probably remove this line now

PR #1236, opened 2016-12-07
Add some metadata to netperf samples

tedsta:
So that we can change the iperf queries to netperf queries.

@ehankland PTAL


gareth-ferneyhough:
LGTM

PR #1242, opened 2016-12-09
Use aes128-ctr if aes128-cbc isn't available

peterfeiner:
The aes128-cbc cipher is deprecated and disabled by default on some
Linux distributions' SSH servers (e.g., on Ubuntu-16.04). This patch
changes copy_throughput to fallback to aes128-ctr if aes128-cbc isn't
available on the VMs.


peterfeiner:
@tedsta Please take a look.

gareth-ferneyhough:
LGTM

PR #1244, opened 2016-12-10
If possible, do not use FlagValues.FlagDict.  Or use FlagValues._flags instead.

yilei:
This is a followup for https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1175/commits/4b6c2dbeac5a2fc2b46686a78fd65da468631ca7

### DIFF
--- a/tests/configs/benchmark_config_spec_test.py
+++ b/tests/configs/benchmark_config_spec_test.py
@@ -38,6 +38,10 @@ _GCP_AWS_VM_CONFIG = {'GCP': {'machine_type': 'n1-standard-1'},
 _GCP_AWS_DISK_CONFIG = {'GCP': {}, 'AWS': {}}
 
 
+def _GetFlagDict(flag_values):
+  return {name: flag_values[name] for name in flag_values}
+
+
 class FlagsDecoderTestCase(unittest.TestCase):
 
   def setUp(self):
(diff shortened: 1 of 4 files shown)


gareth-ferneyhough:
LGTM. @tedsta, can you check this out too?

yilei:
Kindly ping? @tedsta

tests/configs/benchmark_config_spec_test.py
```diff
@@ -38,6 +38,10 @@
 _GCP_AWS_DISK_CONFIG = {'GCP': {}, 'AWS': {}}
 
 
+def _GetFlagDict(flag_values):
+  return {name: flag_values[name] for name in flag_values}
```
tedsta: IIUC, this builds an actual dict out of something that emulates a dict?

tests/flag_util_test.py
```diff
@@ -120,7 +120,11 @@ def testReadAndWrite(self):
     flag_values_copy.test_flag = 1
     self.assertFlagState(flag_values, 0, False)
     self.assertFlagState(flag_values_copy, 1, False)
-    with flag_util.FlagDictSubstitution(flag_values, flag_values_copy.FlagDict):
+    if hasattr(flag_values_copy, '_flags'):
+      flag_dict_func = flag_values_copy._flags
+    else:
+      flag_dict_func = flag_values_copy.FlagDict
```
tedsta: Interesting that there's conditionals in a unit test. Shouldn't there only ever be one path through a unit test?

tests/configs/benchmark_config_spec_test.py
yilei: The flags.FLAGS object is a dict-like object, the tests is comparing a dictionary with _GetFlagDict().

tests/flag_util_test.py
yilei: FlagDict is going away internally, and unfortunately _flags is not available in github.com/google/python-gflags (yet), and the test needs to run in both environments.

I'm going to bring the two flags library to a same state, but it takes sometime, that results in this ugly intermediate state on your code :(

PR #1246, opened 2016-12-12
Fio enhancements

noahl:
Two small enhancements for the fio benchmark:

   1. When retrieving IOPS, latency or bandwidth logs, add timestamps to their names. This keeps us from overwriting old logs when running fio with `--run_stage_time`.
   2. Handle single integers as input to the IntegerList parser. This makes the parser work correctly with config files, which will turn strings of digits into integers on parsing.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
@@ -20,7 +20,6 @@ Quick howto: http://www.bluestop.org/fio/HOWTO.txt
 
 import json
 import logging
-import os
 import posixpath
 import re
 import time
--- a/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
@@ -20,8 +20,10 @@ Quick howto: http://www.bluestop.org/fio/HOWTO.txt
 
 import json
 import logging
+import os
 import posixpath
 import re
+import time
 
 import jinja2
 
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
```diff
@@ -20,8 +20,10 @@
 
 import json
 import logging
+import os
```
meteorfox: `os` module is not being used.

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
```diff
@@ -394,7 +396,9 @@ def GetLogFlags():
                    (FLAGS.fio_iops_log, '--write_iops_log=%(filename)s',),
                    (collect_logs, '--log_avg_msec=%(interval)d',)]
   fio_command_flags = ' '.join([flag for given, flag in fio_log_flags if given])
-  return fio_command_flags % {'filename': PKB_FIO_LOG_FILE_NAME,
+  now = time.time()
+  filename_base = '%s_%s' % (PKB_FIO_LOG_FILE_NAME, str(now))
```
meteorfox: `time.time()` returns a floating point number, which will add a dot to the filename, maybe you want to truncate it to an int?

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
noahl: Thanks! Removed.

perfkitbenchmarker/linux_benchmarks/fio_benchmark.py
noahl: I haven't had problems with having a dot in the filename, and it could potentially cause problems if the sampling interval is small enough. Do you feel strongly about this?

meteorfox: I don't feel strongly about it, I guess one alternative would be to convert it to nanoseconds and use that as an int, that way you don't lose the precision and you can use an integer.

If not, this is good enough.

noahl: Good point about converting to nanoseconds. However, since you don't feel strongly, I would rather leave it as-is because I think I am more likely to recognize a timestamp in seconds, with decimal, if I am looking at filenames six months from now and can't remember the context.

PR #1249, opened 2016-12-13
Use FLAGS.aws_user_name if set for Rhel

cwilkes:
The official CentOS AMI user name is "centos" and not "ec2-user"  This patch enables passing in the AWS user name to be used, ie:
```
--aws_user_name=centos
```
It defaults to ec2-user if none is passed in.

### DIFF
--- a/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
+++ b/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
@@ -545,7 +545,8 @@ class RhelBasedAwsVirtualMachine(AwsVirtualMachine,
 
   def __init__(self, vm_spec):
     super(RhelBasedAwsVirtualMachine, self).__init__(vm_spec)
-    self.user_name = FLAGS.aws_user_name if FLAGS['aws_user_name'].present else 'ec2-user'
+    user_name_set = FLAGS['aws_user_name'].present
+    self.user_name = FLAGS.aws_user_name if user_name_set else 'ec2-user'
 
 
 class WindowsAwsVirtualMachine(AwsVirtualMachine,
--- a/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
+++ b/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
@@ -545,10 +545,8 @@ class RhelBasedAwsVirtualMachine(AwsVirtualMachine,
 
   def __init__(self, vm_spec):
     super(RhelBasedAwsVirtualMachine, self).__init__(vm_spec)
-    if FLAGS['aws_user_name'].present:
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -545,7 +545,10 @@ class RhelBasedAwsVirtualMachine(AwsVirtualMachine,
 
   def __init__(self, vm_spec):
     super(RhelBasedAwsVirtualMachine, self).__init__(vm_spec)
-    self.user_name = 'ec2-user'
+    if FLAGS.aws_user_name:
```
yuyantingzero: Do you mean FLAGS['aws_user_name'].present?
Otherwise aws_user_name always have the default value ubuntu.

yuyantingzero: Also, this can be written in a single line.

cwilkes: Hrm, yes you are right.  Is there a way to see if the user passed in a value or if it is using the default one?  This is trickier than I thought it was going to be.

I could also make a CentOSBasedAwsVirtualMachine that sets its user name to centos, but that sounds a little messy.

yuyantingzero: .present does that

cwilkes:
@yuyantingzero updated.  Using one line and the `present` method

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -545,7 +545,8 @@ class RhelBasedAwsVirtualMachine(AwsVirtualMachine,
 
   def __init__(self, vm_spec):
     super(RhelBasedAwsVirtualMachine, self).__init__(vm_spec)
-    self.user_name = 'ec2-user'
+    user_name_set = FLAGS['aws_user_name'].present
```
yuyantingzero: user_name_set is not reused, prefer just substitute in if.

cwilkes:
Bah, I did the temp variable as to not go above 80 characters.  Didn't think we had that check in github, I'll revisit.

cwilkes:
I'm back to the temporary variable as using the "present" method in a ternary operator makes the line 90 characters long.  Breaking a ternary operator on the "else" part makes it error prone to read and evaluate.

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
cwilkes: Let me know if the temp variable is okay, without it the line is longer than 80 characters and it looks weird to split a ternary operator over multiple lines.

PR #1252, opened 2016-12-19
Add Azure to GPU PCIe bandwidth test

gareth-ferneyhough:


perfkitbenchmarker/linux_benchmarks/gpu_pcie_bandwidth_benchmark.py
```diff
@@ -60,6 +60,10 @@
           machine_type: p2.xlarge
           zone: us-west-2b
```
tedsta: Any reason for different zones for different providers, or is it just random?

PR #1258, opened 2016-12-19
Fix integration test.

yuyantingzero:


yuyantingzero:
@gareth-ferneyhough , ptal

PR #1267, opened 2017-01-05
openstack: Add volume_type option decoder

meteorfox:
This PR introduces the option decoder for the volume_type flag that was added in #1197


gareth-ferneyhough:
LGTM

PR #1274, opened 2017-01-12
Various object storage service benchmark improvements

tedsta:
- Spawn many processes from worker-side script via Python's multiprocessing module instead of SSHing into each machine many times. This improves high VM x Thread runs significantly.
- Created an interactive object timeline. Useful for debugging as well as seeing throughput during different times in the run.
- Make GCS use regional buckets by default, like the other providers already do.


tedsta:
@gareth-ferneyhough @asaksena PTAL

tedsta:
Hey @gareth-ferneyhough sorry could you take another look? Had to make one more last minute fix.

gareth-ferneyhough:
LGTM again!

PR #1278, opened 2017-01-19
Add the capability to publish to InfluxDB

besanradwan:
This code aims to add the capability for PKB to publish directly to InfluxDB

The publishing of data from PKB benchmarks can done either by specifiying a pre-existing Influx DB publisher or with none specified it will create a new one or re-use a pre-existing DB called perfkit. By default if no uri is specified it will publish the data to localhost.

### DIFF
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -15,6 +15,7 @@
 """Classes to collect and publish performance samples to various sinks."""
 
 import abc
+import collections
 import copy
 import csv
 import httplib
@@ -672,7 +673,7 @@ class InfluxDBPublisher(SamplePublisher):
   def _Publish(self, formated_samples):
     try:
       self._CreateDB()
-      body = ' \n '.join(formated_samples)
+      body = '\n'.join(formated_samples)
       self._WriteData(body)
     except (IOError, httplib.HTTPException) as http_exception:
       logging.debug("Error connecting to the database: ", http_exception)
@@ -687,7 +688,8 @@ class InfluxDBPublisher(SamplePublisher):
         tag_set_metadata = ','.join(self._FormatToKeyValue(sample['metadata']))
     tag_keys = ('test', 'official', 'owner', 'run_uri', 'sample_uri',
                 'metric', 'unit')
-    tag_set = ','.join(self._FormatToKeyValue({k: sample[k] for k in tag_keys}))
+    ordered_tags = collections.OrderedDict([(k, sample[k]) for k in tag_keys])
+    tag_set = ','.join(self._FormatToKeyValue(ordered_tags))
     if tag_set_metadata:
       tag_set += ',' + tag_set_metadata
 
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -702,6 +702,9 @@ class InfluxDBPublisher(SamplePublisher):
   def _FormatToKeyValue(self, sample):
     key_value_pairs = []
     for k, v in sample.iteritems():
+      v = str(v)
+      v = v.replace(',', '\,')
+      v = v.replace(' ', '\ ')
       key_value_pairs.append('%s=%s' % (k, v))
     return key_value_pairs
 
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -114,7 +114,7 @@ flags.DEFINE_multistring(
     'by separating each pair by commas.')
 
 flags.DEFINE_string(
-    'influx_uri', 'http://localhost:8086/',
+    'influx_uri', 'localhost:8086',
     'The Influx DB endpoint for queries address and port')
 
 flags.DEFINE_string(
@@ -663,71 +663,78 @@ class InfluxDBPublisher(SamplePublisher):
     self.influx_db_name = influx_db_name
 
   def PublishSamples(self, samples):
-    successful_http_request_codes = [200, 202, 204]
-    sample = list(samples)
-    number_of_samples = 0
+    formated_samples = []
     for sample in samples:
-      number_of_samples += 1
-      timestamp = int((10 ** 9) * sample['timestamp'])
-      measurement = 'perfkitbenchmarker'
-      tag_set_metadata = \
-          ', '.join(self.FormatSampleForInfluxDB(sample['metadata']))
-      tag_set = ','.join((tag_set_metadata, self.
-                          FormatSampleForInfluxDB({k: sample[k]
-                                                   for k in ('test',
-                                                             'official',
-                                                             'owner',
-                                                             'run_uri',
-                                                             'sample_uri')})))
-      field_set = '%s_%s=%s' % (self.FormatSampleForInfluxDB({k: sample[k]
-                                for k in ('metric', 'unit', 'value')}))
-      samples_constructed_body = (' '.join((','.join((measurement,
-                                           tag_set))),
-                                           (' '.join((field_set,
-                                                     timestamp))))) + "\n"
+      formated_samples.append(self._ConstructSample(sample))
+    self._Publish(formated_samples)
+    
+  def _Publish(self, formated_samples):
+    try:
+      self._CreateDB()
+      body = ' \n '.join(formated_samples)
+      self._writeData(body)
+    except (IOError, httplib.HTTPException) as http_exception:
+      logging.debug("Error connecting to the database: ", http_exception)
 
-    create_db_status, create_db_response = self. \
-        _createInfluxDB(self.influx_uri, self.influx_db_name)
-    if create_db_status in successful_http_request_codes:
+  def _ConstructSample(self, sample):
+    timestamp = str(int((10 ** 9) * sample['timestamp']))
+    measurement = 'perfkitbenchmarker'
+    
+    tag_set_metadata = ''
+    if 'metadata' in sample:
+      if sample['metadata']:
+        tag_set_metadata = ','.join(self._FormatToKeyValue(sample['metadata']))
+    tag_keys = ('test','official', 'owner','run_uri','sample_uri', 'metric', 'unit',)
+    tag_set = ','.join(self._FormatToKeyValue({k: sample[k] for k in tag_keys}))
+    if tag_set_metadata:
+      tag_set += ',' + tag_set_metadata
+
+    field_set = '%s=%s' % ('value', sample['value'])
+    
+    sample_constructed_body = '%s,%s %s %s' % (measurement, tag_set,
+                                               field_set, timestamp)
+    return sample_constructed_body
+
+  def _FormatToKeyValue(self, sample):
+    key_value_pairs = []
+    for k, v in sample.iteritems():
+      key_value_pairs.append('%s=%s' % (k, v))
+    return key_value_pairs
+
+  def _CreateDB(self):
+    successful_http_request_codes = [200, 202, 204]
+    header = {'Content-type': 'application/x-www-form-urlencoded',
+               'Accept': 'text/plain'}
+    params = urllib.urlencode({'q': 'CREATE DATABASE ' + self.influx_db_name})
+    conn = httplib.HTTPConnection(self.influx_uri)
+    conn.request('POST', '/query?' + params, headers=header)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
+    if response_status in successful_http_request_codes:
       logging.debug('Success!', self.influx_db_name, ' DB Created')
+    else:
+      logging.debug(response_status,
+                    'Request could not be completed due to: ',
+                    response_response)
+      raise httplib.HTTPException
+
+  def _WriteData(self, data):
+    successful_http_request_codes = [200, 202, 204]
+    params = data
+    header = {"Content-type": "application/octet-stream"}
+    conn = httplib.HTTPConnection(self.influx_uri)
+    conn.request('POST', '/write?' + 'db=' + self.influx_db_name, params, headers=header)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
+    if response_status in successful_http_request_codes:
+      logging.debug('writing samples to publisher: writing samples.')
     else:
       logging.debug(create_db_status,
                     'Request could not be completed due to: ',
-                    create_db_response)
-
-    logging.debug('writing samples to publisher: writing',
-                  number_of_samples, ' samples.')
-    self._writeData(self.influx_uri, self.influx_db_name,
-                    samples_constructed_body)
-    logging.debug(create_db_status,
-                  'Request could not be completed due to: ',
-                  create_db_response)
-
-  def FormatSampleForInfluxDB(self, sample):
-    for k, v in sample.iteritems():
-            return '%s=%s' % (k, v)
-
-  def _CreateDB(self, influx_uri, influx_db_name):
-    headers = {'Content-type': 'application/x-www-form-urlencoded',
-               'Accept': 'text/plain'}
-    body = urllib.urlencode({'q': 'CREATE DATABASE' + influx_db_name})
-    conn = httplib.HTTPConnection(influx_uri)
-    conn.request('POST', '/query?' + body, headers)
-    response = conn.getresponse()
-    response_status = response.status
-    response_response = response.response
-    return response_status, response_response
-
-  def _WriteData(self, influx_uri, influx_db_name, data):
-    body = data
-    headers = {"Content-type": "application/octet-stream"}
-    conn = httplib.HTTPConnection(influx_uri)
-    conn.request('POST', '/write?' + 'db=' + influx_db_name, body, headers)
-    response = conn.getresponse()
-    response_status = response.status
-    response_response = response.response
-    return response_status, response_response
-
+                    response_response)
+      raise httplib.HTTPException
 
 
 class SampleCollector(object):
@@ -789,6 +796,9 @@ class SampleCollector(object):
       publishers.append(ElasticsearchPublisher(es_uri=FLAGS.es_uri,
                                                es_index=FLAGS.es_index,
                                                es_type=FLAGS.es_type))
+    if FLAGS.influx_uri:
+      publishers.append(InfluxDBPublisher(influx_uri=FLAGS.influx_uri,
+                                          influx_db_name=FLAGS.influx_db_name))  
 
     return publishers
 
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -717,7 +717,7 @@ class InfluxDBPublisher(SamplePublisher):
     else:
       logging.debug(response.status,
                     'Request could not be completed due to: ',
-                    response.response)
+                    response.reason)
       raise httplib.HTTPException
 
   def _WriteData(self, data):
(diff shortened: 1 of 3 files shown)


perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+                                                             'official',
+                                                             'owner',
+                                                             'run_uri',
+                                                             'sample_uri')})))
+      field_set = '%s_%s=%s' % (self.FormatSampleForInfluxDB({k: sample[k]
+                                for k in ('metric', 'unit', 'value')}))
+      samples_constructed_body = (' '.join((','.join((measurement,
+                                           tag_set))),
+                                           (' '.join((field_set,
+                                                     timestamp))))) + "\n"
```
meteorfox: Use local variables to decompose these 'joins' and the nested arguments

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+
+  def PublishSamples(self, samples):
+    successful_http_request_codes = [200, 202, 204]
+    sample = list(samples)
+    number_of_samples = 0
+    for sample in samples:
+      number_of_samples += 1
+      timestamp = int((10 ** 9) * sample['timestamp'])
+      measurement = 'perfkitbenchmarker'
+      tag_set_metadata = \
```
meteorfox: nitpick: Prefer hanging indentation for style consistency of the codebase. See: https://www.python.org/dev/peps/pep-0008/#indentation

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+      tag_set = ','.join((tag_set_metadata, self.
+                          FormatSampleForInfluxDB({k: sample[k]
+                                                   for k in ('test',
+                                                             'official',
+                                                             'owner',
+                                                             'run_uri',
+                                                             'sample_uri')})))
+      field_set = '%s_%s=%s' % (self.FormatSampleForInfluxDB({k: sample[k]
+                                for k in ('metric', 'unit', 'value')}))
+      samples_constructed_body = (' '.join((','.join((measurement,
```
meteorfox: Extract the logic to construct the request body for the InfluxDB to its own method.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+    conn.request('POST', '/query?' + body, headers)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
+    return response_status, response_response
+
+  def _WriteData(self, influx_uri, influx_db_name, data):
+    body = data
+    headers = {"Content-type": "application/octet-stream"}
+    conn = httplib.HTTPConnection(influx_uri)
```
meteorfox: Is it possible to re-use the connection instead of building a new one.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
 
+class InfluxDBPublisher(SamplePublisher):
+
+  def __init__(self, influx_uri=None, influx_db_name=None):
+    # set to default above in flags unless changed
+    self.influx_uri = influx_uri
+    self.influx_db_name = influx_db_name
+
+  def PublishSamples(self, samples):
+    successful_http_request_codes = [200, 202, 204]
```
meteorfox: HTTP error handling should probably happen within the method that makes the request.

perfkitbenchmarker/publisher.py
meteorfox: Can you post a real example in the comments of how the final request body looks like? It's hard to follow what is the format of this string.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+    sample = list(samples)
+    number_of_samples = 0
+    for sample in samples:
+      number_of_samples += 1
```
meteorfox: No need to count the samples. Simply using `len(samples)` should return the number of samples.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+                    'Request could not be completed due to: ',
+                    create_db_response)
+
+    logging.debug('writing samples to publisher: writing',
+                  number_of_samples, ' samples.')
+    self._writeData(self.influx_uri, self.influx_db_name,
+                    samples_constructed_body)
+    logging.debug(create_db_status,
+                  'Request could not be completed due to: ',
+                  create_db_response)
```
meteorfox: Won't this always log that the request could not be completed?

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+    self.influx_db_name = influx_db_name
+
+  def PublishSamples(self, samples):
+    successful_http_request_codes = [200, 202, 204]
+    sample = list(samples)
+    number_of_samples = 0
+    for sample in samples:
+      number_of_samples += 1
+      timestamp = int((10 ** 9) * sample['timestamp'])
+      measurement = 'perfkitbenchmarker'
```
vhe182: string.join(measurement) doesn't destroy the measurement string, why not move the initialization of measurement before the for loop?

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,81 @@ def _deDotKeys(self, res):
...
+                                                             'sample_uri')})))
+      field_set = '%s_%s=%s' % (self.FormatSampleForInfluxDB({k: sample[k]
+                                for k in ('metric', 'unit', 'value')}))
+      samples_constructed_body = (' '.join((','.join((measurement,
+                                           tag_set))),
+                                           (' '.join((field_set,
+                                                     timestamp))))) + "\n"
+
+    create_db_status, create_db_response = self. \
+        _createInfluxDB(self.influx_uri, self.influx_db_name)
```
vhe182: _createInfluxDB() isn't a function in this file.  Did you mean _createDB() ?

besanradwan:
@meteorfox @vhe182

perfkitbenchmarker/publisher.py
```diff
@@ -106,6 +108,14 @@
     'of all samples as metadata. Multiple key-value pairs may be specified '
     'by separating each pair by commas.')
 
+flags.DEFINE_string(
+    'influx_uri', 'localhost:8086',
+    'The Influx DB endpoint for queries address and port')
```
meteorfox: nitpick: The help message readability could be improved, I think you can remove 'endpoint for queries'.

Also, I think it needs clarification on what's the format of the string it expects. It's not clear if passing 'http://hostname:port/' is valid, or if only 'hostname:port' is expected.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,90 @@ def _deDotKeys(self, res):
...
+    params = data
+    header = {"Content-type": "application/octet-stream"}
+    conn = httplib.HTTPConnection(self.influx_uri)
+    conn.request('POST', '/write?' + 'db=' + self.influx_db_name, params,
+                 headers=header)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
+    if response_status in successful_http_request_codes:
+      logging.debug('writing samples to publisher: writing samples.')
```
meteorfox: nitpick: Log message should start with capital letter for consistency.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,90 @@ def _deDotKeys(self, res):
...
+    else:
+      logging.debug(response_status,
+                    'Request could not be completed due to: ',
+                    response_response)
+      raise httplib.HTTPException
+
```
meteorfox: I'm not 100% sure if Python cleans up the connection, but I think it wouldn't hurt to close the connection after we are done using it.

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,90 @@ def _deDotKeys(self, res):
...
+    conn = httplib.HTTPConnection(self.influx_uri)
+    conn.request('POST', '/write?' + 'db=' + self.influx_db_name, params,
+                 headers=header)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
+    if response_status in successful_http_request_codes:
+      logging.debug('writing samples to publisher: writing samples.')
+    else:
+      logging.debug(response_status,
```
meteorfox: nitpick: Instead of using local variables `response_status` and `response_response`, simply use the response object directly (i.e. `response.status`, `response.response`)

perfkitbenchmarker/publisher.py
```diff
@@ -640,6 +650,90 @@ def _deDotKeys(self, res):
...
+  def _CreateDB(self):
+    successful_http_request_codes = [200, 202, 204]
+    header = {'Content-type': 'application/x-www-form-urlencoded',
+              'Accept': 'text/plain'}
+    params = urllib.urlencode({'q': 'CREATE DATABASE ' + self.influx_db_name})
+    conn = httplib.HTTPConnection(self.influx_uri)
+    conn.request('POST', '/query?' + params, headers=header)
+    response = conn.getresponse()
+    response_status = response.status
+    response_response = response.response
```
meteorfox: See command below about `response_status` and `response_response`

besanradwan:
@meteorfox @vhe182 please review.

tests/publisher_test.py
```diff
@@ -404,3 +404,94 @@ def testUsesUnionOfMetaKeys(self):
...
+                      'run_uri': 'bba3', 'sample_uri': 'bb', 'timestamp': 55,
+                      'metadata': {}},
+                     {'test': 'testa', 'metric': '3', 'official': 47.0,
+                     'value': 'non', 'unit': 'us', 'owner': 'Rackspace',
+                      'run_uri': '5rtw', 'sample_uri': '5r', 'timestamp': 123}]
+      formated_samples = []
+      for data in sample_data:
+        formated_samples.append(self.mock_db._ConstructSample(sample))
+        self.assertEqual(formated_samples, 'perfkitbenchmarker,metric=1,\
+                                          official=1.0,owner=Rackspace,\
```
meteorfox: Either use multi-line string wrapped with `( )`, instead of breaking the line with `\`, or perhaps consider using a data file for this string since it's very long. See the `tests/linux_benchmarks/unixbench_benchmark_test.py` for an example. https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/tests/linux_benchmarks/unixbench_benchmark_test.py#L29-L33

besanradwan:
@meteorfox @gareth-ferneyhough please review

tests/publisher_test.py
```diff
@@ -404,3 +404,90 @@ def testUsesUnionOfMetaKeys(self):
     rows = list(reader)
     self.assertEqual(['key1', 'key3'], reader.fieldnames[-2:])
     self.assertEqual(3, len(rows))
+
+
+  class InfluxDBPublisherTestCase(unittest.TestCase):
+    def setUp(self):
+      self.db_name = 'test_db'
+      self.db_uri = 'test'
+      self.mock_db = publisher.InfluxDBPublisher(self.db_uri, self.db_name)
```
meteorfox: Since this is the actual instance of a publisher, I think `mock` shouldn't be included in the variable name, since it can be confused for a mock object.

tests/publisher_test.py
```diff
@@ -404,3 +404,90 @@ def testUsesUnionOfMetaKeys(self):
...
+                         '123000000000\nperfkitbenchmarker,metric=2,'
+                         'official=14.0,owner=Rackspace,run_uri=bba3,'
+                         'test=testb,unit=MB,sample_uri=bb value=non '
+                         '55000000000\nperfkitbenchmarker,metric=3,'
+                         'official=47.0,owner=Rackspace,run_uri=5rtw,'
+                         'test=testa,unit=us,sample_uri=5r '
+                         'value=non 123000000000')
+      for data in sample_data:
+        formated_samples.append(self.mock_db._ConstructSample(sample))
+        self.assertEqual(formated_samples, expected_output)
```
meteorfox: I'm not sure I understand how this test works.

Wouldn't this be calling an assertEqual between the `formatted_samples` list and the string `expected_output`? Also, the list `formatted_samples` changes in every iteration as the samples are added but the string `expected output` remains constant. This test would fail every time.

tests/publisher_test.py
```diff
@@ -404,3 +404,90 @@ def testUsesUnionOfMetaKeys(self):
     rows = list(reader)
     self.assertEqual(['key1', 'key3'], reader.fieldnames[-2:])
     self.assertEqual(3, len(rows))
+
+
+  class InfluxDBPublisherTestCase(unittest.TestCase):
```
meteorfox: This is the line why your tests aren't running.

This whole class is nested inside the CSVPublisherTestCase class. Please unindent and recommit. Thanks!

besanradwan:
@meteorfox @vhe182 please review

tests/publisher_test.py
```diff
@@ -404,3 +404,112 @@ def testUsesUnionOfMetaKeys(self):
...
+                    'run_uri': '5rtw', 'sample_uri': '5r', 'timestamp': 123}]
+    formated_samples = []
+    s1 = ('perfkitbenchmarker,sample_uri=33,owner=Rackspace,official=1.0,'
+          'run_uri=323,metric=1,test=testc,unit=MB,bar=foo,more_info=2,'
+          'info=1 value=non 123000000000 \n perfkitbenchmarker,'
+          'sample_uri=bb,owner=Rackspace,official=14.0,run_uri=bba3,metric=2,'
+          'test=testb,unit=MB value=non 55000000000 \n perfkitbenchmarker,'
+          'sample_uri=5r,owner=Rackspace,official=47.0,run_uri=5rtw,metric=3,'
+          'test=testa,unit=us value=non 123000000000')
+    expected_sample_format = self.sortString(s1)
```
meteorfox: I don't think you should sort the expected string. It needs to match the line protocol ordered as defined in Influx here:
https://docs.influxdata.com/influxdb/v1.2/write_protocols/line_protocol_reference/

Also, based on that reference, there shouldn't be whitespace between the newline characters `\n`.

tests/publisher_test.py
```diff
@@ -404,3 +404,112 @@ def testUsesUnionOfMetaKeys(self):
...
+    expected_2 = expected_1.split()
+    expected_3 = ','.join(expected_2)
+    expected_4 = sorted(expected_3.split(','))
+    return expected_4
+
+  def sortList(self, stuff):
+    new_string = []
+    for i in stuff:
+      if "metadata" in i:
+        i = i.replace('metadata={', '').replace('}',
```
meteorfox: Isn't this removing `metadata`?

tests/publisher_test.py
```diff
@@ -404,3 +404,112 @@ def testUsesUnionOfMetaKeys(self):
...
+          'sample_uri=5r']
+    s2 = ['owner=Rackspace', 'unit=MB', 'run_uri=bba3', 'test=testb',
+          'timestamp=55', 'metric=2', 'official=14.0', 'metadata={}',
+          'value=non', 'sample_uri=bb']
+    s3 = ['owner=Rackspace', 'unit=MB', 'run_uri=323', 'test=testc',
+          'timestamp=123', 'metric=1', 'official=1.0',
+          "metadata={'info': '1','bar': 'foo', 'more_info': '2'}", 'value=non',
+          'sample_uri=33']
+
+    self.assertEqual(self.sortList(sorted(no_meta_keys)),
```
meteorfox: Maybe use `assertDictEqual(...)` instead? 
Ref: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertDictEqual

tests/publisher_test.py
```diff
@@ -404,3 +404,112 @@ def testUsesUnionOfMetaKeys(self):
...
+    s1 = ('perfkitbenchmarker,sample_uri=33,owner=Rackspace,official=1.0,'
+          'run_uri=323,metric=1,test=testc,unit=MB,bar=foo,more_info=2,'
+          'info=1 value=non 123000000000 \n perfkitbenchmarker,'
+          'sample_uri=bb,owner=Rackspace,official=14.0,run_uri=bba3,metric=2,'
+          'test=testb,unit=MB value=non 55000000000 \n perfkitbenchmarker,'
+          'sample_uri=5r,owner=Rackspace,official=47.0,run_uri=5rtw,metric=3,'
+          'test=testa,unit=us value=non 123000000000')
+    expected_sample_format = self.sortString(s1)
+    for data in sample_data:
+      formated_samples.append(self.test_db._ConstructSample(data))
```
meteorfox: Since `_ConstructSample(data)` returns a string, why not simply test for that? No need to join them or append them in a list.

If you want to test the expected data format that gets sent to Influx, I'd suggest then to stub out `_CreateDB()` and `_WriteData()` and add a test for `_Publish()` to assert that `_WriteData` gets called with the data format you expect.

besanradwan:
@meteorfox @vhe182 Please review

besanradwan:
@gareth-ferneyhough please review

meteorfox:
@gareth-ferneyhough PTAL, I intend to merge this PR to master. It adds a publisher implementation for InfluxDB. Thanks!

tests/publisher_test.py
```diff
@@ -404,3 +404,143 @@ def testUsesUnionOfMetaKeys(self):
...
+         'run_uri=bba3,sample_uri=bb,metric=2,unit=MB value=non 55000000000'),
+        ('perfkitbenchmarker,test=testa,official=47.0,owner=Rackspace,'
+         'run_uri=5rtw,sample_uri=5r,metric=3,unit=us value=non 123000000000')
+    ]
+
+    mock_publish_method.return_value = None
+    self.test_db.PublishSamples(samples)
+    mock_publish_method.assert_called_once_with(expected)
+
+
```
gareth-ferneyhough: Remove extra blank line.

perfkitbenchmarker/publisher.py
```diff
@@ -645,6 +657,92 @@ def _deDotKeys(self, res):
...
+    for k, v in sample.iteritems():
+      if v == '':
+        v = '\\"\\"'
+      v = str(v)
+      v = v.replace(',', '\,')
+      v = v.replace(' ', '\ ')
+      key_value_pairs.append('%s=%s' % (k, v))
+    return key_value_pairs
+
+  def _CreateDB(self):
```
gareth-ferneyhough: It looks like this is always called? What if you wish to publish to an existing database?

besanradwan: Hey @gareth-ferneyhough I made the requested change to remove the extra blank line, however as for why _CreateDB() is always called: this is because it is idempotent in the sense that if the DB already exists it will simply return a 200 code without re-creating it.

gareth-ferneyhough: Great. Can you just add a comment stating this fact? Then everything should be good to go.

besanradwan: Done! @gareth-ferneyhough thanks for the feedback.

gareth-ferneyhough:
All good. Thank you again.

Issue #1261, opened 2016-12-22
Don't print full usage on usage error


gareth-ferneyhough:
If a user provides an invalid flag to `pkb.py`, it prints a usage error along with all the valid flags. This makes it difficult for the user to see what the actual error was, as he/she must scroll up through the usage instructions to find the error.

It would be better to print minimal usage instructions when the user passes invalid flags, so that the error is clearly visible.

tedsta:
+1

meteorfox:
Also, it would be nice to be able to selectively print the flags info for a particular sub-module (e.g. provider, benchmarks, pkb flags, packages, etc) instead of the wall of text.

Something like `pkb.py --help=perfkitbenchmarker.provders.gcp` which would only show the flags of all the submodules matching the string `perfkitbenchmarker.provders.gcp`, and possibly the string could be a regexp.

Looks like the either method `FLAGS.ModuleHelp(module)`[1] or use one of the internal dicts in gflags that maps module names to a list of flags might do the trick.

[1] https://github.com/google/python-gflags/blob/master/gflags/flagvalues.py#L914-L925

meteorfox:
I have a POC for my suggestion #1268 PTAL

gareth-ferneyhough:
Thanks! I'll take a look early next week.

PR #1280, opened 2017-01-19
Print helpmatch usage instructions on error instead of flags. Closes …

gareth-ferneyhough:
…#1261


perfkitbenchmarker/pkb.py
```diff
@@ -263,7 +263,9 @@ def _ParseFlags(argv=sys.argv):
   try:
     argv = FLAGS(argv)
   except flags.FlagsError as e:
-    logging.error('%s\nUsage: %s ARGS\n%s', e, sys.argv[0], FLAGS)
+    logging.error('%s\nUsage: %s ARGS\n%s', e, sys.argv[0])
```
yuyantingzero: 3 %s in the string

PR #1282, opened 2017-01-20
Add prototype version of benchmark for reading from postgres via JDBC

ssisk:
As discussed, this relies on underlying changes to the dpb_service

Notably, that dpb_service would now have a RunTest method.  The RunTest only takes testname, pipeline options and the job_stdout_file.

RunTest, as we discussed will make a call roughly similar to:
mvn test-compile compile failsafe:integration-test -D beamTestPipelineOptions='[
      "--postgresIp=104.155.130.174",
      "--postgresUsername=postgres",
      "--postgresDatabaseName=yoink",
      "--postgresPassword=sroim3",
      "--postgresSsl=false"
    ]' -Dcheckstyle.skip=true -DskipITs=false

One thing to resolve is that I have been relying on running the tests only inside of a particular module (ie, that runs all the ITs in the module) - I'm not sure if that's still an option when we are using it in this context.

I'll investigate a mvn command line that works when run from the root.


ssisk:
This PR went into the wrong repo - my apologies.

PR #1284, opened 2017-02-03
Fixes a small typo in the anti-affinity FLAG lookup

joelio:
Script will fail when using anti-affinity rules due to an incorrect reference.


joelio:
I signed it!

meteorfox:
Good catch! LGTM

PR #1287, opened 2017-02-07
Fix bug of creating instance with local ssd using nvme interface

tohaowu:
Local-ssds with nvme interface created by GCE use name like nvme0n1, nvme0n2...
And those local ssds are under the /dev .


perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -62,3 +62,5 @@
                      'The boot disk size in GB for GCP VMs.')
 flags.DEFINE_enum('gce_boot_disk_type', None, ['pd-standard', 'pd-ssd'],
                   'The boot disk type for GCP VMs.')
+flags.DEFINE_enum('gce_ssd_interface', 'SCSI', ['SCSI', 'nvme'],
```
ispringer: I suggest making the values consistent with the gcloud CLI - "SCSI" and "NVME", both all caps. You could optionally still support "scsi" and "nvme" as well, since developers dislike caps.

From `gcloud compute instances create --help`:

```
     --local-ssd=[device-name=DEVICE-NAME],[interface=INTERFACE]
        Attaches a local SSD to the instances.

        This flag is currently in BETA and may change without notice.

         device-name
            Optional. A name that indicates the disk name the guest operating
            system will see. If omitted, a device name of the form local-ssd-N
            will be used.

         interface
            Optional. The kind of disk interface exposed to the VM for this
            SSD. Valid values are SCSI and NVME. SCSI is the default and is
            supported by more guest operating systems. NVME may provide higher
            performance.
```

gareth-ferneyhough:
LGTM.

PR #1288, opened 2017-02-09
add support for AWS Spot Instances

dlott:
Added `aws_spot_instances` flag to spin up Spot instances instead of OnDemand.
`aws_spot_price` also needs to be set to control the max price that will be paid.

If a Spot instance cannot be spun up immediately (usually due to your price being too low), the run will abort.

### DIFF
--- a/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
+++ b/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
@@ -542,6 +542,8 @@ class AwsVirtualMachine(virtual_machine.BaseVirtualMachine):
     self.spot_instance_request_id =\
         create_response['SpotInstanceRequests'][0]['SpotInstanceRequestId']
 
+    util.AddDefaultTags(self.spot_instance_request_id, self.region)
+
     while True:
       describe_sir_cmd = util.AWS_PREFIX + [
           'ec2',
--- a/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
+++ b/perfkitbenchmarker/providers/aws/aws_virtual_machine.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import OrderedDict
 
 """Class to represent an AWS Virtual Machine object.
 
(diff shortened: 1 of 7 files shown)


gareth-ferneyhough:
In progress. Do not merge yet.

dlott:
This is ready for review.

tests/aws_test.py
```diff
@@ -129,6 +150,67 @@ def testInstanceDeleted(self):
...
+        ['aws',
+         '--output',
+         'json',
+         'ec2',
+         'request-spot-instances',
+         '--region=us-east-1',
+         '--spot-price=123.45',
+         '--client-token=00000000-1111-2222-3333-444444444444',
+         '--launch-specification='
+         '{"ImageId":"ami-12345","InstanceType":"c3.large",'
```
cwilkes: I would do triplequotes here

``` 
''.join("""
{"ImageId":.....
 "KeyName": ...
}
"""
```

or 
```
json.dumps({'ImageId':....})
```

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -138,9 +152,10 @@ def GetBlockDeviceMap(machine_type, root_volume_size_gb=None,
 
   if machine_type in NUM_LOCAL_VOLUMES:
     for i in xrange(NUM_LOCAL_VOLUMES[machine_type]):
-      mappings.append({
-          'VirtualName': 'ephemeral%s' % i,
-          'DeviceName': '/dev/xvd%s' % chr(ord(DRIVE_START_LETTER) + i)})
+      od = OrderedDict()
+      od['VirtualName'] = 'ephemeral%s' % i
```
cwilkes: While you're in there changing this I would go with the named parameter way:
```
od = OrderedDict(
   VirtualName='ephemeral%s' % i,
  DeviceName= '/dev/xvd%s' % chr(ord(DRIVE_START_LETTER) + i)
)
```

tests/aws_test.py
cwilkes: Nevermind about this -- you'll also have to split the text on newline and with all that it begins to look uglier than the original.

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
dlott: Unfortunately, that loses the ordering.  From https://docs.python.org/2.7/library/collections.html#collections.OrderedDict

> The OrderedDict constructor and update() method both accept keyword arguments, but their order is lost because Python’s function call semantics pass-in keyword arguments using a regular unordered dictionary.

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -465,6 +501,69 @@ def _Create(self):
...
+        ('DeviceIndex', 0),
+        ('AssociatePublicIpAddress', True),
+        ('SubnetId', self.network.subnet.id)])]
+    launch_specification = OrderedDict([
+        ('ImageId', self.image),
+        ('InstanceType', self.machine_type),
+        ('KeyName', 'perfkit-key-%s' % FLAGS.run_uri),
+        ('Placement', placement)])
+    if block_device_map:
+      print 'RAW block_device_map'
```
dlott: whoops that was for debugging.  removed.

dlott: FYI the vm_util.IssueCommand logs the actual command that is run.

perfkitbenchmarker/providers/aws/aws_virtual_machine.py
```diff
@@ -465,6 +501,69 @@ def _Create(self):
...
+      assert len(sir_response) == 1, 'Expected exactly 1 SpotInstanceRequest'
+
+      status_code = sir_response[0]['Status']['Code']
+
+      if status_code in SPOT_INSTANCE_REQUEST_HOLDING_STATUSES or \
+         status_code in SPOT_INSTANCE_REQUEST_TERMINAL_STATUSES:
+        message = sir_response[0]['Status']['Message']
+        raise errors.Resource.CreationError(message)
+      elif status_code == "fulfilled":
+        self.id = sir_response[0]['InstanceId']
```
dlott: when this succeeds, the next (logged) command is an `aws ec2 describe-instances` so you should know it happens

tests/aws_test.py
dlott: triple quotes brings in a bunch of white space.  json.dumps doesn't order things consistently.  ugly as it is, i think this way is best?

cwilkes:
LGTM

PR #1290, opened 2017-02-09
Object storage service benchmark updates

tedsta:
- Fix object_storage_service benchmark for Azure
- Add option to create latency histogram samples
- Add support for cold reads tests via `--object_storage_objects_written_file`, `--object_storage_read_objects`, and `--object_storage_dont_delete_bucket`

To create objects for a cold reads test, you run the `api_multistream_writes` (`api_multistream` will work too) scenario with the `--object_storage_objects_written_file` flag specified. It'll dump a json structure storing the bucket name as well as all the objects' names and sizes.

To perform a cold reads test, you run the `api_multistream_reads` scenario with the `--object_storage_read_objects` flag pointed at the file that was created with the `--object_storage_written_file`. You can specify `--object_storage_dont_delete_bucket` if you want to keep the bucket around for another cold reads test later on. Otherwise, the bucket will be deleted at the end of the `api_multistream_reads` run.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -49,6 +49,7 @@ from perfkitbenchmarker import flags
 from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import object_storage_service
 from perfkitbenchmarker import sample
+from perfkitbenchmarker import temp_dir
 from perfkitbenchmarker import units
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.providers.gcp import gcs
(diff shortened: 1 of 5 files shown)


perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -1197,7 +1330,10 @@ def Cleanup(benchmark_spec):
 
   vm_util.RunThreaded(lambda vm: CleanupVM(vm, service), vms)
 
-  for bucket in buckets:
-    service.DeleteBucket(bucket)
+  # Only clean up buckets if we're not saving the objects for a later run
+  if FLAGS.object_storage_objects_written_file is None and \
+     not FLAGS.object_storage_dont_delete_bucket:
```
tedsta: create meaningful boolean variable for this condition so the code reads nicer

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -921,6 +987,55 @@ def MultiStreamWriteBenchmark(results, metadata, vms, command_builder,
...
+  """A benchmark for multi-stream read latency and throughput.
+
+  Args:
+    results: the results array to append to.
+    metadata: a dictionary of metadata to add to samples.
+    vms: the VMs to run the benchmark on.
+    command_builder: an APIScriptCommandBuilder.
+    bucket_name: the primary bucket to benchmark.
+    read_objects: List of lists of [object_name, object_size]. In the outermost
+      list, each element corresponds to a VM's worker process.
```
tedsta: s/VM's worker process/VM/

PR #1299, opened 2017-02-22
Let publisher.py republish already-collected samples.

noahl:
I often collect data and then realize I wish I had exported to
CSV. This change lets publisher.py run independently. In this mode, it
will read the JSON sample file that it always generates by default
(inside the run directory) and run the publish step again, in the same
way that it does during normal PKB runs. This gives the flexibility to
re-publish samples to anywhere they could be published initially.

### DIFF
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -767,24 +767,14 @@ def RepublishJSONSamples(path):
     path: the path to the JSON file.
   """
 
-  samples = []
   with open(path, 'r') as file:
-    for line in file:
-      try:
-        raw_dict = json.loads(line)
-        labels = raw_dict.pop('labels')
-        metadata = dict()
-        # labels is a comma-seprated list of key, value pairs
-        for pair in labels.split(','):
-          # Need to strip leading and trailing '|' from each pair
-          key, _, value = pair[1:-1].partition(':')
-          metadata[key] = value
-        raw_dict['metadata'] = metadata
-        samples.append(raw_dict)
-      except Exception as e:
-        logging.info('Exception processing sample %s', line)
-        logging.info('Exception was: %s', e)
-        raise e
+    samples = [json.loads(s) for s in file if s]
+  for sample in samples:
+    # Chop '|' at the beginning and end of labels and split labels by '|,|'
+    fields = sample.pop('labels')[1:-1].split('|,|')
+    # Turn the fields into [[key, value], ...]
+    key_values = [field.split(':', 1) for field in fields]
+    sample['metadata'] = {k: v for k, v in key_values}
 
   # We can't use a SampleCollector because SampleCollector.AddSamples depends on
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/publisher.py
```diff
@@ -736,3 +758,56 @@ def PublishSamples(self):
...
+  with open(path, 'r') as file:
+    for line in file:
+      try:
+        raw_dict = json.loads(line)
+        labels = raw_dict.pop('labels')
+        metadata = dict()
+        # labels is a comma-seprated list of key, value pairs
+        for pair in labels.split(','):
+          # Need to strip leading and trailing '|' from each pair
+          key, _, value = pair[1:-1].partition(':')
```
tedsta: Might I suggest this for loading samples?

```
samples = [json.loads(s) for s in file.read().split('\n') if s]
for sample in samples:
  # Chop '|' at the beginning and end of labels and split labels by '|,|'
  fields = sample['labels'][1:-1].split('|,|')
  # Turn the fields into a [[key, value], ...]
  key_values = [field.split(':', 1) for field in fields]
  sample['metadata'] = { k: v for k, v in key_values }
```

It's concise and very robust. Allows commas and even '|' in metadata values.

noahl: Very nice! I like your code. I pushed a very small variation on it - just have to pop 'labels' so that there's not an extra field in the result. Thanks a lot!

noahl:
@tedsta I think I've responded to your comment. PTAL.

perfkitbenchmarker/publisher.py
tedsta: Right, thanks!

perfkitbenchmarker/publisher.py
```diff
@@ -655,37 +658,56 @@ class SampleCollector(object):
...
     if metadata_providers is not None:
       self.metadata_providers = metadata_providers
     else:
       self.metadata_providers = DEFAULT_METADATA_PROVIDERS
 
-    if publishers is not None:
-      self.publishers = publishers
-    else:
-      self.publishers = SampleCollector._DefaultPublishers()
+    self.publishers = publishers.copy()
```
tedsta: Looks like this should change to
```python
self.publishers = publishers[:]
```
To make the tests happy

noahl: Ah shoot, I forgot to run the tests. Thank you! I wonder why this error didn't trigger when I ran the code.

I copy the list because we're modifying it. If someone passed in a list and then later reused that list for something else, they might be surprised to find the extra elements in it. I admit it's an unlikely scenario here, but I try to keep my functions clean because bugs like that can be really difficult to track down.

noahl:
@tedsta just uploaded a new version, please take a look. Thanks for your reviews!

tedsta:
LGTM thanks!

PR #1302, opened 2017-02-28
Fixed exception caused by publishers being None

tedsta:


perfkitbenchmarker/publisher.py
```diff
@@ -776,7 +776,7 @@ class SampleCollector(object):
       the run directory to the publishers list.
     run_uri: A unique tag for the run.
   """
-  def __init__(self, metadata_providers=None, publishers=None,
+  def __init__(self, metadata_providers=None, publishers=[],
```
ehankland: I'd prefer checking if publishers is none below and initializing it to an empty list - I like staying away from mutable default args.

Also, I notice that publishers (and not self.publishers) is getting extended online 790 and 792 - is that what's intended?

tedsta: Done

Good good catch, I believe it's supposed to modify self.publishers

perfkitbenchmarker/publisher.py
```diff
@@ -785,11 +785,14 @@ def __init__(self, metadata_providers=None, publishers=None,
     else:
       self.metadata_providers = DEFAULT_METADATA_PROVIDERS
 
-    self.publishers = publishers[:]
+    if publishers is not None:
+      self.publishers = publishers[:]
+    else:
+      self.publishers = []
```
ehankland: ```self.publishers = publishers[:] if publishers else []```
is more compact, but up to you

tedsta: Cool, I like it. Done

PR #1327, opened 2017-03-30
Update hadoop version.

maggiejiao:
Hadoop version 2.5.2 no longer exists. Updated to newer version 2.7.2.
http://www.us.apache.org/dist/hadoop/common/


asaksena:
LGTM. Thanks for the update

PR #1337, opened 2017-04-10
Replace flag matrices with one matrix/config

ehankland:


s-deitz:
This PR is almost a year old and it looks like we lost track of it. Is this still valid? If so, please reply and somebody will review in the next couple of weeks.

PR #1338, opened 2017-04-12
Install sysbench05plus package in a separate path from sysbench.

maggiejiao:
Sysbench 0.5 contains breaking changes from previous versions. All existing oltp benchmarks depending on older version of sysbench will break if we install 0.5 or later for them. Therefore, it's necessary that we have a separate installation path here for 0.5 and later so that when we install both sysbench and sybench 0.5, the test knows in which path to run the sysbench command.


maggiejiao:
@tedsta Please take a look. Thanks,

gareth-ferneyhough:
LGTM

PR #1340, opened 2017-04-13
Change --openstack_additional_flags to be a list flag.

yilei:
Otherwise when a string value is passed, each individual character in the string will be added here:  https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/ba0e0aa67bd6c8e4a4a92be2e2f6dad10113627e/perfkitbenchmarker/providers/openstack/utils.py#L112

If this indeed is a string flag, then the default value should be changed to None or an empty string.


tedsta:
LGTM, thanks

PR #1341, opened 2017-04-14
Move gflags installation from netperf benchmark to package.

maggiejiao:
Separate the package installation from prepare stage so that the test could run using a prebaked image under no-internet connection environment.


maggiejiao:
@tedsta Please take a look. Thanks

PR #1345, opened 2017-04-21
Always create VM_TMP_DIR before scp-ing files to it. Fixes #1344

### LINKED ISSUE #1344: RobustRemoteCommand fails is run after the VM is restarted
`RobustRemoteCommand` fails if it is run on a VM that has been rebooted, because the `/tmp/pkb` directory has been deleted. `linux_virtual_machine.py` should recreate the tmp directory if necessary.

gareth-ferneyhough:


gareth-ferneyhough:
Don't merge. This should be refactored by adding a `Reboot` method to virtual_machine that takes care of ensuring that the VM is in the appropriate state after reboot.

PR #1350, opened 2017-04-30
Fixed wrong argument in readme and added note on Amazon ES publisher.

Lirt:
Just small update to README. 

I mentioned that argument --zone is not working, there is only defined argument --zones. 

And I added one note on using AWS ElasticSearch Publisher from, where port 9200 is closed and only accessible port is 80. 
https://forums.aws.amazon.com/thread.jspa?messageID=681938


Lirt:
I signed it!

gareth-ferneyhough:
Thanks, LGTM.

PR #1357, opened 2017-05-09
Don't wait in wait_for_command if the command has already succeeded

gareth-ferneyhough:
* wait_for_command would always wait for the entire specified timeout, even if the command already returned. This PR fixes that by checking for the return code in the wait loop.

### DIFF
--- a/perfkitbenchmarker/scripts/wait_for_command.py
+++ b/perfkitbenchmarker/scripts/wait_for_command.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python2
 #
-# Copyright 2015 PerfKitBenchmarker Authors. All rights reserved.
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/scripts/wait_for_command.py
```diff
@@ -67,7 +67,7 @@ def main():
 
   start = time.time()
   return_code_str = None
-  while (time.time() < WAIT_TIMEOUT_IN_SEC + start):
+  while (time.time() < WAIT_TIMEOUT_IN_SEC + start and return_code_str is None):
```
tedsta: Consider `break`ing after `status.read()` so not to add another condition? LGTM

PR #1360, opened 2017-05-09
Storage flag

sspano:
@stfeng @gareth-ferneyhough 
Adapt mysql_service_benchmark  to handle storage size flag. Remove conflicting comments.

FYI @gareth-ferneyhough I did not change RDS side of this other than replacing the hard coded storage size with the flag value. This means that volume size is 1 so IOPS is capped will be capped by this.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
@@ -22,6 +22,25 @@ managed MySQL services.
 
 As other cloud providers deliver a managed MySQL service, we will add it here.
 
+As of May 2017 to make this benchmark run for GCP you must install the
+gcloud beta component. This is necessary because creating a Cloud SQL instance
+with a non-default storage size is in beta right now. This can be removed when
+this feature is part of the default components.
+See https://cloud.google.com/sdk/gcloud/reference/beta/sql/instances/create
+for more information.
+To run this benchmark for GCP it is required to install a non-default gcloud
+component. Otherwise this benchmark will fail.
+
+To ensure that gcloud beta is installed, type
+        'gcloud components list'
+into the terminal. This will output all components and status of each.
+Make sure that
+  name: gcloud Beta Commands
+  id:  beta
+has status: Installed.
+If not, run
+        'gcloud components install beta'
+to install it. This will allow this benchmark to properly create an instance.
 """
 import json
 import logging
@@ -61,6 +80,8 @@ flags.DEFINE_integer('sysbench_latency_percentile', 99,
 flags.DEFINE_integer('sysbench_report_interval', 2,
                      'The interval, in seconds, we ask sysbench to report '
                      'results.')
+flags.DEFINE_integer('storage_size', 100,
+                     'Storage size for SQL instance in GB.')
 
 BENCHMARK_NAME = 'mysql_service'
 BENCHMARK_CONFIG = """
@@ -89,14 +110,6 @@ RDS_CORE_TO_DB_CLASS_MAP = {
 RDS_DB_ENGINE = 'MySQL'
 RDS_DB_ENGINE_VERSION = '5.6.23'
 RDS_DB_STORAGE_TYPE_GP2 = 'gp2'
-# Storage IOPS capacity of the DB instance.
-# Currently this is fixed because the cloud provider GCP does not support
-# changing this setting. As soon as it supports changing the storage size, we
-# will expose a flag here to allow caller to select a storage size.
-# Default GCP storage size is 1TB PD-SSD which supports 10K Read or 15K Write
-# IOPS (12.5K mixed).
-# To support 12.5K IOPS on EBS-GP, we need 4170 GB disk.
-RDS_DB_STORAGE_GP2_SIZE = '4170'
 
 # A list of status strings that are possible during RDS DB creation.
 RDS_DB_CREATION_PENDING_STATUS = frozenset(
@@ -128,6 +141,10 @@ RESPONSE_TIME_TOKENS = ['min', 'avg', 'max', 'percentile']
 
 def GetConfig(user_config):
   return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+
+
+class StorageSizeFlagError(Exception):
+  pass
 
 
 class DBStatusQueryError(Exception):
@@ -440,6 +457,7 @@ class RDSMySQLBenchmark(object):
 
     # Get a list of zones and pick one that's different from the zone VM is in.
     new_subnet_zone = None
+    self._ValidateSize(FLAGS.storage_size)
     get_zones_cmd = util.AWS_PREFIX + ['ec2', 'describe-availability-zones']
     stdout, _, _ = vm_util.IssueCommand(get_zones_cmd)
     response = json.loads(stdout)
@@ -504,7 +522,7 @@ class RDSMySQLBenchmark(object):
         '--engine', RDS_DB_ENGINE,
         '--engine-version', RDS_DB_ENGINE_VERSION,
         '--storage-type', RDS_DB_STORAGE_TYPE_GP2,
-        '--allocated-storage', RDS_DB_STORAGE_GP2_SIZE,
+        '--allocated-storage', FLAGS.storage_size,
         '--vpc-security-group-ids', vm.group_id,
         '--master-username', vm.db_instance_master_user,
         '--master-user-password', vm.db_instance_master_password,
@@ -558,6 +576,18 @@ class RDSMySQLBenchmark(object):
     logging.info('Successfully created an RDS DB instance. Address is %s',
                  vm.db_instance_address)
     logging.info('Complete output is:\n %s', response)
+
+  def _ValidateSize(self, size):
+    """Validate flag for storage size and throw exception if invalid.
+
+    AWS supports storage sizes from 1GB to 16TB currently.
+
+    Args:
+      size: (GB).
+    """
+    if size < 1 or size > 16000:
+      raise StorageSizeFlagError('Storage size flag given is not valid. '
+                                 'Must be between 1 and 16000 GB for AWS.')
 
   def Cleanup(self, vm):
     """Clean up RDS instances, cleanup the extra subnet created for the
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -22,6 +22,25 @@
 
 As other cloud providers deliver a managed MySQL service, we will add it here.
 
+To make this benchmark run for GCP:
+To run this benchmark for GCP if is required to install a non-default gcloud
```
stfeng: typo: "if"

specify reasons why it's required to install a non-default gcloud component (because ...), also specify the time/date of this requirement ("As of May 2017")

stfeng: I see the reasons are specified below, properly better to pull that up to here for clarity.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -61,6 +80,7 @@
 flags.DEFINE_integer('sysbench_report_interval', 2,
                      'The interval, in seconds, we ask sysbench to report '
                      'results.')
+flags.DEFINE_integer('storage_size', 3400, 'Storage size for SQL instance.')
```
stfeng: specify the units for storage size, e.g., in GB

Also, does the default has to be this large? It has cost consequence. Probably we should specify a smaller default for those who just want to try out the benchmarks. 100?

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -668,13 +680,23 @@ def Prepare(self, vm):
     logging.info('Preparing MySQL Service benchmarks for Google Cloud SQL.')
 
     vm.db_instance_name = 'pkb%s' % FLAGS.run_uri
+    storage_size = self._validate_size(FLAGS.storage_size)
```
stfeng: Can we have a simple validation for RDS too? (for completeness).

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -751,6 +770,20 @@ def Prepare(self, vm):
...
+    """Validates flag for storage size, returns default if invalid.
+
+    Args:
+      size: (GB).
+    Returns:
+      size: (GB).
+
+    GCP supports storage sizes from 0 to 64TB currently.
+    """
+    if size < 0 or size > 64000:
```
stfeng: 0 is not supported either.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -751,6 +770,20 @@ def Prepare(self, vm):
...
+      size = 3334
+    return size
```
stfeng: If size is invalid we should throw an exception and fail the benchmark run. We can't silently change user's input to a default value....

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Handled in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Handled in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Handled in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Handled in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Handled in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -558,6 +572,21 @@ def Prepare(self, vm):
     logging.info('Successfully created an RDS DB instance. Address is %s',
                  vm.db_instance_address)
     logging.info('Complete output is:\n %s', response)
+     
+  def _validate_size(self, size):
```
gareth-ferneyhough: For some reason, PKB's style is CamelCase for function names, so please change it to be consistent with the rest of the file. You can leave the underscore.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -558,6 +572,21 @@ def Prepare(self, vm):
...
+    """Validates flag for storage size, throws exception if invalid.
```
gareth-ferneyhough: Change to "Validate flag for storage size and throw exception if invalid." The style guide requests that summary doc strings are written as a command, not a description. Also, move  "AWS supports storage sizes from 1GB to 16TB currently."  Up so that it is one line below the summary and above the args and returns.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -22,11 +22,31 @@
 
 As other cloud providers deliver a managed MySQL service, we will add it here.
 
+As of May 2017 to make this benchmark run for GCP you must install the
+gcloud beta component. This is necessary because creating a SQL instance with
```
stfeng: s/SQL/"Cloud SQL"/

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -559,6 +574,21 @@ def Prepare(self, vm):
...
+
+    Args:
+      size: (GB).
+    Returns:
+      size: (GB).
+
+    AWS supports storage sizes from 1GB to 16TB currently.
+    """
+    if size < 1 or size > 16000:
+      sys.exit('Storage size flag given is not valid. Must be between '
```
stfeng: instead of doing sys.exit() can you throw an exception instead? (follow the example of error handling logic in other places of this code).

By throwing exception here, the parent classes of PKB may decide to catch those and initiate a cleanup if necessary, however, doing sys.exit() would void all that possibility.

(As FYI: Currently PKB doesn't usually do cleanup upon benchmark failure, so if a user sees an exception here, they will need to call pkb.py --run_stage=cleanup --run_uri == this run uri, but the key here by throwing an exception up from here, we all the possibility to implement cleanup at the parent class in the future).

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -751,6 +788,21 @@ def Prepare(self, vm):
...
+
+    Args:
+      size: (GB).
+    Returns:
+      size: (GB).
+
+    GCP supports storage sizes from 1GB to 64TB currently.
+    """
+    if size < 1 or size > 64000:
+      sys.exit('Storage size flag given is not valid. Must be between '
```
stfeng: same as above.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -440,6 +454,7 @@ def Prepare(self, vm):
 
     # Get a list of zones and pick one that's different from the zone VM is in.
     new_subnet_zone = None
+    storage_size = self._ValidateSize(FLAGS.storage_size)
```
stfeng: It doesn't seem necessary to return anything from the validatesize() function, if it never changes the size.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -559,6 +574,21 @@ def Prepare(self, vm):
...
+
+    AWS supports storage sizes from 1GB to 16TB currently.
+
+    Args:
+      size: (GB).
+    Returns:
+      size: (GB).
+    """
+    if size < 1 or size > 16000:
+      sys.exit('Storage size flag given is not valid. Must be between '
```
stfeng: change sys.exit()

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -751,6 +791,18 @@ def Prepare(self, vm):
     logging.info('Set root password completed. Stdout:\n%s\nStderr:\n%s',
                  stdout, stderr)
 
+  def _ValidateSize(self, size):
+    """Validate flag for storage size and throw exception if invalid
```
gareth-ferneyhough: add "." to the end. The google linter complains if there is no punctuation at the end of the summary.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -440,6 +458,8 @@ def Prepare(self, vm):
 
     # Get a list of zones and pick one that's different from the zone VM is in.
     new_subnet_zone = None
+    storage_size = FLAGS.storage_size
```
stfeng: local variable storage_size is unnecessary, you can just use FLAGS.storage_size here and below.

Same comment for GCP.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -751,6 +791,18 @@ def Prepare(self, vm):
...
+  def _ValidateSize(self, size):
+    """Validate flag for storage size and throw exception if invalid
+
+    GCP supports storage sizes from 1GB to 64TB currently.
+
+    Args:
+      size: (GB).
+    """
+    if size < 1 or size > 64000:
+      sys.exit('Storage size flag given is not valid. Must be between '
```
stfeng: forgot this one? (sys.exit)

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Fixed in new commit.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Yes haha. Fixed in new commit.

stfeng:
LGTM from me, please wait for LGTM from Gareth as well. Once you have that, please do an end-to-end test with the final code, reply back when that is done.

sspano:
End to end testing with valid and invalid cases, acts as expected.

PR #1362, opened 2017-05-10
Adds the gce_remote_access_firewall_rule flag 

dlott:
PKB will rely on this rule to access VMs rather than create individual RemoteAccessPort firewall rules.


perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -40,6 +40,9 @@
 flags.DEFINE_string('gce_subnet_addr', '10.128.0.0/20', 'Address range to the '
                     'subnet, given in CDR notation. Not used unless '
                     '--gce_subnet_region is given.')
+flags.DEFINE_string('gce_remote_access_firewall_rule', None, 'The name of an '
```
ehankland: Why do we need the name of the rule? It seems sufficient to know whether such a rule exists (and it doesn't look like you use the name for anything)

dlott: I wanted to keep the flag parallel to gce_network_name.  It is unused, but if we want to get fancy in the future and validate that the rule actually exists, we can.

PR #1363, opened 2017-05-11
Optionally create failed run sample

gareth-ferneyhough:
## Release Notes:
* Added a command line flag, `create_failed_run_samples` which, if `True`, will instruct PKB to create 'Run Failed' samples for failed run stages. If a publisher is used, then these failed run samples will be published as well.

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -594,12 +594,33 @@ def RunBenchmark(spec, collector):
 
 
 def MakeFailedRunSample(error_message, run_stage_that_failed):
-    metadata = {
-        'error_message': error_message,
-        'run_stage': run_stage_that_failed,
-        'flags': str(flag_util.GetProvidedCommandLineFlags())
-    }
-    return [sample.Sample('Run Failed', 1, 'Run Failed', metadata)]
+  """Create a sample.Sample representing a failed run stage.
+
+  The sample metric will have the name 'Run Failed';
+  the value will be 1 (has to be convertable to a float),
+  and the unit will be 'Run Failed' (for lack of a better idea).
+
+  The sample metadata will include the error message from the
+  Exception, the run stage that failed, as well as all PKB
+  command line flags that were passed in.
+
+  Args:
+    error_message: error message that was caught, resulting in the
+      run stage failure.
+    run_stage_that_failed: run stage that failed by raising an Exception
+
+  Returns:
+    a sample.Sample representing the run stage failure.
+  """
+  # Note: currently all provided PKB command line flags are included in the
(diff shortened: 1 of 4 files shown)


perfkitbenchmarker/pkb.py
```diff
@@ -576,6 +593,33 @@ def RunBenchmark(spec, collector):
...
+
+  Returns:
+    a sample.Sample representing the run stage failure.
+  """
+  metadata = {
+      'error_message': error_message,
+      'run_stage': run_stage_that_failed,
+      'flags': str(flag_util.GetProvidedCommandLineFlags())
+  }
+  return [sample.Sample('Run Failed', 1, 'Run Failed', metadata)]
```
gareth-ferneyhough: Maybe this should be 'Run Stage Failed'?

tests/flag_util_test.py
```diff
@@ -292,5 +293,32 @@ def testBadYAML(self):
...
+        'flag2': MockFlag('flag2', '2', True),
+        'flag3': MockFlag('flag3', '3', False)
+    }
+    patcher = mock.patch(flag_util.__name__ + '.FLAGS')
+    flags_mock = patcher.start()
+    flags_mock.FlagDict.return_value = flag_dict
+    self.addCleanup(patcher.stop)
+
+  def testGetProvidedCommandLineFlags(self):
+    self.assertDictEqual({
```
dlott: Do you need to make this an ordered dictionary so that it always passes unit tests? See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/commit/6bd92986fa35e9456ef40a872c749292f66ded23#diff-44ccde18eb382a76acff381351dc2ef2R156

tests/pkb_test.py
```diff
@@ -0,0 +1,123 @@
...
+    self.make_failed_run_sample_mock = self.patchPkbFunction(
+        'MakeFailedRunSample')
+
+    self.flags_mock.run_stage = [
+        stages.PROVISION, stages.PREPARE, stages.RUN, stages.CLEANUP,
+        stages.TEARDOWN
+    ]
+
+  def testCreateProvisionFailedSamples(self):
+    spec = mock.MagicMock()
```
dlott: worth creating these in setup instead of each test method?

tests/flag_util_test.py
gareth-ferneyhough: No, `assertDictEqual` takes care of that for you. Order doesn't matter.

tests/pkb_test.py
gareth-ferneyhough: sure

dlott:
LGTM

PR #1367, opened 2017-05-16
Move data load from run to prepare phase

sspano:
@stfeng @gareth-ferneyhough 
Move data load into prepare phase of Sysbench (previously in run phase). This allows multiple runs on the same instance.
Moved tables count and table size to vm attribute so they persist over multiple runs.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
@@ -279,7 +279,7 @@ def ParseSysbenchOutput(sysbench_output, results, metadata):
         metadata))
 
 
-def _IssueSysbenchCommand(vm, duration):
+def _IssueSysbenchCommand(vm, duration, metadata):
   """Issues a sysbench run command given a vm and a duration.
 
       Does nothing if duration is <= 0
@@ -287,25 +287,29 @@ def _IssueSysbenchCommand(vm, duration):
   Args:
     vm: The test VM to issue command to.
     duration: the duration of the sysbench run.
+    metadata: The PKB metadata to be passed along to the final results.
 
   Returns:
     stdout, stderr: the result of the command.
   """
   stdout = ''
   stderr = ''
+  num_threads = metadata['sysbench_thread_count']
+  tables_count = metadata['mysql_svc_oltp_tables_count']
+  table_size = metadata['mysql_svc_oltp_tables_size']
   oltp_script_path = sysbench05plus.OLTP_SCRIPT_PATH
   if duration > 0:
     run_cmd_tokens = ['%s' % sysbench05plus.SYSBENCH05PLUS_PATH,
                       '--test=%s' % oltp_script_path,
                       '--mysql_svc_oltp_tables_count=%d' %
-                      FLAGS.mysql_svc_oltp_tables_count,
+                      tables_count,
                       '--oltp-table-size=%d' %
-                      FLAGS.mysql_svc_oltp_table_size,
+                      table_size,
                       '--rand-init=%s' % RAND_INIT_ON,
                       '--db-ps-mode=%s' % DISABLE,
                       '--oltp-dist-type=%s' % UNIFORM,
                       '--oltp-read-only=%s' % OFF,
-                      '--num-threads=%d' % FLAGS.sysbench_thread_count,
+                      '--num-threads=%d' % num_threads,
                       '--percentile=%d' % FLAGS.sysbench_latency_percentile,
                       '--report-interval=%d' %
                       FLAGS.sysbench_report_interval,
@@ -336,6 +340,44 @@ def _RunSysbench(vm, metadata):
   Returns:
     Results: A list of results of this run.
   """
+  results = DATA_LOAD_RESULTS
+
+  if not hasattr(vm, 'db_instance_address'):
+    logging.error(
+        'Prepare has likely failed, db_instance_address is not found.')
+    raise DBStatusQueryError('RunSysbench: DB instance address not found.')
+
+  # Now run the sysbench OLTP test and parse the results.
+  # First step is to run the test long enough to cover the warmup period
+  # as requested by the caller. Second step is the 'real' run where the results
+  # are parsed and reported.
+
+  warmup_seconds = FLAGS.sysbench_warmup_seconds
+  if warmup_seconds > 0:
+    logging.info('Sysbench warm-up run, duration is %d', warmup_seconds)
+    _IssueSysbenchCommand(vm, warmup_seconds, metadata)
+
+  run_seconds = FLAGS.sysbench_run_seconds
+  logging.info('Sysbench real run, duration is %d', run_seconds)
+  stdout, _ = _IssueSysbenchCommand(vm, run_seconds, metadata)
+  logging.info('\n Parsing Sysbench Results...\n')
+  ParseSysbenchOutput(stdout, results, metadata)
+
+  return results
+
+
+def _PrepareSysbench(vm, metadata):
+  """Prepare the Sysbench OLTP test with data loading stage.
+
+  Data loaded on the DB instance indicated by the vm.db_instance_address.
+
+  Args:
+    vm: The client VM that will issue the sysbench test.
+    metadata: The PKB metadata to be passed along to the final results.
+
+  Returns:
+    results: A list of results of the data loading step.
+  """
   results = []
 
   if not hasattr(vm, 'db_instance_address'):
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -336,8 +336,49 @@ def _RunSysbench(vm, metadata):
...
   """
-  results = []
+  results = DATA_LOAD_RESULTS
+
+  if not hasattr(vm, 'db_instance_address'):
+    logging.error(
+        'Prepare has likely failed, db_instance_address is not found.')
+    raise DBStatusQueryError('RunSysbench: DB instance address not found.')
+
+  # Now run the sysbench OLTP test and parse the results.
```
gareth-ferneyhough: This is a bit awkward. At the least, I suggest moving 'warm-up' and 'run' into constants so that you aren't performing logic on magic strings.
Even better (IMO) would be to lose the for loop entirely:

 ```
# Now run the sysbench OLTP test and parse the results.
# First step is to run the test long enough to cover the warmup period
# as requested by the caller. Then we do the "real" run, parse and report
# the results.

if FLAGS.sysbench_warmup_seconds > 0:
  warmup_seconds = FLAGS.sysbench_warmup_seconds
  logging.info('Sysbench warm-up run, duration is %d', warmup_seconds)
  _IssueSysbenchCommand(vm, warmup_seconds)
 
run_seconds = FLAGS.sysbench_run_seconds
logging.info('Sysbench real run, duration is %d', run_seconds)
stdout, _ = _IssueSysbenchCommand(vm, run_seconds)
logging.info('\n Parsing Sysbench Results...\n')
ParseSysbenchOutput(stdout, results, metadata)

return results
```

Seems more readable to me.

File: perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py [hunk 2]
```diff
@@ -840,9 +863,26 @@ def Prepare(benchmark_spec):
   # Setup common test tools required on the client VM
   vms[0].Install('sysbench05plus')
 
+
+  vms[0].mysql_svc_oltp_tables_count = FLAGS.mysql_svc_oltp_tables_count
```
gareth-ferneyhough: We traditionally store state on the benchmark_spec itself, not the vm object. I would suggest doing that here, since these flag values are not vm-specific.

gareth-ferneyhough: So, `benchmark_spec.mysql_svc_oltp_tables_count = FLAGS.mysql_svc_oltp_tabes_count`

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -336,8 +336,49 @@ def _RunSysbench(vm, metadata):
...
+def _PrepareSysbench(vm, metadata):
+  """Prepares the Sysbench OLTP test with data loading stage.
+
+  Data loaded on the DB instance indicated by the vm.db_instance_address.
+
+  Args:
+    vm: The client VM that will issue the sysbench test.
+    metadata: The PKB metadata to be passed along to the final results.
 
+  Results:
```
gareth-ferneyhough: Returns: not Results:

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -336,8 +336,49 @@ def _RunSysbench(vm, metadata):
...
+    if phase == 'run':
+      # We only need to parse the results for the "real" run.
+      logging.info('\n Parsing Sysbench Results...\n')
+      ParseSysbenchOutput(stdout, results, metadata)
+
+  return results
+
+
+def _PrepareSysbench(vm, metadata):
+  """Prepares the Sysbench OLTP test with data loading stage.
```
gareth-ferneyhough: Change Prepares to Prepare. Docstring summaries should be a command, not a description.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Agreed but we don't pass the benchmark spec to PrepareSysbench or IssueSysbenchCommand..

As per offline conversation will leave as is unless @stfeng has a different opinion.

stfeng: +1 

I think these do belong to the benchmark_spec object, table count and table size are exactly "benchmark spec". ;)

(Sam: it's my bad on suggesting putting these to the VM object, benchmark_spec is much better.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Agreed. Changed!

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Done.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
sspano: Done.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
stfeng: chatted offline, we can simply change these two functions to pass in the benchmark_spec or the metadata object which also captures required information from benchmark_spec.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -818,6 +840,7 @@ def Cleanup(self, vm):
 MYSQL_SERVICE_BENCHMARK_DICTIONARY = {
     providers.GCP: GoogleCloudSQLBenchmark(),
     providers.AWS: RDSMySQLBenchmark()}
+DATA_LOAD_RESULTS = []
```
stfeng: 1. Add a comment explaining why this variable needs to be global.

2. perhaps rename it to "DATA_LOADING_RESULTS"? Similarly for all phrases below that says "data load results", consider renaming them to "data loading results".

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -840,9 +863,26 @@ def Prepare(benchmark_spec):
...
+      'mysql_svc_oltp_table_size': vms[0].mysql_svc_oltp_table_size,
+      'mysql_svc_db_instance_cores': FLAGS.mysql_svc_db_instance_cores,
+      'sysbench_warm_up_seconds': FLAGS.sysbench_warmup_seconds,
+      'sysbench_run_seconds': FLAGS.sysbench_run_seconds,
+      'sysbench_thread_count': FLAGS.sysbench_thread_count,
+      'sysbench_latency_percentile': FLAGS.sysbench_latency_percentile,
+      'sysbench_report_interval': FLAGS.sysbench_report_interval
+  }
+  DATA_LOAD_RESULTS.append(_PrepareSysbench(vms[0], metadata))
+  logging.info('Data Load Results: \n%s', DATA_LOAD_RESULTS)
```
stfeng: change this to 

print DATA_LOAD_RESULTS

which will output to stdout, which is much easier to read and consume. (logging.info will log to stderr which will mix this up with all other output, very noisy...)

sspano: Done

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -26,7 +26,7 @@
 @functools32.lru_cache()
 def GetDefaultProject():
   """Get the default project."""
-  cmd = ['gcloud', 'config', 'list', '--format=json']
+  cmd = [FLAGS.gcloud_path, 'config', 'list', '--format=json']
```
stfeng: This change doesn't below to here. We need to figure out what happened.

Please find out the latest state of this file on GCP/master and then revert this one to that state.

gareth-ferneyhough: It is from https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/commit/3ff5327958c0b16c38b7685927f35b6767cb20a3

You just need to rebase on master I believe.

perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py
```diff
@@ -840,9 +887,25 @@ def Prepare(benchmark_spec):
...
+      'mysql_svc_oltp_tables_count': benchmark_spec.mysql_svc_oltp_tables_count,
+      'mysql_svc_oltp_table_size': benchmark_spec.mysql_svc_oltp_table_size,
+      'mysql_svc_db_instance_cores': FLAGS.mysql_svc_db_instance_cores,
+      'sysbench_warm_up_seconds': FLAGS.sysbench_warmup_seconds,
+      'sysbench_run_seconds': FLAGS.sysbench_run_seconds,
+      'sysbench_thread_count': FLAGS.sysbench_thread_count,
+      'sysbench_latency_percentile': FLAGS.sysbench_latency_percentile,
+      'sysbench_report_interval': FLAGS.sysbench_report_interval
+  }
+  DATA_LOAD_RESULTS.append(_PrepareSysbench(vms[0], metadata))
```
stfeng: [note] As found in test, this should be assignment ('='). Otherwise DATA_LOAD_RESULTS becomes a list of list.

sspano:
E2E test passed.

PR #1374, opened 2017-05-18
Add script to automagically drive a new release for PKB

tedsta:
In the beginning it'll ask you for your github credentials and verify them. It'll use them to push as well as to use the github REST API.

After that, some manual intervention is still required at the following points:
- After side-by-side tests (human has to squint at results)
- After PR is created (human still has to merge it)


s-deitz:
This PR is almost a year old and it looks like we lost track of it. Is this still valid? If so, please reply and somebody will review in the next couple of weeks.

tedsta:
If I had to guess the release process has probably changed and this is invalid by now.

PR #1375, opened 2017-05-18
Fix type of --beam_it_timeout flag (was string, needs to be integer)

tedsta:


asaksena:
LGTM

dlott:
Add a test?

tedsta:
@dlott Looks like this test would've caught it if gflags validated default parameters: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/tests/config_test.py

PR #1384, opened 2017-05-24
Fixed non-managed memcached ycsb benchmark

tedsta:
There's no such flag `FLAGS.ycsb_server_vms`

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/memcached_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/memcached_ycsb_benchmark.py
@@ -69,6 +69,9 @@ memcached_ycsb:
     installation. Specify the number of YCSB client VMs with
     --ycsb_client_vms and the number of YCSB server VMS with
     --num_vms.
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_benchmarks/memcached_ycsb_benchmark.py
```diff
@@ -69,6 +69,9 @@
     installation. Specify the number of YCSB client VMs with
     --ycsb_client_vms and the number of YCSB server VMS with
     --num_vms.
+  flags:
+    ycsb_client_vms: 1
```
asaksena: Is this flag specific to memcached ? or a generic ycsb flag ?

tedsta: It's a [YCSB flag](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_packages/ycsb.py#L101)

PR #1385, opened 2017-05-24
Set default azure_lib_version to 1.0.3

tedsta:
`object_storage_service` benchmark was breaking with the latest version on pip. Something about TLS errors and wanting us to upgrade to python 3.4. This is a quick fix for now.


s-deitz:
This should be done already.

PR #1387, opened 2017-05-27
PKB to support Cloud Spanner.

haih-g:
Added a new benchmark cloud_spanner_ycsb, and a new provider gcp/gcp_spanner.py

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
@@ -42,16 +42,13 @@ cloud_spanner_ycsb:
       https://www.googleapis.com/auth/spanner.admin
       https://www.googleapis.com/auth/spanner.data"""
 
-# As of May 2017, cloud spanner is not included in the latest YCSB release
-# 0.12.0. You can build your own release binary and feed it in with
-# --cloud_spanner_custom_ycsb_release=<url>. You are welcome to use a pre-build
-# binary located at
-# https://storage.googleapis.com/cloud-spanner-ycsb-custom-release/
-# ycsb-cloudspanner-binding-0.13.0.tar.gz
+# As of May 2017, cloud Spanner is not included in the latest YCSB release
+# 0.12.0. The URL below points to a custom YCSB build for Cloud Spanner.
+# You can override it by flag --cloud_spanner_custom_ycsb_release.
 #
 # TODO: Remove this once cloud spanner is included in YCSB release.
-YCSB_BINDING_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases'
-                        '/download/0.13.0/'
+YCSB_BINDING_TAR_URL = ('https://storage.googleapis.com/'
+                        'cloud-spanner-ycsb-custom-release/'
                         'ycsb-cloudspanner-binding-0.13.0.tar.gz')
 REQUIRED_SCOPES = (
     'https://www.googleapis.com/auth/spanner.admin',
@@ -67,9 +64,15 @@ flags.DEFINE_integer('cloud_spanner_zeropadding',
 flags.DEFINE_string('cloud_spanner_host',
                     None,
                     'The Cloud Spanner host used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_project',
+                    None,
+                    'The gcloud project to use in the YCSB benchmark. Leave '
+                    'it unset to use the default gcloud project.')
 flags.DEFINE_string('cloud_spanner_instance',
                     'ycsb-' + getpass.getuser(),
-                    'The Cloud Spanner instance used in the YCSB benchmark.')
+                    'The Cloud Spanner instance used in the YCSB benchmark. '
+                    'Use distinct instances if you want to run multiple '
+                    'benchmarks in parallel.')
 flags.DEFINE_string('cloud_spanner_database',
                     'ycsb',
                     'The Cloud Spanner database used in the YCSB benchmark.')
@@ -87,7 +90,7 @@ flags.DEFINE_integer('cloud_spanner_batchinserts',
 flags.DEFINE_string('cloud_spanner_custom_ycsb_release',
                     None,
                     'If provided, the URL of a custom YCSB release')
-flags.DEFINE_string('cloud_spanner_desp',
+flags.DEFINE_string('cloud_spanner_description',
                     'YCSB',
                     'The description of the Cloud Spanner instance.')
 flags.DEFINE_integer('cloud_spanner_nodes',
@@ -129,10 +132,14 @@ def Prepare(benchmark_spec):
   benchmark_spec.always_call_cleanup = True
 
   benchmark_spec.spanner_instance = gcp_spanner.GcpSpannerInstance(
-      FLAGS.cloud_spanner_instance, FLAGS.cloud_spanner_desp,
-      FLAGS.cloud_spanner_nodes, FLAGS.cloud_spanner_config,
-      FLAGS.cloud_spanner_database, FLAGS.cloud_spanner_ddl)
-  benchmark_spec.spanner_instance.Delete()
+      FLAGS.cloud_spanner_project, FLAGS.cloud_spanner_instance,
+      FLAGS.cloud_spanner_description, FLAGS.cloud_spanner_nodes,
+      FLAGS.cloud_spanner_config, FLAGS.cloud_spanner_database,
+      FLAGS.cloud_spanner_ddl)
+  if benchmark_spec.spanner_instance._Exists():
+    logging.warning('Cloud Spanner instance %s exists, delete it first.' %
+                    FLAGS.cloud_spanner_instance)
+    benchmark_spec.spanner_instance.Delete()
   benchmark_spec.spanner_instance.Create()
   if not benchmark_spec.spanner_instance._Exists():
     logging.warning('Failed to create Cloud Spanner instance and database.')
@@ -147,7 +154,7 @@ def Prepare(benchmark_spec):
   else:
     ycsb.YCSB_TAR_URL = YCSB_BINDING_TAR_URL
 
-  logging.warning('YCSB tar url: ' + ycsb.YCSB_TAR_URL)
+  logging.info('YCSB tar url: ' + ycsb.YCSB_TAR_URL)
 
   vms = benchmark_spec.vms
 
@@ -187,6 +194,8 @@ def Run(benchmark_spec):
     load_kwargs['threads'] = 1
   samples = list(benchmark_spec.executor.LoadAndRun(
       vms, load_kwargs=load_kwargs, run_kwargs=run_kwargs))
+  # TODO: Figure out a common set of properties and update the benchmark
+  # meta data here.
   return samples
 
 
--- a/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
@@ -41,7 +41,21 @@ cloud_spanner_ycsb:
     gcloud_scopes: >
       https://www.googleapis.com/auth/spanner.admin
       https://www.googleapis.com/auth/spanner.data"""
-
+BENCHMARK_SCHEMA = """
+CREATE TABLE usertable (
+  id     STRING(MAX),
+  field0 STRING(MAX),
+  field1 STRING(MAX),
+  field2 STRING(MAX),
+  field3 STRING(MAX),
+  field4 STRING(MAX),
+  field5 STRING(MAX),
+  field6 STRING(MAX),
+  field7 STRING(MAX),
+  field8 STRING(MAX),
+  field9 STRING(MAX),
+) PRIMARY KEY(id)
+"""
 # As of May 2017, cloud Spanner is not included in the latest YCSB release
 # 0.12.0. The URL below points to a custom YCSB build for Cloud Spanner.
 # You can override it by flag --cloud_spanner_custom_ycsb_release.
@@ -55,58 +69,54 @@ REQUIRED_SCOPES = (
     'https://www.googleapis.com/auth/spanner.data')
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string('cloud_spanner_table',
-                    'usertable',
-                    'The table name used in the YCSB benchmark.')
-flags.DEFINE_integer('cloud_spanner_zeropadding',
-                     12,
-                     'The zero padding used in the YCSB benchmark.')
-flags.DEFINE_string('cloud_spanner_host',
+flags.DEFINE_integer('cloud_spanner_ycsb_batchinserts',
+                     1,
+                     'The Cloud Spanner batch inserts used in the YCSB '
+                     'benchmark.')
+flags.DEFINE_integer('cloud_spanner_ycsb_boundedstaleness',
+                     0,
+                     'The Cloud Spanner bounded staleness used in the YCSB '
+                     'benchmark.')
+flags.DEFINE_string('cloud_spanner_ycsb_config',
+                    'regional-us-central1',
+                    'The config for the Cloud Spanner instance.')
+flags.DEFINE_string('cloud_spanner_ycsb_custom_release',
                     None,
-                    'The Cloud Spanner host used in the YCSB benchmark.')
-flags.DEFINE_string('cloud_spanner_project',
+                    'If provided, the URL of a custom YCSB release')
+flags.DEFINE_string('cloud_spanner_ycsb_database',
+                    'ycsb',
+                    'The Cloud Spanner database used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_ycsb_ddl',
+                    BENCHMARK_SCHEMA,
+                    'The schema DDL for the Cloud Spanner database.')
+flags.DEFINE_string('cloud_spanner_ycsb_description',
+                    'YCSB',
+                    'The description of the Cloud Spanner instance.')
+flags.DEFINE_string('cloud_spanner_ycsb_end_point',
                     None,
-                    'The gcloud project to use in the YCSB benchmark. Leave '
-                    'it unset to use the default gcloud project.')
-flags.DEFINE_string('cloud_spanner_instance',
+                    'The Cloud Spanner end point used in the YCSB benchmark. '
+                    'Leave it unset to use the default end point.')
+flags.DEFINE_string('cloud_spanner_ycsb_instance',
                     'ycsb-' + getpass.getuser(),
                     'The Cloud Spanner instance used in the YCSB benchmark. '
                     'Use distinct instances if you want to run multiple '
                     'benchmarks in parallel.')
-flags.DEFINE_string('cloud_spanner_database',
-                    'ycsb',
-                    'The Cloud Spanner database used in the YCSB benchmark.')
-flags.DEFINE_enum('cloud_spanner_readmode',
-                  'query', ['query', 'read'],
-                  'The Cloud Spanner read mode used in the YCSB benchmark.')
-flags.DEFINE_integer('cloud_spanner_boundedstaleness',
-                     0,
-                     'The Cloud Spanner bounded staleness used in the YCSB '
-                     'benchmark.')
-flags.DEFINE_integer('cloud_spanner_batchinserts',
-                     1,
-                     'The Cloud Spanner batch inserts used in the YCSB '
-                     'benchmark.')
-flags.DEFINE_string('cloud_spanner_custom_ycsb_release',
-                    None,
-                    'If provided, the URL of a custom YCSB release')
-flags.DEFINE_string('cloud_spanner_description',
-                    'YCSB',
-                    'The description of the Cloud Spanner instance.')
-flags.DEFINE_integer('cloud_spanner_nodes',
+flags.DEFINE_integer('cloud_spanner_ycsb_nodes',
                      1,
                      'The number of nodes for the Cloud Spanner instance.')
-flags.DEFINE_string('cloud_spanner_config',
-                    'regional-us-central1',
-                    'The config for the Cloud Spanner instance.')
-flags.DEFINE_string('cloud_spanner_ddl',
-                    """CREATE TABLE usertable (
-                    id STRING(MAX), field0 STRING(MAX), field1 STRING(MAX),
-                    field2 STRING(MAX), field3 STRING(MAX), field4 STRING(MAX),
-                    field5 STRING(MAX), field6 STRING(MAX), field7 STRING(MAX),
-                    field8 STRING(MAX), field9 STRING(MAX),
-                    ) PRIMARY KEY(id)""",
-                    'The schema DDL for the Cloud Spanner database.')
+flags.DEFINE_string('cloud_spanner_ycsb_project',
+                    None,
+                    'The gcloud project to use in the YCSB benchmark. Leave '
+                    'it unset to use the default gcloud project.')
+flags.DEFINE_enum('cloud_spanner_ycsb_readmode',
+                  'query', ['query', 'read'],
+                  'The Cloud Spanner read mode used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_ycsb_table',
+                    'usertable',
+                    'The table name used in the YCSB benchmark.')
+flags.DEFINE_integer('cloud_spanner_ycsb_zeropadding',
+                     12,
+                     'The zero padding used in the YCSB benchmark.')
 
 
 def GetConfig(user_config):
@@ -132,13 +142,17 @@ def Prepare(benchmark_spec):
   benchmark_spec.always_call_cleanup = True
 
   benchmark_spec.spanner_instance = gcp_spanner.GcpSpannerInstance(
-      FLAGS.cloud_spanner_project, FLAGS.cloud_spanner_instance,
-      FLAGS.cloud_spanner_description, FLAGS.cloud_spanner_nodes,
-      FLAGS.cloud_spanner_config, FLAGS.cloud_spanner_database,
-      FLAGS.cloud_spanner_ddl, FLAGS.cloud_spanner_host)
-  if benchmark_spec.spanner_instance._Exists():
+      name=FLAGS.cloud_spanner_ycsb_instance,
+      description=FLAGS.cloud_spanner_ycsb_description,
+      nodes=FLAGS.cloud_spanner_ycsb_nodes,
+      database=FLAGS.cloud_spanner_ycsb_database,
+      ddl=FLAGS.cloud_spanner_ycsb_ddl,
+      project=FLAGS.cloud_spanner_ycsb_project,
+      config=FLAGS.cloud_spanner_ycsb_config,
+      end_point=FLAGS.cloud_spanner_ycsb_end_point)
+  if benchmark_spec.spanner_instance._Exists(instance_only=True):
     logging.warning('Cloud Spanner instance %s exists, delete it first.' %
-                    FLAGS.cloud_spanner_instance)
+                    FLAGS.cloud_spanner_ycsb_instance)
     benchmark_spec.spanner_instance.Delete()
   benchmark_spec.spanner_instance.Create()
   if not benchmark_spec.spanner_instance._Exists():
@@ -149,8 +163,8 @@ def Prepare(benchmark_spec):
 
   # TODO: figure out a less hacky way to override.
   # Override so that we only need to download the required binding.
-  if FLAGS.cloud_spanner_custom_ycsb_release:
-    ycsb.YCSB_TAR_URL = FLAGS.cloud_spanner_custom_ycsb_release
+  if FLAGS.cloud_spanner_ycsb_custom_release:
+    ycsb.YCSB_TAR_URL = FLAGS.cloud_spanner_ycsb_custom_release
   else:
     ycsb.YCSB_TAR_URL = YCSB_BINDING_TAR_URL
 
@@ -178,16 +192,17 @@ def Run(benchmark_spec):
   """
   vms = benchmark_spec.vms
   run_kwargs = {
-      'table': FLAGS.cloud_spanner_table,
-      'zeropadding': FLAGS.cloud_spanner_zeropadding,
-      'cloudspanner.instance': FLAGS.cloud_spanner_instance,
-      'cloudspanner.database': FLAGS.cloud_spanner_database,
-      'cloudspanner.readmode': FLAGS.cloud_spanner_readmode,
-      'cloudspanner.boundedstaleness': FLAGS.cloud_spanner_boundedstaleness,
-      'cloudspanner.batchinserts': FLAGS.cloud_spanner_batchinserts,
+      'table': FLAGS.cloud_spanner_ycsb_table,
+      'zeropadding': FLAGS.cloud_spanner_ycsb_zeropadding,
+      'cloudspanner.instance': FLAGS.cloud_spanner_ycsb_instance,
+      'cloudspanner.database': FLAGS.cloud_spanner_ycsb_database,
+      'cloudspanner.readmode': FLAGS.cloud_spanner_ycsb_readmode,
+      'cloudspanner.boundedstaleness':
+          FLAGS.cloud_spanner_ycsb_boundedstaleness,
+      'cloudspanner.batchinserts': FLAGS.cloud_spanner_ycsb_batchinserts,
   }
-  if FLAGS.cloud_spanner_host:
-    run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_host
+  if FLAGS.cloud_spanner_ycsb_end_point:
+    run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_ycsb_end_point
 
   load_kwargs = run_kwargs.copy()
   if not FLAGS['ycsb_preload_threads'].present:
@@ -212,7 +227,7 @@ def Cleanup(benchmark_spec):
 def _Install(vm):
   vm.Install('ycsb')
 
-  if FLAGS.cloud_spanner_host is not None:
+  if FLAGS.cloud_spanner_ycsb_end_point is not None:
     _, _ = vm.RemoteCommand(
         'gcloud config set api_endpoint_overrides/spanner %s' %
-        FLAGS.cloud_spanner_host)
+        FLAGS.cloud_spanner_ycsb_end_point)
--- a/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
@@ -14,12 +14,8 @@
 
 """Run YCSB benchmark against Google Cloud Spanner
 
-Before running this benchmark, you have to download your JSON
-service account credential file to local machine, and pass the path
-via 'cloud_spanner_credential_file' parameters to PKB.
-
 By default, this benchmark provision 1 single-CPU VM and spawn 1 thread
-to test Spanner.
+to test Spanner. Configure the number of VMs via --ycsb_client_vms.
 """
 
 import getpass
@@ -36,7 +32,7 @@ BENCHMARK_CONFIG = """
 cloud_spanner_ycsb:
   description: >
       Run YCSB against Google Cloud Spanner.
-      Configure the number of VMs via --num-vms.
+      Configure the number of VMs via --ycsb_client_vms.
   vm_groups:
     default:
       vm_spec: *default_single_core
@@ -52,6 +48,8 @@ cloud_spanner_ycsb:
 # binary located at
 # https://storage.googleapis.com/cloud-spanner-ycsb-custom-release/
 # ycsb-cloudspanner-binding-0.13.0.tar.gz
+#
+# TODO: Remove this once cloud spanner is included in YCSB release.
 YCSB_BINDING_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases'
                         '/download/0.13.0/'
                         'ycsb-cloudspanner-binding-0.13.0.tar.gz')
@@ -60,10 +58,10 @@ REQUIRED_SCOPES = (
     'https://www.googleapis.com/auth/spanner.data')
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string('table',
+flags.DEFINE_string('cloud_spanner_table',
                     'usertable',
                     'The table name used in the YCSB benchmark.')
-flags.DEFINE_integer('zeropadding',
+flags.DEFINE_integer('cloud_spanner_zeropadding',
                      12,
                      'The zero padding used in the YCSB benchmark.')
 flags.DEFINE_string('cloud_spanner_host',
@@ -75,9 +73,9 @@ flags.DEFINE_string('cloud_spanner_instance',
 flags.DEFINE_string('cloud_spanner_database',
                     'ycsb',
                     'The Cloud Spanner database used in the YCSB benchmark.')
-flags.DEFINE_string('cloud_spanner_readmode',
-                    'query',
-                    'The Cloud Spanner read mode used in the YCSB benchmark.')
+flags.DEFINE_enum('cloud_spanner_readmode',
+                  'query', ['query', 'read'],
+                  'The Cloud Spanner read mode used in the YCSB benchmark.')
 flags.DEFINE_integer('cloud_spanner_boundedstaleness',
                      0,
                      'The Cloud Spanner bounded staleness used in the YCSB '
@@ -122,6 +120,12 @@ def CheckPrerequisites(benchmark_config):
 
 
 def Prepare(benchmark_spec):
+  """Prepare the virtual machines to run cloud spanner benchmarks.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
   benchmark_spec.always_call_cleanup = True
 
   benchmark_spec.spanner_instance = gcp_spanner.GcpSpannerInstance(
@@ -156,10 +160,19 @@ def Prepare(benchmark_spec):
 
 
 def Run(benchmark_spec):
+  """Spawn YCSB and gather the results.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+
+  Returns:
+    A list of sample.Sample instances.
+  """
   vms = benchmark_spec.vms
   run_kwargs = {
-      'table': FLAGS.table,
-      'zeropadding': FLAGS.zeropadding,
+      'table': FLAGS.cloud_spanner_table,
+      'zeropadding': FLAGS.cloud_spanner_zeropadding,
       'cloudspanner.instance': FLAGS.cloud_spanner_instance,
       'cloudspanner.database': FLAGS.cloud_spanner_database,
       'cloudspanner.readmode': FLAGS.cloud_spanner_readmode,
@@ -170,14 +183,20 @@ def Run(benchmark_spec):
     run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_host
 
   load_kwargs = run_kwargs.copy()
-  if FLAGS['ycsb_preload_threads'].present:
-    load_kwargs['threads'] = FLAGS.ycsb_preload_threads
+  if not FLAGS['ycsb_preload_threads'].present:
+    load_kwargs['threads'] = 1
   samples = list(benchmark_spec.executor.LoadAndRun(
       vms, load_kwargs=load_kwargs, run_kwargs=run_kwargs))
   return samples
 
 
 def Cleanup(benchmark_spec):
+  """Cleanup.
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
   benchmark_spec.spanner_instance.Delete()
 
 
--- a/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
@@ -205,8 +205,6 @@ def Run(benchmark_spec):
     run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_ycsb_end_point
 
   load_kwargs = run_kwargs.copy()
-  if not FLAGS['ycsb_preload_threads'].present:
-    load_kwargs['threads'] = 1
   samples = list(benchmark_spec.executor.LoadAndRun(
       vms, load_kwargs=load_kwargs, run_kwargs=run_kwargs))
   # TODO: Figure out a common set of properties and update the benchmark
--- a/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
@@ -135,7 +135,7 @@ def Prepare(benchmark_spec):
       FLAGS.cloud_spanner_project, FLAGS.cloud_spanner_instance,
       FLAGS.cloud_spanner_description, FLAGS.cloud_spanner_nodes,
       FLAGS.cloud_spanner_config, FLAGS.cloud_spanner_database,
-      FLAGS.cloud_spanner_ddl)
+      FLAGS.cloud_spanner_ddl, FLAGS.cloud_spanner_host)
   if benchmark_spec.spanner_instance._Exists():
     logging.warning('Cloud Spanner instance %s exists, delete it first.' %
                     FLAGS.cloud_spanner_instance)
@@ -211,3 +211,8 @@ def Cleanup(benchmark_spec):
 
 def _Install(vm):
   vm.Install('ycsb')
+
+  if FLAGS.cloud_spanner_host is not None:
+    _, _ = vm.RemoteCommand(
+        'gcloud config set api_endpoint_overrides/spanner %s' %
+        FLAGS.cloud_spanner_host)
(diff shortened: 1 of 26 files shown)


perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,126 @@
...
+    super(GcpSpannerInstance, self).__init__()
+    self._nodes = nodes
+    self._name = name
+    self._desp = desp
+    self._config = config
+    self._database = database
+    self._ddl = ddl
+
+    # Cloud Spanner is not using the following common flags.
+    self.project = None
```
yuyantingzero: it is project specific

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,126 @@
...
+      return False
+    result = json.loads(stdout)
+    instances = {instance['name'].split('/')[-1] for instance in result}
+    if self._name not in instances:
+      return False
+
+    cmd = util.GcloudCommand(self, 'spanner', 'databases', 'list')
+    cmd.flags['filter'] = self._database
+    cmd.flags['instance'] = self._name
+    stdout, stderr, retcode = cmd.Issue(suppress_warning=True)
```
yuyantingzero: why?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,126 @@
...
+    if retcode != 0:
```
yuyantingzero: if you didn't use suppress_warning, it is already logged

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,126 @@
...
+    cmd = util.GcloudCommand(self, 'spanner', 'instances', 'list')
+    cmd.flags['filter'] = self._name
+    stdout, stderr, retcode = cmd.Issue(suppress_warning=True)
+    if retcode != 0:
+      # This is not ideal, as we're returning false not because we know
+      # the instance isn't there, but because we can't figure out whether
+      # it is there.  This behavior is consistent without other
+      # _Exists methods.
+      logging.error('Unable to list GCP Spanner instances. Return code %s '
+                    'STDOUT: %s\nSTDERR: %s', retcode, stdout, stderr)
```
yuyantingzero: same

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,126 @@
...
+    if retcode != 0:
+      logging.error('Update GCP Spanner database schema failed. Return code %s '
+                    'STDOUT: %s\nSTDERR: %s', retcode, stdout, stderr)
+
+  def _Delete(self):
+    """Deletes the instance."""
+    cmd = util.GcloudCommand(self, 'spanner', 'instances', 'delete',
+                             self._name)
+    stdout, stderr, retcode = cmd.Issue()
+    if retcode != 0:
```
yuyantingzero: should already logged

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+      'cloudspanner.database': FLAGS.cloud_spanner_database,
+      'cloudspanner.readmode': FLAGS.cloud_spanner_readmode,
+      'cloudspanner.boundedstaleness': FLAGS.cloud_spanner_boundedstaleness,
+      'cloudspanner.batchinserts': FLAGS.cloud_spanner_batchinserts,
+  }
+  if FLAGS.cloud_spanner_host:
+    run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_host
+
+  load_kwargs = run_kwargs.copy()
+  if FLAGS['ycsb_preload_threads'].present:
```
yuyantingzero: already set in ycsb package

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+    load_kwargs['threads'] = FLAGS.ycsb_preload_threads
+  samples = list(benchmark_spec.executor.LoadAndRun(
+      vms, load_kwargs=load_kwargs, run_kwargs=run_kwargs))
+  return samples
+
+
+def Cleanup(benchmark_spec):
```
yuyantingzero: add docstring

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+  # Install required packages and copy credential files
+  vm_util.RunThreaded(_Install, vms)
+
+  # Restore YCSB_TAR_URL
+  ycsb.YCSB_TAR_URL = default_ycsb_tar_url
+  benchmark_spec.executor = ycsb.YCSBExecutor('cloudspanner')
+
+
+def Run(benchmark_spec):
```
yuyantingzero: add docstring

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+  return config
+
+
+def CheckPrerequisites(benchmark_config):
+  for scope in REQUIRED_SCOPES:
+    if scope not in FLAGS.gcloud_scopes:
+      raise ValueError('Scope {0} required.'.format(scope))
+
+
+def Prepare(benchmark_spec):
```
yuyantingzero: add docstring

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def GetConfig(user_config):
+  config = configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+  if FLAGS['ycsb_client_vms'].present:
+    config['vm_groups']['default']['vm_count'] = FLAGS.ycsb_client_vms
+  return config
+
+
+def CheckPrerequisites(benchmark_config):
```
yuyantingzero: check credential?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+                    'regional-us-central1',
+                    'The config for the Cloud Spanner instance.')
+flags.DEFINE_string('cloud_spanner_ddl',
+                    """CREATE TABLE usertable (
+                    id STRING(MAX), field0 STRING(MAX), field1 STRING(MAX),
+                    field2 STRING(MAX), field3 STRING(MAX), field4 STRING(MAX),
+                    field5 STRING(MAX), field6 STRING(MAX), field7 STRING(MAX),
+                    field8 STRING(MAX), field9 STRING(MAX),
+                    ) PRIMARY KEY(id)""",
+                    'The schema DDL for the Cloud Spanner database.')
```
yuyantingzero: seems like all cloud_spanner_.. flags belongs to gcp_spanner

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+      https://www.googleapis.com/auth/spanner.admin
+      https://www.googleapis.com/auth/spanner.data"""
+
+# As of May 2017, cloud spanner is not included in the latest YCSB release
+# 0.12.0. You can build your own release binary and feed it in with
+# --cloud_spanner_custom_ycsb_release=<url>. You are welcome to use a pre-build
+# binary located at
+# https://storage.googleapis.com/cloud-spanner-ycsb-custom-release/
+# ycsb-cloudspanner-binding-0.13.0.tar.gz
+YCSB_BINDING_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases'
```
yuyantingzero: add todo to install ycsb instead of self build ones.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Run YCSB benchmark against Google Cloud Spanner
+
+Before running this benchmark, you have to download your JSON
+service account credential file to local machine, and pass the path
+via 'cloud_spanner_credential_file' parameters to PKB.
```
yuyantingzero: how

haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+# ycsb-cloudspanner-binding-0.13.0.tar.gz
+YCSB_BINDING_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases'
+                        '/download/0.13.0/'
+                        'ycsb-cloudspanner-binding-0.13.0.tar.gz')
+REQUIRED_SCOPES = (
+    'https://www.googleapis.com/auth/spanner.admin',
+    'https://www.googleapis.com/auth/spanner.data')
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('table',
```
haih-g: This is spanner specific. Some storage, such as bigtable, are just key-value-pair storage and don't have a concept of table. Will rename this to cloud_spanner_table.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+                    'usertable',
+                    'The table name used in the YCSB benchmark.')
+flags.DEFINE_integer('zeropadding',
```
haih-g: Will rename this to cloud_spanner_zeropadding

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+flags.DEFINE_string('cloud_spanner_host',
+                    None,
+                    'The Cloud Spanner host used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_instance',
+                    'ycsb-' + getpass.getuser(),
+                    'The Cloud Spanner instance used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_database',
+                    'ycsb',
+                    'The Cloud Spanner database used in the YCSB benchmark.')
+flags.DEFINE_string('cloud_spanner_readmode',
```
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+flags.DEFINE_string('cloud_spanner_desp',
+                    'YCSB',
+                    'The description of the Cloud Spanner instance.')
+flags.DEFINE_integer('cloud_spanner_nodes',
+                     1,
+                     'The number of nodes for the Cloud Spanner instance.')
+flags.DEFINE_string('cloud_spanner_config',
+                    'regional-us-central1',
+                    'The config for the Cloud Spanner instance.')
+flags.DEFINE_string('cloud_spanner_ddl',
```
haih-g: It seems the benchmark.py should be self-contained. The BENCHMARK_CONFIG is also in-lined as a string.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: This follows the same pattern that all the flags are include in cloud_bigtable_ycsb_benchmark.py and gcp_bigtable contains no flags.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: cloud spanner use gcp default credential. It doesn't use any special key file / credential files.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def CheckPrerequisites(benchmark_config):
+  for scope in REQUIRED_SCOPES:
+    if scope not in FLAGS.gcloud_scopes:
+      raise ValueError('Scope {0} required.'.format(scope))
+
+
+def Prepare(benchmark_spec):
+  benchmark_spec.always_call_cleanup = True
```
haih-g: This is needed to delete the cloud spanner instance after the test is done. Otherwise I'll need to provide a way to bypass instance creation in the Prepare() method.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+
+# As of May 2017, cloud spanner is not included in the latest YCSB release
+# 0.12.0. You can build your own release binary and feed it in with
+# --cloud_spanner_custom_ycsb_release=<url>. You are welcome to use a pre-build
+# binary located at
+# https://storage.googleapis.com/cloud-spanner-ycsb-custom-release/
+# ycsb-cloudspanner-binding-0.13.0.tar.gz
+#
+# TODO: Remove this once cloud spanner is included in YCSB release.
+YCSB_BINDING_TAR_URL = ('https://github.com/brianfrankcooper/YCSB/releases'
```
asaksena: Sorry, but this looks to be a non existent URL, which is going to fail 100% of the time.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: There is a data folder which might be the more appropriate location for the ddl. Please see example https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/data/cassandra/create-ycsb-table.cql.j2

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+                     'The Cloud Spanner bounded staleness used in the YCSB '
+                     'benchmark.')
+flags.DEFINE_integer('cloud_spanner_batchinserts',
+                     1,
+                     'The Cloud Spanner batch inserts used in the YCSB '
+                     'benchmark.')
+flags.DEFINE_string('cloud_spanner_custom_ycsb_release',
+                    None,
+                    'If provided, the URL of a custom YCSB release')
+flags.DEFINE_string('cloud_spanner_desp',
```
asaksena: Sorry but is 'desp' an acronym for something ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -170,14 +183,20 @@ def Run(benchmark_spec):
     run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_host
 
   load_kwargs = run_kwargs.copy()
-  if FLAGS['ycsb_preload_threads'].present:
-    load_kwargs['threads'] = FLAGS.ycsb_preload_threads
+  if not FLAGS['ycsb_preload_threads'].present:
```
yuyantingzero: this flag already set in ycsb package.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -170,14 +183,20 @@ def Run(benchmark_spec):
...
+    load_kwargs['threads'] = 1
```
yuyantingzero: why hardcode to 1?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+        required to run the benchmark.
+  """
+  benchmark_spec.always_call_cleanup = True
+
+  benchmark_spec.spanner_instance = gcp_spanner.GcpSpannerInstance(
+      FLAGS.cloud_spanner_instance, FLAGS.cloud_spanner_desp,
+      FLAGS.cloud_spanner_nodes, FLAGS.cloud_spanner_config,
+      FLAGS.cloud_spanner_database, FLAGS.cloud_spanner_ddl)
+  benchmark_spec.spanner_instance.Delete()
+  benchmark_spec.spanner_instance.Create()
```
asaksena: Are the 2 line above Delete() and then Create() to make sure that any previous instances with the same name are cleaned up ? What is the expected behavior if 2 users of the same project are trying to run the same benchmark with default instance name ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+  benchmark_spec.always_call_cleanup = True
+
+  benchmark_spec.spanner_instance = gcp_spanner.GcpSpannerInstance(
+      FLAGS.cloud_spanner_instance, FLAGS.cloud_spanner_desp,
+      FLAGS.cloud_spanner_nodes, FLAGS.cloud_spanner_config,
+      FLAGS.cloud_spanner_database, FLAGS.cloud_spanner_ddl)
```
asaksena: Sorry, but is there a reason why all these flags are local to the benchmark ? If I wanted to create another benchmark using the cloud spanner resource, how do I access these flags ?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -45,8 +45,8 @@ def __init__(self, name, desp, nodes, config, database, ddl):
     self._database = database
     self._ddl = ddl
 
-    # Cloud Spanner is not using the following common flags.
-    self.project = None
+    # Cloud Spanner does not explicitly set the following common flags.
+    self.project = util.GetDefaultProject()
```
yuyantingzero: https://cloud.google.com/spanner/docs/gcloud-spanner#set_a_default_project

How do you specify a project different from default one?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
yuyantingzero: dont think it will pollute the output

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+  benchmark_spec.spanner_instance.Delete()
+  benchmark_spec.spanner_instance.Create()
+  if not benchmark_spec.spanner_instance._Exists():
+    logging.warning('Failed to create Cloud Spanner instance and database.')
+    benchmark_spec.spanner_instance.Delete()
+
+  default_ycsb_tar_url = ycsb.YCSB_TAR_URL
```
asaksena: Can you please explain the reasoning for capturing the default value, not using it and the restoring it ?
Are you expecting the ycsb module's TAR_URL property to be used somewhere else ? How are multiple processes going to co-ordinate the value ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+
+  # TODO: figure out a less hacky way to override.
+  # Override so that we only need to download the required binding.
+  if FLAGS.cloud_spanner_custom_ycsb_release:
+    ycsb.YCSB_TAR_URL = FLAGS.cloud_spanner_custom_ycsb_release
+  else:
+    ycsb.YCSB_TAR_URL = YCSB_BINDING_TAR_URL
```
asaksena: Non existent URL, see comment above.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -55,50 +55,51 @@ def _Create(self):
...
     if retcode != 0:
-      logging.error('Delete GCP Spanner instances failed. Return code %s '
-                    'STDOUT: %s\nSTDERR: %s', retcode, stdout, stderr)
+      logging.error('Delete GCP Spanner instance failed.')
+    else:
+      logging.info('Deleted GCP Spanner instance.')
 
   def _Exists(self):
     """Returns true if the instance and the database exists."""
     cmd = util.GcloudCommand(self, 'spanner', 'instances', 'list')
```
yuyantingzero: describe doesnt work?

https://cloud.google.com/spanner/docs/gcloud-spanner#list_a_specific_instance

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -109,13 +110,14 @@ def _Exists(self):
     cmd = util.GcloudCommand(self, 'spanner', 'databases', 'list')
     cmd.flags['filter'] = self._database
     cmd.flags['instance'] = self._name
+    # Suppress warning to prevent it from pollute stdout.
```
yuyantingzero: it will not pollute stdout.
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L258

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -109,13 +110,14 @@ def _Exists(self):
     cmd = util.GcloudCommand(self, 'spanner', 'databases', 'list')
```
yuyantingzero: what are you trying to do here? retry if cannot list?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+
+  logging.warning('YCSB tar url: ' + ycsb.YCSB_TAR_URL)
```
asaksena: Is this warning or info ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,204 @@
...
+      'cloudspanner.boundedstaleness': FLAGS.cloud_spanner_boundedstaleness,
+      'cloudspanner.batchinserts': FLAGS.cloud_spanner_batchinserts,
+  }
+  if FLAGS.cloud_spanner_host:
+    run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_host
+
+  load_kwargs = run_kwargs.copy()
+  if not FLAGS['ycsb_preload_threads'].present:
+    load_kwargs['threads'] = 1
+  samples = list(benchmark_spec.executor.LoadAndRun(
```
asaksena: Do you need to update the metadata with any run properties ?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
yuyantingzero: nvm, didnt notice it is database.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,128 @@
...
+                             self._name)
+    _, _, retcode = cmd.Issue()
+    if retcode != 0:
+      logging.error('Delete GCP Spanner instance failed.')
+    else:
+      logging.info('Deleted GCP Spanner instance.')
+
+  def _Exists(self):
+    """Returns true if the instance and the database exists."""
+    cmd = util.GcloudCommand(self, 'spanner', 'instances', 'list')
```
asaksena: $ gcloud spanner instances describe [MY_INSTANCE_ID]

Is there a reason for not using the specific api for the instance ?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,128 @@
...
+    cmd.flags['filter'] = self._name
+    # Suppress warning to prevent it from pollute stdout.
```
asaksena: Sorry I don't understand the pollution issue ? Can you please elaborate ?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,128 @@
...
+    if self._name not in instances:
+      return False
+
+    cmd = util.GcloudCommand(self, 'spanner', 'databases', 'list')
+    cmd.flags['filter'] = self._database
+    cmd.flags['instance'] = self._name
+    # Suppress warning to prevent it from pollute stdout.
+    stdout, stderr, retcode = cmd.Issue(suppress_warning=True)
+    if retcode != 0:
+      # This is not ideal, as we're returning false not because we know
```
asaksena: Is the API response non deterministic ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: Referencing a non existent tar

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: Added a check to see if the instance exists and log a warning if so.

By default cloud_spanner_instance = ycsb-$USER. So it already isolates among users.

 Added a doc string in FLAGS.cloud_spanner_instance so if one user wants to run multiple benchmarks in parallel, he/she understands what to do.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: This was discussed (with yuyantingzero) in my previous comments. This follows the pattern of gcp_bigtable.py and cloud_bigtable_ycsb_benchmark.py. 

And I think it's a bad design if gcp_spanner.py owns all the flags. gcp_spanner.py is a class library (a lower layer) and cloud-spanner_ycsb_benchmark.py is the driver (the top layer). If gcp_spanner.py owns all the flags, it should use the flags to set up the cloud spanner instance, and shouldn't expose them as parameters of the class constructor. Then cloud_spanner_ycsb_benchmark.py has no way to override it at the driver layer, which is the the top layer. Maybe we can still do it, but then we need to implement extra logic in gcp_spanner.py to override the flags, which complicates the class and reduces readability.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: This follows the pattern in cloud_datastore_ycsb_benchmark.py. I'm not sure why but I think there is a reason. 

Why multiple processes needs to co-ordinate the value? If you are talking about the VMs, they should all use the same value as defined here. If you are talking about different runs of the benchmarks, I don't think there is need to co-ordinate it.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: This follows the pattern in cloud_bigtable_ycsb_benchmark.py, and I think there is a reason for that.

If I have to guess, I think this was not set at some previous version of YCSB. If so, I think we should keep it here, because user can use any YCSB version in the test.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: Again, this follows the pattern in cloud_bigtable_ycsb_benchmark.py. If I have to guess the reason, I think we only want to have one thread to load the database, otherwise there will be conflict of inserting the same key from different threads.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: Added one more flag cloud_spanner_project, which is default to None, and in which case, it will use the default project.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: Change to describe, and simplified the logic.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: Yes, but not in this pull request. We are going to determine a common set of properties shared by cloud spanner, bigtable, datastore, and then we'll update all the benchmarks.

Added a TODO

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
haih-g: DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
yuyantingzero: yeah, we discussed, but i still dont agree. any flag value referenced in gcp_spanner.py should be independent of this benchmark

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I am sorry but the pattern in the datastore benchmark you referenced predates me. That being said, I I have not seen any guidance to indicate that it should be used as a standard template. My 2 cents, unless we can come with a good reason for following a pattern that does not seem right, we should not repeat it, as it adds unnecessary complexity to the codebase. Please let me know if you disagree.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I do not understand your response.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I do not understand your response.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I am sorry, but you still did not answer my question:
"How do 2 benchmarks both of which use the cloud spanner resource access the configuration flags" ? Are you recommending importing one benchmark into another benchmark for the sake of accessing flags ?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I do not understand your response.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: I do not understand your response.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
asaksena: What is the sample your benchmark publishes ?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
asaksena: I do not understand your response.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
asaksena: I do not understand your response.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
asaksena: I do not understand your response.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: I'm not sure but I think this is for running multiple benchmarks in sequence. PKB supports benchmark sets that includes multiple benchmarks. If a previous benchmark changes the default TAR_URL without restoring it, the next benchmark may end up with a wrong URL.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE means I've changed the code following the suggestion. You can see it in the code I've just committed.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: The regular YCSB ones. latency (p50, p90, etc.) and throughput.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: Sorry for the late reply. This comment was hidden as 'outdated' on the github webpage.

Re: How do 2 benchmarks both of which use the cloud spanner resource access the configuration flags

They should each have their own set of flags. 

Most of these flags are logically isolated between benchmarks, because their default values in YCSB benchmark are different from the default value in other benchmarks:

instance - in ycsb benchmark, it's default value is ycsb-$USER, while in another benchmark, e.g., tpcc, it should be tpcc-$USER

database - in ycsb benchmark, it's default value is ycsb, while in another benchmark, e.g., tpcc, it should be tpcc.

ddl - this ddl is only for ycsb benchmark, for other benchmark, e.g., tpcc, it should be a much complicated one and many others are yscb-specific.

As I went through the list of flags, only 3 of them *might* be moved to gcp_spanner.py:
cloud_spanner_project,
cloud_spanner_endpoint
cloud_spanner_config

But I still want to argue that moving them to gcp_spanner.py is likely a bad design. The benefit is that all of them *might* share the same default values among benchmarks. But moving them to gcp_spanner.py creates trouble when a benchmark wants to override those values. What if cloud spanner team creates a tuple of project / config / endpoint just for YCSB benchmarking? We would like to have cloud_spanner_ycsb_benchmark.py own those default values, instead of gcp_spanner.py. 

I'm making a change to rename all the flags to cloud_spanner_ycsb_xxx so it's clear that all those are ycsb-specific flags.

asaksena: I am sorry but renaming will just violate DRY in my opinion, since these flags are relevant when defining the desired spanner instance and should be independent of whether that instance is used in a ycsb benchmark or another benchmark like TPC-D, etc. Making spanner instance configuration specific to a test like ycsb means people who write a different benchmark against spanner, will now have to recreate the flags.

haih-g: IMHO, DRY is irrelevant here. gcp_spanner.py is a class library. It's sufficient for a class library to expose configurable parameters at class constructor level. At the class library level, we don't necessarily know what a user will do. A user of the class library may choose to use FLAGS, which happens to be the case of cloud_spanner_ycsb_benchmarks.py, or it may also choose to use a configuration file, or simply define them as constants should they be fixed for a specific benchmark. As said, these parameters may change from benchmarks to benchmarks. And it makes sense that each benchmark owns the default values for those parameters (again, benchmarks can own the default values as FLAG, configuration file, constants, etc). Otherwise, when a user runs a benchmark, he/she is now forced to always pass in FLAGS, which is not convenient.

haih-g:
Friendly ping. 

Just want to understand what's the status of this pull request. Is there still comment I need to address?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,233 @@
...
+      'cloudspanner.boundedstaleness':
+          FLAGS.cloud_spanner_ycsb_boundedstaleness,
+      'cloudspanner.batchinserts': FLAGS.cloud_spanner_ycsb_batchinserts,
+  }
+  if FLAGS.cloud_spanner_ycsb_end_point:
+    run_kwargs['cloudspanner.host'] = FLAGS.cloud_spanner_ycsb_end_point
+
+  load_kwargs = run_kwargs.copy()
+  if not FLAGS['ycsb_preload_threads'].present:
+    load_kwargs['threads'] = 1
```
yuyantingzero: not necessary

asaksena:
There are still unaddressed comments. Please clarify your responses so we can go ahead

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
```diff
@@ -0,0 +1,233 @@
...
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+
+  Returns:
+    A list of sample.Sample instances.
+  """
+  vms = benchmark_spec.vms
+  run_kwargs = {
+      'table': FLAGS.cloud_spanner_ycsb_table,
```
yuyantingzero: as mentioned in previous comment. this flag should belong to ycsb.
Otherwise, are you going to make a datastore_ycsb_table, bigtable_ycsb_table, ...?

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
yuyantingzero: If you *really* believe there is a reason, be specific

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import resource
+from perfkitbenchmarker.providers.gcp import util
+
+FLAGS = flags.FLAGS
+
+
+DEFAULT_CLOUD_SPANNER_END_POINT = 'https://spanner.googleapis.com/'
+DEFAULT_CLOUD_SPANNER_CONFIG = 'regional-us-central1'
```
haih-g: This is the default value to the class constructor. Benchmarks can override it by passing a different value to the constructor.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+    self._nodes = nodes
+    self._name = name
+    self._description = description
+    self._config = config
+    self._database = database
+    self._ddl = ddl
+    self._end_point = end_point
+
+    # Cloud Spanner may not explicitly set the following common flags.
+    self.project = project if project is not None else util.GetDefaultProject()
```
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+    self.zone = None
+
+    self._OverrideEndPoint()
```
haih-g: My understanding is that PKB only supports serial execution of benchmarks, so this won't cause race condition.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+        logging.error('Reset Cloud Spanner end point failed.')
+
+  def _Create(self):
+    """Creates the instance, the database, and update the schema."""
+    cmd = util.GcloudCommand(self, 'spanner', 'instances', 'create', self._name)
+    cmd.flags['description'] = self._description
+    cmd.flags['nodes'] = self._nodes
+    cmd.flags['config'] = self._config
+    _, _, retcode = cmd.Issue()
+    if retcode != 0:
```
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+      logging.error('Create GCP Spanner instance failed.')
+      return
+
+    cmd = util.GcloudCommand(self, 'spanner', 'databases', 'create',
+                             self._database)
+    cmd.flags['instance'] = self._name
+    _, _, retcode = cmd.Issue()
+    if retcode != 0:
```
haih-g: DONE

perfkitbenchmarker/providers/gcp/gcp_spanner.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if instance_only:
+      return True
+
+    cmd = util.GcloudCommand(self, 'spanner', 'databases', 'describe',
+                             self._database)
+    cmd.flags['instance'] = self._name
+
+    # Do not log error or warning when checking existence.
+    _, _, retcode = cmd.Issue(suppress_warning=True)
```
haih-g: Same

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: Datastore doesn't have a concept of table, it's an 'Entity'. So it should have been datastore_ycsb_entity anyway.

YCSB is a benchmarks for No SQL database, not traditional database. Basically a No SQL database is a key value store, and may not have the concept of 'Table'.

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
haih-g: DONE

haih-g:
Friendly ping. Is there anything else I missed?

perfkitbenchmarker/providers/gcp/gcp_spanner.py
yuyantingzero: it support parallel execution.

perfkitbenchmarker/providers/gcp/gcp_spanner.py
yuyantingzero: Not Done!~

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
yuyantingzero: Not DONE

perfkitbenchmarker/linux_benchmarks/cloud_spanner_ycsb_benchmark.py
yuyantingzero: https://github.com/brianfrankcooper/YCSB/wiki/Core-Properties
table is an argument for ycsb.
I assume you will need a flag for every "table"/"entity" for SQL/No-SQL database.

haih-g:
To recapture a few key DOs & DON'Ts that benchmark contributors need to understand.

On a single machine, there might be multiple benchmarks running. They may run against the same storage products under different configurations. Each of the benchmark is running from a process. One process runs only one benchmark at a time. Thus

- DON'T change the environment shared by all the processes. It will affect other benchmarks.
- DO isolate the backends. There might be two processes both running cloud spanner banchmarks at the same time. If they pointing to the same project / instance / database / table, they will interfere with one another. Use FLAGS.run_uri to differentiate them.

It would be nice to include this in the CONTRIBUTING.md

haih-g:
3768166 contains following major changes:

1. Remove the logic to override end points from the code. This contradicts "DON'T change the share denvironment". If needed, user should manually set end point before invoking pkb. Also, added a flag to run custom vm install commands. User can override end points or doing other uncommon setups on VM.

2. Instance name is ycsb-<run_uri>. This should distinguish all ycsb-benchmarks. Since we separate at the instance layer, everything else (database, table) can be a constant.

3. Although I still think it's an awkward design, moving flags into gcp_spanner.py might make the code easier to maintain. User may need to always pass in flags, which can be mitigated by user side scripts.

asaksena:
Thank you for making changes as per the recommendations. 
Based on our conversation, as you understand, there are still a few issues which don't fit the pkb paradigm, however since current usage of the spanner resource in pkb is isolated, we can defer any further modifications till a different benchmark needs to use the resource.

asaksena:
merging

haih-g:
Thanks.

PR #1388, opened 2017-05-31
[script] A wrapper script to drive step-load runs of the managed_mysql_service benchmark.

sspano:
Wrapper file that drives mysql_service_benchmark sysbench runs with different thread counts.

@stfeng @gareth-ferneyhough

### DIFF
--- a/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
+++ b/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
@@ -91,6 +91,7 @@ gflags.DEFINE_string(RUN_URI, None,
 gflags.DEFINE_list(ADDITIONAL_FLAGS, None,
                    'List of additional PKB mysql_service valid flags (strings).'
                    'For example: ["--storage_size=100"].')
+
 # TODO: Implement flag for STDOUT/STDERR file paths.
 
 
@@ -193,16 +194,11 @@ def _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename):
   start_time = time.time()
   p = subprocess.Popen(pkb_cmd_list, stdout=stdout_file, stderr=stderr_file)
   logging.info('Waiting for PKB call to finish.')
-  # TODO: implement better timeout. Currently this call will continue executing.
-  # In other words, unsure that p.terminate() is the best way to execute the
-  # timeout.
-  while p.returncode is None:
-    elapsed_time = time.time() - start_time
-    if elapsed_time > PKB_TIMEOUT:
-      p.terminate()
-      raise OperationTimeoutError(
-          'PKB execution timed out at %s seconds.' % str(int(elapsed_time)))
-  logging.info('PKB call finished.')
+  # TODO: implement timeout. Currently this call will wait unbounded.
+  # Will probably have to implement with threading.
+  p.wait()
+  elapsed_time = time.time() - start_time
+  logging.info('PKB call finished in %d seconds.', elapsed_time)
 
 
 def _get_run_uri(filename):
--- a/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
+++ b/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
@@ -60,7 +60,7 @@ DATETIME_FORMAT = '{:%m_%d_%Y_%H_%M_}'
 URI_REGEX = r'run_uri=([a-z0-9]{8})'
 ADDITIONAL_FLAGS = 'additional_flags'
 SLEEP_TIME_BETWEEN_RUNS = 20  # seconds
-TAIL_LINE_NUM = 20
+TAIL_LINE_NUM = '20'
 
 MAX_SLEEP_ITER = 24  # max wait time for a run is max_sleep_iter x sleep_time
 SLEEP_TIME = 1800  # seconds
--- a/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
+++ b/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
@@ -81,7 +81,7 @@ FLAGS = gflags.FLAGS
 gflags.DEFINE_integer(SYSBENCH_RUN_SECONDS, 480,
                       'The duration, in seconds, of each run phase with varying'
                       'thread count.')
-gflags.DEFINE_list(THREAD_COUNT_LIST, [1, 2, 5, 8],
+gflags.DEFINE_list(THREAD_COUNT_LIST, [1, 2, 4, 8],
                    'The number of test threads on the client side.')
 gflags.DEFINE_integer(SYSBENCH_REPORT_INTERVAL, 2,
                       'The interval, in seconds, we ask sysbench to report '
@@ -157,7 +157,6 @@ def _run(run_uri):
     stdout_filename, stderr_filename = _generate_filenames(RUN, t)
     logging.info('Executing PKB run with thread count: %i', t)
     _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
-#     _wait_for_run(stderr_filename)
     logging.info('Finished executing PKB with thread count: %i', t)
     time.sleep(SLEEP_TIME_BETWEEN_RUNS)
 
@@ -172,7 +171,6 @@ def _cleanup_teardown_pkb(run_uri):
   pkb_cmd = (PKB + STAGE_FLAG + CLEANUP + ',' + TEARDOWN + URI_FLAG + run_uri)
   [stdout_filename, stderr_filename] = _generate_filenames(CLEANUP, None)
   _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
-#   _wait_for_run(stderr_filename)
   logging.info('Finished executing PKB cleanup and teardown.')
 
 
@@ -190,7 +188,7 @@ def _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename):
   logging.info('pkb command list: %s', str(pkb_cmd_list))
   p = subprocess.Popen(pkb_cmd_list, stdout=stdout_file, stderr=stderr_file)
   logging.info('Waiting for PKB call to finish.')
-  p.wait()  # add timeout time
+  p.wait()
   logging.info('PKB call finished.')
 
 
--- a/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
+++ b/perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
@@ -31,8 +31,9 @@ All requirements for mysql_service_benchmark still apply. See
 perkfitbenchmarker.linuxbenchmarks.mysql_service_benchmark for more details.
 
 Any additional flags not specifically outlined in the flags below can be added
-as a list of strings under the 'additional_flags' flag. For example, a possible
-call with additional flags could be:
+as a list of strings under the 'additional_flags' flag. These flags will be
+passed through as is to the underlying benchmarking code. For example,
+a possible call with additional flags could be:
 ./launch_mysql_service --run_uri=2u2u2u3i
   --additional_flags=['--storage_size=100', '--cloud=GCP'].
 """
@@ -50,7 +51,6 @@ import gflags
 SYSBENCH_RUN_SECONDS = 'sysbench_run_seconds'
 SYSBENCH_THREAD_COUNT = 'sysbench_thread_count'
 SYSBENCH_REPORT_INTERVAL = 'sysbench_report_interval'
-STORAGE_BUCKET = 'storage_bucket'
 THREAD_COUNT_LIST = 'thread_count_list'
 RUN_URI = 'run_uri'
 DEFAULT_RUN_TIME = 60
@@ -59,6 +59,8 @@ STDERR = 'STDERR'
 DATETIME_FORMAT = '{:%m_%d_%Y_%H_%M_}'
 URI_REGEX = r'run_uri=([a-z0-9]{8})'
 ADDITIONAL_FLAGS = 'additional_flags'
+SLEEP_TIME_BETWEEN_RUNS = 20  # seconds
+TAIL_LINE_NUM = 20
 
 MAX_SLEEP_ITER = 24  # max wait time for a run is max_sleep_iter x sleep_time
 SLEEP_TIME = 1800  # seconds
@@ -69,7 +71,6 @@ STAGE_FLAG = ' --run_stage='
 URI_FLAG = ' --run_uri='
 THREAD_FLAG = ' --sysbench_thread_count='
 RUN_TIME = ' --sysbench_run_seconds='
-STORAGE = ' --cloud_storage_bucket='
 PROVISION = 'provision'
 PREPARE = 'prepare'
 RUN = 'run'
@@ -78,8 +79,8 @@ TEARDOWN = 'teardown'
 
 FLAGS = gflags.FLAGS
 gflags.DEFINE_integer(SYSBENCH_RUN_SECONDS, 480,
-                      'The duration of the actual run in which results are '
-                      'collected, in seconds.')
+                      'The duration, in seconds, of each run phase with varying'
+                      'thread count.')
 gflags.DEFINE_list(THREAD_COUNT_LIST, [1, 2, 5, 8],
                    'The number of test threads on the client side.')
 gflags.DEFINE_integer(SYSBENCH_REPORT_INTERVAL, 2,
@@ -88,14 +89,12 @@ gflags.DEFINE_integer(SYSBENCH_REPORT_INTERVAL, 2,
 gflags.DEFINE_string(RUN_URI, None,
                      'Run identifier, if provided, only run phase '
                      'will be completed.')
-gflags.DEFINE_string(STORAGE_BUCKET, None,
-                     'GCS bucket to upload records to. Bucket must exist.')
 gflags.DEFINE_list(ADDITIONAL_FLAGS, None,
                    'List of additional PKB mysql_service valid flags (strings).'
                    'For example: ["--storage_size=100"].')
 
 
-class TimeOutError(Exception):
+class UnexpectedFileOutputError(Exception):
   pass
 
 
@@ -112,7 +111,8 @@ def driver(argv):
     sys.exit(1)
   run_uri = FLAGS.run_uri
   if not run_uri:
-    logging.info('No run_uri given. Will run full sysbench test.')
+    logging.info('No run_uri given. Will run full mysql_service_benchmark '
+                 'test.')
     run_uri = _provision_prepare_pkb()
     logging.info('Provision and prepare completed. Run uri assigned: %s',
                  run_uri)
@@ -122,7 +122,7 @@ def driver(argv):
 
 
 def _provision_prepare_pkb():
-  """Run cleanup stage of PKB benchmark.
+  """Run provision and prepare stage of PKB benchmark.
 
   Returns:
     run_uri: (string)
@@ -135,7 +135,7 @@ def _provision_prepare_pkb():
                pkb_cmd)
   [stdout_filename, stderr_filename] = _generate_filenames(PROVISION, None)
   _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
-  return _wait_for_run(stderr_filename)
+  return _get_run_uri(stderr_filename)
 
 
 def _run(run_uri):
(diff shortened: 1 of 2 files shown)


stfeng:
@sspano

FYI - I changed the description of this PR to properly reflect what this is really about.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python
```
stfeng: script location: does it have to be in a sub-dir under the scripts dir? consider one of these two options:

1. put it directly under scripts dir.
2. rename the "step_load_driver" to database_scripts. We will then put all future database benchmarking related scripts to this dir.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+phase. Below are examples of possible ways to call this class.
+
+Possible call:
+./launch_mysql_service --thread_count_list=[1,2,5] --sysbench_run_seconds=20
+./launch_mysql_service --run_uri=12e4s6t8 --thread_count_list=[1,2,4,8,10]
+
+All requirements for mysql_service_benchmark still apply. See
+perkfitbenchmarker.linuxbenchmarks.mysql_service_benchmark for more details.
+
+Any additional flags not specifically outlined in the flags below can be added
```
stfeng: clarify that "additional_flags" will be passed through as is to the underlying benchmarking code.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+import subprocess
+import sys
+import time
+import gflags
+
+# GLOBAL STRINGS
+SYSBENCH_RUN_SECONDS = 'sysbench_run_seconds'
+SYSBENCH_THREAD_COUNT = 'sysbench_thread_count'
+SYSBENCH_REPORT_INTERVAL = 'sysbench_report_interval'
+STORAGE_BUCKET = 'storage_bucket'
```
stfeng: can you pass this through as part of "additional flags" rather than defining a new flag for it?

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+MAX_SLEEP_ITER = 24  # max wait time for a run is max_sleep_iter x sleep_time
+SLEEP_TIME = 1800  # seconds
+
+# FLAG STRINGS
+PKB = './pkb.py --benchmarks=mysql_service'
+STAGE_FLAG = ' --run_stage='
+URI_FLAG = ' --run_uri='
+THREAD_FLAG = ' --sysbench_thread_count='
+RUN_TIME = ' --sysbench_run_seconds='
+STORAGE = ' --cloud_storage_bucket='
```
stfeng: same as above.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+PROVISION = 'provision'
+PREPARE = 'prepare'
+RUN = 'run'
+CLEANUP = 'cleanup'
+TEARDOWN = 'teardown'
+
+FLAGS = gflags.FLAGS
+gflags.DEFINE_integer(SYSBENCH_RUN_SECONDS, 480,
```
stfeng: clarify in comment that this is the run duration for EACH thread count.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+                      'collected, in seconds.')
+gflags.DEFINE_list(THREAD_COUNT_LIST, [1, 2, 5, 8],
+                   'The number of test threads on the client side.')
+gflags.DEFINE_integer(SYSBENCH_REPORT_INTERVAL, 2,
+                      'The interval, in seconds, we ask sysbench to report '
+                      'results.')
+gflags.DEFINE_string(RUN_URI, None,
+                     'Run identifier, if provided, only run phase '
+                     'will be completed.')
+gflags.DEFINE_string(STORAGE_BUCKET, None,
```
stfeng: same as above, should we just pass through storage_bucket flag as additional flags?

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+    argv: system arguments (command line flags).
+  """
+  try:  # Parse command line flags
+    argv = FLAGS(argv)
+  except gflags.FlagsError as e:
+    logging.error('%s\nUsage: %s ARGS\n%s', e, sys.argv[0], FLAGS)
+    sys.exit(1)
+  run_uri = FLAGS.run_uri
+  if not run_uri:
+    logging.info('No run_uri given. Will run full sysbench test.')
```
stfeng: minor correction in the log string: run full "mysql_service_benchmark" test.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+    run_uri = _provision_prepare_pkb()
+    logging.info('Provision and prepare completed. Run uri assigned: %s',
+                 run_uri)
+  _run(run_uri)
+  if not FLAGS.run_uri:
+    _cleanup_teardown_pkb(run_uri)
+
+
+def _provision_prepare_pkb():
+  """Run cleanup stage of PKB benchmark.
```
stfeng: comment is wrong (opposite to what the code does. ;)

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+      run_iterations)
+  for t in FLAGS.thread_count_list:
+    pkb_cmd = (PKB + STAGE_FLAG + RUN + URI_FLAG + run_uri + THREAD_FLAG +
+               str(t) + RUN_TIME + str(FLAGS.sysbench_run_seconds) + ' &')
+    if FLAGS.additional_flags:
+      pkb_cmd = _append_additional_flags(pkb_cmd)
+    stdout_filename, stderr_filename = _generate_filenames(RUN, t)
+    logging.info('Executing PKB run with thread count: ' + str(t))
+    _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
+    _wait_for_run(stderr_filename)
```
stfeng: Usually we have a small wait period in between the runs with different thread count, e.g.
a typical step load run includes:

run thread = 1 for 20 minutes, wait for it to finish.
wait for 20 seconds - 30 seconds.
run thread = 2 for 20 minutes, wait for it to finish, 

so on so forth.

Can you please add a new flag "sleep_time_between_runs" or something like that? And then add a sleep right here (right after _wait_for_run).

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+    logging.info('Finished executing PKB with thread count: ' + str(t))
```
stfeng: str(t): here, below and above - is it really necessary to do str(integer) transition?

Typical convention as seen in the rest of PKB code base for string formatting is following:

string_result = "template string: %d and %s" % (123, "xyz")

this will put 123 in place of %d and "xyz" in place of %s

So for this line, it'd look like "finished executing... %d" % t

But meanwhile, I can see that str(t) fits naturally when you need to concatenate multiple strings together using + sign, it is also quite readable.

So your call here: whether you want to convert some of the str(t) here to fit more consistently with the rest of the PKB code base.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+def _cleanup_teardown_pkb(run_uri):
+  """Run cleanup stage of PKB benchmark.
+
+  Args:
+    run_uri: (string)
+  """
+  logging.info('Run phase complete. Starting cleanup/teardown.')
+  pkb_cmd = (PKB + STAGE_FLAG + CLEANUP + ',' + TEARDOWN + URI_FLAG + run_uri)
+  [stdout_filename, stderr_filename] = _generate_filenames(CLEANUP, None)
+  _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
```
stfeng: as chatted in person, add a wait_for_completion here.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+  """
+  stdout_file = open(stdout_filename, 'w+')
+  stderr_file = open(stderr_filename, 'w+')
+  pkb_cmd_list = shlex.split(pkb_cmd)
+  logging.info('pkb command list:' + str(pkb_cmd_list))
+  subprocess.Popen(pkb_cmd_list, stdout=stdout_file, stderr=stderr_file)
+
+
+def _wait_for_run(filename):
+  """Given a filename, wait for it to be populated, return run_uri.
```
stfeng: document the algorithm/method you are using here to implement the wait, point out the limitation of that method (e..g, if pkb output changes then the wait would be broken).

add a TODO here for improvement on the wait. (check out the "RobustRemoteCommand" implementation in PKB, it solves the same problem of "how to wait for a process to finish").

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+    run_uri: (string) Run identifier from file.
+
+  Raises:
+    TimeOutError: Filename not populated after considerable sleep iterations.
+  """
+  for _ in range(MAX_SLEEP_ITER):
+    logging.info('Checking %s for run completion.', filename)
+    r = re.compile(URI_REGEX)
+    for line in open(filename):
+      matches = r.search(line)
```
stfeng: Can you read the last N lines from the file, and then do the search?
(where N >= the number of lines between the pattern and the end of the result file)

f = open(filename)
lines = read_last_N_lines(f)
for line in lines:
   matches = r.seach(line)

The current method will read the entire output file every time which is unnecessary. please check to  see if there is examples in python to read the last N lines of a file? (I can see quite a few after doing a quick google search)

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
```diff
@@ -0,0 +1,258 @@
...
+SYSBENCH_THREAD_COUNT = 'sysbench_thread_count'
+SYSBENCH_REPORT_INTERVAL = 'sysbench_report_interval'
+STORAGE_BUCKET = 'storage_bucket'
+THREAD_COUNT_LIST = 'thread_count_list'
+RUN_URI = 'run_uri'
+DEFAULT_RUN_TIME = 60
+STDOUT = 'STDOUT'
+STDERR = 'STDERR'
+DATETIME_FORMAT = '{:%m_%d_%Y_%H_%M_}'
+URI_REGEX = r'run_uri=([a-z0-9]{8})'
```
stfeng: looks like this string is used to check for run completion.

but I don't think run_uri=*** is a unique (enough) signal to indicate a particular stage is completed.

It's entirely possible that PKB code may emit some messages like "I am starting a run with run_uri=*** " at the beginning of a run.

I think we need to look for more specific strings to robustly catch the completion of a particular run, can you do more research on this? The string might even be different depending on the phase, if so, that's fine, we just need to define the strings separately.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
stfeng: Note above, if you can find a way to wait for the run to finish without relying on file parsing, then this comment becomes moot. 

(see my other comment about where to explore for examples of waiting for process to finish).

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: I chose to rename file to database_scripts.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Makes sense. Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Can you tell I copied and pasted a bit? haha. Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done with string formatting in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Realized the subprocess call I'm making has a clean way of way of waiting for process to terminate. Testing it now. Now the wait_for_run will only grab the run_uri for provision/prepare once the process has completed.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Using another way to test run completion so URI_REGEX will only be used to grab the run_uri after the provision/prepare phase is finished.

perfkitbenchmarker/scripts/step_load_driver/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
```diff
@@ -0,0 +1,261 @@
...
+    stdout_filename: (str) filename string.
+    stderr_filename: (str) filename_str
+  """
+  stdout_file = open(stdout_filename, 'w+')
+  stderr_file = open(stderr_filename, 'w+')
+  pkb_cmd_list = shlex.split(pkb_cmd)
+  logging.info('pkb command list: %s', str(pkb_cmd_list))
+  p = subprocess.Popen(pkb_cmd_list, stdout=stdout_file, stderr=stderr_file)
+  logging.info('Waiting for PKB call to finish.')
+  p.wait()  # add timeout time
```
stfeng: Is it hard to support a timeout value here for p.wait()? If not, let's get it done with this PR, otherwise, please add a TODO here to indicate that will be addressed later.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
```diff
@@ -0,0 +1,261 @@
...
+  """Run cleanup stage of PKB benchmark.
+
+  Args:
+    run_uri: (string)
+  """
+  logging.info('Run phase complete. Starting cleanup/teardown.')
+  pkb_cmd = (PKB + STAGE_FLAG + CLEANUP + ',' + TEARDOWN + URI_FLAG + run_uri)
+  [stdout_filename, stderr_filename] = _generate_filenames(CLEANUP, None)
+  _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
+#   _wait_for_run(stderr_filename)
```
stfeng: remove this line, it's no longer necessary now.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
```diff
@@ -0,0 +1,261 @@
...
+      run_iterations)
+  for t in FLAGS.thread_count_list:
+    pkb_cmd = (PKB + STAGE_FLAG + RUN + URI_FLAG + run_uri + THREAD_FLAG +
+               str(t) + RUN_TIME + str(FLAGS.sysbench_run_seconds) + ' &')
+    if FLAGS.additional_flags:
+      pkb_cmd = _append_additional_flags(pkb_cmd)
+    stdout_filename, stderr_filename = _generate_filenames(RUN, t)
+    logging.info('Executing PKB run with thread count: %i', t)
+    _execute_pkb_cmd(pkb_cmd, stdout_filename, stderr_filename)
+#     _wait_for_run(stderr_filename)
```
stfeng: remove it, no need to leave deadcode here.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
```diff
@@ -0,0 +1,261 @@
...
+PREPARE = 'prepare'
+RUN = 'run'
+CLEANUP = 'cleanup'
+TEARDOWN = 'teardown'
+
+FLAGS = gflags.FLAGS
+gflags.DEFINE_integer(SYSBENCH_RUN_SECONDS, 480,
+                      'The duration, in seconds, of each run phase with varying'
+                      'thread count.')
+gflags.DEFINE_list(THREAD_COUNT_LIST, [1, 2, 5, 8],
```
stfeng: 1, 2, 5, 8 as default seems unusual, did you mean 1, 2, 4, 8?

stfeng:
Some additional (minor) comments.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
sspano: Yes. Good catch.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
sspano: Outdated comment. Removed in next commit. 

p.wait() holds until the process completes.

As discussed, run phase has a sleep 
 line 161 : "time.sleep(SLEEP_TIME_BETWEEN_RUNS)"
after the pkb cmd is finished executing.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
sspano: Done in next commit.

perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
stfeng: The question is about timeout value for the p.wait() command. Is there a way to specify a timeout  for p.wait()? (Note: the sleep in-between the individual runs is not relevant to this discussion). 

This discussion is more about how we guard against a run that's hung/take too long. Please at least take a look into this, if it's simple to add a timeout, we can do it now in this PR; otherwise, we can add a TODO here and address it later.

stfeng:
One final comment about the p.wait() timeout value. 

@gareth-ferneyhough do you want to also take a look at this? This PR is entirely about tooling/utility, it doesn't affect PKB benchmarking code. Thanks.

sspano:
Current implementation has a wait which throws an exception but I think the PKB call continues to run in the background. Added TODO for better implementation in the future but works for now.

stfeng:
LGTM. Merging...

PR #1394, opened 2017-06-12
Change default Azure disk type

ehankland:


dlott:
Do you also want to update CHANGES.next.md?

ehankland:
@dlott This just makes the default work again - its the same type as before

PR #1396, opened 2017-06-13
Support Beam Python in beam_integration_benchmark

markflyhigh:
Currently beam_integration_benchmark can only run Beam Java pipeline. Improve it so that people can use it to run integration test with Python SDK.

### DIFF
--- a/perfkitbenchmarker/beam_benchmark_helper.py
+++ b/perfkitbenchmarker/beam_benchmark_helper.py
@@ -63,6 +63,7 @@ BEAM_REPO_LOCATION = 'https://github.com/apache/beam.git'
 INSTALL_COMMAND_ARGS = ["clean", "install", "-DskipTests",
                         "-Dcheckstyle.skip=true"]
 
+
 def InitializeBeamRepo(benchmark_spec):
   """Ensures environment is prepared for running Beam benchmarks.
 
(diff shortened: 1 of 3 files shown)


markflyhigh:
+R: @asaksena

perfkitbenchmarker/beam_benchmark_helper.py
```diff
@@ -137,7 +171,65 @@ def BuildMavenCommand(benchmark_spec, classname, job_arguments):
...
+
+def _GetBeamDir():
   # TODO: This is temporary, find a better way.
-  beam_dir = FLAGS.beam_location if FLAGS.beam_location else os.path.join(
+  return FLAGS.beam_location if FLAGS.beam_location else os.path.join(
       vm_util.GetTempDir(), 'beam')
-  return cmd, beam_dir
+
+
+def _FindFiles(base_path, pattern):
```
asaksena: Do you think this can be used ? 
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/4db490efb48ebb33584ac159580c271f5e1d37f5/perfkitbenchmarker/data/__init__.py#L103

markflyhigh: I'm afraid I need more that check exist. It basically search for the a file that matching the given pattern under a given directory. The tar ball name contains version number against HEAD and `os.path.exists` doesn't support glob.

asaksena: Got it

asaksena:
LGTM

PR #1405, opened 2017-06-19
Add managed_relational_db base classes and tests

gareth-ferneyhough:


perfkitbenchmarker/managed_relational_db.py
```diff
@@ -0,0 +1,107 @@
+from abc import ABCMeta, abstractmethod
+import uuid
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import resource
+
+# TODO (ferneyhough): change to enum
+flags.DEFINE_string('database', None,
+                    'Managed database flavor to use (mysql, postgres)')
+flags.DEFINE_string('database_name', None,
```
stfeng: Is database name dependent on the test tool? e.g., sysbench requires the database to be named as sbtest, I believe. (correct me if I am wrong here).

gareth-ferneyhough: That is up to the implementation of the benchmark in PKB. If the sysbench benchmark requires a certain name, then that will be reflected in the spec.

gareth-ferneyhough: And ill be sure to watch out for that. I don't think it is appropriate for this class to have any notion of benchmark-specific requirements, however.

PR #1407, opened 2017-06-20
Fix gce gpu benchmarks

gareth-ferneyhough:


perfkitbenchmarker/providers/gcp/gce_virtual_machine.py
```diff
@@ -408,7 +478,8 @@ def _GenerateCreateCommand(self, ssh_keys_path):
     cmd.flags['metadata'] = ','.join(
         ['%s=%s' % (k, v) for k, v in metadata.iteritems()])
 
-    if not FLAGS.gce_migrate_on_maintenance:
+    # TODO: If GCE one day supports live migration on GPUs this can be revised.
+    if not FLAGS.gce_migrate_on_maintenance or self.gpus:
```
dlott: Should we raise an error if `FLAGS.gce_migrate_on_maintenance and self.gpus` since that's a bad config option?  Better to tell the user they can't migrate rather than them thinking they can and failing?

PR #1410, opened 2017-06-21
[gcp_managed_relational_db] Added GCP functionality and unit tests.

sspano:
@stfeng2 @gareth-ferneyhough 
Added methods and unit tests for GCP managed relational db.

### DIFF
--- a/perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
+++ b/perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
@@ -36,6 +36,7 @@ to install it. This will allow this benchmark to properly create an instance.
 """
 
 import json
+import logging
 
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import providers
@@ -67,10 +68,13 @@ class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
     return self.port
 
   @staticmethod
-  # TODO: implement for real
+  # TODO: Implement so values aren't hard-coded. Retrieve latest versions
+  # from outside resource.
   def GetLatestDatabaseVersion(database):
     if database == managed_relational_db.MYSQL:
       return '5.7'
+    elif database == managed_relational_db.POSTGRES:
+      return '9.6'
 
   def __init__(self, managed_relational_db_spec):
     super(GCPManagedRelationalDb, self).__init__(managed_relational_db_spec)
@@ -145,10 +149,13 @@ class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
     """
     cmd = util.GcloudCommand(self, 'sql', 'instances', 'describe',
                              self.instance_id)
-    stdout, _, _ = cmd.Issue()
-    json_output = json.loads(stdout)
-    return (json_output[0]['kind'] ==
-            'sql#instance')
+    stdout, _, _ = cmd.Issue
+    try:
+      json_output = json.loads(stdout)
+      exists = (json_output[0]['kind'] == 'sql#instance')
+    except:
+      exists = False
+    return exists
 
   def _IsReady(self):
     """Return true if the underlying resource is ready.
@@ -163,20 +170,31 @@ class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
     cmd = util.GcloudCommand(self, 'sql', 'instances', 'describe',
                              self.instance_id)
     stdout, _, _ = cmd.Issue()
-    json_output = json.loads(stdout)
-    is_ready = (json_output[0]['state']
-                == 'RUNNABLE')
+    try:
+      json_output = json.loads(stdout)
+      is_ready = (json_output[0]['state'] == 'RUNNABLE')
+    except:
+      logging.exception('Error attempting to read stdout. Creation failure.')
+      is_ready = False
     if not is_ready:
       return False
     self.endpoint = self._ParseEndpoint(json_output)
     self.port = self.MYSQL_DEFAULT_PORT
     return True
 
-  def _ParseType(self, describe_instance_json):
-    return describe_instance_json[0]['kind'] == 'sql#instance'
-
   def _ParseEndpoint(self, describe_instance_json):
-    return describe_instance_json[0]['selfLink']
+    """Return the URI of the resource given the metadata as JSON.
+
+    Args:
+      describe_instance_json: JSON output.
+    Returns:
+      resource URI (string)
+    """
+    try:
+      selflink = describe_instance_json[0]['selfLink']
+    except:
+      selflink = ""
+    return selflink
 
   def _PostCreate(self):
     """Method that will be called once after _CreateReource is called.
--- a/perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
+++ b/perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
@@ -45,6 +45,11 @@ from perfkitbenchmarker.providers.gcp import util
 
 FLAGS = flags.FLAGS
 
+LATEST_MYSQL_VERSION = '5.7'
+LATEST_POSTGRES_VERSION = '9.6'
+DEFAULT_GCP_MYSQL_VERSION = 'MYSQL_5_7'
+DEFAULT_GCP_POSTGRES_VERSION = 'POSTGRES_9_6'
+
 
 class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
   """An object representing a GCP managed relational database.
@@ -57,24 +62,8 @@ class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
   CLOUD = providers.GCP
   SERVICE_NAME = 'managed_relational_db'
   # These are the constants that should be specified in GCP's cloud SQL command.
-  DEFAULT_GCP_MY_SQL_VERSION = 'MYSQL_5_7'
   GCP_PRICING_PLAN = 'PACKAGE'
   MYSQL_DEFAULT_PORT = 3306
-
-  def GetEndpoint(self):
-    pass
-
-  def GetPort(self):
-    return self.port
-
-  @staticmethod
-  # TODO: Implement so values aren't hard-coded. Retrieve latest versions
-  # from outside resource.
-  def GetLatestDatabaseVersion(database):
-    if database == managed_relational_db.MYSQL:
-      return '5.7'
-    elif database == managed_relational_db.POSTGRES:
-      return '9.6'
 
   def __init__(self, managed_relational_db_spec):
     super(GCPManagedRelationalDb, self).__init__(managed_relational_db_spec)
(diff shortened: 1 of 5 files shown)


perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -57,12 +58,13 @@ class GCPManagedRelationalDb(managed_relational_db.BaseManagedRelationalDb):
...
 
   def GetEndpoint(self):
     pass
 
   def GetPort(self):
-    pass
+    return self.port
 
   @staticmethod
   # TODO: implement for real
```
stfeng2: remove this TODO? the implementation is here now.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -124,7 +133,7 @@ def _Delete(self):
     deleted.
     """
     cmd = util.GcloudCommand(self, 'sql', 'instances', 'delete',
-                             self.instance_id)
+                             self.instance_id, '--quiet')
```
stfeng2: Would --quiet also stop logging for the delete operation? If so, is there a different option that would skip the confirmation but still do the logging? (like -y or such)

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -149,8 +158,21 @@ def _IsReady(self):
     Returns:
       True if the resource was ready in time, False if the wait timed out.
     """
+    cmd = util.GcloudCommand(self, 'sql', 'instances', 'describe',
```
stfeng2: Question: when is _IsReady() being called? During prepare phase or provision phase?

The reason it's important is becuase we also have a data loading step at the prepare phase, if "is ready" implies the data is loaded and prepare is done, then we are not ready here yet. I think it's for the provision phase only, but I could be wrong here.

@gareth-ferneyhough FYI.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -149,8 +158,21 @@ def _IsReady(self):
...
+                             self.instance_id)
+    stdout, _, _ = cmd.Issue()
+    json_output = json.loads(stdout)
```
stfeng2: a few lines above, for the _Exists() method, we used the retcode of gcloud describe being 0 to imply the resource exists, but that is fragile, it's possible the return code is for the execution status of the glcoud command itself, not for the output of the gcloud. If the resource does not exist but gcloud successfully determined it does not exist, it might just return 0 and output an error message on command line.

I'd consider using a different way of detecting existence. Please look at the AWS side of implementation to see what should be used.

@gareth-ferneyhough fyi

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: Good catch.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: I don't think so. 
From gcloud docs: 
```
--quiet, -q
Disable all interactive prompts when running gcloud commands.
If input is required, defaults will be used, or an error will be raised.
Overrides the default core/disable_prompts property value for this command invocation.
```
Further stdout is directed in the[ util command](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L293). 

And I noted that we did the same for [mysql_service_benchmark](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/linux_benchmarks/mysql_service_benchmark.py#L841).

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: [During create.](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/resource.py#L155)

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: You're right. I changed it to something similar to the isready method of parsing json output.

@gareth-ferneyhough Do you have any issue with this?

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: The goal of this method is to have a way of having the most up to date version of postgres or mysql without it being hardcoded. It is used by the [spec](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/2931812f6e1817d525347d3da7c5115cb65c016f/perfkitbenchmarker/configs/benchmark_config_spec.py#L339) if no version is specified.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
gareth-ferneyhough: Either is fine, but gcloud does return 1 on describe if the instance does not exist, so that is why I wrote it that way. May be more fragile than looking at the json.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -149,8 +160,24 @@ def _IsReady(self):
...
+    is_ready = (json_output[0]['state']
+                == 'RUNNABLE')
+    if not is_ready:
+      return False
+    self.endpoint = self._ParseEndpoint(json_output)
+    self.port = self.MYSQL_DEFAULT_PORT
     return True
 
+  def _ParseType(self, describe_instance_json):
```
gareth-ferneyhough: This isn't used anywhere - what is its purpose? Also the naming and intention is unclear, because it isn't parsing the type - it is returning whether or not the instance is a sql instance.

gareth-ferneyhough: Also, please add docs to all all your methods.

sspano: You're right. I forgot to delete it. I just transferred the logic to _Exists(). I created it because of earlier comment about retcode.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
stfeng2: ok, please clarify in the TODO comment, such as: "implement the function for postgres".

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
stfeng2: sounds good. thanks.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
stfeng2: SG.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
stfeng2: thanks, I am leaning towards return string parsing.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -136,8 +145,10 @@ def _Exists(self):
     """
     cmd = util.GcloudCommand(self, 'sql', 'instances', 'describe',
                              self.instance_id)
-    _, _, retcode = cmd.Issue()
-    return retcode == 0
+    stdout, _, _ = cmd.Issue()
+    json_output = json.loads(stdout)
```
stfeng2: if the command fails, then most likely the stdout is empty or it would not be a JSON result, in that case, we should not try to do json.loads; alternatively, we can always do json.loads here but wrap it in try_except. On any exception, we consider the call failed and return "not exist".

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
```diff
@@ -149,8 +160,24 @@ def _IsReady(self):
     Returns:
       True if the resource was ready in time, False if the wait timed out.
     """
+    cmd = util.GcloudCommand(self, 'sql', 'instances', 'describe',
+                             self.instance_id)
+    stdout, _, _ = cmd.Issue()
```
stfeng2: Here and everywhere in this file where we try to do json.loads(stdout), it's quite fragile and needs to be wrapped inside try except and errors need to be handled. The reason being, stdout can be absent or does not contain a JSON, depending on the status of the command and such. We need to handle the error here and return false/true accordingly to the caller.

tests/data/gcloud-describe-db-instances-available.json
```diff
@@ -0,0 +1,52 @@
...
+        "type": "PRIMARY"
+      }
+    ],
+    "kind": "sql#instance",
+    "name": "pkb-db-instance-123",
+    "project": "sql-benchmarking",
+    "region": "us-central1",
+    "selfLink": "https://www.googleapis.com/sql/pkb-db-instance-123",
+    "serverCaCert": {
+      "cert": "-----BEGIN CERTIFICATE-----\n9n\nbJ\nBG\ns\nC\ng\nCeA==\n-----END CERTIFICATE-----",
```
stfeng2: just to make sure, the "cert" and sha1fingerprint here are all MOCK-UPs right?

tests/data/gcloud-describe-db-instances-available.json
```diff
@@ -0,0 +1,52 @@
...
+      "certSerialNumber": "0",
+      "commonName": "C=US,O=Google\\, Inc,CN=Google Cloud SQL Server CA",
+      "createTime": "2017-06-21T17:54:45.430000+00:00",
+      "expirationTime": "2019-06-21T17:55:45.430000+00:00",
+      "instance": "pkb-db-instance-123",
+      "kind": "sql#sslCert",
+      "sha1Fingerprint": "58ca6"
+    },
+    "serviceAccountEmailAddress": "vb7e5otoij2hnwsi@speckle-umbrella-7.iam.gserviceaccount.com",
```
stfeng2: use a mock up for this serviceAccountEmailAddress.

tests/data/gcloud-describe-db-instances-available.json
sspano: Yes.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: Okay will do.

perfkitbenchmarker/providers/gcp/gcp_managed_relational_db.py
sspano: String parsing it is.

tests/data/gcloud-describe-db-instances-available.json
sspano: Done.

stfeng2:
Last two minor comments, otherwise LGTM.

After you address these comments, I will let Gareth do the final merge (because this is his branch. ;)

stfeng2:
@gareth-ferneyhough I've approved this. Please feel free to merge after your approval. Thanks!

PR #1436, opened 2017-07-12
[launch_mysql_service] 

sspano:
[launch_mysql_service] Can drive a benchmark phase (prepare,provision or run a pre-provisioned instance) instead of a full run. Can declare db instance and client vm specifications such as boot disk size.

@stfeng2 @gareth-ferneyhough


perfkitbenchmarker/scripts/database_scripts/launch_mysql_service.py
```diff
@@ -115,15 +160,23 @@ def driver(argv):
...
-  _run(run_uri)
-  if not FLAGS.run_uri:
+    if run_stage == 'provision,prepare':
+      print run_uri
+      return run_uri
+  if not run_stage or run_stage == RUN:
+    _run(run_uri)
+  if not run_stage or run_stage == 'cleanup,teardown':
     _cleanup_teardown_pkb(run_uri)
+  print(run_uri)
```
gareth-ferneyhough: remove print, or specify what you are printing

PR #1439, opened 2017-07-14
Add model-specific default batch sizes for some models

gareth-ferneyhough:


perfkitbenchmarker/linux_benchmarks/tensorflow_benchmark.py
```diff
@@ -91,6 +103,17 @@ def GetConfig(user_config):
   return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
 
 
+def _GetDefaultBatchSizeByModel(model):
+  try:
+    return DEFAULT_BATCH_SIZES_BY_MODEL[model]
```
tohaowu: we can also wirte like this DEFAULT_BATCH_SIZES_BY_MODEL.get(model, DEFAULT_BATCH_SIZE)

PR #1442, opened 2017-07-18
Log stack trace

dlott:


perfkitbenchmarker/vm_util.py
```diff
@@ -249,7 +249,8 @@ def WrappedFunction(*args, **kwargs):
             raise e
           else:
             if log_errors:
-              logging.error('Got exception running %s: %s', f.__name__, e)
+              logging.error('Got exception running %s: %s', f.__name__, e,
```
ehankland: Maybe just logging.exception()?
https://docs.python.org/2/library/logging.html#logging.Logger.exception

asaksena: Can this be made optional ? something akin to setting a log level. The whole stack trace may not be of interest in certain expected cases.

dlott: How about this version - full stack is only printed after you've timeout/max-retried.  So if it succeeds, the logs are small. If it fails, you have what you need to debug (eventually).  And no configs to muck with.

perfkitbenchmarker/vm_util.py
```diff
@@ -246,10 +246,10 @@ def WrappedFunction(*args, **kwargs):
           sleep_time = poll_interval * fuzz_multiplier
           if ((time.time() + sleep_time) >= deadline or
               (max_retries >= 0 and tries > max_retries)):
-            raise e
+            raise
```
asaksena: Sorry, I dont fully understand the implications of this change. Ideally we would want to raise the same root cause exception without masking it, and make it clear what we are doing. Does removing the exact exception have any effect ?

perfkitbenchmarker/vm_util.py
```diff
@@ -246,10 +246,10 @@ def WrappedFunction(*args, **kwargs):
...
           else:
             if log_errors:
-              logging.error('Got exception running %s: %s', f.__name__, e)
+              logging.error('Retrying exception running %s: %s', f.__name__, e)
```
asaksena: Is this simply a wording change ?

perfkitbenchmarker/vm_util.py
dlott: http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html

Basically `raise e` gets a stack trace rooted here, whereas `raise` gives you the original stack trace, which is where you actually need to go fix something.

perfkitbenchmarker/vm_util.py
dlott: Yes, to make it clearer that you're in the retry logic.

perfkitbenchmarker/vm_util.py
asaksena: Sorry, I am interested because I see this pattern being used in spanner work, and I thought 'e' was the original exception. Is that not the case. Will 'e' and the original exception from the function being retried ever differ ? If so I may have to request some changes elsewhere. Thanks for the link

perfkitbenchmarker/vm_util.py
asaksena: Cool

perfkitbenchmarker/vm_util.py
dlott: I am not a python expert, but from reading in https://docs.python.org/2/library/logging.html#logger-objects, it may be that the `e` is the same but the `exc_info` that is logged by the higher catch has been updated to the second raise site?

asaksena: Good trace.

asaksena:
LGTM

PR #1453, opened 2017-07-27
Add support for p100 gpus. Changed default clock speeds to base, rather

gareth-ferneyhough:
than max clocks.


perfkitbenchmarker/linux_packages/cuda_toolkit_8.py
```diff
@@ -51,21 +70,62 @@ class NvidiaSmiParseOutputException(Exception):
...
+  stdout, _ = vm.RemoteCommand('nvidia-smi -L', should_log=True)
+  try:
+    gpu_types = [line.split(' ')[3] for line in stdout.splitlines() if line]
+  except:
+    raise NvidiaSmiParseOutputException('Unable to parse gpu type')
+  if gpu_types[1:] != gpu_types[:-1]:
+    raise HeterogeneousGpuTypesException(
+        'PKB only supports one type of gpu per VM')
+
+  if gpu_types[0].find('P100') != -1:
```
ehankland: Think it might be more readable to use 'in' instead of find - especially because you don't need the index

perfkitbenchmarker/linux_packages/cuda_toolkit_8.py
```diff
@@ -51,21 +70,62 @@ class NvidiaSmiParseOutputException(Exception):
...
+  Raises:
+    NvidiaSmiParseOutputException: if nvidia-smi output cannot be parsed
+    HeterogeneousGpuTypesException: if more than one gpu type is detected
+  """
+  stdout, _ = vm.RemoteCommand('nvidia-smi -L', should_log=True)
+  try:
+    gpu_types = [line.split(' ')[3] for line in stdout.splitlines() if line]
+  except:
+    raise NvidiaSmiParseOutputException('Unable to parse gpu type')
+  if gpu_types[1:] != gpu_types[:-1]:
```
ehankland: I think
```
if  any(type != gpu_types[0] for type in gpu_types):
```
might be more readable

perfkitbenchmarker/linux_packages/cuda_toolkit_8.py
gareth-ferneyhough: yea, i always forget about that. will do.

perfkitbenchmarker/linux_packages/cuda_toolkit_8.py
gareth-ferneyhough: agree

PR #1467, opened 2017-08-09
Allow newer versions of OpenStack

charliejllewellyn:
The version of the openstack client that is pinned here breaks when running against our cloud. We should allow newer versions which have many bug fixes applied.


charliejllewellyn:
I signed it!

meteorfox:
Looks good. Which OpenStack release did you test this against? As long as it works with the 'Maintained' releases, in this case Newton, should be OK. My general advice is that If there are breaking changes between two versions, we should try to accommodate both until the old one goes into EOL.

charliejllewellyn:
Hey, this has been run against Mitaka and Newton

gareth-ferneyhough:
Carlons, is this still good with you since it works with Newton?

PR #1469, opened 2017-08-14
Testdfsio benchmark on dpb

asaksena:
Commit of the testdfsio benchmark and implementation of read on emr and dataproc dpb implementations.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
@@ -74,53 +74,49 @@ SUPPORTED_DPB_BACKENDS = [dpb_service.DATAPROC, dpb_service.EMR]
 
 
 def GetConfig(user_config):
-    return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
+  return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
 
 
 def CheckPrerequisites(benchmark_config):
-    """Verifies that the required resources are present.
+  """Verifies that the required resources are present.
 
     Raises:
-      perfkitbenchmarker.errors.Config.InvalidValue: On encountering invalid
-      configuration.
-    """
-    dpb_service_type = benchmark_config.dpb_service.service_type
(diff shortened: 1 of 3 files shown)


perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
```diff
@@ -0,0 +1,162 @@
...
+          machine_type: n1-standard-4
+          boot_disk_size: 1500
+        AWS:
+          machine_type: m4.xlarge
+      disk_spec:
+        GCP:
+          disk_type: nodisk
+        AWS:
+          disk_size: 1500
+          disk_type: gp2
```
s-deitz: Do we typically specify these disks differently? That is, is this the AWS boot disk?

perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
```diff
@@ -0,0 +1,162 @@
...
+        update_source_default_fs = True
+
+    source_dir = '{}{}'.format(source, '/dfsio')
+
+    combos = ((file_size, num_files)
+              for file_size in FLAGS.dfsio_file_sizes_list
+              for num_files in FLAGS.dfsio_num_files_list)
+
+    results = []
+    for file_size, num_files in combos:
```
s-deitz: Suggestion. Instead of pre-computing combos (which is not used elsewhere), just use a nested loop.
https://google.github.io/styleguide/pyguide.html?showone=List_Comprehensions#List_Comprehensions

perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
asaksena: I cannot speak to other resources but for dpb in particular, the difference in how the worker node disk (type and size) is specified for the different clouds is due to the difference in their cli specifications. Dataproc allows direct specification of boot disks which the workers use for local computation while EMR requires EBS specification attached to worker's instance group.

perfkitbenchmarker/linux_benchmarks/dpb_testdfsio_benchmark.py
asaksena: Done.

PR #1471, opened 2017-08-15
[database_scripts] Update per second graph functionality

sspano:
Added try/catch around _execute_pkb_cmd so per second graph functionality isn't affected by 512 thread failure.
This change might lead to per second graphs having a different amount of thread counts (iterations). Added this functionality as well.
@gareth-ferneyhough @stfeng2


stfeng2:
LGTM, merging...

PR #1472, opened 2017-08-16
Fix PrepareVMEnvironment to work on container-based images

SaiedKazemi:
This commit makes changes to PrepareVMEnvironment() to work with
ubuntu_container operating system types (like COS).  It also adds
a flag (--gce_leave_running) to leave the GCE VM running for
troubleshooting.

Signed-off-by: Saied Kazemi <saied@google.com>

### DIFF
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -925,10 +925,17 @@ class ContainerizedDebianMixin(DebianMixin):
 
   def PrepareVMEnvironment(self):
     """Initializes docker before proceeding with preparation."""
-    self._CreateVmTmpDir()
     if not self._CheckDockerExists():
       self.Install('docker')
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/resource.py
```diff
@@ -166,6 +170,10 @@ def WaitUntilReady():
 
   def Delete(self):
     """Deletes a resource and its dependencies."""
+    if FLAGS.gce_leave_running:
```
ehankland: GCE specific stuff doesn't belong here since the resource class should be provider agnostic. If you want to leave resources running, you can simply specify --run_stage=provision,prepare,run (and omit cleanup,teardown). Let me know if this approach doesn't work for your use case.

SaiedKazemi:
@ehankland I reverted the changes for gce_leave_running flag.  Now, the only change in this commit is the fix to PrepareVMEnvironment.  Please take another look.

PR #1476, opened 2017-08-18
Fixing a minor bug during the bucket creation prep step for a single …

asaksena:
…dfsio iteration. Without this fix the subsequent iterations will revert to using non regional buckets, even if the zones is specified.


asaksena:
Just a heads up. Committing this bug fix.

PR #1477, opened 2017-08-20
Make YCSB_WORKLOAD_DIR_NAME path dynamic

kirecek:
When using `pkb.py` outside of repository root, *FileResourceLoader* is unable to find `/data/ycsb` since it's defined as relative path.

This PR updates `YCSB_WORKLOAD_DIR_NAME` path's base to perkitbenchamarker package location.


gareth-ferneyhough:
LGTM

PR #1486, opened 2017-08-25
Set up host firewall on VMs running a COS image

SaiedKazemi:
By default, the Container-Optimized OS (COS) host firewall allows only
outgoing connections and incoming SSH connections. To allow other incoming
connections, we must open the ports.

This commit uses gcloud commands to get the VM's subnet prefix and adds
iptables rules on the host to accept connections on all ports.

Signed-off-by: Saied Kazemi <saied@google.com>

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
@@ -27,6 +27,7 @@ import csv
 import io
 import json
 import logging
+import re
 
 from collections import Counter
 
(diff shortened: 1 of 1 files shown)


SaiedKazemi:
@ehankland Based on your suggestion, I reverted linux_virtual_machine.py and moved a much simpler version of the code to netperf_benchmark.py.  Successfully tested it with both internal and external IP addresses.  Please let me know what you think.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -133,6 +144,30 @@ def Prepare(benchmark_spec):
...
+
+  Args:
+    benchmark_spec: The benchmark specification. Contains all data that is
+        required to run the benchmark.
+  """
+
+  client_vm = benchmark_spec.vms[0]
+  server_vm = benchmark_spec.vms[1]
+
+  if vm_util.ShouldRunOnExternalIpAddress():
```
ehankland: Running on external ips doesn't exclude internal ips - they can both be tested in the same run

SaiedKazemi:
@ehankland Please take another look.

Issue #1490, opened 2017-08-29
Update Azure object storage benchmark to use latest version.


s-deitz:
The azure_blob_storage API hard-codes an old version so that it continues to work. We should update to the latest version.

yuyantingzero:
Closing, feel free to open

PR #1491, opened 2017-08-29
Azure object storage

s-deitz:
This PR makes changes to the Azure provider so that the object storage service benchmark runs successfully. A future PR should be completed to update the Azure object storage API.

See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/1490


perfkitbenchmarker/linux_packages/azure_sdk.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2016 PerfKitBenchmarker Authors. All rights reserved.
```
ehankland: s/2016/2017/

s-deitz: Done

PR #1504, opened 2017-10-03
Adding MNIST benchmark that can run on TPU and CPU.

tohaowu:


perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
```diff
@@ -0,0 +1,155 @@
...
+
+BENCHMARK_NAME = 'mnist'
+BENCHMARK_CONFIG = """
+mnist:
+  description: Runs MNIST Benchmark.
+  vm_groups:
+    default:
+      vm_spec:
+        GCP:
+          image: debian-9-tf-v20171002
```
gareth-ferneyhough: does this require debian?

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
```diff
@@ -0,0 +1,155 @@
...
+          image_project: ml-images
+          machine_type: n1-standard-8
+          zone: us-central1-c
+  flags:
+    gcloud_scopes: https://www.googleapis.com/auth/cloud-platform
+"""
+
+flags.DEFINE_string('mnist_train_file',
+                    'gs://tfrc-test-bucket/mnist-records/train.tfrecords',
```
gareth-ferneyhough: Is this path public? If not, it shouldn't be in PKB.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
```diff
@@ -0,0 +1,155 @@
...
+        required to run the benchmark.
+
+  Returns:
+    A list of sample.Sample objects.
+  """
+  _UpdateBenchmarkSpecWithFlags(benchmark_spec)
+  vms = benchmark_spec.vms
+  master_vm = vms[0]
+  mnist_benchmark_dir = 'tpu-demos/cloud_tpu/models/mnist'
+  tpu_master = 'grpc://{}:8470'.format(benchmark_spec.tpu.tpu_ip)
```
gareth-ferneyhough: does this require grpc to be installed on the client vm?

perfkitbenchmarker/providers/gcp/gcp_tpu.py
```diff
@@ -0,0 +1,106 @@
...
+"""
+
+import json
+import logging
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import resource
+from perfkitbenchmarker.providers.gcp import util
+
+DEFAULT_CLOUD_SPANNER_CONFIG = 'regional-us-central1'
```
gareth-ferneyhough: Delete this

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
gareth-ferneyhough: Also, is there anyway we can be consistent and install tensorflow from source, as we do in the other tensorflow benchmarks?

perfkitbenchmarker/providers/gcp/gcp_tpu.py
```diff
@@ -0,0 +1,106 @@
...
+TPU_TIMEOUT = 1200
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('tpu_version', 'nightly',
+                    'The version for the Cloud TPU.')
+
+
+class GcpCloudTpu(resource.BaseResource):
```
gareth-ferneyhough: I'm considering creating an 'Accelerator' base resource so that we can specify both GPUs and TPUs in the same manner, but I'm not completely sold. If not, then we should add TPUs to GceVmSpec, so that we specify TPUs in the benchmark config, like we do with GPUs.

perfkitbenchmarker/providers/gcp/gcp_tpu.py
```diff
@@ -0,0 +1,106 @@
...
+  """class representing a GCP cloud TPU.
+
+  Attributes:
+    name: Name of the cloud TPU to create.
+    project: the GCP project.
```
gareth-ferneyhough: Would you ever create a TPU in a different project than what the rest of PKB is using?

perfkitbenchmarker/providers/gcp/gcp_tpu.py
```diff
@@ -0,0 +1,106 @@
...
+    self.version = FLAGS.tpu_version
+    self.zone = FLAGS.zones[0]
+    self.tpu_ip = '10.240.{}.2'
+
+  def ListIpAndCidr(self):
+    """List Cloud TPUs."""
+
+  def _Create(self):
+    """Create Cloud TPU."""
+    cmd = util.GcloudCommand(self, 'alpha', 'compute', 'tpus', 'list')
```
gareth-ferneyhough: Why are you listing TPUs before you create it?

perfkitbenchmarker/providers/gcp/gcp_tpu.py
```diff
@@ -0,0 +1,106 @@
...
+    stdout, _, retcode = cmd.Issue()
+    if retcode != 0:
+      logging.error('List GCP cloud TPU failed.')
+      return
+
+    result = json.loads(stdout)
+    tpu_ips_cidr = [tpu['ipAddress'] for tpu in result]
+    cmd = util.GcloudCommand(self, 'alpha', 'compute', 'tpus', 'create',
+                             self.name)
+    for i in range(2 ** 8):
```
gareth-ferneyhough: This is a little confusing. Can you clarify what you are doing here, or break it into a method so that it is more clear?

gareth-ferneyhough:
Also, please please add unit tests for each file, even if it is very basic. It makes it easier to make changes down the road.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
tohaowu: I want to use preinstall tensorflow images from ml-images. And currently, all the tf-nightly images is debian images

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
tohaowu: Yes, this path is public.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
tohaowu: I didn't install grpc. I think it default installed

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
tohaowu: I want to use preinstalled tensorflow image from ml-images. Now all the tf-nightly images is debian images in gcp

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: Thanks

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: I don't know. I think creating TPU is isolated with creating a GCE. 
Creating GCE using command "gcloud compute instances create"
Creating TPU using command "gcloud alpha compute tpus create"
If the TPU is created successfully, then any GCE VMs can use this TPU, just using the IP of this TPU.
That's why I created this TPU like this.

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: The GCE and TPU use the same same project.

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: Because TPU_IP=10.240.i.2, where i is an integer not already in use.
So I listed all the TPUs to find which one is not in use.

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: Thanks, I will break this into two functions.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
gareth-ferneyhough: Ok cool.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
gareth-ferneyhough: Alright, seems fine since this benchmark can only run on GCP.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
gareth-ferneyhough: Ok

perfkitbenchmarker/providers/gcp/gcp_tpu.py
gareth-ferneyhough: Ok, let's leave it out of GceVmSpec. I still think we should add it to the benchmark_config_spec to be consistent with PKB. I can help with this.

perfkitbenchmarker/providers/gcp/gcp_tpu.py
gareth-ferneyhough: Ok

perfkitbenchmarker/providers/gcp/gcp_tpu.py
tohaowu: Thanks. I will add it to benchmark_config_spec.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
```diff
@@ -0,0 +1,156 @@
...
+  description: Runs MNIST Benchmark.
+  vm_groups:
+    default:
+      vm_spec:
+        GCP:
+          image: debian-9-tf-v20171009
+          image_project: ml-images
+          machine_type: n1-standard-8
+          zone: us-central1-c
+  cloud_tpu:
```
gareth-ferneyhough: I notice that you don't specify tpu_accelerator_type here, and there is no default in the flags or the spec. What default is gcloud using? Maybe we should add a default value in PKB to be clear.

perfkitbenchmarker/cloud_tpu.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2017 PerfKitBenchmarker Authors. All rights reserved.
```
gareth-ferneyhough: This file needs a GetResourceMetadata method, which should be called by publisher.py.

gareth-ferneyhough: Additionally, the child classes can implement this method too, to add cloud-specific metadata.

perfkitbenchmarker/cloud_tpu.py
```diff
@@ -0,0 +1,96 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+
+from perfkitbenchmarker import flags
+from perfkitbenchmarker import resource
+
+
+flags.DEFINE_string('tpu_cidr_range', None, """CIDR Range for the TPU. The IP
```
gareth-ferneyhough: Is it reasonable to have a default here? Can you at least explain how to choose a value that works?

tohaowu: No, this is a required flag in tpu creating command.

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
tohaowu: The default tpu_accelerator_type is tpu-v2. If not specified in "gcloud compute tpus create", the default is tpu-v2. We can give a value in PKB. If we don't give, it will use the default tpu_accelerator_type value from this command "gcloud compute tpus create", which is tpu-v2

perfkitbenchmarker/cloud_tpu.py
tohaowu: Thanks

perfkitbenchmarker/linux_benchmarks/mnist_benchmark.py
gareth-ferneyhough: Ok, let's just be explicit and set the default flag value in PKB to tpu-v2.

PR #1507, opened 2017-10-05
Added cloudspanner.channels flag for GRPC channels

hostirosti:
The default channel count is set to 4 in the Spanner Java client library. For benchmarking a higher channel count is recommended, setting it here to 100 by default and adding the capability to change it via a runtime flag.


hostirosti:
@gareth-ferneyhough What is the best way to get PRs reviewed and merged into the repo?

PR #1514, opened 2017-10-11
add --network_enable_BBR flag to Perfkitbenchmarker

NathanTeeuwen:
When network_enable_BBR is added on the command line, PKB will check to
make sure the linux kernel is at least verion 4.9. During machine
prepare, before benchmaks are run, This flag will set

'net.core.default_qdisc = fq'
'net.ipv4.tcp_congestion_control=bbr'

by making changes to /etc/syctrl.conf and then performing a reboot.
Metadata now includes 'tcp_congestion_control' so we can track which
runs have BBR enabled.


NathanTeeuwen:
I signed it!

PR #1519, opened 2017-10-12
Fix temp dir used by object_storage_service_benchmark

s-deitz:
The temporary directory was found with temp_dir.GetRunDirPath rather
than vm_util.GetTempDir. The latter function uses an LRU cache in case
the run_uri changes. Without this fix, the temporary directory used to
store the list of objects written would be incorrect when using the
--object_storage_read_objects flag.

This appears to be the only usage of temp_dir.GetRunDirPath that does
not call it via vm_util.GetTempDir.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
@@ -49,7 +49,6 @@ from perfkitbenchmarker import flags
 from perfkitbenchmarker import flag_util
 from perfkitbenchmarker import object_storage_service
 from perfkitbenchmarker import sample
-from perfkitbenchmarker import temp_dir
 from perfkitbenchmarker import units
 from perfkitbenchmarker import vm_util
 from perfkitbenchmarker.providers.gcp import gcs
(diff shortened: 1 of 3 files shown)


s-deitz:
Good idea. I moved the caching to temp_dir.GetRunDirPath, but still changed the benchmark to call vm_util.GetTempDir. PTAL. I'm running a test that appears to be working.

perfkitbenchmarker/linux_benchmarks/object_storage_service_benchmark.py
```diff
@@ -1037,13 +1036,16 @@ def MultiStreamReadBenchmark(results, metadata, vms, command_builder,
...
+      # don't overwrite the same local file. They are pushed to the VM to a file
+      # named OBJECTS_WRITTEN_FILE.
+      tmp_objects_written_path = os.path.join(vm_util.GetTempDir(),
+                                              '%s-%s' % (OBJECTS_WRITTEN_FILE,
+                                                         vm.name))
       with open(tmp_objects_written_path, 'w') as objects_written_file:
         objects_written_file.write(json.dumps(vm_objects_written))
-      vm.PushFile(tmp_objects_written_path, '/tmp/pkb/')
+      vm.PushFile(tmp_objects_written_path,
+                  os.path.join('/tmp/pkb/', OBJECTS_WRITTEN_FILE))
```
ehankland: Looks like it was this way before this change, but I think using the vm_util.VM_TMP_DIR constant is preferred to '/tmp/pkb/' when referring to the temp directory on remote machines. Also, since this is a benchmark for linux VMs, it should use posixpath.join instead of os.path.join for paths on remote hosts since os.path.join will use "\\" if PKB is running on a windows machine

s-deitz: Done.

PR #1523, opened 2017-10-13
Add support to record numa_node_count in metadata. Also ... when retrying remote commands, retries appear as "INFO" intead of "ERROR" in the output.

NathanTeeuwen:
Add support to record numa_node_count in metadata. Also ... when retrying remote commands, retries appear as "INFO" intead of "ERROR" in the output


NathanTeeuwen:
FYI:  this is a backport from an internal change

PR #1529, opened 2017-10-17
Fix blazemark install within docker container.

HugoGuiroux:
The ubuntu docker image neither contains apt-add-repository nor wget needed
to install blazemark dependencies.


HugoGuiroux:
I signed it!

PR #1551, opened 2017-12-29
Feat/remote-json-publisher

### LINKED ISSUE #1550: Implement universal JSON publisher
For specific use cases, where user needs to get clean JSON data from results there is needed simple JSON publisher which will connect to remote endpoint and send results data using HTTP POST request. 

This is needed when developers wants to have different (custom) database schema in comparison to ones used in **InfluxDB** or **ElasticSearch** publishers, where there is a lack of standardization across schemas in different tests.

Lirt:
This PR implements universal remote JSON publisher which connects to URI given by argument `--remote_json_uri` and sends pure JSON data using HTTP POST request using python library `requests`. 

More info in #1550.

Closes #1550

### DIFF
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -127,6 +127,12 @@ flags.DEFINE_string(
     'influx_db_name', 'perfkit',
     'Name of Influx DB database that you wish to publish to or create')
 
+flags.DEFINE_string(
+    'remote_json_uri', None,
+    'URI to remote listener for JSON data. Expects format '
+    'hostname:port or format accepted by python requests library e.g. '
+    'http://localhost:9000/endpoint')
+
 DEFAULT_JSON_OUTPUT_NAME = 'perfkitbenchmarker_results.json'
 DEFAULT_CREDENTIALS_JSON = 'credentials.json'
 GCS_OBJECT_NAME_LENGTH = 20
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -468,6 +468,7 @@ class NewlineDelimitedJSONPublisher(SamplePublisher):
           sample['labels'] = GetLabelsFromDict(sample.pop('metadata', {}))
         fp.write(json.dumps(sample) + '\n')
 
+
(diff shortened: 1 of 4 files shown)


tests/publisher_test.py
```diff
@@ -104,6 +104,21 @@ def testJSONRecordPerLine(self):
                          result)
 
 
+class RemoteJSONPublisherTestCase(unittest.TestCase):
+
+  @mock.patch.object(publisher.RemoteJSONPublisher, 'PublishSamples')
```
s-deitz: Instead of mocking RemoteJSONPublisher, I'd suggest mocking requests and then asserting that the right post function is made with the expected sample output.

perfkitbenchmarker/publisher.py
```diff
@@ -18,7 +18,7 @@
 
 import abc
 import collections
-import copy
+from copy import copy, deepcopy
```
s-deitz: Use import for modules only. See this style guide that we try to follow: https://google.github.io/styleguide/pyguide.html#Imports

perfkitbenchmarker/publisher.py
```diff
@@ -33,6 +33,7 @@
 import time
 import urllib
 import uuid
+import requests
```
dlott: please alphabetize

Lirt: Updated
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1551/commits/9186b0390ec186544a0ec63755a60618800f1499

perfkitbenchmarker/publisher.py
Lirt: Updated https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1551/commits/b3edfe0eba294ec506f773b9a513fd353e0c3fcf

Lirt:
Hello again,
I was thinking about this PR with my colleague and we found out, that this wouldn't be very usable, because most of endpoints will be using some kind of authentication. There are many types of authentication and we cannot make this work universally, so I am closing this PR. 

Thank you for time spent on review.

Issue #1555, opened 2018-01-12
Dynamic_pipeline_options should allow to specify format.


szewi:
This will be very useful for performance testing of hadoop clusters and any other tests where complex options are needed. Format will define pattern that will be used to create final option. In example: ` hdfsConfiguration` option will be created by substitution `LoadBalancerIp` in pattern.

```
 static_pipeline_options:
 dynamic_pipeline_options:
  - name: hdfsConfiguration
    format: '[{\"fs.defaultFS\":\"hdfs://{{LoadBalancerIp}}:9000\",\"dfs.replication\":1}]'
    type: LoadBalancerIp
    serviceName: hadoop-external
```

yuyantingzero:
Seems like it is already resolved. Closing, feel free to reopen if more work needs to be done.

PR #1554, opened 2018-01-09
add --autodetect option for bq load

DariuszAniszewski:
This PR uses `--autodetect` option (suggested in #1552) to create schema based on uploaded data if table doesn't exists yet. 

It's also seem to be connected to #229. This PR doesn't follow idea from that issue, but final effect is the same - if table doesn't exists, it's being created. Advantage is that it doesn't require additional information and creates schema based on the data.

PR #1556, opened 2018-01-12
added support for custom format for dynamicPipelineOptions

DariuszAniszewski:
this PR solves #1555. It allows PerfKit to return dynamic pipeline option with custom, user-defined format.

### DIFF
--- a/perfkitbenchmarker/beam_pipeline_options.py
+++ b/perfkitbenchmarker/beam_pipeline_options.py
@@ -46,13 +46,16 @@ def EvaluateDynamicPipelineOptions(dynamic_options):
   """
   Takes the user's dynamic args and retrieves the information to fill them in.
 
-  dynamic_args is a python map of argument name -> {type, kubernetesSelector}
+  dynamic_args is a python map of argument name -> {type, kubernetesSelector, *format}
   returns a list of tuples containing (argName, argValue)
+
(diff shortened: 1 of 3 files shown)


szewi:
Hi, we need this to create complex pipeline options in nice way using PerfKit. This change doesn't brake anything just adds new functionality to beam pipelineoptions parser. 
Example of usage could be seen in `pkb-config.yml` file in [#4401](https://github.com/apache/beam/pull/4401). 
@asaksena @jasonkuster could you help us with that ?

asaksena:
https://travis-ci.org/GoogleCloudPlatform/PerfKitBenchmarker/jobs/328084429 is failing, can you please fix this after which I can merge.

DariuszAniszewski:
@asaksena I found the problem with builds on Travis. It was dependency problem - `numpy` version wasn't fixed and there was recently a bigger update 1.14. I forced older version (1.13.3) and builds are now passing.

The same story is in my other PR #1554 wich was failing due to same issue. After fixing numpy dependency it's green.

This however seems more critical, as not only tests but other parts of system might fail on this. I'm going to file an issue about `numpy 1.14` support and separate PR for fixing the version until full compatibility is proven.

chamikaramj:
@asaksena appreciate if you can merge this.

Issue #1558, opened 2018-01-22
numpy v1.14 compatibility


DariuszAniszewski:
[NumPy 1.14](https://github.com/numpy/numpy/releases/tag/v1.14.0) was released some time ago and it seems that current version of PerfKit isn't compatible with it.

There were some test failures, but it *might* be that problem is more serious. For now I propose to fix version of numpy at `1.13.3` which seems to work and keep it until v1.14 is proven to work.

DariuszAniszewski:
Update: Along with #1556 version of NumPy was set to 1.13.3

PR #1559, opened 2018-01-22
Fix NumPy version to 1.13.3

DariuszAniszewski:
Walk-around PR for #1558.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ blinker>=1.3
 futures>=3.0.3
 PyYAML==3.12
 pint>=0.7
-numpy
+numpy==1.13.3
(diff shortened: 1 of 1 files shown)


DariuszAniszewski:
@asaksena this is the PR I was talking in #1556 - it doesn't solve compatibility issues, but fixing versions in `requirements.txt` is good practise anyways.
PTAL, thanks

asaksena:
I think this can be closed now ?

DariuszAniszewski:
@asaksena in general yes, fix to NumPy was already merged with #1556.

But I'd suggest to investigate NumPy v1.14 - there seems to be some problem with it.

PR #1561, opened 2018-02-13
Moe writing branch from f15cc343183cd1dfe85054ab55b5014b1c97fc7d

s-deitz:


perfkitbenchmarker/linux_benchmarks/sysbench_benchmark.py
```diff
@@ -356,8 +352,6 @@ def Prepare(benchmark_spec):
   prepare_results = _PrepareSysbench(vm, metadata, benchmark_spec)
   print prepare_results
```
dlott: Are these just lost?  We probably should not be printing them regardless.

s-deitz: The assignment to the global was not doing anything. Printing the data could still be useful. Nathan, any comment?

dlott:
print is usually for debug. if it's useful to see the data it should be
logged.

PR #1461, opened 2017-08-01
Cloudspanner reuse instance and database

larryxiao:
Separate cloud spanner instnace lifecycle from perfkit resource lifecycle.

PR #1562, opened 2018-02-16
Updates for YCSB on Cloud Spanner

hostirosti:
- Added additional meta-data for publisher
- Enabled reuse of preloaded DB
- Added GCE Instance Reset method
- Updated GCE Ubuntu image to 'ubuntu-16-04'

Includes changes from #1461

### DIFF
--- a/perfkitbenchmarker/linux_packages/ycsb.py
+++ b/perfkitbenchmarker/linux_packages/ycsb.py
@@ -95,7 +95,7 @@ flags.DEFINE_boolean('ycsb_include_individual_results', False,
                      'Include results from each client VM, rather than just '
                      'combined results.')
 flags.DEFINE_boolean('ycsb_reload_database', True,
-                     'Reload database, othewise skip load stage. '
+                     'Reload database, otherwise skip load stage. '
                      'Note, this flag is only used if the database '
                      'is already loaded.')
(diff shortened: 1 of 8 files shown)


perfkitbenchmarker/linux_packages/ycsb.py
```diff
@@ -771,10 +769,9 @@ def Load(self, vms, workloads=None, load_kwargs=None):
     workloads = workloads or _GetWorkloadFileList()
     load_samples = []
     assert workloads, 'no workloads'
-    if FLAGS.ycsb_reload_database or not self.loaded:
```
asaksena: If the flag value for ycsb_reload_database is set to be false, will this change mean that the ycsb load is never triggered ?

hostirosti: Correct.

The use case is if you want to run multiple tests against a DB that takes very long to load with different benchmark setups (number of machines, location of machines, etc.) reloading the DB everytime in this case is just not feasible.

asaksena: I discussed this internally, I think this will cause issues when the db is not loaded at the onset, and but no reload is required. Please support that behavior in your change.

Issue #1570, opened 2018-02-26
Running PerfKitBenchmarker on RHEL fails on STDERR: /usr/bin/time: unrecognized option '--quiet'


akrzos:
Running on RHEL 7, with rpm time-1.7-45.el7.x86_64 install, /usr/bin/time does not include a --quiet option which results in failure to run PerfKitBenchmarker

```
(undercloud) [stack@undercloud ~]$ /usr/bin/time --help
Usage: /usr/bin/time [-apvV] [-f format] [-o file] [--append] [--verbose]
       [--portability] [--format=format] [--output=file] [--version]
       [--help] command [arg...]
(undercloud) [stack@undercloud ~]$ /usr/bin/time --quiet
/usr/bin/time: unrecognized option '--quiet'
Usage: /usr/bin/time [-apvV] [-f format] [-o file] [--append] [--verbose]
       [--portability] [--format=format] [--output=file] [--version]
       [--help] command [arg...]
(undercloud) [stack@undercloud ~]$ /usr/bin/time --version
GNU time 1.7
```

PerfKitBenchmarker output snippet:
```
2018-02-26 08:42:46,043 browbeat Thread-1 fio(1/1) os_virtual_machine.py:153 INFO     Testing OpenStack CLI command is installed and working
2018-02-26 08:42:46,043 browbeat Thread-1 fio(1/1) vm_util.py:295 INFO     Running: openstack image list --format json
2018-02-26 08:42:46,047 browbeat Thread-1 fio(1/1) vm_util.py:344 INFO     Ran: {openstack image list --format json}  ReturnCode:1
STDOUT: 
STDERR: /usr/bin/time: unrecognized option '--quiet'
Usage: /usr/bin/time [-apvV] [-f format] [-o file] [--append] [--verbose]
       [--portability] [--format=format] [--output=file] [--version]
       [--help] command [arg...]
```

akrzos:
To add to this, it only occurs in time is installed as you can see in the code, [vm_util](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/vm_util.py#L302) won't use the time command if it is not installed.

dlott:
Thanks for the updated fix @akrzos!

dlott:
Fix was in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/1571 since all the above references seem to be commits.

PR #1571, opened 2018-02-26
Fix for #1570

akrzos:
Create flag for timing issued commands.  Disable it if you machine has
time installed which does not include the --quiet flag.


akrzos:
@dlott Ok changes made.

dlott:
Manually flipping the CLA back on since @akrzos and I are good to commit.

PR #1588, opened 2018-03-09
fix remote destination copy path for k8s provider

zmgit:


zmgit:
I signed it!

perfkitbenchmarker/providers/kubernetes/kubernetes_virtual_machine.py
```diff
@@ -391,7 +391,7 @@ def RemoteHostCopy(self, file_path, remote_path='', copy_to=True):
       RemoteCommandError: If there was a problem copying the file.
     """
     if copy_to:
-      src_spec, dest_spec = file_path, '%s:' % (self.name,)
+      src_spec, dest_spec = file_path, '%s:%s' % (self.name, remote_path)
```
ehankland: This won't actually change the cp command if remote_path is an empty string - I think instead of remote_path we probably want 
```
file_name = posixpath.basename(file_path)
src_spec, dest_spec = file_path, '%s:%s' % (self.name, file_name)
```
It ends up getting moved to remote path below (line 407-411)

ehankland:
Thanks for fixing this

Issue #1595, opened 2018-03-14
Increase MAX_NUM_WAITS_FOR_K8S_GET value


lgajowy:
It is used by beam benchmarks when trying to obtain certain Kubernetes resources - for example LoadBalancerIP from a service. The IP address sometimes is available after more than 60 seconds which makes the number of trials insufficient.

Let's increase the number of trials - it is harmless (it only gets resources).

PR #1596, opened 2018-03-14
Increase number of kubernetes resource retires

lgajowy:
This fixes [1595](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/1595) issue. 

@asaksena could you take a look?


lgajowy:
Wow, that was fast... Thanks! :)

PR #1597, opened 2018-03-15
Fix yum proxy bug

kopecpiotr:
This removes unnecessary quotation marks and semicolon. Before that Perfkit fails with yum error
```
Options error: Error parsing "proxy = '\'"http://10.10.10.1:8080";\''": URL must be http, ftp, https, socks4, socks4a, socks5 or socks5h not ""
```
We don't need those quotation marks and a semicolon at the end in ```/etc/yum.conf```


kopecpiotr:
I have used wrong e-mail in commits, let me fix that.

PR #1624, opened 2018-04-03
Update Elasticsearch publisher to work with current version of Elasticsearch

dPhanekham:
Updated ElasticsearchPublisher in publisher.py to support current version of Elasticsearch. Also maintains backwards compatibility to older versions of Elasticsearch. 

The problem was that after Elasticsearch 5.0.0, they removed the 'string' type and replaced it with the types 'text' and 'keyword', as explained in this [blog post](https://www.elastic.co/blog/strings-are-dead-long-live-strings)

This means the mappings needed to be changed slightly to work in newer versions, but I kept in the old mappings as well in case anyone wants to use an older version of elasticsearch for whatever reason.

### DIFF
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -675,7 +675,7 @@ class ElasticsearchPublisher(SamplePublisher):
       else:
         es.indices.create(index=self.es_index, body=self.mapping_old)
         logging.info('Create index %s and default mappings for'
-                     ' elasticsearch version < 5.0.0', 
+                     ' elasticsearch version < 5.0.0',
                      self.es_index)
     for s in samples:
       sample = copy.deepcopy(s)
--- a/perfkitbenchmarker/publisher.py
+++ b/perfkitbenchmarker/publisher.py
@@ -609,6 +609,37 @@ class ElasticsearchPublisher(SamplePublisher):
                     "strings": {
                         "match_mapping_type": "string",
                         "mapping": {
+                            "type": "text",
+                            "fields": {
+                                "raw": {
+                                    "type": "keyword",
+                                    "ignore_above": 256
+                                }
+                            }
+                        }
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/publisher.py
```diff
@@ -593,6 +593,37 @@ def __init__(self, es_uri=None, es_index=None, es_type=None):
     self.es_index = es_index.lower()
     self.es_type = es_type
     self.mapping = {
```
yuyantingzero: Clarify name of the mappings, e.g. mapping_5_plus?

perfkitbenchmarker/publisher.py
```diff
@@ -593,6 +593,37 @@ def __init__(self, es_uri=None, es_index=None, es_type=None):
...
+                                }
+                            }
+                        }
+                    }
+                }]
+            }
+        }
+    }
+
+    self.mapping_old = {
```
yuyantingzero: Clarify naming, e.g. mapping_before_5?

perfkitbenchmarker/publisher.py
```diff
@@ -634,8 +665,18 @@ def PublishSamples(self, samples):
...
     es = Elasticsearch([self.es_uri])
     if not es.indices.exists(index=self.es_index):
-      es.indices.create(index=self.es_index, body=self.mapping)
-      logging.info('Create index %s and default mappings', self.es_index)
+      # choose whether to use old or new mapings based on
+      # the version of elasticsearch that is being used
+      if int(es.info()['version']['number'].split('.')[0]) >= 5:
+        es.indices.create(index=self.es_index, body=self.mapping)
+        logging.info('Create index %s and default mappings for'
+                     ' elasticsearch version > 5.0.0',
```
yuyantingzero: \>= 5.0.0

PR #1647, opened 2018-04-19
Migrate Beam benchmark to use Gradle for Java runs

lgajowy:
Since there is an ongoing maven -> gradle migration in beam, the mvn commands located in beam_benchmark_helper.py should be migrated to gradle's equivalents. 

More on that matter here: https://issues.apache.org/jira/browse/BEAM-3942. The beam's part was done, now it's time to change what's necessary in PerfKit.

@asaksena could you take a look? If you don't mind, I'd also like @DariuszAniszewski to take a look before merging this, as he also works very closely with this benchmark.


lgajowy:
Sorry, this PR was incorrect. I'll submit another one.

Issue #1567, opened 2018-02-23
Azure Stack PerfKitBenchmarker


gsee2000:
Hi All,

Possibly not the right place, but as a question to the community, I wondered if anyone had made a branch or version of this for Azure Stack?

Cheers

G

s-deitz:
We are not aware of any branch for Azure Stack. If anybody is interested in doing so, we'd be happy to fold this into PKB.

gsee2000:
I have now created and put in a merge request for the Azure Stack provider. Currently awaiting this to be verified, and hopefully approved!

PR #1652, opened 2018-04-20
Azurestack provider

gsee2000:
Off the back of:

https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/1567#issuecomment-368158444

I have created an azurestack provider branch. Please can we publish this, so others in the community are able to add to, or improve. Cheers G


perfkitbenchmarker/linux_benchmarks/tensorflow_benchmark.py
```diff
@@ -97,6 +97,9 @@
 flags.DEFINE_boolean('tf_distortions', True,
                      '''Enable/disable distortions during image preprocessing.
                      These include bbox and color distortions.''')
+flags.DEFINE_string('tf_benchmarks_commit_hash',
+                    'bab8a61aaca3d2b94072ae2b87f0aafe1797b165',
+                    'git commit hash of desired tensorflow benchmark commit.')
```
s-deitz: I'm not sure what's going on here with the diff. Odd merge or sync?

perfkitbenchmarker/providers/azurestack/azurestack_virtual_machine.py
```diff
@@ -0,0 +1,446 @@
...
+                                        is_image=True)
+
+  def _CreateDependencies(self):
+    """Create VM dependencies."""
+    self.public_ip.Create()
+    self.nic.Create()
+
+  def _Create(self):
+    create_cmd = (
+        [azurestack.AZURESTACK_PATH, 'vm', 'create', '--use-unmanaged-disk',
```
s-deitz: There are very few changes here as compared to azure_virtual_machine. This is one such change where we have the extra argument --use-unmanged-disk.

We're going to have trouble keeping this in sync with azure_virtual_machine unless we're able to share more code, or make this inherit from azure virtual machine.

gsee2000:
Thank you for coming back to me. More than happy to expand a bit more, apologise for not detailing this earlier. 

So our current use case for this is; we have a multi cloud environment consisting of OpenStack, vCloud, Oracle and AzureStack. Using a tool such as this has allowed us to benchmark our various platforms, firstly against each other, so that could be platform vs platform, region vs region, site vs site, etc….. We are also using it to compare against the likes of public cloud providers such as Azure, AWS, etc…… 

Going forward, and in the community, I envisage this being a very useful tool for providers of private AzureStack offerings to performance test and capacity manage effectively. The tool would allow them to see performance from day one, and allow them to trend from there on once more workloads come on. There are of course different make ups of Azure Stacks, different vendors, cpu and ram amounts, storage types (all flash vs hybrid). If a provider was to have different offering levels, this would also help in producing stats on how much the offerings differ. As you suggested, it would also allow clouds vs azure hybrid deployments. Hope this clears it up a little more, let me know if you require any more on this.

perfkitbenchmarker/linux_benchmarks/tensorflow_benchmark.py
gsee2000: Regarding the merge, I am unsure on this, and happy to re upload if required.

perfkitbenchmarker/providers/azurestack/azurestack_virtual_machine.py
gsee2000: That is correct that there isn’t too much in the way of change against the current Azure provider. Due to Azure Stacks nature of it being a scaled down version of public Azure, it does have its limitations. I say this in the ability to port command sets from Azure to Azure Stack. Currently Azure CLI expects a certain profile to be present when interacting with Azure Stack:

https://docs.microsoft.com/en-us/cli/azure/?view=azure-cli-2017-03-09-profile

Azure Stack also lacks the complete and full functionality of public Azure, so like you picked up on, the use of ‘managed disks’ is not present in Azure Stack. So this needs to be tweaked to explicitly state --use-unmanaged-disk, as you can appreciate there are many other commands that expect things in a different order. Or are simply not yet present. As Azure Stack catches up more and functions become available, I expect people to be able to contribute and update the AzureStack provider to include/remove as it matures.

dlott: I agree with Steve that refactoring to share more code, probably inheriting from azure virtual machine and perhaps overriding unsupported functions is better than two parallel providers.

gsee2000: Thanks @dlott, I agree this approach would be far better. Is this something you are able to fold in or compile to the current Azure provider?

perfkitbenchmarker/linux_benchmarks/tensorflow_benchmark.py
gsee2000: Ok, let me know if you would like me to attempt this again

perfkitbenchmarker/providers/azurestack/azurestack_virtual_machine.py
s-deitz: I'm not sure I follow your question, but folding this in to the current Azure provider seems reasonable. If there aren't too many non-trivial changes, could you add a flag to the Azure provider that would switch to using Azure Stack instead of Azure?

gsee2000: Would you be able to provide an example of this, and I will look to amend accordingly?

PR #1657, opened 2018-04-26
Fix Kubernetes StorageClass deletion

AndriiIlika:
Use _kubectl --ignore-not-found_ flag to delete resources.
In particular, this fixes _copy_throughput_ benchmark on Kubernetes which creates the same StorageClass twice and gets stuck trying to delete it the second time.


gareth-ferneyhough:
LGTM, thanks.

PR #1742, opened 2018-07-31
Cleanup and fix Beam bechmark.

markflyhigh:
Some cleanup and value fix.

- Cleanup maven related flags that's no longer used in Beam benchmarks.
- Specify default python benchmark to single one: `test_wordcount_it`, otherwise, some unnecessary pipelines will be invoked accidentally as well.
- Remove hard code of PythonSDK tar file path. As long as it's built under root of Python sdk, benchmark can find and use it for pipeline execution.


markflyhigh:
doubled checked CLA

markflyhigh:
signed CLA

markflyhigh:
@asaksena can you please take a look at this? thanks!

PR #1753, opened 2018-08-10
Moe writing branch from 208265278

NathanTeeuwen:


NathanTeeuwen:
Confirmed with flint via chat that his changes are ok to push.

PR #1795, opened 2018-10-17
Update requirements.txt

voellm:
Adding missing dependency for mac


s-deitz:
Thanks for the PR. Can you describe why this is necessary for Mac? I tried adding this to the requirements.txt section on my workstation, but I see a segfault when I run the command:
pip install -r "/tmp/github/PerfKitBenchmarker/requirements.txt"

PR #1806, opened 2018-11-02
Improve Beam benchmark logging

markflyhigh:
Add few commandline flags to enable detail logging of Gradle run.

We want this fix in to help us triage current benchmark failures.


markflyhigh:
+R: @asaksena

PR #1818, opened 2018-11-26
Import latest via MOE

dlott:


dlott:
Trying again to get a cleaner merge.

PR #1929, opened 2019-06-10
Add Docker Provider type

dPhanekham:
Useful for container to container testing and on-prem to cloud testing

Includes support for

Docker as a provider type
Docker volumes as scratch disks
Local Docker image builds
I will gladly modify or add anything that is needed.


perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+    self._DeleteVolumes()
+
+  def _Create(self):
+    """Create a Docker instance"""
+
+    # Locally build docker container
+    with self.docker_build_lock:
+      image_exists = self._LocalImageExists(self.container_image)
+      if image_exists is False:
+        self._BuildImageLocally()
```
s-deitz: Is the docker image read-only? I'm curious about the case where we might have two Docker virtual machines. Is there any reason to support that?

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+
+    remove_command = ['docker', 'rm', self.name]
+    output = vm_util.IssueCommand(remove_command)
+    logging.info(output[STDOUT].rstrip())
+
+    return
+
+  @vm_util.Retry(poll_interval=10, max_retries=10)
+  def _Exists(self):
+    """
```
s-deitz: There is a bit of inconsistency in the docstring style. See https://www.python.org/dev/peps/pep-0257/ for formatting of single line and multiline docstrings and when to use either. In particular, single line docstring comments should be on the same line as the """ delimiters.

A single line here might be:
"""Returns whether the container is up and running."""

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+    if there is any and a return code. 0
+    """
+    logging.info("Checking Container Information")
+    inspect_cmd = ['docker', 'inspect', self.name]
+    info, _, returnCode = vm_util.IssueCommand(inspect_cmd, suppress_warning=True)
+    info = json.loads(info)
+    return info, returnCode
+
+  def _ConfigureProxy(self):
+    """
```
s-deitz: Single line summary and then elaborate.

https://www.python.org/dev/peps/pep-0257/

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+  DEFAULT_IMAGE = UBUNTU_IMAGE
+
+  def ApplySysctlPersistent(self, sysctl_params):
+    """Override ApplySysctlPeristent function for Docker provider
+    Parent function causes errors with Docker because it shutdowns container
+    Args:
+      sysctl_params: dict - the keys and values to write
+    """
+    logging.warn("Sysctl flags NOT applied to Docker container. "
+                 "Please use --docker_sysctl_flags if sysctl need to be applied")
```
s-deitz: Ah, this is the override I was thinking of above. Can you use the same flag and disable applying it here?

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+      return True
+    logging.info("Image does not exist")
+    return False
+
+  def _BuildImageLocally(self):
+    """
+    Build Container Image Locally
+    """
+
+    # Dockerfiles located at PerfKitBenchmarker/docker/<containerImage>/Dockerfile
```
ehankland: we already have perfkitbenchmarker/data/docker/<containerImage>/ - probably don't want two directories so either move the existing image/related functions to the new directory or switch to the /data/ directory

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+    directory = os.path.dirname(
+        data.ResourcePath(os.path.join('docker', self.container_image, 'Dockerfile')))
+    build_cmd = [
```
ehankland: Would be nice if these build functions could be shared with https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/container_service.py#L321

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
dPhanekham: A docker image is read-only. This bit of code is to prevent a specific case where Docker image doesn't yet exist on the local machine and you are running a benchmark that uses two containers, then it will attempt to build the Docker image in two threads at the same time, leading to an untagged duplicate image that is never used and just taking up space.

My specific use case for using multiple docker containers was when I was developing/testing a networking benchmark and containers were faster to test on. Other than testing, inter-container communication could be a reasonable thing to benchmark

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+from perfkitbenchmarker.providers.docker import docker_resource_spec
+from perfkitbenchmarker.vm_util import OUTPUT_STDOUT as STDOUT
+
+FLAGS = flags.FLAGS
+
+UBUNTU_IMAGE = 'ubuntu:xenial'
+DEFAULT_DOCKER_IMAGE = 'pkb/ubuntu16_ssh'
+
+
+class DockerVirtualMachine(virtual_machine.BaseVirtualMachine):
```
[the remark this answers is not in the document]
dPhanekham: I think this is a good idea. I've switched to a naming convention more like the Mesos provider.

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
```diff
@@ -0,0 +1,303 @@
...
+    self.cpus = vm_spec.docker_provider_cpus
+    self.memory_mb = vm_spec.docker_provider_memory_mb
+    self.privileged = vm_spec.privileged_docker
+    self.container_image = DEFAULT_DOCKER_IMAGE
+    self.docker_sysctl_flags = ''
+    # apply flags
+    if FLAGS.docker_custom_image:
+      self.container_image = FLAGS.docker_custom_image
+    if FLAGS.docker_sysctl_flags:
+      self.docker_sysctl_flags = FLAGS.docker_sysctl_flags
```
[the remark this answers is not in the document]
dPhanekham: Switched to use --sysctl flag already present, updated dependent functions accordingly

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
dPhanekham: switched to use --sysctl flag. removed --docker_sysctl_flags

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
dPhanekham: moved Dockerfile into /perfkitbenchmarker/data/docker/

docker/pkb/ubuntu16_ssh/Dockerfile
```diff
@@ -0,0 +1,20 @@
+#Dockerfile to use with Docker provider
```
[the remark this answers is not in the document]
dPhanekham: moved to perfkitbenchmarker/data/docker

perfkitbenchmarker/providers/docker/docker_resource_spec.py
```diff
@@ -0,0 +1,53 @@
...
+  """
+
+  CLOUD = providers.DOCKER
+
+  @classmethod
+  def _GetOptionDecoderConstructions(cls):
+    result = super(DockerSpec, cls)._GetOptionDecoderConstructions()
+    result.update({
+        'docker_provider_cpus': (option_decoders.FloatDecoder, {'default': 0}),
+        'docker_provider_memory_mb': (option_decoders.IntDecoder, {'default': 0}),
```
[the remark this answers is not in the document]
dPhanekham: I have removed these flags in favor custom_virtual_machine_spec stuff. Can now use --machine_type flag to supply custom cpu/memory.

perfkitbenchmarker/providers/docker/docker_virtual_machine.py
dPhanekham: I agree that using those build functions would be nice, and I looked into it when I first wrote this. But these are part of BaseContainerRegistry, and the Docker provider type doesn't really have a registry since the image it uses is just built locally.

dPhanekham:
Thank you for all of your feedback. It has been very useful for improving my code. I think I have now addressed all of your comments. For the simpler comments I just made the changes, and for others I both made changes and responded to the comment directly. Let me know if there are any further changes that need to be made.

perfkitbenchmarker/providers/docker/docker_container_instance.py
```diff
@@ -0,0 +1,319 @@
...
+  def _RebootIfNecessary(self):
+    """Override RebootIfNecessary for Docker Provider"""
+    logging.warn("Docker Containers cannot be rebooted to apply flags")
+
+
+class Ubuntu1604BasedDockerContainer(
+    DebianBasedDockerContainer, linux_virtual_machine.Ubuntu1604Mixin):
+  DEFAULT_IMAGE = UBUNTU_IMAGE
+
+# class Ubuntu1404BasedDockerVirtualMachine(
```
s-deitz: Suggest dropping this commented out code and perhaps adding a note about what it will take to add support for Ubuntu 14 and Ubuntu 18.

Drop Ubuntu 17.10 since it is not a long-term support version and is no longer supported.

s-deitz:
There are conflicts in this PR. If you are able to resolve those conflicts, we can merge this change.

dPhanekham:
I've done a rebase with master. I think it should work now.

s-deitz:
Thanks. I have merged these changes.

PR #1939, opened 2019-06-27
add helpmatchmd flag for help markdown

mjz-smu:
Adds markdown formatted help messages


perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+    be slower, and the output format might change over time.
+    (default: 'false')
+  --[no]use_cprofile_for_profiling: Use cProfile instead of the profile module
+    for profiling. This has no effect unless --run_with_profiling is set.
+    (default: 'true')
+"""
+
+  # normal helpmatch search from above
+  if not matches:
+    print(FLAGS)
```
s-deitz: It seems like if you don't specify a match, you would not generate the markdown and produce the same results as with --helpmatch. That said, it looks like matches is never None (based on the call sites), so this can probably be dropped.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+        # matches module name
+        module_regex = r'^\s+?(.*?):.*'
+        # matches each flag
+        flags_regex = r'(^\s\s--.*?(?=^\s\s--|\Z))+?'
+        # matches flag name in each flag
+        flagsmd_regex = re.compile(r'^\s+?(--.*?)(:.*\Z)',
+                                   re.MULTILINE | re.DOTALL)
+        # matches triple quoted comments
+        docstring_regex = r'"""(.*?|$)"""'
+        # output standard help to string for markdown processing
```
s-deitz: These comments are helpful. Please take a couple minutes to use complete sentences so as to be consistent with the rest of the code base. Change this to:

# Output standard help to string for markdown processing.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+        txt = FLAGS.module_help(module_name)
+        # converts module name to github linkable string
+        # eg: perfkitbenchmarker.linux_benchmarks.iperf_vpn_benchmark ->
+        # perfkitbenchmarker/linux_benchmarks/iperf_vpn_benchmark.py
+        module = re.search(module_regex, txt, ).group(1)
+        module_link = module.replace('.', '/') + '.py'
+        # puts flag name in a markdown code block for visibility
```
s-deitz: What's this mean?

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+        flags[:] = [flagsmd_regex.sub(r"`\1`\2", flag) for flag in flags]
+       # get the docstring for the module without importing everything into our
+       # namespace. Probably a better way to do this
+        docstring = 'No description available'
+        # only pull doststrings from inside pkb source files
+        if (isfile(module_link)):
+          with open(module_link, "r") as f:
+            source = f.read()
+            # get the triple quoted matches
+            dsm = re.search(docstring_regex, source, re.MULTILINE | re.DOTALL)
```
s-deitz: Minor nit: dsm comes out of the blue as an abbrevation for some docstring. Perhaps call this docstring_match or just match.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+            # some modules don't have docstrings
+            # eg perfkitbenchmarker/providers/alicloud/flags.py
+            if dsm is not None:
+              docstring = dsm.group(1)
+           # print docstring
+        # format output
+        print('### [' + module, '](' + BASE_RELATIVE + module_link + ')\n')
+        print('#### Description:\n\n' + docstring + '\n\n#### Flags:\n')
+        print('\n'.join(flags) + '\n')
```
s-deitz: Can you share an example of what the final text looks like?

mjz-smu: Sorry - should have included this in the original comment: [helpmd output](https://github.com/SMU-ATT-Center-for-Virtualization/PerfKitBenchmarker/tree/helpmd/testsuite_docs) has the results for the examples listed above. The link writer isn't smart, so there are a couple places where it assumes a local source path that doesn't exist - the `absl` module and things like `perfkit.configs` for example.

mjz-smu: Added a check so no bad links are written (there were 4 in all).

perfkitbenchmarker/pkb.py
mjz-smu: Fixed.

perfkitbenchmarker/pkb.py
mjz-smu: Lines 494-496 convert module names to source file names that will get written as links later.
Line 497 adds back-ticks (\`) around each flag name in the module... so '--flag' will show up as `--flag` in the markdown viewer.

perfkitbenchmarker/pkb.py
mjz-smu: Fixed.

perfkitbenchmarker/pkb.py
mjz-smu: Fixed (removed). This check is in _PrintHelp too and seems to suggest that calling `--helpmatch` with no arguments would print(FLAGS) which was the expected behaviour here as well (without markdown). Running `./pkb.py --helpmatch` doesn't produce anything helpful though... should these flags have a default value of `pkb` or `.*` instead of the empty string?

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+
+* GCP provider:
+`./pkb.py --helpmatchmd providers.gcp.* >> testsuite_docs/providers_gcp.md`
+
+  Args:
+    matches: regex string or None. Filters help to only those whose name
+      matched the regex. If None then all flags are printed.
+  """
+
+  """ example docstring:
```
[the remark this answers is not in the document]
mjz-smu: Fixed (removed). It was example output.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+      if regex.search(module_name):
+        # where are we relative to pkb home?
+        BASE_RELATIVE = '../'
+        # matches module name
+        module_regex = r'^\s+?(.*?):.*'
+        # matches each flag
+        flags_regex = r'(^\s\s--.*?(?=^\s\s--|\Z))+?'
+        # matches flag name in each flag
+        flagsmd_regex = re.compile(r'^\s+?(--.*?)(:.*\Z)',
+                                   re.MULTILINE | re.DOTALL)
```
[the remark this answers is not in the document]
mjz-smu: Fixed.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+  else:
+    flags_by_module = FLAGS.flags_by_module_dict()
+    modules = sorted(flags_by_module)
+    regex = re.compile(matches)
+    for module_name in modules:
+      if regex.search(module_name):
+        # where are we relative to pkb home?
+        BASE_RELATIVE = '../'
+        # matches module name
+        module_regex = r'^\s+?(.*?):.*'
```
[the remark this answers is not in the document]
mjz-smu: Fixed.

perfkitbenchmarker/pkb.py
```diff
@@ -417,6 +422,102 @@ def _PrintHelp(matches=None):
...
+`./pkb.py --helpmatchmd linux_benchmarks.*`
+
+* specific modules
+`./pkb.py --helpmatchmd iperf`
+
+* windows packages
+`./pkb.py --helpmatchmd windows_packages.*`
+
+* GCP provider:
+`./pkb.py --helpmatchmd providers.gcp.* >> testsuite_docs/providers_gcp.md`
```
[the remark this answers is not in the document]
mjz-smu: Fixed. Intended use is to redirect output to a file for a markdown viewer to render. We wanted to have module documentation available in github project boards for sprint planning/reviews without having to fire up PKB.

perfkitbenchmarker/pkb.py
```diff
@@ -296,6 +296,11 @@ def GetCurrentUser():
     'helpmatch', '',
     'Shows only flags defined in a module whose name matches the given regex.',
     allow_override_cpp=True)
+flags.DEFINE_string(
+    'helpmatchmd', '',
+    'Markdown friendly help output'
```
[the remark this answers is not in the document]
mjz-smu: Fixed.

perfkitbenchmarker/pkb.py
s-deitz: I think you could do this, but I think we skip it on the empty string. So that's probably fine.

perfkitbenchmarker/pkb.py
s-deitz: Thanks for the link. This is very nice! It should be really helpful.

s-deitz:
There are a couple issues around formatting with tox, but this looks good to me. Thanks for making the changes.

mjz-smu:
Fixed formatting for tox.

PR #1948, opened 2019-07-12
Moe

gareth-ferneyhough:


perfkitbenchmarker/providers/docker/docker_container_instance.py
```diff
@@ -152,15 +152,15 @@ def _Delete(self):
...
+    info, return_code = self._GetContainerInfo()
 
-    logging.info("Checking if Docker Container Exists")
-    if len(info) > 0 and returnCode == 0:
+    logging.info('Checking if Docker Container Exists')
+    if info and return_code == 0:
       status = info[0]['State']['Running']
-      if status == "True" or status is True:
-        logging.info("Docker Container %s is up and running.", self.name)
+      if status:
```
dlott: @dPhanekham just wanted to give you a heads up about this change: I've only seen status as a bool, but let me know if you've seen otherwise (and maybe add a unit test with the mocked actual response you're seeing)

perfkitbenchmarker/providers/docker/docker_container_instance.py
```diff
@@ -278,8 +280,8 @@ def GetResourceMetadata(self):
       dict mapping string property key to value.
     """
     result = super(DockerContainer, self).GetResourceMetadata()
-    logging.warn("GET RESOURCE METADATA")
-    print(result)
+    logging.warn('GET RESOURCE METADATA')
+
```
dlott: @dPhanekham was this leftover debug code?  I've removed it.  You could add it as to the above logging statement if it's useful.  But it does seem to be more info than warning.

dPhanekham: Yes, sorry, that was debug code that I didn't catch when cleaning up the file

PR #1949, opened 2019-07-12
docker provider - small fix

dPhanekham:
the ip address here is a string, not a number, so this line was throwing a (non-fatal) error


dlott:
Sorry about that.  Thanks for fixing!

PR #1997, opened 2019-09-18
Implement support for Clear Linux

florinpapa:
This diff implements support for Clear Linux (`--os_type=clear`) in PerfKitBenchmarker.

I have tested this with sample_benchmark, by adding the following lines in `perfkitbenchmarker/linux_benchmarks/sample_benchmark.py`, after the BENCHMARK_DATA dict:
```
BENCHMARK_DATA_URL = {
    'preprovisioned_data.txt':
        'https://github.com/florinpapa/preprovisioned_data/raw/master/preprovisioned_data.txt'}
```
Then, I used this command to run the benchmark:
```
./pkb.py --cloud=AWS --benchmarks=sample --image=ami-0e6e70a1c30534bd2 --os_type=clear --machine_type=m5.4xlarge --zone=us-west-2a
```

### DIFF
--- a/CHANGES.next.md
+++ b/CHANGES.next.md
@@ -133,6 +133,7 @@
 - Added a benchmark to run spark application on a cluster that computes an approximation to pi.
 - Added a benchmark to run spark io workload on a cluster that provisions a database and queries it.
 - Added flag --record_lscpu=True to record lscpu data as its own metric "lscpu"
+- Added support for the Clear Linux OS type for AWS and static VMs (e.g. --os-type=clear)
 - Added flag --aws_placement_group_style to allow different AWS placement groups options for ec2 VM benchmarks.
 
 ### Enhancements:
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -315,8 +315,7 @@ class BaseLinuxMixin(virtual_machine.BaseOsMixin):
       if self.is_static:
         self.SnapshotPackages()
       self.SetupPackageManager()
-      if (self.OS_TYPE != 'clear'):
-        self.InstallPackages('python')
+      self.Install('python')
     self.SetFiles()
     self.DoSysctls()
     self._DoAppendKernelCommandLine()
@@ -461,19 +460,19 @@ class BaseLinuxMixin(virtual_machine.BaseOsMixin):
       self._lscpu_cache = LsCpuResults(lscpu)
     return self._lscpu_cache
 
+  def GetOsInfo(self):
+    """Returns information regarding OS type and version"""
+    self.Install('lsb_release')
+    stdout, _ = self.RemoteCommand('lsb_release -d')
+    return regex_util.ExtractGroup(LSB_DESCRIPTION_REGEXP, stdout)
+
   @property
   def os_info(self):
     """Get distribution-specific information."""
     if self.os_metadata.get('os_info'):
       return self.os_metadata['os_info']
     else:
-      if (self.OS_TYPE == 'clear'):
-        stdout, _ = self.RemoteCommand('swupd info | grep Installed')
-        return "Clear Linux build: {0}".format(regex_util.ExtractGroup(CLEAR_BUILD_REGEXP, stdout))
-      else:
-        self.Install('lsb_release')
-        stdout, _ = self.RemoteCommand('lsb_release -d')
-        return regex_util.ExtractGroup(LSB_DESCRIPTION_REGEXP, stdout)
+      return self.GetOsInfo()
 
   @property
   def kernel_release(self):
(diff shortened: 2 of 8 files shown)


perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -314,7 +315,8 @@ def PrepareVMEnvironment(self):
       if self.is_static:
         self.SnapshotPackages()
       self.SetupPackageManager()
-      self.InstallPackages('python')
+      if (self.OS_TYPE != 'clear'):
+        self.InstallPackages('python')
```
s-deitz: I think it would be nicer to add a SwupdInstall in linux_packages/python. Even if it is unimplemented and simply logs a warning.

We do need this for most benchmarks. Is there a reason we can't install Python for Clear, or is this just left for future work. Leaving it for future work is fine.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -465,9 +467,13 @@ def os_info(self):
     if self.os_metadata.get('os_info'):
       return self.os_metadata['os_info']
     else:
-      self.Install('lsb_release')
-      stdout, _ = self.RemoteCommand('lsb_release -d')
-      return regex_util.ExtractGroup(LSB_DESCRIPTION_REGEXP, stdout)
+      if (self.OS_TYPE == 'clear'):
```
s-deitz: minor: You shouldn't need the parentheses here.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -465,9 +467,13 @@ def os_info(self):
...
+        stdout, _ = self.RemoteCommand('swupd info | grep Installed')
```
s-deitz: Let's call a method GetOsInfo that installs and runs lsb_release. Then we can override this in the Mixin class for Linux Clear.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -1091,6 +1097,97 @@ def _DoAppendKernelCommandLine(self):
...
+
+  def GetPathToConfig(self, package_name):
+    """Returns the path to the config file for PerfKit packages.
+
+    This function is mostly useful when config files locations
+    don't match across distributions (such as mysql). Packages don't
+    need to implement it if this is not the case.
+    """
+    package = linux_packages.PACKAGES[package_name]
+    return package.YumGetPathToConfig(self)
```
s-deitz: Is there any reason this uses YumGetPathToConfig, or should we add something like a SwupdGetPathToConfig?

CHANGES.next.md
```diff
@@ -133,6 +133,7 @@
 - Added a benchmark to run spark application on a cluster that computes an approximation to pi.
 - Added a benchmark to run spark io workload on a cluster that provisions a database and queries it.
 - Added flag --record_lscpu=True to record lscpu data as its own metric "lscpu"
+- Added support for the Clear Linux OS type (e.g. --os-type=clear)
```
s-deitz: To note what is supported, add "for AWS and static VMs."

perfkitbenchmarker/linux_virtual_machine.py
florinpapa: @s-deitz done. Changed with `self.Install('python')` here. Also added `Swupdinstall` to linux_packages/python and the `python` package to `AptInstall`, so that the behavior remains the same as before.

perfkitbenchmarker/linux_virtual_machine.py
florinpapa: @s-deitz good catch. No, there was no reason for that to be there. Changed with Swupd.

florinpapa:
@s-deitz thank you for the feedback. Please see the new changes.

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -314,7 +315,7 @@ def PrepareVMEnvironment(self):
       if self.is_static:
         self.SnapshotPackages()
       self.SetupPackageManager()
-      self.InstallPackages('python')
+      self.Install('python')
```
s-deitz: With this change, we may not install the 'python' package through YumInstall, if I've followed the code through. I think this may be okay since we will install a version of Python. To be safe, we may want to add python to the YumInstall alongside the AptInstall.

florinpapa: I added "python" to YumInstall alongside what is retrieved by getattr, to be consistent with Apt. I am not adding anything in the situation that getattr returns exactly "python".

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -1091,6 +1096,101 @@ def _DoAppendKernelCommandLine(self):
...
+    """Returns the service name of a PerfKit package.
+
+    This function is mostly useful when service names don't
+    match across distributions (such as mongodb). Packages don't
+    need to implement it if this is not the case.
+    """
+    package = linux_packages.PACKAGES[package_name]
+    return package.SwupdGetServiceName(self)
+
+  def GetOsInfo(self):
```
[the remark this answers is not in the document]
florinpapa: @s-deitz good point, thanks for the suggestion. I have added the comment for GetOsInfo and have simplified a few of the other comments.

florinpapa:
@s-deitz implemented your latest feedback and rebased. Please take a look.

s-deitz:
Merged. Thanks.

PR #2001, opened 2019-09-20
Update .travis.yml and requirements.txt for Python 3

bvliu:
Edit travis.yml so that it runs tests for Python 3.6.


bvliu:
Tox failing "flake8" when running locally in a virtualenv with py36 but not with py27 - going to do more investigating.

PR #2035, opened 2019-11-07
update README to azure cli 2.0 instructions

dPhanekham:
I've updated the README to reflect that PerfKitBenchmarker has been upgraded to use the azure cli 2.0 since this pull request: #1359 

The old azure cli uses the 'azure' command and the azure cli 2.0+ uses the 'az' command

The old cli no longer works with pkb, as the [azure provider uses the command for the new cli](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/perfkitbenchmarker/providers/azure/__init__.py) 

I have tested this both with the master branch and with the latest release

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ Follow the instructions at https://docs.microsoft.com/en-us/cli/azure/install-az
 on Linux, run the following commands:
 
 ```bash
-$ curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+$ curl -L https://aka.ms/InstallAzureCli | bash
 $ az login
 ```
 Test that `azure` is installed correctly:
--- a/README.md
+++ b/README.md
@@ -381,31 +381,27 @@ $ aws configure
 
 ### Windows Azure CLI and credentials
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -381,31 +381,27 @@ $ aws configure
...
-Go [here](https://nodejs.org/download/), and follow the setup instructions.
-
-Next, run the following (omit the `sudo` on Windows):
+Follow the instructions at https://docs.microsoft.com/en-us/cli/azure/install-azure-cli or 
+on Linux, run the following commands:
 
 ```bash
-$ sudo npm install azure-cli -g
-$ azure login
+$ curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
```
dlott: We should probably suggest the more generic install from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux
`curl -L https://aka.ms/InstallAzureCli | bash` 
But the reference to the main docs is good.

dPhanekham: I've updated to use the more generic install instruction

Issue #2022, opened 2019-10-28
ERROR  Exception occurred while calling CreateAndBootVm(<BaseVirtualMachine [ip=None, internal_ip=None]>):


ashutoshbhatt1:
New to PKB as well as Openstack. I have been trying to run this command and see if basic PKB things are working or not with multiple Openstack instances I am running but endup getting the same error. 

running this command from my local mac machine where I have performance kitbench mark is installed. Here

> pkb.py  --cloud=OpenStack --os_type=centos7 --machine_type=m1.medium   --openstack_network=public --openstack_network=private --benchmarks=ping

> 2019-10-28 15:22:12,466 a853139b MainThread INFO     Setting --max_concurrent_threads=200.
> 2019-10-28 15:22:12,486 a853139b MainThread WARNING  The key "flags" was not in the default config, but was in user overrides. This may indicate a typo.
> 2019-10-28 15:22:12,602 a853139b MainThread ping(1/1) INFO     Provisioning resources for benchmark ping
> 2019-10-28 15:22:12,603 a853139b Thread-1 ping(1/1) INFO     Testing OpenStack CLI command is installed and working
> 2019-10-28 15:22:12,603 a853139b Thread-1 ping(1/1) INFO     Running: openstack image list --format json
> 2019-10-28 15:22:14,867 a853139b Thread-1 ping(1/1) INFO     Validating prerequisites.
> 2019-10-28 15:22:14,867 a853139b Thread-1 ping(1/1) INFO     Running: openstack image show centos7 --format json
> 2019-10-28 15:22:16,363 a853139b Thread-1 ping(1/1) INFO     Running: openstack flavor show m1.medium --format json
> 2019-10-28 15:22:18,098 a853139b Thread-1 ping(1/1) INFO     Running: openstack network show private --format json
> 2019-10-28 15:22:19,447 a853139b Thread-1 ping(1/1) INFO     Prerequisites validated.
> 2019-10-28 15:22:19,447 a853139b Thread-1 ping(1/1) INFO     Running: openstack security group show perfkit_sc_group --format json
> 2019-10-28 15:22:20,745 a853139b Thread-1 ping(1/1) INFO     Running: openstack keypair create perfkit_key_a853139b --format json --public-key /var/folders/sc/l_g86zr57qg0pr7jbxj3c3340000gn/T/perfkitbenchmarker/runs/a853139b/perfkitbenchmarker_keyfile.pub
[...]

PR #2058, opened 2019-12-13
Fix Openstack provider subnet creation issue

ichbinfrog:
Added checks for return type of `openstack network show {{ network_name }}`.

This change is meant to address this specific issue #2022 where `openstack network show {{ network_name }}` returns a `list<string>` rather than a `string`, thus breaking the openstack provider. 

This occurs with the openstack CLI 4.0.0 (although the specific version responsible for the change in return type is hard to identify in the changelogs).


ichbinfrog:
@googlebot I signed it!

PR #2075, opened 2020-01-07
Add netperf UDP stream test

dPhanekham:
Add support for netperf UDP stream test.
Add flags to set size of stream tests
Also add same flag to container_netperf benchmark

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
@@ -27,6 +27,9 @@ from perfkitbenchmarker.linux_benchmarks import netperf_benchmark
 
 FLAGS = flags.FLAGS
 
+# Set default as 128KB (131072 bytes)
+# the Linux default of 16K cannot achieve
+# "link rate" and that the higher default value will let us
 flags.DEFINE_integer('container_netperf_tcp_stream_send_size_in_bytes', 131072,
                      'Send size to use for TCP_STREAM tests (netperf -m flag)')
 
--- a/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
@@ -27,6 +27,9 @@ from perfkitbenchmarker.linux_benchmarks import netperf_benchmark
 
 FLAGS = flags.FLAGS
 
+flags.DEFINE_integer('container_netperf_tcp_stream_send_size_in_bytes', 131072,
+                     'Send size to use for TCP_STREAM tests (netperf -m flag)')
+
 BENCHMARK_NAME = 'container_netperf'
 BENCHMARK_CONFIG = """
 container_netperf:
@@ -87,6 +90,7 @@ def Run(benchmark_spec):
                   '-H', container_0.ip_address,
                   '-l', '100',
                   '--',
+                  '-m', FLAGS.container_netperf_tcp_stream_send_size_in_bytes,
                   '-o', netperf_benchmark.OUTPUT_SELECTOR]
   cluster.DeployContainer('netperf', benchmark_spec.container_specs['netperf'])
   container_1 = cluster.containers['netperf'][1]
--- a/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
@@ -27,9 +27,8 @@ from perfkitbenchmarker.linux_benchmarks import netperf_benchmark
 
 FLAGS = flags.FLAGS
 
-# Set default as 128KB (131072 bytes)
-# the Linux default of 16K cannot achieve
-# "link rate" and that the higher default value will let us
+# We set the default to 128KB (131072 bytes) to override the Linux default
+# of 16K so that we can achieve the "link rate".
 flags.DEFINE_integer('container_netperf_tcp_stream_send_size_in_bytes', 131072,
                      'Send size to use for TCP_STREAM tests (netperf -m flag)')
 
--- a/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
@@ -65,8 +65,12 @@ flags.DEFINE_integer('netperf_thinktime_array_size', 0,
 flags.DEFINE_integer('netperf_thinktime_run_length', 0,
                      'The number of contiguous numbers to sum at a time in the '
                      'thinktime array.')
+flags.DEFINE_integer('netperf_udp_stream_send_size_in_bytes', 1024,
+                     'Send size to use for UDP_STREAM tests (netperf -m flag)')
+flags.DEFINE_integer('netperf_tcp_stream_send_size_in_bytes', 131072,
+                     'Send size to use for TCP_STREAM tests (netperf -m flag)')
 
-ALL_BENCHMARKS = ['TCP_RR', 'TCP_CRR', 'TCP_STREAM', 'UDP_RR']
+ALL_BENCHMARKS = ['TCP_RR', 'TCP_CRR', 'TCP_STREAM', 'UDP_RR', 'UDP_STREAM']
 flags.DEFINE_list('netperf_benchmarks', ALL_BENCHMARKS,
                   'The netperf benchmark(s) to run.')
 flags.register_validator(
@@ -356,7 +360,7 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
   enable_latency_histograms = FLAGS.netperf_enable_histograms or num_streams > 1
   # Throughput benchmarks don't have latency histograms
   enable_latency_histograms = enable_latency_histograms and \
-      benchmark_name != 'TCP_STREAM'
+      (benchmark_name != 'TCP_STREAM' and benchmark_name != 'UDP_STREAM')
   # Flags:
   # -o specifies keys to include in CSV output.
   # -j keeps additional latency numbers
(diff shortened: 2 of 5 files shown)


perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
```diff
@@ -27,6 +27,9 @@
 
 FLAGS = flags.FLAGS
 
+flags.DEFINE_integer('container_netperf_tcp_stream_send_size_in_bytes', 131072,
```
s-deitz: Please add a comment as to why this default value was selected.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -65,8 +65,12 @@
 flags.DEFINE_integer('netperf_thinktime_run_length', 0,
                      'The number of contiguous numbers to sum at a time in the '
                      'thinktime array.')
+flags.DEFINE_integer('netperf_udp_stream_send_size_in_bytes', 1024,
+                     'Send size to use for UDP_STREAM tests (netperf -m flag)')
+flags.DEFINE_integer('netperf_tcp_stream_send_size_in_bytes', 131072,
+                     'Send size to use for TCP_STREAM tests (netperf -m flag)')
```
s-deitz: Comment here as well.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -367,6 +371,16 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
   confidence = ('-I 99,5 -i {0},3'.format(FLAGS.netperf_max_iter)
                 if FLAGS.netperf_max_iter else '')
   verbosity = '-v2 ' if enable_latency_histograms else ''
+
+  netperf_cmd = ""
+
+  remote_cmd_timeout = \
```
s-deitz: nit: Avoid backslash for line continuation. Use parentheses instead:

remote_cmd_timeout = (
    FLAGS.netperf_test_length * (FLAGS.netperf_max_iter or 1) + 300)

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -377,7 +391,20 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
                      server_ip=server_ip,
                      length=FLAGS.netperf_test_length,
                      output_selector=OUTPUT_SELECTOR,
-                     confidence=confidence, verbosity=verbosity)
+                     confidence=confidence,
+                     verbosity=verbosity)
+
+  if benchmark_name.upper() == 'UDP_STREAM':
+    netperf_cmd += (' -R 1 -m {send_size} '.format(
```
s-deitz: Is this a change in response size as well?

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
dPhanekham: This value, 131072 bytes (128KB) was selected at the recommendation of @jonesrick. 

"The 128K send size is deliberate - otherwise netperf will default to 16K based on Linux send socket buffer size initial values, which has enough overhead to preclude achieving "link rate" as it were.  And the default receive size will be 87380 bytes for similar reasons on the receive side. That would be a good thing to change in the netperf command-lines in PKB.  And it would match the default send size in iperf3.  64K may be sufficient but 128K would give additional cushion.  The need for receive side (-M) isn't as clear-cut but it won't hurt." 

Or, if preferred, I could give it a default of None and let netperf select the size unless this flag is set.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -367,6 +371,16 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
   confidence = ('-I 99,5 -i {0},3'.format(FLAGS.netperf_max_iter)
                 if FLAGS.netperf_max_iter else '')
   verbosity = '-v2 ' if enable_latency_histograms else ''
+
+  netperf_cmd = ""
```
cwilkes: Did you mean to set this to an empty string up here?  It is set on line 370/384

dPhanekham: That line has now been removed

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dPhanekham: Changed to parentheses

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dPhanekham: I have also added the -M flag to set the remote receive size as well. I do not believe it changes that response size, as UDP_STREAM and TCP_STREAM test throughput unidirectionally.

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
s-deitz: I think setting the default is important. I'd just like a comment in the code so that we know why we selected this default. Noting that the Linux default of 16K cannot achieve "link rate" and that the higher default value will let us would be sufficient.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -377,7 +389,19 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
                      server_ip=server_ip,
                      length=FLAGS.netperf_test_length,
                      output_selector=OUTPUT_SELECTOR,
-                     confidence=confidence, verbosity=verbosity)
+                     confidence=confidence,
+                     verbosity=verbosity)
+
+  if benchmark_name.upper() == 'UDP_STREAM':
+    netperf_cmd += (' -R 1 -m {send_size} -M {send_size} '.format(
```
dlott: What is `-R 1` for?  I don't see it in https://hewlettpackard.github.io/netperf/doc/netperf.html#UDP_005fSTREAM

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
dlott: Do we also need to set `-s`?

Per https://hewlettpackard.github.io/netperf/doc/netperf.html#UDP_005fSTREAM
If the value of the -m option is larger than the local send socket buffer size (-s option) netperf will likely abort with an error message about how the send call failed:

     netperf -t UDP_STREAM -H 192.168.2.125
     UDP UNIDIRECTIONAL SEND TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 192.168.2.125 (192.168.2.125) port 0 AF_INET
     udp_send: data send error: Message too long

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -356,7 +360,7 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
   enable_latency_histograms = FLAGS.netperf_enable_histograms or num_streams > 1
   # Throughput benchmarks don't have latency histograms
   enable_latency_histograms = enable_latency_histograms and \
-      benchmark_name != 'TCP_STREAM'
+      (benchmark_name != 'TCP_STREAM' and benchmark_name != 'UDP_STREAM')
```
dlott: More readable as
(benchmark_name not in ['TCP_STREAM' , 'UDP_STREAM'])

Also if you could fix the \ continue as well, that'd be great.

jonesrick:
The "-R  1"  as a test-specific option for a UDP_STREAM test is needed to
disable setting SO_DONTROUTE on the data socket, which is set by default to
address issues where various entities/groups in the have used the
UDP_STREAM test as part of system link up/down testing and taken-out a
corporate security video feed as a result.  The way netmasks and such are
set in GCP mean that all destinations are not "local" (need to be routed
to) thus the "-R 1" is needed to successfully run a netperf UDP_STREAM test
in GCP.

jonesrick:
A test-specific -s option is not needed with TCP_STREAM and a 131072 byte
send size because TCP is a byte-stream rather than message-oriented
protocol.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dPhanekham: when testing UDP_STREAM, i found the default send socket buffer size to be 212,992 bytes, which is larger than the max possible UDP message size of 65,507 bytes, so it should not be a problem and we don't need the -s flag. I can add in an upper limit of 65507 bytes, because it will cause an error if you try to use a larger size than that.

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
dPhanekham: Added a comment to clarify default size. Addressed -s flag in a comment below

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dPhanekham: comment added

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
s-deitz: Thanks for adding the comment. Since comments are easier to parse if they contain punctuation and consist of complete sentences, perhaps edit it so that it reads more like the following:

# We set the default to 128KB (131072 bytes) to override the Linux default
# of 16K so that we can achieve the "link rate".

s-deitz: Hmm. Looks like the prefix of "#" changed my font size. I just meant to write a comment.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
s-deitz: I see it here as response size: http://www.cs.kent.edu/~farrell/dist/ref/Netperf.html

jonesrick:
I don't mean to "bikeshed" but since netperf will "process" the values
given to the -m (and -M and -s and -S) option as occasion warrants,  you
could make this a string and then when someone says 128k they will get
128000 and when they say 128K they will get 131072.  Netperf will interpret
a suffix of  k|m|g as powers of ten, and K|M|G as powers of two.  Of
course, then if they give a rather bogus string, Metcalf only knows what
netperf will do with it :)

jonesrick:
In the version of the netperf manual at kent.edu that specific case the -R
appears to be in a section on DLPI-test-specific command-line options.
That part of the manual does not pertain to the UDP_STREAM test.

perfkitbenchmarker/linux_benchmarks/container_netperf_benchmark.py
dPhanekham: changed the comment

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
```diff
@@ -377,7 +392,19 @@ def RunNetperf(vm, benchmark_name, server_ip, num_streams):
...
+                     confidence=confidence,
+                     verbosity=verbosity)
+
+  if benchmark_name.upper() == 'UDP_STREAM':
+    netperf_cmd += (' -R 1 -m {send_size} -M {send_size} '.format(
+                    send_size=FLAGS.netperf_udp_stream_send_size_in_bytes))
+    metadata['netperf_send_size_in_bytes'] = FLAGS.netperf_udp_stream_send_size_in_bytes
+
+  elif benchmark_name.upper() == 'TCP_STREAM':
+    netperf_cmd += (' -R 1 -m {send_size} -M {send_size} '.format(
```
dlott: Per rick's comment it seems that we can drop the `-m` and `-M` here.

As well as the `-R`

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dlott: What platforms have you tested this on?  My reading is that there may be some where `-s` is required (no idea which) but I don't see a downside to specifying both `-s` and `-S`

dlott:
If you move the additions in CHANGES.next.md up a line or two, they'll won't conflict as people add things.

tests/data/netperf_results.json
```diff
@@ -38,5 +38,15 @@
     "MIGRATED UDP REQUEST/RESPONSE TEST from 0.0.0.0 (0.0.0.0) port 20001 AF_INET to 10.240.31.117 () port 20001 AF_INET : +/-2.500% @ 99% conf.  : first burst 0",
     "Throughput,Throughput Units,Throughput Confidence Width (%),Confidence Iterations Run,Stddev Latency Microseconds,50th Percentile Latency Microseconds,90th Percentile Latency Microseconds,99th Percentile Latency Microseconds,Minimum Latency Microseconds,Maximum Latency Microseconds,Local Transport Retransmissions,Remote Transport Retransmissions",
     "3313.49,Trans/s,7.546,20,214.64,295,330,406,200,500,0,0"
+  ],
+  [
+    "MIGRATED UDP STREAM TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 104.198.67.251 () port 20001 AF_INET : +/-2.500% @ 99% conf.",
+    "Throughput,Throughput Units,50th Percentile Latency Microseconds,90th Percentile Latency Microseconds,99th Percentile Latency Microseconds,Stddev Latency Microseconds,Minimum Latency Microseconds,Maximum Latency Microseconds,Confidence Iterations Run,Throughput Confidence Width (%),Local Transport Retransmissions,Remote Transport Retransmissions",
+    "1102.42,10^6bits/s,3,3,11,46.14,1,15144,1,-1.000,-1,-1"
```
dlott: Is this only a send throughput?  Should there also be a receive throughput?

https://hewlettpackard.github.io/netperf/doc/netperf.html#UDP_005fSTREAM
> The biggest of these implications is the data which is sent might not be received by the remote. For this reason, the output of a UDP_STREAM test shows both the sending and receiving throughput. On some platforms, it may be possible for the sending throughput to be reported as a value greater than the maximum rate of the link. This is common when the CPU(s) are faster than the network and there is no intra-stack flow-control.

Is do you have a pkb.log that shows the raw output of a run?

jonesrick:
Yes, for netperf TCP_STREAM, a test-specific -R is not necessary.

I suspect though we still want the -m option because we want netperf
TCP_STREAM to use a send size of 128KiB and not the on-Linux default of
16KiB.  If no send size (-m option) is specified on the command line,
netperf uses the value of SO_SNDBUF at the time of socket creation.  if
there is no send socket buffer size option (test-specific -s) specified,
that will be the default the system has configured for socket buffer size.
Under Linux that is 16 KiB (net.ipv4.tcp_wmem, middle value).  Thus,
without a -m option, netperf will have a send size of 16 KiB.  And that is
insufficient to drive more than about 20 Gbit/s.  So, we do want to pass a
-m option, with a send size of 128KiB, to netperf via the command line for
a TCP_STREAM test.  This also matches the default send size for TCP in
iperf3.

There is similar logic for the receive size used by netperf, looking at the
-M and possibly the -S option. net.ipv4.tcp_rmem, again middle value.
There though, depending on the version of the kernel, netperf will default
to either 86KiB or 128KiB because those are the default initial sizes for
the SO_RCVBUF of a TCP socket.

Whether one wants to have a -s or -S option will depend on circumstances
(eg bandwidthXdelay product) and whether or not, under Linux,
net.core.[rw]mem_max has been tuned. And/or net.ipv4.tcp{rw}mem.

jonesrick:
For a UDP_STREAM test the "throughput" output selector will report what the
remote netserver received.  It will be equivalent to specifying
remote_recv_throughput (or is supposed to be).

$ netperf -- -O ? | grep -i through
THROUGHPUT
THROUGHPUT_UNITS
THROUGHPUT_CONFID
LOCAL_SEND_THROUGHPUT
LOCAL_RECV_THROUGHPUT
REMOTE_SEND_THROUGHPUT
REMOTE_RECV_THROUGHPUT

Local would be the netperf side and remote the netserver side.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dPhanekham: I've test on ubuntu18, debian9, and centos8. So far I haven't needed to specify the -s/-S flags. But I can certainly certainly add them in as flags and just give them the default values netperf has given them so far in my testing.

jonesrick:
There can be an entire discussion around whether or not a TCP_STREAM test
requires a test-specific -s or -S option.  It would center on when Linux
uses autotuning (no -s or -S option) and the settings for
net.ipv4.tcp_[rw]mem on both sides, and when the application makes an
explicit setsockopt() call for SO_SNDBUF or SO_RCVBUF and where
net.core.[rw]mem_max are set.  At this point I would recommend against
using -s and/or -S with default values.  Unless that default value is "-1"
in which case netperf will parse the option(s) but not actually make any
setsockopt() calls to set socket buffer size.
UDP_STREAM on the other hand possibly wants a -S option but without tuning
sysctls (net.core.rmem_max) it likely won't be effective anyway.  A
UDP_STREAM test (under Linux at least) does *not* want a -s option.

perfkitbenchmarker/linux_benchmarks/netperf_benchmark.py
dlott: For posterity, it is documented in https://github.com/HewlettPackard/netperf/blob/bcb868bde7f0203bbab69609f65d4088ba7398db/Release_Notes#L212

PR #2090, opened 2020-01-21
Add SoftLayer as Provider

antigenius0910:
This adds support for SoftLayer Linux benchmarks with current master


antigenius0910:
@googlebot I fixed it.

PR #2092, opened 2020-01-21
add softlayer as provider

antigenius0910:
This add softlayer as provider for current master


dlott:
Thanks for your pull request!

Given it's large size, if you could break it down into several smaller requests, that would help expedite the review.  For example, `perfkitbenchmarker/providers/softlayer/softlayer_virtual_machine.py` is probably a good place to start to get the `cluster_boot` benchmark working.  And `perfkitbenchmarker/providers/softlayer/coss3.py` could be a follow on PR.

Also if you could add some unit tests, that would be great.  For example, S3Service doesn't implement the ObjectStorageService.CopyToBucket abstract method.

Finally, you should be able to run the unit tests and tox locally to verify that Travis will pass.

antigenius0910:
Closing this pull request and will submit as @dlott suggested into several smaller requests

PR #2096, opened 2020-01-24
Add window size tuning parameters

dPhanekham:
I have added in four flags to set the following sysctl settings:
        net.ipv4.tcp_rmem
        net.ipv4.tcp_wmem
        net.core.rmem_max
        net.core.wmem_max

tcp_rmem and tcp_wmem each consist of 3 values (min default max). I only added flags to change the max value, as this is what the automatic tcp window scaling in linux uses as is maximum.

rmem_max and wmem_max set the maximum send/receive buffers for non TCP connections.

Tuning these parameters are essential for achieve decent throughput over links with a high RTT

### DIFF
--- a/perfkitbenchmarker/linux_virtual_machine.py
+++ b/perfkitbenchmarker/linux_virtual_machine.py
@@ -154,13 +154,13 @@ flags.DEFINE_string(
 
 flags.DEFINE_integer(
     'tcp_max_receive_buffer', None, 
-    'The maximum receive buffer for TCP socket connections. '
+    'The maximum receive buffer for TCP socket connections in bytes. '
     'Increasing this value may increase single stream TCP throughput '
     'for high latency connections')
 
 flags.DEFINE_integer(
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -152,6 +152,27 @@
     'non-empty string will cause a reboot to occur after VM prepare. '
     'If unspecified, the kernel command line will be unmodified.')
 
+flags.DEFINE_integer(
```
dlott: Include the sysctl which is modified in the description.

dlott:
Also update the README with the new flags.

Otherwise looks good.

dlott:
Can you pull master so I can merge?

dlott:
Needs another pull.

dPhanekham:
Rebased on master. Should be up to date now

dlott:
Thanks!

PR #2109, opened 2020-02-10
Openstack provider to handle JSON format properly

keerockl:
external network value returns True or false instead of 'External' string.


perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -211,7 +211,7 @@ def _CheckNetworks(self):
 
   def _CheckFloatingIPNetworkExists(self, floating_network_name_or_id):
     network = self._CheckNetworkExists(floating_network_name_or_id)
-    if network['router:external'] != 'External':
+    if network['router:external'] != 'External' and not 'True':
```
jellyfishcake: Do you mean _if network['router:external'] not in ('External', 'True')_?
because _not 'True'_ just evaluates to false

keerockl: The old version returns as 'External', and the new version returns as 'True', so if value is neither 'External' or 'True', then we need to let it fail.

jellyfishcake: you will need to do _if network['router:external'] not in ('External', 'True')_ or  _if network['router:external'] != 'External' or network['router:external'] != 'True'_
What you wrote there currently means _if network['router:external'] != 'External' and if not 'True'_.
You can open an interactive python interpreter to fiddle and test. For example, set network['router:external'] = 'foobar' and see what result you get with each. :)
Since this can be confusing, you should also add a unit test.

keerockl: Thank you @jellyfishcake , it's my mistake. let me update patch.

keerockl: @jellyfishcake, Fix is updated and added unittest. It would be great if you can suggest better idea on creating benchmarkspec in unittest.

keerockl: Hi @jellyfishcake, could you please approve this change? This feature needs to be merged in order to have pkb openstack working on current API version.

tests/providers/openstack/os_virtual_machine_test.py
```diff
@@ -0,0 +1,103 @@
...
+    spec.gpu_count = "1"
+    spec.gpu_type = "test-gpu-type"
+    spec.image = "test-image"
+    spec.install_packages = "None"
+    spec.background_cpu_threads = "None"
+    spec.background_network_mbits_per_sec = "1"
+    spec.background_network_ip_type = "None"
+    spec.vm_metadata = {}
+    openstackVM = os_virtual_machine.OpenStackVirtualMachine(spec)
+    network = openstackVM._CheckFloatingIPNetworkExists('External')
```
jellyfishcake: Do 'True' here to align with your test.

tests/providers/openstack/os_virtual_machine_test.py
```diff
@@ -0,0 +1,103 @@
...
+    self.assertEqual(_network_external, network)
+
+
+if __name__ == '__main__':
+  unittest.main()
```
jellyfishcake: Add a negative test, where you expect the test to fail, such as {'router:external': 'Foobar'}

keerockl: Done :)

PR #2111, opened 2020-02-10
Add a flag openstack_post_provisioning_script

keerockl:


keerockl:
Could you please kindly let me know which metadata file you are meaning?
I have already specified default value as below flags.DEFINE_string('openstack_post_provisioning_script', None, '...

jellyfishcake:
Ah, what I mean is you want to add a method in OpenStackVirtualMachine class that adds onto GetResourceMetadata in virtual_machine.BaseVirtualMachine. something along the lines of:  
_result = super(OpenStackVirtualMachine, self).GetResourceMetadata()_
_result[use_provisioning_script] = true_
Any custom flags that was used should be recorded with this method.

perfkitbenchmarker/providers/openstack/os_virtual_machine.py
```diff
@@ -391,6 +394,17 @@ def CreateScratchDisk(self, disk_spec):
...
     self._CreateScratchDiskFromDisks(disk_spec, disks)
 
+  def GetResourceMetadata(self):
+    """Returns a dict containing metadata about the VM.
+
+    Returns:
+      dict mapping string property key to value.
+    """
+    result = super(OpenStackVirtualMachine, self).GetResourceMetadata()
+    if FLAGS.openstack_post_provisioning_script:
```
jellyfishcake: flag values should not be reused. Store the value as a object attribute, i.e. self.post_provisioning_script = FLAGS.openstack_post_provisioning_script during __init__ and then use self.post_provisioning_script in later calls.

keerockl: Thank you @jellyfishcake, I have updated patchset.

PR #2112, opened 2020-02-10
Add proxy setting in Clearlinux VM

keerockl:
Clearlinux is not using /etc/environment as environment file. This patch
adds SetupProxy, ProxyCleanup under Clearlinux VM class, and applys
/etc/profile to RemoteCommand as it's not picking up environment
variable in non-interactive shell.


perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -1309,6 +1309,87 @@ def GetOsInfo(self):
     stdout, _ = self.RemoteCommand('swupd info | grep Installed')
     return "Clear Linux build: {0}".format(regex_util.ExtractGroup(CLEAR_BUILD_REGEXP, stdout))
 
+  def SetupProxy(self):
+    """Sets up proxy configuration variables for the cloud environment."""
+    super(ClearMixin, self).SetupProxy()
+    profile_file = '/etc/profile'
+    commands = []
+    command_template = "grep -qxF '{0}' {1} || echo '{0}' | sudo tee -a {1}"
```
dlott: Why are we searching in /etc/profile for these values?

perfkitbenchmarker/linux_virtual_machine.py
```diff
@@ -1309,6 +1309,87 @@ def GetOsInfo(self):
...
+    if FLAGS.no_proxy:
+      commands.append(command_template.format(
+          'export no_proxy=' + FLAGS.no_proxy, profile_file))
+      commands.append(command_template.format(
+          'export NO_PROXY=' + FLAGS.no_proxy, profile_file))
+
+    if commands:
+      self.RemoteCommand(';'.join(commands))
+
+  def ProxyCleanup(self):
```
dlott: Where is this called?

dlott: Where is it defined?

https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/search?q=ProxyCleanup&type=Code
We couldn’t find any code matching 'ProxyCleanup' in GoogleCloudPlatform/PerfKitBenchmarker
You could try an advanced search.

perfkitbenchmarker/linux_virtual_machine.py
dlott: In the [main method](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/53e5ce22442512dc26c49729eecae291365d6315/perfkitbenchmarker/linux_virtual_machine.py#L319) the flag values are appended to `/etc/environment` which I could see happening here but to `/etc/profile` instead.  But I don't understand why you'd grep to see if they're there and still append?  If a value is set multiple times in the flag, does the last one win?

keerockl: I have modified to make it the same style as base class.

perfkitbenchmarker/linux_virtual_machine.py
keerockl: sorry my apology, ProxyCleanup function only exists in Internal repository, not in upstream. I have removed it.

Issue #2105, opened 2020-02-05
iperf 2.0.14 new features


rjmcmahon:
[Iperf 2.0.14](https://sourceforge.net/projects/iperf2/) has a lot of [new features](https://youtu.be/pfzwrcTDc38) that help with measuring latency, network power, etc.  This might be of of interest.

Bob

mjz-smu:
Is there a good mirror of this repo to fall back on? Sourceforge disappeared on me mid-run yesterday.

![image](https://user-images.githubusercontent.com/35934517/77579225-f3607480-6ea7-11ea-9669-199f796217f0.png)

![image](https://user-images.githubusercontent.com/35934517/77579335-230f7c80-6ea8-11ea-86f0-f03effffc1bf.png)

lealog:
Is available using the following URL:
- https://sourceforge.net/p/iperf2/code/ci/master/tree/

rjmcmahon:
Hi,

The current head on sourceforge is broken and won't compile. There is also a mutex problem that's being fixed.  I hope to have this cleaned up soon.

Bob

dlott:
Let's track consolidate to https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/2545

PR #2115, opened 2020-02-14
Update iperf2

dPhanekham:
In reference to issue #2105
Updates iperf2 version from 2.0.4 to 2.0.13.
Installs from source, as 2.0.13 is not in package managers currently
Otherwise, functionality of the benchmark remains the same.


jellyfishcake:
Looks good, needs a sync for us to pull.

dPhanekham:
Rebased on master. Should be in sync now

PR #2184, opened 2020-04-08
Implement AllowIcmp() for AWS and GCP provider.

dPhanekham:
I found that the ping benchmark with --ping_also_run_using_external_ip=True would fail when trying to perform tests with external IP addresses.

The benchmark makes a call to AllowIcmp(), which is not implemented for multiple cloud providers. So I have implemented that function for Google Cloud and now ping tests using external IPs function as intended.

external ping tests are also failing on AWS and I'm working on a fix for those

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -103,7 +103,9 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
     sending_vm: The VM issuing the ping request.
     receiving_vm: The VM receiving the ping.  Needed for metadata.
     receiving_ip: The IP address to be pinged.
-    ip_type: The type of 'receiving_ip' (either 'internal' or 'external')
+    ip_type: The type of 'receiving_ip',
+        (either 'vm_util.IpAddressSubset.INTERNAL 
+         or vm_util.IpAddressSubset.EXTERNAL')
   Returns:
     A list of samples, with one sample for each metric.
   """
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -114,7 +114,7 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
     logging.warn('%s is not reachable from %s', receiving_vm, sending_vm)
     return []
 
-  logging.info('Ping results (ip_type = %s):' % ip_type)
+  logging.info('Ping results (ip_type = %s):', ip_type)
   ping_cmd = 'ping -c 100 %s' % receiving_ip
   stdout, _ = sending_vm.RemoteCommand(ping_cmd, should_log=True)
   stats = re.findall('([0-9]*\\.[0-9]*)', stdout.splitlines()[-1])
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -80,13 +80,13 @@ def Run(benchmark_spec):
       ip_type = vm_util.IpAddressMetadata.EXTERNAL
       results = results + _RunPing(sending_vm,
                                    receiving_vm,
-                                   receiving_vm.internal_ip,
+                                   receiving_vm.ip_address,
                                    ip_type)
     if vm_util.ShouldRunOnInternalIpAddress(sending_vm, receiving_vm):
       ip_type = vm_util.IpAddressMetadata.INTERNAL
       results = results + _RunPing(sending_vm,
                                    receiving_vm,
-                                   receiving_vm.ip_address,
+                                   receiving_vm.internal_ip,
                                    ip_type)
   return results
 
@@ -105,7 +105,7 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if ip_type == vm_util.IpAddressSubset.INTERNAL and not sending_vm.IsReachable(receiving_vm):
+  if ip_type == vm_util.IpAddressMetadata.INTERNAL and not sending_vm.IsReachable(receiving_vm):
     logging.warn('%s is not reachable from %s', receiving_vm, sending_vm)
     return []
 
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -103,7 +103,7 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if not sending_vm.IsReachable(receiving_vm) and ip_type == 'internal':
+  if ip_type == 'internal' and not sending_vm.IsReachable(receiving_vm):
     logging.warn('%s is not reachable from %s', receiving_vm, sending_vm)
     return []
 
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -22,6 +22,7 @@ import logging
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import sample
+from perfkitbenchmarker import vm_util
 import re
 
 flags.DEFINE_boolean('ping_also_run_using_external_ip', False,
@@ -79,16 +80,19 @@ def Run(benchmark_spec):
   vms = benchmark_spec.vms
   results = []
   for sending_vm, receiving_vm in vms, reversed(vms):
+    ip_type = vm_util.IpAddressSubset.INTERNAL
     results = results + _RunPing(sending_vm,
                                  receiving_vm,
                                  receiving_vm.internal_ip,
-                                 'internal')
+                                 ip_type)
+
   if FLAGS.ping_also_run_using_external_ip:
     for sending_vm, receiving_vm in vms, reversed(vms):
+      ip_type = vm_util.IpAddressSubset.EXTERNAL
       results = results + _RunPing(sending_vm,
                                    receiving_vm,
                                    receiving_vm.ip_address,
-                                   'external')
+                                   ip_type)
   return results
 
 
@@ -103,11 +107,11 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if ip_type == 'internal' and not sending_vm.IsReachable(receiving_vm):
+  if ip_type == vm_util.IpAddressSubset.INTERNAL and not sending_vm.IsReachable(receiving_vm):
     logging.warn('%s is not reachable from %s', receiving_vm, sending_vm)
     return []
 
-  logging.info('Ping results (ip_type = %s):', ip_type)
+  logging.info('Ping results (ip_type = %s):' % ip_type)
   ping_cmd = 'ping -c 100 %s' % receiving_ip
   stdout, _ = sending_vm.RemoteCommand(ping_cmd, should_log=True)
   stats = re.findall('([0-9]*\\.[0-9]*)', stdout.splitlines()[-1])
--- a/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
@@ -120,7 +120,13 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   stats = re.findall('([0-9]*\\.[0-9]*)', stdout.splitlines()[-1])
   assert len(stats) == len(METRICS), stats
   results = []
-  metadata = {'ip_type': ip_type,
+
+  if ip_type == vm_util.IpAddressSubset.INTERNAL:
+    ip_string = 'internal'
+  else:
+    ip_string = 'external'
+
+  metadata = {'ip_type': ip_string,
               'receiving_zone': receiving_vm.zone,
               'sending_zone': sending_vm.zone}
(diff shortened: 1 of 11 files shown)


dPhanekham:
Also added a fix for aws ping tests

perfkitbenchmarker/providers/aws/aws_network.py
```diff
@@ -64,6 +64,28 @@ def __init__(self):
     self.firewall_set = set()
     self._lock = threading.Lock()
 
+  def AllowIcmp(self, vm):
+    """Opens the ICMP protocol on the firewall.
+    Args:
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    print("ALLOWING ICMP")
```
miketruty, line 72: Why do you print status here and not in gce_network.py?

dPhanekham: Forgot to remove a debug statement. It is gone now

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -103,7 +103,7 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if not sending_vm.IsReachable(receiving_vm):
+  if not sending_vm.IsReachable(receiving_vm) and ip_type == 'internal':
```
miketruty, line 106: Since IsReachable() is only valid on ip_type internal, consider inverting order of clauses?

dPhanekham: changed

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -103,7 +103,7 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if not sending_vm.IsReachable(receiving_vm):
+  if ip_type == 'internal' and not sending_vm.IsReachable(receiving_vm):
```
dlott, line 106: Why isn't this cased the same as `IpAddressSubset.INTERNAL`

perfkitbenchmarker/providers/aws/aws_network.py
```diff
@@ -64,6 +64,27 @@ def __init__(self):
     self.firewall_set = set()
     self._lock = threading.Lock()
 
+  def AllowIcmp(self, vm):
+    """Opens the ICMP protocol on the firewall.
+    Args:
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    entry = (-1, -1, vm.region, vm.group_id)
```
dlott, line 72: Other `firewall_set` entries include cidr as the fifth tuple.  Should we have different firewall_set for tcp/udp and icmp?

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -586,6 +586,30 @@ def DisallowAllPorts(self):
...
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    if vm.is_static:
+      return
+    with self._lock:
+      firewall_name = ('perfkit-firewall-%s-%d-%d' %
+                       (FLAGS.run_uri, -1, -1))
+
+      key = (vm.project, -1, -1)
+      if key in self.firewall_rules:
```
dlott, line 602: similar question about comingling tcp/udp and icmp rules in the same dictionary.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: It's just a string passed by the Run() function directly above it, but I can change it to INTERNAL/EXTERNAL if that is more consistent

perfkitbenchmarker/providers/aws/aws_network.py
dPhanekham: Changed to add cidr as a fifth tuple. I don't think we need a separate firewall set for ICMP, they are all firewall rules and with the way the entries are formatted, there shouldn't be duplicates

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dlott: Hardcoding the same constant in two different places is not as good as using the same shared constant.  You could define a new one here, but if there's already a enum that does what you want, seems better to use it.

perfkitbenchmarker/providers/aws/aws_network.py
dlott: They both (int, int, string, string, string) Seems like there could be duplicates.  Please either use separate sets or add the protocol to the tuple, or add tests to prevent future maintainers from breaking the implicit formatting assumptions.

dPhanekham: I see where the overlap could be now. I've moved icmp firewall rules to their own sets for both aws and gcp

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: Changed to use existing enum

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -103,11 +107,11 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   Returns:
     A list of samples, with one sample for each metric.
   """
-  if not sending_vm.IsReachable(receiving_vm):
+  if ip_type == vm_util.IpAddressSubset.INTERNAL and not sending_vm.IsReachable(receiving_vm):
```
dlott, line 110: Please update the method documentation.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -103,11 +107,11 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
...
     logging.warn('%s is not reachable from %s', receiving_vm, sending_vm)
     return []
 
-  logging.info('Ping results (ip_type = %s):', ip_type)
+  logging.info('Ping results (ip_type = %s):' % ip_type)
```
dlott, line 114: Please revert this change (we have a different internal linter that complains about `logging-not-lazy`)  We're working to align both of them...

perfkitbenchmarker/providers/aws/aws_network.py
```diff
@@ -62,8 +62,32 @@ class AwsFirewall(network.BaseFirewall):
...
+    self.firewall_icmp_set = set()
     self._lock = threading.Lock()
 
+  def AllowIcmp(self, vm):
+    """Opens the ICMP protocol on the firewall.
+    Args:
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    source = '0.0.0.0/0'
+    # type, code, region, group_id, cidr
```
dlott, line 74: This doesn't match the fields of the tuple?

dPhanekham: In the aws cli, when specifying an ICMP firewall rule, the from-port and to-port numbers actually specify the type and code of ICMP packets allowed, with -1 meaning all types/codes.

https://docs.aws.amazon.com/cli/latest/reference/ec2/authorize-security-group-ingress.html
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: updated the documentation to reflect the enum being used for ip_type

perfkitbenchmarker/providers/aws/aws_network.py
dlott: Ahhh. Both the API is confusing (can you include a note that port=type and link to the CLI) and you've got `code` in the tuple, but not in `authorize_cmd`.  Since we've got separate sets now, you should just be able to drop it from the tuple, right?

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -585,6 +586,30 @@ def DisallowAllPorts(self):
...
+
+    Args:
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    if vm.is_static:
+      return
+    with self._lock:
+      firewall_name = ('perfkit-firewall-icmp-%s-%d-%d' %
+                       (FLAGS.run_uri, -1, -1))
+      key = (vm.project, -1, -1)
```
dlott, line 603: There's no specifying of ports/types/codes here?  Can we just remove the -1 from the key and the name?

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: change reverted

perfkitbenchmarker/providers/aws/aws_network.py
dPhanekham: I've inserted a comment explaining the port=type connection and a included a link. 

I was going to leave the type and code in the tuple in case anyone wanted to add the ability to specify the type and code in the future, but I've removed it now for the sake of simplicity.

perfkitbenchmarker/providers/gcp/gce_network.py
dPhanekham: Yes. The icmp firewall stuff in the gcloud cli doesn't have the same quirks as the aws one, so I've removed the ports from the key. 

I also added the vm cidr to the key, similar to the AllowPort() method so it doesn't have any collisions when there are multiple networks

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -22,6 +22,7 @@
 from perfkitbenchmarker import configs
 from perfkitbenchmarker import flags
 from perfkitbenchmarker import sample
+from perfkitbenchmarker import vm_util
 import re
 
 flags.DEFINE_boolean('ping_also_run_using_external_ip', False,
```
dlott, line 28: Per conversation, please use the ShouldRunOnExternalIpAddress() style.

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -113,7 +120,13 @@ def _RunPing(sending_vm, receiving_vm, receiving_ip, ip_type):
   stats = re.findall('([0-9]*\\.[0-9]*)', stdout.splitlines()[-1])
   assert len(stats) == len(METRICS), stats
   results = []
-  metadata = {'ip_type': ip_type,
+
+  if ip_type == vm_util.IpAddressSubset.INTERNAL:
+    ip_string = 'internal'
```
dlott, line 125: Can you unify this with the constants in netperf_benchmark?  Defining them in vm_util seems reasonable for now.

perfkitbenchmarker/providers/gcp/gce_network.py
```diff
@@ -585,6 +586,37 @@ def DisallowAllPorts(self):
...
+      vm: The BaseVirtualMachine object to open the ICMP protocol for.
+    """
+    if vm.is_static:
+      return
+    with self._lock:
+      if vm.cidr:  # Allow multiple networks per zone.
+        cidr_string = network.BaseNetwork.FormatCidrString(vm.cidr)
+        firewall_name = ('perfkit-firewall-icmp-%s-%s' %
+                         (cidr_string, FLAGS.run_uri))
+        key = (vm.project, vm.cidr)
```
cwilkes, line 605: It looks like GCEFirewall doesn't use vm.cidr but vm.network.network_resource.name when generating a rule.
I take it there's a 1:1 mapping between cidr and the network name?

dPhanekham: Yes, I tried to implement this in the same way as in the AllowPort method.

If the cidr string is 10-0-1-0-24, then the network resource name would be something like
pkb-network-multi-10-0-1-0-24-(run-uri).

I have also tested and confirmed that this works with all of the network configs in gcp_network_test

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: I've made a new enum in vm_util and changed netperf,iperf,ping to use it

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dPhanekham: Changed to use the style we discussed

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
```diff
@@ -25,10 +25,6 @@
 from perfkitbenchmarker import vm_util
 import re
 
-flags.DEFINE_boolean('ping_also_run_using_external_ip', False,
```
dlott, line 27: Looks like @miketruty 's new tutorials used this flag.  
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/d486aa9ba587c44dc3edc43f75b8fa0d7222f1f7/tutorials/inter_region_reports/data/simple_region_latency.yaml#L16

Are they working currently given that we don't have the firewalls in place?  
Preferences on updating these all together or seperately?

dPhanekham: There is currently an error when using external IPs in the tutorial, which is what originally prompted this PR. I'll also update the tutorial to use the new stuff

perfkitbenchmarker/linux_benchmarks/ping_benchmark.py
dlott: Please update the breaking changes section with the deleted flag and new usage.  Otherwise we'll let @cwilkes take one last look.

dPhanekham: I've added an entry in the breaking changes section

PR #2211, opened 2020-04-16
Remove "default" keyword from AWS and Azure boot disk size

florinpapa:
The default value for the boot_disk_size field in the AWS and Azure providers is incompatible with the comments in their VmSpec:
```
`boot_disk_size: None or int. The size of the boot disk in GB.`
```
Instead of `None`, the default value is set to `'default'` in the metadata. This mismatch between the default value type (string) and the value when a custom size is provided (int) can cause issues when trying to insert the metadata to a database.

Tested on both AWS and Azure using the sample_benchmark, by adding the following lines in perfkitbenchmarker/linux_benchmarks/sample_benchmark.py, after the BENCHMARK_DATA dict:
```
BENCHMARK_DATA_URL = {
    'preprovisioned_data.txt':
        'https://github.com/florinpapa/preprovisioned_data/raw/master/preprovisioned_data.txt'}
```
Then, I used the following commands to start the benchmark (which worked):
```
./pkb.py --cloud=Azure --benchmarks=sample
./pkb.py --cloud=AWS --benchmarks=sample --machine_type=m5.xlarge
```

### DIFF
--- a/CHANGES.next.md
+++ b/CHANGES.next.md
@@ -101,4 +101,5 @@
         (to 20GB).
     -   The current defaults leave other Linux OSes at 10GB and Windows OSes at
         50GB.
+-   Remove 'default' keyword from AWS and Azure boot_disk_size
 -   Explicitly setting Netperf to Python 3.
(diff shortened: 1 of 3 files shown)


florinpapa:
@pmkc @dlott thanks for approving.

@pmkc I am converting the pkb output to JSON, then inserting the JSON fields to the DB. I am expecting an int for `boot_disk_size`, but getting the `default` string instead, causing a conflict. There is no conflict if the value in metadata is kept as None, because this will get converted to null in JSON, which is allowed.

florinpapa:
@pmkc @dlott I see that the `feedback/copybara` check is failing for this, but I don't have access to details, in order to fix it.

Can you please let me know what is needed from my side in order to get it working?

dlott:
@bvliu for copybara troubleshooting.

PR #2224, opened 2020-04-27
Check ycsb proportion explicitly against none.

justinuang:
When I set ycsb_readproportion to 0.0, it ended up not setting the readproportion on the underlying ycsb command. This is because we are using if FLAGS.ycsb_readproportion to check if the flag is set, but the flag could be set explicitly to 0.0 and still fail the check.


justinuang:
@googlebot I fixed it.

PR #2225, opened 2020-04-28
Use PreprovisionedData for downloading generic content and apply it for the Maven package

florinpapa:
Using the `PreprovisionedData` feature was tied to the implementation of `ShouldDownloadPreprovisionedData`, which was not implemented in the base class and prevented from using it on static instances or cloud providers that did not implement it. This feature is useful because it establishes a uniform way of downloading benchmark/package content (source code, deb files, rpm files etc.), that also includes a checksum verification. This also provides an easy way to pre-download these resources on the PerfKitBenchmarker host and prevent them from being downloaded every time on the system under test.

This PR adds the following changes:
*  Add the `--preprovision_ignore_checksum` flag that can be used when downloading content that is not registered in `PREPROVISIONED_DATA` (e.g. a version different than the default)
*  Return `False` in the `ShouldDownloadPreprovisionedData` method in virtual_machine.py instead of raising a `NotImplemented` error, to allow static instances and cloud providers that don't implement this method to use the `PreprovisionedData` feature
*  Implement this feature for the Maven package
*  Add a few other changes to the Maven package to make it more robust (retrieving JAVA_HOME, adding the option to provide a custom mirror via the `--maven_mirror_url` flag, fixing a java lib issue observed on CentOS)

Tested using the following commands:
```
./pkb.py --cloud=AWS --benchmarks=redis_ycsb --os_type=centos7 --machine_type=m5.2xlarge --ycsb_version=0.17.0 --openjdk_version=8
./pkb.py --cloud=AWS --benchmarks=redis_ycsb --os_type=ubuntu1804 --machine_type=m5.2xlarge --ycsb_version=0.17.0 --openjdk_version=8
```


perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -70,22 +96,48 @@ def _Install(vm):
...
+    vm.RemoteCommand('mkdir -p ~/.m2')
+    vm.RenderTemplate(settings_local_path, settings_remote_path, context)
+
 
 def Uninstall(vm):
+  java_home = _GetJavaHome(vm)
+  java_lib_sym = posixpath.join(java_home, '..', 'lib')
+  vm.RemoteCommand('sudo rm -f {0}'.format(java_lib_sym))
   vm.Uninstall('openjdk')
+  vm.Uninstall('curl')
```
pmkc, line 141: What if the OS came with curl? seems better to leave it.

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -70,22 +96,48 @@ def _Install(vm):
...
+                        MVN_DIR, maven_remote_path))
+
+  java_home = _GetJavaHome(vm)
 
   # Set env variables for maven
   maven_env = MVN_ENV.format(java_home=java_home, maven_home=MVN_DIR)
   cmd = 'echo "{0}" | sudo tee -a {1}'.format(maven_env, MVN_ENV_PATH)
   vm.RemoteCommand(cmd)
 
+  # On CentOS, the mvn command looks for tools.jar in $JAVA_HOME/../lib
```
pmkc, line 120: This seems a little inelegant, and in my testing unnecessary.
 
On A CentOS 8 VM mvn 3.6.3 found the tools jar of OpenJDK 8 regardless of whether JAVA_HOME was unset, set to /usr/lib/jvm/java/ or /usr/lib/jvm/java/jre

perfkitbenchmarker/virtual_machine.py
```diff
@@ -1135,7 +1143,7 @@ def ShouldDownloadPreprovisionedData(self, module_name, filename):
     Returns:
       A boolean indicates if preprovisioned data is available.
     """
-    raise NotImplementedError()
+    return False
```
pmkc, line 1146: I have no objection to this, but I don't see how it helps. If you are using a static VM, how can you trigger preprovisioned data (I'm not completely familiar with the mechanism)?

florinpapa: The `InstallPreprovisionedData` functionality will search for preprovisioned data using the cloud-specific implementation for `DownloadPreprovisionedData`. If the resource is not found and a URL is provided for the resource, it will default to downloading it using `wget`.

My goal is to allow the execution to default to downloading from a URL, even if static VMs or the current cloud provider does not implement `DownloadPreprovisionedData`. This will offer a uniform way of downloading resources, alongside checksum verification.

The issue is that static VMs or cloud providers need to implement `ShouldDownloadPreprovisionData`, otherwise this method here is called, throwing a `NotImplemented` error. Static VMs currently don't implement this method and only a few cloud providers do. The alternative to my current change is to implement `ShouldDownloadPreprovisionedData` and `DownloadPreprovisionedData` for all possible targets, which seems a bit excessive to me.

@pmkc please let me know if this answers your question.

perfkitbenchmarker/linux_packages/maven.py
florinpapa: @pmkc fair enough. I have removed this line.

perfkitbenchmarker/linux_packages/maven.py
florinpapa: @pkmc this is something I experienced on a particular CentOS 7 static VM, with maven 3.6.1. I haven't been able to reproduce it on any cloud provider.

florinpapa: I have removed this, we can revisit in the future if anyone has issues with the lib path.

florinpapa:
@pmkc I have made the required changes. Please let me know if there is anything else needed from my side to get this in.

florinpapa:
Can I please get additional feedback on this?

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -16,10 +16,14 @@
...
 
 from perfkitbenchmarker import flags
+from perfkitbenchmarker import data
 from perfkitbenchmarker.linux_packages import INSTALL_DIR
 from six.moves.urllib.parse import urlparse
 
-flags.DEFINE_string('maven_version', '3.6.3', 'The version of maven')
+flags.DEFINE_string('maven_version', '3.6.3',
+                    'The version of maven')
+flags.DEFINE_string('maven_mirror_url', '',
```
dlott, line 25: Why empty string instead of None as the default?

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -70,20 +96,37 @@ def _Install(vm):
...
+                        MVN_DIR, maven_remote_path))
+
+  java_home = _GetJavaHome(vm)
 
   # Set env variables for maven
   maven_env = MVN_ENV.format(java_home=java_home, maven_home=MVN_DIR)
   cmd = 'echo "{0}" | sudo tee -a {1}'.format(maven_env, MVN_ENV_PATH)
   vm.RemoteCommand(cmd)
 
+  if FLAGS.maven_mirror_url != '':
```
dlott, line 120: if FLAGS.maven_mirror_url:

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -70,20 +96,37 @@ def _Install(vm):
...
+    settings_local_path = data.ResourcePath(posixpath.join(
```
dlott, line 121: You want ospath.join since this is on the machine you're running PKB from.

perfkitbenchmarker/virtual_machine.py
dlott: @yuyantingzero any thoughts on preprovisioned data changes?

perfkitbenchmarker/linux_packages/maven.py
dlott: Also, @yuyantingzero can/should we require this to be https for security?

florinpapa: @dlott good point. I have made the default to be `None`.

As far as requiring https, I see that all the mirrors specified in the [maven documentation](https://repo.maven.apache.org/maven2/.meta/repository-metadata.xml) are actually http. Mirrors are also often used internally, on private networks, and setting something like this up with https would probably be a bit complicated.

florinpapa:
@dlott thank you for reviewing. I have addressed all the comments and re-tested the changes. Looking forward to additional feedback.

perfkitbenchmarker/linux_packages/maven.py
yuyantingzero: HTTPs is preferred, but IIRC, some packages is only available from HTTP mirror.

One hack I did when not able of find HTTPs source is building from source code directly.

Correct me if I am wrong, maven is available from github: https://github.com/apache/maven/tree/maven-3.6.3. 
You can potentially skip the compilation and download the binary zip directly from github. https://github.com/apache/maven/archive/maven-3.6.3.zip

florinpapa: @yuyantingzero the `maven_mirror_url` does not refer to a location for downloading maven, but for downloading certain Maven artifacts and plugins, as specified by the target project (being built using Maven) in their pom.xml file. This is a Maven setting, and more details on Maven mirrors can be found [here](http://maven.apache.org/guides/mini/guide-mirror-settings.html).

Maven has a default mirror from which it downloads these artifacts/plugins, but sometimes the user might want to use a different (internal) mirror, to speed things up, hence this flag.

Please let me know if you have additional changes and if I should change the name of the flag to be more descriptive.

yuyantingzero: I see. Thanks for the explanation.
Looks good to me.
If user choose a non-default mirror, it is their responsibility to ensure security.

florinpapa: Sounds good, thanks for the feedback.

perfkitbenchmarker/virtual_machine.py
florinpapa: @yuyantingzero makes sense. I am now catching the `NotImplemented` exception, instead of setting the default to False.

florinpapa:
@yuyantingzero  thank you for the feedback. I have added the requested changes. Please let me know if there is anything else I should address.

florinpapa:
I see that the `feedback/copybara` check is failing, and I cannot see anything when I go to Details. Please let me know if there is any additional change I need to make in order to pass that check.

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -70,20 +97,37 @@ def _Install(vm):
...
-      ('mkdir {0} && curl -L {1} | '
-       'tar -C {0} --strip-components=1 -xzf -').format(MVN_DIR, maven_url))
-
-  # Get JAVA_HOME
-  out, _ = vm.RemoteCommand('readlink -f `which java`')
-  out = out.strip()
-  java_home = out[:out.index('/jre')]
+  maven_tar = maven_url.split('/')[-1]
+  if maven_tar not in PREPROVISIONED_DATA:
+    PREPROVISIONED_DATA[maven_tar] = ''  # will only work with preprovision_ignore_checksum
```
yuyantingzero, line 102: this line is too long, move comments to a separate line.

perfkitbenchmarker/linux_packages/maven.py
```diff
@@ -14,12 +14,17 @@
 """Module containing maven installation functions."""
 
 import posixpath
+import os
```
yuyantingzero, line 17: Reorder the import:

import os
import posixpath

from perfkitbenchmarker import data
from perfkitbenchmarker.linux_packages import INSTALL_DIR
from six.moves.urllib.parse import urlparse

tests/linux_packages/maven_test.py
```diff
@@ -16,7 +16,9 @@
 
 import unittest
 import mock
+import posixpath
```
yuyantingzero, line 19: import posixpath
import unittest

import mock
from perfkitbenchmarker.linux_packages import INSTALL_DIR
from perfkitbenchmarker.linux_packages import maven
from perfkitbenchmarker.tests import pkb_common_test_case

florinpapa: This
```
from perfkitbenchmarker.tests import pkb_common_test_case
```
will give me the error
```
ModuleNotFoundError: No module named 'perfkitbenchmarker.tests'
```
because the tests are not under perfkitbenchmarker, but in the root dir of the repo.

Also, why is mock not part of the first import block? Like:
```
import mock
import posixpath
import unittest

from perfkitbenchmarker.linux_packages import INSTALL_DIR
from perfkitbenchmarker.linux_packages import maven
from tests import pkb_common_test_case
```

florinpapa: @yuyantingzero I am guessing you are using `flake8-import-order` for internal tests, which explains the ordering issues in my PR. Since I am not aware of the ordering rules you use, I made the exact changes you requested, except the last module that was causing the import error. Hope it works now.

florinpapa:
Still failing. I will be happy to make the necessary changes if you can point me in the right direction.

yuyantingzero:
Don't worry, the failure this time is caused by known issue of internal tool we use.
It is now merged!

Issue #2233, opened 2020-05-05
tempdir generation breaks SSH on macos


bleything:
The default `temp_dir` path on macos is too long and causes SSH to fail. This manifests as PKB looping forever on the connection check step. Copying/pasting the command shows the error:

```
$ ssh -A -p 22 perfkit@[redacted] -2 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o IdentitiesOnly=yes -o PreferredAuthentications=publickey -o PasswordAuthentication=no -o ConnectTimeout=5 -o GSSAPIAuthentication=no -o ServerAliveInterval=30 -o ServerAliveCountMax=10 -i /var/folders/tp/vv4yshsd70qfn9qlfrsnl7_000lbtv/T/perfkitbenchmarker/runs/609d5015/perfkitbenchmarker_keyfile -o   ControlPath="/var/folders/tp/vv4yshsd70qfn9qlfrsnl7_000lbtv/T/perfkitbenchmarker/runs/609d5015/ssh/%h" -o ControlMaster=auto -o ControlPersist=30m hostname
unix_listener: path "/var/folders/tp/vv4yshsd70qfn9qlfrsnl7_000lbtv/T/perfkitbenchmarker/runs/609d5015/ssh/[redacted].YVBXsmnvZVRPpdlR" too long for Unix domain socket
```

This is mostly macos's fault and not PKB's but it does break PKB on the mac. One possible solution is shortening the suffix from 'perfkitbenchmarker' to 'pkb'... this works for me but I'm not sure it's the best solution. I'll put in a PR for it just in case.

yuyantingzero:
Can you try --temp_dir?

dlott:
Recommend adding check on the length of control path and throw an error on MacOS if it's too long and suggest using `--temp_dir`.

PR #2234, opened 2020-05-05
shorten tempdir suffix

bleything:
tl;dr: tempdir paths are too long on macos, which breaks SSH. See #2233 for details.

This solution works on my machine but it might not be the ideal fix. I don't know enough about macos internals to suggest a better solution, though.

### DIFF
--- a/perfkitbenchmarker/temp_dir.py
+++ b/perfkitbenchmarker/temp_dir.py
@@ -30,7 +30,7 @@ if six.PY2:
 else:
   import functools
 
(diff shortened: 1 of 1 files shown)


yuyantingzero:
Commented in issue, maybe use --temp_dir instead.
The change is still fragile on MacOS. In addition, I am not sure MacOS is fully supported.

bleything:
@yuyantingzero: that's legit. I'm no longer sure this is a good change. I cc'd you on the internal bug, we can discuss further there. I'm going to close this.

PR #2327, opened 2020-06-19
Debian9 SpecCPU 2017 ISO support

mikeweng:
Added support for SpecCPU2017 with ISO support that targets to run on Debian9. 

* ISO mount is missing `config/` and `config/flags` directories and `config/flags/clang.xml` file. Added the mkdir and push file command during prepare. 
* Added Debian9 Static VM to allow staticvm config to specify os_type=debian9. And skips reinstall build_tools when targeting version 6.
* Respect user specified mount directory when trigger install.sh.


mikeweng:
@yuyantingzero

PR #2848, opened 2021-04-07
Cloudharmony network benchmark uplink test fix

dPhanekham:
These changes to the nginx server configuration make the uplink test function properly.
I set a static max size of 100 Mbits, but I could change this to being based on the --ch_network_throughput_size flag

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
@@ -141,7 +141,7 @@ def _PrepareServer(vm):
   vm.Install(NGINX)
 
   # Required to make uplink test function properly
-  vm.RemoteCommand('sudo sed -i "/server_name _;/a error_page  405 =200 $uri;" /etc/nginx/sites-enabled/default')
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
```diff
@@ -139,6 +139,12 @@ def _PrepareServer(vm):
     vm: the server virtual machine.
   """
   vm.Install(NGINX)
+
+  # Required to make uplink test function properly
+  vm.RemoteCommand('sudo sed -i "/server_name _;/a error_page  405 =200 $uri;" /etc/nginx/sites-enabled/default')
```
jellyfishcake, line 144: why do you need to set the error_page flag? Also does this set error to return 200? (not error)

jellyfishcake (commented):
I think it is fine to leave client max body size as a default and not a flag.

perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
dPhanekham: This was the solution I found in several places for enabling POST requests on static pages with nginx. Honestly, it seems kinda hacky, but it works to allow the uplink test to function properly.

jellyfishcake (approved):

PR #2896, opened 2021-05-03
Add support for DNS tests in cloudharmony_network benchmark

dPhanekham:
This PR adds support to run DNS test in cloudharmony_network benchmark.

Basically, DNS tests would only run if ch_network_test_service_type = dns (or if this argument is missing completely), so I added DNS as a test_service_type and put in a few errors that tell the user both ch_network_test_service_type and ch_network_test flags have to be set to 'dns' for it to work.

I've added the 'dns_recursive' flag to the cloudharmony run.sh command when running a dns test. It fails without this unless you explicitly hand it the IP of your DNS server, which we could alternatively do.

I also changed the test_endpoint from the ip address to the internal domain name for dns tests. This is necessary, as the whole point of the test is to time how long the DNS query for the server's domain name takes. This is currently only implemented for GCP.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
@@ -47,6 +47,7 @@ cloudharmony_network:
 # network test service types
 COMPUTE = 'compute'
 STORAGE = 'storage'
+DNS = 'dns'
 # network compute server type
 NGINX = 'nginx'
 # tests accepted by the --test flag of the gartner network benchmark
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
```diff
@@ -110,6 +111,12 @@ def CheckPrerequisites(_):
   # TODO(user): add AWS & Azure support for object storage
   if FLAGS.ch_network_test_service_type == STORAGE and FLAGS.cloud != 'GCP':
     raise NotImplementedError('Benchmark only supports GCS object storage.')
+  elif FLAGS.ch_network_test_service_type == DNS and FLAGS.ch_network_test != 'dns':
```
shyamsabhaya, line 114: Combine with below if statement: `if (`FLAGS.ch_network_test_service_type == DNS xor FLAGS.ch_network_test == DNS):`

perfkitbenchmarker/linux_benchmarks/cloudharmony_network_benchmark.py
```diff
@@ -306,6 +315,15 @@ def _Run(benchmark_spec, test):
     http_url = f'http://{benchmark_spec.bucket}.storage.googleapis.com/probe'
     endpoints = f'--test_endpoint={http_url}'
     _AddStorageMetadata(client, metadata)
+  elif FLAGS.ch_network_test_service_type == DNS:
+    vms = vm_groups['server']
+    # use GCP zonal internal DNS, but maybe should add domain to vm's data attributes?
```
shyamsabhaya, line 320: Hmm. yeah we probably want to set up a Domain. I don't know if GCP zonal internal DNS is what cloud harmony is interested in. @jellyfishcake  - Thoughts?

shyamsabhaya (changes requested):

PR #3499, opened 2022-04-22
Adding support for newer versions of bonnie++

elsonrodriguez:
There has been a schema change in the bonnie++ csv format, this adds support for the newer schema while maintaining support for older schemas.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/bonnie_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/bonnie_benchmark.py
@@ -39,6 +39,8 @@ bonnieplusplus:
 """
 
 LATENCY_REGEX = r'([0-9]*\.?[0-9]+)(\w+)'
+
+
 # Bonnie++ result fields mapping, see man bon_csv2txt for details.
 BONNIE_RESULTS_MAPPING_1_96 = {
     'format_version': 0,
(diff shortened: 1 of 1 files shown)


yuyantingzero:
Thanks for the contribution. The code looks fine. 

However, since bonnie++ is a really old test, imo it is reasonable to drop support for older version. Is it possible to update the installation package to use a fixed version instead.

elsonrodriguez:
> Thanks for the contribution. The code looks fine.
> 
> However, since bonnie++ is a really old test, imo it is reasonable to drop support for older version. Is it possible to update the installation package to use a fixed version instead.

The error came up while using a newer version of ubuntu. The default version of Debian that is spun up on GCE still uses the old bonnie. It will be necessary to keep compatibility with both until all distros ship with newer bonnie.

perfkitbenchmarker/linux_benchmarks/bonnie_benchmark.py
```diff
@@ -223,10 +277,23 @@ def ParseCSVResults(results):
         If a 4th element is included, it is a dictionary of sample
         metadata.
   """
+  results = results.split(',')
+
+  if results[0] == "1.98":
```
yuyantingzero, line 282: suggest make a  dictionary. e.g.
supported_parser = {'1.96': BONNIE_RESULTS_MAPPING_1_96, '1.97': ..._1_97, ..}

Such that you can simply here, rather than copy-pasting the log lines 
version = results[0]
if version in supported_parser:
  logging...
  BONNIE_RESULTS_MAPPING = supported_parser[version]
else:
  raise ...

elsonrodriguez: That's a better approach, done.

elsonrodriguez:
@yuyantingzero Addressed previous feedback, ready for another review?

yuyantingzero (approved):

yuyantingzero:
Seems great, thanks!

PR #3676, opened 2022-07-30
Add iPerf interval reporting

dPhanekham:
This adds to option for the iPerf benchmark to output throughput and other stats at periodic intervals during the test. 
This allows us to see how throughput changes over the course of an iPerf test. 
The interval outputs are parsed and stored as lists in the metadata. When not using interval reporting, the output is unchanged from how is was previously. 

I've also added a iperf_sleep_time flag to optionally sleep for a period of time after each iPerf test. This can help ensure that one test does not effect a subsequent test.


perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
```diff
@@ -171,74 +184,190 @@ def _RunIperf(sending_vm, receiving_vm, receiving_ip_address, thread_count,
...
+        total_stats['rtt'] = sum(rtt_avg_list)/len(rtt_avg_list)
+        total_stats['cwnd'] = sum(cwnd_avg_list)/len(cwnd_avg_list)
+        total_stats['netpwr'] = sum(netpwr_sum_list)/len(netpwr_sum_list)
+        total_stats['rtt_unit'] = thread_results_for_interval[0]['rtt_unit']
+
+        total_throughput = total_stats['throughput']
+
+      elif thread_count == 1:
+        
+        r = re.compile((
```
andyz422, line 258: This seems to be duplicated from the single-thread regex. I would suggest putting this before 'if thread_count > 1'.

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
```diff
@@ -171,74 +184,190 @@ def _RunIperf(sending_vm, receiving_vm, receiving_ip_address, thread_count,
...
+        
+        total_stats = interval_sums[len(interval_sums)-1]
+        total_stats['rtt'] = sum(rtt_avg_list)/len(rtt_avg_list)
+        total_stats['cwnd'] = sum(cwnd_avg_list)/len(cwnd_avg_list)
+        total_stats['netpwr'] = sum(netpwr_sum_list)/len(netpwr_sum_list)
+        total_stats['rtt_unit'] = thread_results_for_interval[0]['rtt_unit']
+
+        total_throughput = total_stats['throughput']
+
+      elif thread_count == 1:
```
andyz422, line 256: I suggest adding a check for thread_count < 1, so that the error won't occur when creating tcp_metadata.

andyz422 (approved):
I left a couple suggestions for minor changes. Looks good to me.

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
dPhanekham: I've moved the single thread logic to before the multithread logic. The regex statements are all similar, but they are slightly different for using intervals single/multi thread and not using intervals single/multithread.

perfkitbenchmarker/linux_benchmarks/iperf_benchmark.py
dPhanekham: I have added a check in _Run to ignore thread_count that are less than 1

bvliu (approved):

PR #3703, opened 2022-08-11
Add stats and cost calculator to GCP Dataflow provider

rarsan:
Include sample config file for wordcount benchmark


perfkitbenchmarker/configs/dataflow_wordcount.yaml
```diff
@@ -0,0 +1,85 @@
...
+    dpb_service: { service_type: dataflow, worker_count: 1, worker_group: *four_core },
+    flags: *myflags
+  }
+- dpb_wordcount_benchmark: {
+    dpb_service: { service_type: dataflow, worker_count: 1, worker_group: *two_core },
+    flags: *myflags
+  }
+
+
+# dpb_wordcount_benchmark:
```
dorellang, line 57: Remove commented lines.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+        "name": project_name,
+        "filter": api_filter,
+        "interval": interval,
+        "view": monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+      }
+    )
+
+    return self._GetLastValueFromTimeSeries(results)
+
+  def _GetLastValueFromTimeSeries(self, time_series):
```
dorellang, line 428: Rewrite the method:
```python
  def _GetLastValueFromTimeSeries(self, time_series):
    try:
      return time_series[0].points[0].value.int64_value
    except IndexError:
      return None
```

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+    interval = TimeInterval(
+        {
+            "end_time": {"seconds": seconds},
+            "start_time": {"seconds": seconds - interval_length * 60},
+        }
+    )
+
+    api_filter = (
+      'metric.type = "pubsub.googleapis.com/subscription/num_undelivered_messages" '
+      'AND resource.labels.subscription_id = "' + subscription_name + '" '
```
dorellang, line 414: prefer f-strings:  https://google.github.io/styleguide/pyguide.html#310-strings

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+        "view": monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+        "aggregation": aggregation,
+      }
+    )
+
+    # if not results:
+    #   raise Exception('No monitoring data found. Unable to calculate avg CPU utilization')
+    return self._GetAvgValueFromTimeSeries(results)
+
+  def GetMaxOutputThroughput(self, ptransform: str, start_time: datetime, end_time: datetime):
```
dorellang, line 339: Does this actually get used in this class? If not I suggest moving it to the subclass that will be in the subsequent PR.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+      }
+    )
+
+    if not results:
+      logging.warn('No monitoring data found. Unable to calculate max throughput.')
+      return None
+
+    return self._GetMaxValueFromTimeSeries(results)
+
+  def GetSubscriptionBacklogSize(self, subscription_name, interval_length=4):
```
dorellang, line 398: Does this actually get used in this class? If not I suggest moving it to the subclass that will be in the subsequent PR.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+      averaged = sum(points) / len(points)
+      # If metric unit is a fractional number between 0 and 1 (e.g. CPU utilization metric)
+      # multiply by 100 to display a percentage usage.
+      if time_series.unit == "10^2.%":
+        averaged = round(averaged * 100, 2)
+      return averaged
+    
+    return None 
+
+  def _GetMaxValueFromTimeSeries(self, time_series):
```
dorellang, line 463: Does this actually get used in this class? If not I suggest moving it to the subclass that will be in the subsequent PR.

(Noticed this is only used in GetMaxOutputThroughput, which is unused in this PR).

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: Does this actually get used in this class? If not I suggest moving it to the subclass that will be in the subsequent PR.

(Noticed this is only used in GetSubscriptionBacklogSize, which is unused in this PR).

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
     basic_data = super(GcpDpbDataflow, self).GetMetadata()
     basic_data['dpb_dataflow_runner'] = FLAGS.dpb_dataflow_runner
     basic_data['dpb_dataflow_sdk'] = FLAGS.dpb_dataflow_sdk
+    basic_data['dpb_job_id'] = self.job_id
     return basic_data
+
+  def GetJobStatus(self):
```
dorellang, line 192: Remove unused no-op method.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+    cmd = util.GcloudCommand(self, 'dataflow', 'jobs', 'show', self.job_id)
+    cmd.flags = {
+        'project': self.project,
+        'format': 'json',
+    }
+
+  def GetStats(self):
```
dorellang, line 199: Do you need to actually refresh self.job_stats at some point in the lifetime of this class? If not, I would recomend rewriting this method as a [cached property](https://docs.python.org/3/library/functools.html#functools.cached_property).

E.g:
```python
@functools.cached_property
def job_stats(self):
  ...
```
Then update all the references

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+    """Collect series of relevant performance and cost stats."""
+    stats = {}
+    stats['total_vcpu_time'] = self.GetMetricValue('TotalVcpuTime')/3600         # vCPU-hr
+    stats['total_mem_usage'] = self.GetMetricValue('TotalMemoryUsage')/1024/3600 # GB-hr
+    stats['total_pd_usage'] = self.GetMetricValue('TotalPdUsage')/3600           # GB-hr
+    # TODO(rarsan): retrieve BillableShuffleDataProcessed
+    # and/or BillableStreamingDataProcessed when applicable
+    self.job_stats = stats
```
dorellang, line 207: If taking the cached_property route, you would want to remove this.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+    
+    cost += total_pd_usage * PD_PER_GB_HR
+    # TODO(rarsan): Add cost related to per-GB data processed by Dataflow Shuffle
+    # (for batch) or Streaming Engine (for streaming) when applicable
+    return cost
+
+  def _PullJobMetrics(self, force_refresh=False):
+    """Retrieve and cache all job metrics from Dataflow API"""
+    # Skip if job metrics is already populated unless force_refresh is True
+    if self.job_metrics is not None and not force_refresh:
```
dorellang, line 237: Also return if self.job_id is None

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -152,4 +186,303 @@ def GetMetadata(self):
...
+    value = None
+    for i, time_interval in enumerate(time_series):
+      if i != 0: break
+      for j, snapshot in enumerate(time_interval.points):
+        if j != 0: break
+        value = snapshot.value.int64_value
+
+    return value
+
+  def _GetAvgValueFromTimeSeries(self, time_series):
```
dorellang, line 438: Add typing annotation to time_series

perfkitbenchmarker/configs/dataflow_wordcount.yaml
```diff
@@ -0,0 +1,85 @@
+#################################
```
dorellang, line 1: Does this get used somewhere?

dorellang (changes requested):
Make all strings single quotes in gcp_dpb_dataflow.py (unless you use single quotes inside the string).

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Makes sense. Removing unused method in this PR.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Updated to cached property. However, that impacts the [reference in benchmark](https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/3703/files#diff-e1531d19e0eadfb8770d6c74873dec725e2eb5627da62dfa72e74dd635dddaefR184) as well. So every `dpb_service_instance` type should be aware of the `job_stats` property (vs `GetStats()` method) and ideally override it, is that a reasonable expectation/contract?

perfkitbenchmarker/configs/dataflow_wordcount.yaml
rarsan: This is the config YAML I use to run the benchmark e.g. `./pkb.py --cloud=GCP --project=$PROJECT_ID --benchmark_config_file=dataflow_wordcount.yaml` . I found it's a learning curve and tricky to get these flags right for new PKB users. Therefore providing this as a self-contained example which was referenced in the talk.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Should we raise an exception if job_id is None?
When self.job_id is None (which is not expected), self.job_metrics will be None, and then `GetMetricValue` will error out. I'm tempted to fail fast, rather than having a sequence of job runs proceed without any stats collected. What do you think?

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: While I do agree with your point, GetStats is not a standard dpb_service_instance method (i.e. is unique to the Dataflow implementation), so you if you leave it as-is you would face the same problem. This doesn't have any runtime consequences as the part of the benchmark code that access this property is guarded by a service_type check.

I know this is not pretty, but for me it's good enough if you leave a TODO in the benchmark code.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: I think that failing on job_id is None would make the code simpler so I support the decision, I am just worried that parsing it from a stderr human readable message is kinda brittle, isn't there a better way to do it?

perfkitbenchmarker/configs/dataflow_wordcount.yaml
dorellang: Awesome! Thank you for the addition.

perfkitbenchmarker/configs/dataflow_wordcount.yaml
dorellang: From the recent commit I noticed you added a comment saying this is actually an alternative set up. Sounds good.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Sounds good! Will leave the new `job_stats`

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Sounds good. Updated.
I agree job_id parsing is bit brittle since parent dataflow provider originally relies on java executable. For the subclass'ed dataflow template provider, it's more robust as job_is is returned as part of stdout response (JSON object) by gcloud command.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -160,4 +195,172 @@ def GetMetadata(self):
...
+    # (for batch) or Streaming Engine (for streaming) when applicable
+    return cost
+
+  def _PullJobMetrics(self, force_refresh=False):
+    """Retrieve and cache all job metrics from Dataflow API"""
+    # Skip if job metrics is already populated unless force_refresh is True
+    if self.job_metrics is not None and not force_refresh:
+      return
+    # Raise exception if job id not available
+    if self.job_id is None:
```
dorellang, line 239: Let's better raise this error in the SubmitJob method, where you parse the ID (to make it fail fast). Also I wouldn't use ValueError here, it implies that the error is raised because a bad argument was passed (which I don't think it's necessarily the case), a generic Exception would do fine here.

rarsan: As discussed, we'll leave as is: one can submit the job and let it be acceptable not to have a job id after that, but if they try to get metrics then they get the error.

dorellang: As per our internal discussion we'll leave this alone.

dorellang (approved):
LGTM

dorellang (approved):

dorellang (approved):

dorellang (approved):

dorellang (approved):

PR #3727, opened 2022-08-23
New GCP provider for dataflow template jobs (batch and streaming)

rarsan:
- New provider `gcp_dpb_dataflow_template.py`
- New benchmark for testing streaming job template `dpb_dataflow_template_benchmark.py`
- Sample YAML config `dataflow_template.yaml`
- Update existing provider `gcp_dpb_dataflow.py` with additional stats collection e.g. max throughput


perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    stdout, _, _ = cmd.Issue()
+
+    # Parse output to retrieve submitted job ID
+    try:
+      result = json.loads(stdout)
+      self.job_id = result['id']
+    except Exception as err:
+      logging.error('Failed to parse Dataflow job ID: {}'.format(err))
```
dorellang, line 103: Don't interpolate variables directly in strings to be logged. Refer to https://google.github.io/styleguide/pyguide.html#3101-logging

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+      job_poll_interval=None,
+      job_arguments=None,
+      job_input_sub = None):
+
+    worker_machine_type = self.spec.worker_group.vm_spec.machine_type
+    num_workers = self.spec.worker_count
+    max_workers = self.spec.worker_count
+
+    now = datetime.datetime.now()
+    job_name = template_gcs_location.split('/')[-1] \
```
dorellang, line 78: Prefer f-strings.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+
+  def _GetCompletedJob(self, job_id):
+    """See base class."""
+    job_input_sub = job_id
+
+    # Job completed if input subscription is empty *and* job is drained;
+    # otherwise keep waiting
+    if not self.input_sub_empty:
+      backlog_size = self.GetSubscriptionBacklogSize(job_input_sub)
+      logging.info('Polling: Backlog size of subscription {} is {}'
```
dorellang, line 119: Don't interpolate variables directly in strings to be logged. Refer to https://google.github.io/styleguide/pyguide.html#3101-logging

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+        self.input_sub_empty = True
+        # Start draining job once input subscription is empty
+        cmd = util.GcloudCommand(
+            self, 'dataflow', 'jobs', 'drain', self.job_id)
+        cmd.flags = {
+            'project': self.project,
+            'region': util.GetRegionFromZone(FLAGS.dpb_service_zone),
+            'format': 'json',
+        }
+        logging.info('Polling: Draining job {} ...'.format(self.job_id))
```
dorellang, line 131: Don't interpolate variables directly in strings to be logged. Refer to https://google.github.io/styleguide/pyguide.html#3101-logging

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -351,12 +357,94 @@ def GetAvgCpuUtilization(self, start_time: datetime, end_time: datetime):
...
+    )
+
+    if not results:
+      logging.warning(
+          'No monitoring data found. Unable to calculate max throughput.')
+      return None
+
+    return self._GetMaxValueFromTimeSeries(results)
+
+  def GetSubscriptionBacklogSize(self, subscription_name, interval_length=4):
```
dorellang, line 418: Move this method to the child class. It doesn't belong here conceptually.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -200,6 +199,13 @@ def GetMetadata(self):
     basic_data['dpb_job_id'] = self.job_id
     return basic_data
 
+  def GetJobStatus(self):
```
dorellang, line 202: Move this method to the child class. It doesn't belong here conceptually.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+      stdout, _, _ = cmd.Issue()
+      job_state = json.loads(stdout)['state']
+      logging.info('Polling: Job state is {} '.format(job_state))
+      if job_state == "Drained":
+          self.job_drained = True
+      else:
+          return None
+
+    # TODO: calculate run_time, pending_time as args for JobResult()
+    return True
```
dorellang, line 153: Add empty line at the end.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+from perfkitbenchmarker import errors
+from perfkitbenchmarker import providers
+from perfkitbenchmarker import vm_util
+from perfkitbenchmarker.providers.gcp import gcp_dpb_dataflow
+from perfkitbenchmarker.providers.gcp import util
+
+# Refer to flags with prefix 'dpb_df_template' defined in gcp provider flags.py
+FLAGS = flags.FLAGS
+
+class GcpDpbDataflowTemplate(gcp_dpb_dataflow.GcpDpbDataflow):
```
dorellang, line 40: Rename to GcpDpbDataflowStreamingTemplate, we might add in the future support for batch templates.

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
```diff
@@ -0,0 +1,200 @@
+# Copyright 2022 PerfKitBenchmarker Authors. All rights reserved.
```
dorellang, line 1: Kind of pedantic comment: I'd rather name this file `dpb_dataflow_streaming_template_benchmark.py`.

It's a long name, but here are my reasons:
- `df` is confusing, if you don't have context on Dataflow.
- This uses a streaming template, we might add support for non batch templates in the future.

Pleas update all referrences accordingly.

dorellang (changes requested):
Please rebase on top of the current master. I did see some old lines that were changed when I finally merged the previous PR.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
```diff
@@ -0,0 +1,153 @@
...
+
+  def Create(self):
+    """See base class."""
+    pass
+
+  def Delete(self):
+    """See base class."""
+    pass
+
+  def SubmitJob(
```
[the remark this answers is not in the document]
rarsan: How would you add additional arguments like `template_gcs_location`? Also, some of the arguments in the base class are not applicable like `pyspark_file`, should they still be defined here?
Note that in the original parent provider gcp_dpb_dataflow.py, the arguments of JobSubmit() also diverge from the base class.

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
```diff
@@ -0,0 +1,200 @@
...
+  input_pubsub_name = benchmark_spec.input_subscription_name
+  additional_args = FLAGS.dpb_df_template_additional_args
+
+  # Get handle to the dpb service
+  dpb_service_instance = benchmark_spec.dpb_service
+  # TODO: set input pubsub name and id as job attributes
+
+  # Create a file handle to contain the response from running the job on
+  # the dpb service
+  stdout_file = tempfile.NamedTemporaryFile(suffix='.stdout',
```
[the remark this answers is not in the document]
rarsan: You're right. This is a remnant from dpb_wordcount_benchmark, where stdout_file is passed as argument to SubmitJob. However it's not used in neither SubmitJob implementation.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: It was added here because retrieving job status applies to any dataflow provider (parent or child). It's the same reason I added job_id in this parent class. Depending on the benchmark, one might need to get job status to perform other operations like cleanup.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
rarsan: Similar to GetJobStatus, this was placed in parent class since it applies here as well e.g. in case of a streaming job (with pubsub input) that is executed from jar file.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
rarsan: Same dataflow template provider can be used to run [batch or streaming template](https://cloud.google.com/dataflow/docs/guides/templates/provided-templates), which is why I named it as such without specificity. However, I can see us extend GetCompletedJob to add support for different input types (PubSub, BQ, GCS)...I'll add a TODO there. SG?

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
rarsan: Makes sense. will update benchmark name.

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
```diff
@@ -0,0 +1,200 @@
...
+  cmd = util.GcloudCommand(None, 'pubsub', 'snapshots', 'create',
+                          benchmark_spec.input_subscription_snapshot_name)
+  cmd.flags = {
+      'project':  util.GetDefaultProject(),
+      'subscription': benchmark_spec.input_subscription_name,
+      'format': 'json',
+  }
+  stdout, _, _ = cmd.Issue()
+  snapshot_id = json.loads(stdout)[0]['snapshotId']
+  logging.info('Prepare: Created snapshot {} for input subscription data'
```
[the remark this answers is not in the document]
rarsan: Got it. Thx for the pointer.

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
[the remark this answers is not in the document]
dorellang: Please delete 🙏

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: Sorry I got confused, since we don't launch the jobs using the gcloud CLI I thought you couldn't interact with them in this way, but you are right.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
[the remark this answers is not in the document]
dorellang: > Note that in the original parent provider gcp_dpb_dataflow.py, the arguments of JobSubmit() also diverge from the base class.

This is an antipattern. The arguments shouldn't change in any child class.  

But, I do think the Dataflow classes stick out like a sore thumb in the DpbService hierarchy, because they are the only ones that run non Hadoop/Spark workloads.

I will let this pass, since there's not any quick solution for this. Just add a TODO for me, reminding me to address the change in signature.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: I think we shouldn't polute this class with PubSub stuff. Let's better leave that role for the subclass.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
dorellang: My point for making this class explicitly for streaming would be that you have to add logic to check for the stream's end. Whereas if you run a batch, it just ends. I'd rather have all the complexity of dealing with the "Stream end" on its own separate class.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow_template.py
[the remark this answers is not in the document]
rarsan: Sounds good. TODO added. After this merge, I can also help make the signature consistent across Dataflow classes and BaseDpbService class.

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
```diff
@@ -372,3 +460,42 @@ def _GetAvgValueFromTimeSeries(self, time_series: ListTimeSeriesResponse):
...
+  def _GetLastValueFromTimeSeries(self, time_series):
+    """Parses time series data and returns last value in last interval.
+
+    Args:
+      time_series: time series data returned by monitoring.
+
+    Returns:
+      Last value across intervals
+    """
+    value = None
```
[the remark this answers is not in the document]
rarsan: Method rewritten. This is much cleaner. The list returned is in reverse chronological order, that's why it's retrieving the first value in the list, which is the 'last value recorded in that time interval'

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
dorellang: Resolving as per our conversation

perfkitbenchmarker/providers/gcp/gcp_dpb_dataflow.py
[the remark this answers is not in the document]
dorellang: Awesome, thank you

perfkitbenchmarker/linux_benchmarks/dpb_df_template_benchmark.py
```diff
@@ -0,0 +1,200 @@
...
+  # Pass template parameters
+  template_params = []
+  template_params.append('inputSubscription={}'.format(input_pubsub_id))
+  for arg in additional_args:
+    template_params.append(arg)
+
+  results = []
+
+  start_time = datetime.datetime.now()
+  dpb_service_instance.SubmitJob(
```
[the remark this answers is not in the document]
dorellang: Will be addressed in a future refactor.

dorellang (approved):
LGTM.

There's one place I would use an f-string, but I can handle it in the merge.

Let's try to

dorellang (approved):

PR #3969, opened 2023-01-23
Option to set disk size for external volume

dchung-us:
Option to set disk size for external volume to override defaults,
also to support ubuntu2004.
This is for IBMCloud.

### DIFF
--- a/perfkitbenchmarker/providers/ibmcloud/ibmcloud_virtual_machine.py
+++ b/perfkitbenchmarker/providers/ibmcloud/ibmcloud_virtual_machine.py
@@ -402,6 +402,11 @@ class Ubuntu1804BasedIbmCloudVirtualMachine(
     super(Ubuntu1804BasedIbmCloudVirtualMachine, self).PrepareVMEnvironment()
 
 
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/providers/ibmcloud/ibmcloud_virtual_machine.py
```diff
@@ -359,7 +359,7 @@ def DownloadPreprovisionedData(self, install_path, module_name, filename):
 
   def ShouldDownloadPreprovisionedData(self, module_name, filename):
     """Returns whether or not preprovisioned data is available."""
-    return True
+    return False
```
bvliu, line 362: Why was this set to False?

dchung-us: It was set to True for a test before the code was submitted before, however that should have been set to False in the first place as we don't keep any pre provisioned data.

bvliu: Gotcha, thanks.

bvliu (approved):

Issue #3947, opened 2022-12-28
Running the tutorials "inter_region_ reports" creates a VM in only one region, us-central1.


lbj0412:
I am testing while seeing the tutorial in "Reproducting inter-Region Test with Benchmarker" (1).
Proceed through the tutorial, but create an instance only in us-central1-a and run a ping test.

As instructed in "Task 3. Start a benchmark test for latency", I proceeded without changing a single letter, but instances were created only in us-central1-a and not in other regions.
The ping test runs 190 times, but all tests run only on us-central1-a.

Any workaround?

(1)https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/tutorials/inter_region_reports/README.md

dlott:
@dPhanekham is this on your radar?

dPhanekham:
Yes, I think I have a solution and I'll submit an update soon

bvliu:
Should be fixed with 4fe1f18b0eaf51cbee11194f163ac8515775d93c.

Issue #3746, opened 2022-09-01
Removal of extra_zones flag breaks tutorials


dPhanekham:
Removal of the extra zones flag breaks all the .yaml configs in 
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/tree/master/tutorials/inter_region_reports/data

This was very useful for allowing two zone axes for multi-zone tests.
It is unclear how to achieve a similar result now without having several dozen configs files.

miketruty:
It looks like the resolution is some use of flag_zip and zone no?

raymond13513:
Working on this

PR #4037, opened 2023-02-17
inter-region tutorial fix

dPhanekham:
This updates the inter_region_reports tutorial to use the zone flag instead of zones and extra_zones.

This addresses issue #3947 and issue #3746 

It also updates references to Data Studio to instead reference Looker Studio, the new branding.


tutorials/inter_region_reports/README.md
```diff
@@ -236,17 +236,16 @@ other options for each VM.
...
...west3-a,us-west1-a],[europe-west3-a,us-west2-a],[europe-west4-a,europe-west6-a],[europe-west4-a,northamerica-northeast1-a],[europe-west4-a,southamerica-east1-a],[europe-west4-a,us-central1-a],[europe-west4-a,us-east1-b],[europe-west4-a,us-east4-a],[europe-west4-a,us-west1-a],[europe-west4-a,us-west2-a],[europe-west6-a,northamerica-northeast1-a],[europe-west6-a,southamerica-east1-a],[europe-west6-a,us-central1-a],[europe-west6-a,us-east1-b],[europe-west6-a,us-east4-a],[europe-west6-a,us-west1-a],[europe-west6-a,us-west2-a],[northamerica-northeast1-a,southamerica-east1-a],[northamerica-northeast1-a,us-central1-a],[northamerica-northeast1-a,us-east1-b],[northamerica-northeast1-a,us-east4-a],[northamerica-northeast1-a,us-west1-a],[northamerica-northeast1-a,us-west2-a],[southamerica-east1-a,us-central1-a],[southamerica-east1-a,us-east1-b],[southamerica-east1-a,us-east4-a],[southamerica-east1-a,us-west1-a],[southamerica-east1-a,us-west2-a],[us-central1-a,us-east1-b],[us-central1-a,us-east4-a],[us-central1-a,us-west1-a],[us-central1-a,us-west2-a],[us-east1-b,us-east4-a],[us-east1-b,us-west1-a],[us-east1-b,us-west2-a],[us-east4-a,us-west1-a],[us-east4-a,us-west2-a],[us-west1-a,us-west2-a]]
```
s-deitz, line 242: Too bad we're making you do this with the deprecation of those flags..

We used to have a way of creating zip and cross products across flags. Is there some support in YAML that we can use to create this list of lists?

If we can't do that, is there a bash command we can add to make this easier to generate?

tutorials/inter_region_reports/README.md
```diff
@@ -308,18 +307,17 @@ other options for each VM.
...
```
s-deitz, line 313: If this is the same list, is there a way to define it in one place?

tutorials/inter_region_reports/data/simple_region_iperf.yaml
```diff
@@ -1,14 +1,10 @@
...
   flag_matrix: inter_region
-  flag_matrix_filters:
-    # For each zone:extra_zone pair where zone != extra_zone.
-    inter_region: "zones < extra_zones"
   flag_matrix_defs:
     inter_region:
       gce_network_tier: [premium]
-      zones: [us-central1-a,us-east1-b]
-      extra_zones: [us-central1-a,us-east1-b]
+      zone: [[us-central1-a,us-east1-b]]
```
s-deitz, line 7: Well, this looks better. Is there a way to do this without the flag matrix? A list in flags perhaps, or does that not work?

bvliu:
Sorry about the churn, looks like we were working on a fix at the same time. I have a general fix for the tutorials in progress 
here which should allow flag_matrix_defs to work with multi string flags: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/4038. I'll look to submit this by early next week in which case this change may not be needed. Thanks!

dPhanekham:
No, worries, if there is a better fix on the way, I can close this PR.

PR #4663, opened 2024-01-11
P95 latency stats

PedroRivera125:
Allow users to specify which latency percentiles to calculate/report for Netperf benchmarks as a flag.
Previous behavior only calculated 50th, 90th, and 99th percentiles.


jellyfishcake:
Can this be deleted?

PR #4664, opened 2024-01-11
Mpi percentiles

PedroRivera125:
Added functionality to produce latency percentile information from MPI benchmarks like PingPong and Multi-PingPong

- moved _HistogramStatsCalculator from netperf_benchmark.py to separate utility file
- add imports for _HistogramStatsCalculator in netperf_benchmark and mpi_benchmark to produce/report user-specified latency percentiles


jellyfishcake:
Can this be deleted?

PR #5257, opened 2024-10-04
Worked on sample

harneet862:
Worked with creating a loop to get the data for sample object

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/cryptsetup_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/cryptsetup_benchmark.py
@@ -58,8 +58,17 @@ def Run(benchmark_spec: bm_spec.BenchmarkSpec) -> list[sample.Sample]:
   vm = benchmark_spec.vms[0]
   # TODO(user): Implement me
   samples = []
-  stdout, _ = vm.RemoteCommand('cryptsetup benchmark --help')
+  stdout, _ = vm.RemoteCommand('cryptsetup benchmark')
   logging.info('Benchmark output: %s', stdout)
+  line = stdout.splitlines()
+  line = line[1:5]
+  for line in stdout:
+      list = line.split()
+      metric = list[0] 
+      iteration = list[1]
+      unit = ''.join(list[2:4])
+      metadata = list[6]
+      samples.append(Sample(metric, iteration, unit, metadata))
   return samples
 
 


pmkc:
Hi Harneet sorry we got cut off:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork is what I was going to suggest for pulling changes to your repo (and then pulling from your own GitHub)

jellyfishcake (commented):
It would be helpful to add a unit test. To do so, I recommend:
1) factor out the parsing into a new function. i.e.
def ParseOutput(stdout):
  samples = []
  ...
  return samples

Your unit test should be a new file in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/tree/master/tests/linux_benchmarks called cryptsetup_test.py

In this file, you should include the expected output as a string, and call your parse function. 
See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/master/tests/linux_benchmarks/cassandra_stress_benchmark_test.py for an example.

perfkitbenchmarker/linux_benchmarks/cryptsetup_benchmark.py
```diff
@@ -58,8 +58,17 @@ def Run(benchmark_spec: bm_spec.BenchmarkSpec) -> list[sample.Sample]:
...
-  stdout, _ = vm.RemoteCommand('cryptsetup benchmark --help')
+  stdout, _ = vm.RemoteCommand('cryptsetup benchmark')
   logging.info('Benchmark output: %s', stdout)
+  line = stdout.splitlines()
+  line = line[1:5]
+  for line in stdout:
+      list = line.split()
+      metric = list[0] 
+      iteration = list[1]
+      unit = ''.join(list[2:4])
```
pmkc, line 69: I think this needs to be `[2:5]`, because the second one is inclusive.
We also want `'  '.join` to not have "iterationspersecond" smooshed together.

perfkitbenchmarker/linux_benchmarks/cryptsetup_benchmark.py
```diff
@@ -58,8 +58,17 @@ def Run(benchmark_spec: bm_spec.BenchmarkSpec) -> list[sample.Sample]:
...
+      metadata = list[6]
```
pmkc, line 70: Oh and this needs to be a dictionary.
Something like:
`metadata = {'key_size': list[6]}`

perfkitbenchmarker/linux_benchmarks/cryptsetup_benchmark.py
```diff
@@ -58,8 +58,17 @@ def Run(benchmark_spec: bm_spec.BenchmarkSpec) -> list[sample.Sample]:
   vm = benchmark_spec.vms[0]
   # TODO(user): Implement me
   samples = []
-  stdout, _ = vm.RemoteCommand('cryptsetup benchmark --help')
+  stdout, _ = vm.RemoteCommand('cryptsetup benchmark')
   logging.info('Benchmark output: %s', stdout)
+  line = stdout.splitlines()
+  line = line[1:5]
+  for line in stdout:
+      list = line.split()
```
pmkc, line 66: could this be renamed to `words` or something  besides list to avoid colliding with the python builtin `list`?

pmkc (changes requested):

pmkc:
I'll tweak this and merge it.

PR #5323, opened 2024-10-25
small fixes for GCP group-placement creation & intel repo installation

PedroRivera125:
## GCP group-placement bug

Currently, setting the `gce_placement_group_max_distance` flag with `placement_group_style=COLLOCATED`, as recommended in https://cloud.google.com/compute/docs/instances/use-compact-placement-policies#create-compact-policy, results in following error
```
perfkitbenchmarker.errors.Resource.CreationError: Failed to create placement group: ERROR: (gcloud.alpha.compute.resource-policies.create.group-placement) Could not fetch resource:
 - Required 'Alpha Access' permission for 'Compute API'
```
Current implementation uses alpha API creation command if `gce_placement_group_max_distance` flag is set when beta command can/should be used
https://cloud.google.com/sdk/gcloud/reference/alpha/compute/resource-policies/create/group-placement
https://cloud.google.com/sdk/gcloud/reference/beta/compute/resource-policies/create/group-placement

This change adds separate checks for when `gce_placement_group_max_distance` flag is set or `placement_group_style=CLUSTERED` and uses Alpha or Beta commands as appropriate

## Intel Repo bug

Currently, attempting to use a newer intelmpi_version, such as setting `intelmpi_version=2021.11.0`, results in following error
`STDOUT: STDERR: sed: can't read /etc/yum.repos.d/intelproducts.repo: No such file or directory`

Current implementation attempts to fix GPG related issue by disabling a GPG check. However, newer Intel Repo installations (2021.X.X+) use OneAPI, resulting in different .repo files

'regular' installation produces a intelproducts.repo
OneAPI installation produces a oneAPI.repo

This change checks if OneAPI was used, and disables the GPG check for the appropriate .repo file

### DIFF
--- a/perfkitbenchmarker/providers/gcp/gce_placement_group.py
+++ b/perfkitbenchmarker/providers/gcp/gce_placement_group.py
@@ -122,11 +122,21 @@ class GcePlacementGroup(placement_group.BasePlacementGroup):
         'group-placement',
         self.name,
     )
-    if (
-        self.style == CLUSTERED
-        or gcp_flags.GCE_PLACEMENT_GROUP_MAX_DISTANCE.value
-    ):
-      # Only alpha API supported for CLUSTERED and max-distance.
+    # alpha API for CLUSTERED
+    # beta API for max-distance
+    # https://cloud.google.com/sdk/gcloud/reference/alpha/compute/resource-policies/create/group-placement
+    # https://cloud.google.com/sdk/gcloud/reference/beta/compute/resource-policies/create/group-placement
+    if gcp_flags.GCE_PLACEMENT_GROUP_MAX_DISTANCE.value:
+      cmd = gcp_util.GcloudCommand(
+          self,
+          'beta',
+          'compute',
+          'resource-policies',
+          'create',
+          'group-placement',
+          self.name,
+      )
+    if self.style == CLUSTERED:
       cmd = gcp_util.GcloudCommand(
           self,
           'alpha',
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/providers/gcp/gce_placement_group.py
```diff
@@ -122,11 +122,21 @@ def _Create(self):
...
+          self,
+          'beta',
+          'compute',
+          'resource-policies',
+          'create',
+          'group-placement',
+          self.name,
+      )
+    if self.style == CLUSTERED:
       cmd = gcp_util.GcloudCommand(
```
bvliu, line 140: Sorry I could be misunderstanding the change: if both flags are set, wouldn't this still result in using the alpha command whereas we want it to be beta? 

Also I'm curious about the effects of just switching to use the beta gcloud command and not having to deal with alpha.

PedroRivera125: Unless I'm misunderstanding, the alpha API supports the following values for style: clustered, collocated, unspecified-collocation. 
The beta & general APIs only support values of collocated/unspecified-collocation for style.

Additionally, the max-distance flag is only supported by the alpha/beta APIs. 

I think this means that we should use the beta API if max-distance is set. However, we should switch to the alpha API if style=clustered since the alpha API is the only one which supports this setting.

Hope I could clarify my thoughts - let me know if the change is still unclear!

bvliu: Ah thanks for the clarification - that makes sense to me.

jellyfishcake (approved):

bvliu (approved):

PR #5725, opened 2025-05-06
 Fix incorrect subnet name from vpc GCP

jimmycgz:
This PR fixes an issue where subnet names were incorrectly using the network name instead of being properly configured separately.
 
## Changes made:
- Modified GceNetwork class to properly handle subnet names
- Ensured network name and subnet names are managed separately
- Fixed references to subnet resources to use the correct network name
 
This change ensures that when using GCP VPC networks with custom subnets, the subnet names are correctly configured and don't default to using the network name in all places.

Test command to reproduce and verify this fix
```
source ../../venv/bin/activate
cd $HOME/projects/PerfKitBenchmarker
export LOG_FOLDER=../pkb-logs
export MACHINE_TYPE=c4d-standard-4

python3 ./pkb.py \
  --benchmarks=nginx \
  --temp_dir=$LOG_FOLDER \
  --cloud=GCP \
  --project=your-gcp-proj1 \
  --zone=us-central1-a \
  --machine_type=$MACHINE_TYPE \
  --run_uri=standalone \
  --nginx_server_machine_type=$MACHINE_TYPE \
  --nginx_upstream_server_machine_type=$MACHINE_TYPE \
  --nginx_client_machine_type=$MACHINE_TYPE \
  --nginx_content_size=1024 \
  --nginx_load_configs="32:300:32:1000,512:300:512:1000" \
  --gce_network_name=your-custom-vpc \
  --gce_subnet_name=your-custom-subnet \
  --owner=owner-name \
  --accept_licenses

### DIFF
--- a/perfkitbenchmarker/providers/gcp/gce_network.py
+++ b/perfkitbenchmarker/providers/gcp/gce_network.py
@@ -903,13 +903,21 @@ class GceNetwork(network.BaseNetwork):
     self.is_existing_network = True
     self.subnet_names = []
     self.primary_subnet_name = None
+    self.network_name = None
+    
+    # Determine the network name first
+    if gcp_flags.GCE_NETWORK_NAMES.value:
+      self.network_name = gcp_flags.GCE_NETWORK_NAMES.value[0]
+    else:
+      self.network_name = self._MakeGceNetworkName()
+      self.is_existing_network = False
+      
+    # Then handle subnet names
     if network_spec.subnet_names:
       self.subnet_names = network_spec.subnet_names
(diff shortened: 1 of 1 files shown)


pmkc:
There was a conscious decision to always pass subnet and never pass network.
We recommend only passing --gce_subnet_name and never passing --gce_network_name

In the case of a an auto network, the subnet and network names coincide. In the case of a custom network, the network name is insufficient.

I'm not sure the exact behavior of the code today, when you pass both network and subnet, but I would recommend only passing subnet. Is that the case that you think PKB handles incorrectly?

PR #5875, opened 2025-07-07
Add support for AKS Automatic cluster

vofish:
Adds support for creating Azure Kubernetes Service (AKS) clusters using the Automatic SKU: `--sku automatic`

### DIFF
--- a/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
@@ -391,9 +391,10 @@ class AksCluster(container_service.KubernetesCluster):
     stdout, _, _ = vm_util.IssueCommand(cmd)
     nodepools = json.loads(stdout)
     return [nodepool['name'] for nodepool in nodepools]
-
+  
 class AksAutomaticCluster(AksCluster):
   """Class representing an AKS Automatic cluster, which has managed node pools."""
+  # https://learn.microsoft.com/en-us/azure/aks/automatic/quick-automatic-managed-network
 
   CLOUD = provider_info.AZURE
   CLUSTER_TYPE = 'Auto'
@@ -422,7 +423,7 @@ class AksAutomaticCluster(AksCluster):
         'automatic',
         '--tags',
     ] + tags_list
-    vm_util.Retry(timeout=300)(vm_util.IssueCommand)(
+    vm_util.IssueCommand(
         cmd,
         # Half hour timeout on creating the cluster.
         timeout=1800,
@@ -431,20 +432,16 @@ class AksAutomaticCluster(AksCluster):
   def _IsReady(self):
     """Returns True if the cluster is ready."""
     # Check provisioning state
-    show_cmd = [
-        azure.AZURE_PATH,
-        'aks',
-        'show',
-        '--name',
-        self.name,
-    ] + self.resource_group.args
-    stdout, _, _ = vm_util.IssueCommand(show_cmd, raise_on_failure=False)
-    try:
-        cluster = json.loads(stdout)
-        if cluster.get('provisioningState') != 'Succeeded':
-            return False
-    except Exception:
-        return False
+    vm_util.IssueCommand(
+        [
+          azure.AZURE_PATH,
+          'aks',
+          'show',
+          '--name',
+          self.name,
+        ]
+        + self.resource_group.args
+    )
 
     vm_util.IssueCommand(
         [
@@ -473,8 +470,50 @@ class AksAutomaticCluster(AksCluster):
     stdout, _, _ = vm_util.IssueCommand(get_cmd)
     return 'default' in stdout
   
+  def _CreateDependencies(self):
+    """Creates the resource group, service principal, and registers required AKS features."""
+    super()._CreateDependencies()
+
+    # Register the AutomaticSKUPreview feature
+    vm_util.IssueCommand(
+        [
+          azure.AZURE_PATH,
+          'feature',
+          'register',
+          '--namespace',
+          'Microsoft.ContainerService',
+          '--name',
+          'AutomaticSKUPreview',
+        ]
+    )
+
+    # Show the feature registration status
+    vm_util.IssueCommand(
+        [
+          azure.AZURE_PATH,
+          'feature',
+          'show',
+          '--namespace',
+          'Microsoft.ContainerService',
+          '--name',
+          'AutomaticSKUPreview',
+        ]
+    )
+    
+    # Register the provider
+    vm_util.IssueCommand(
+        [
+          azure.AZURE_PATH,
+          'provider',
+          'register',
+          '--namespace',
+          'Microsoft.ContainerService',
+        ]
+    )
+
   def _PostCreate(self):
     pass
+
 
 def _AzureNodePoolName(pkb_nodepool_name: str) -> str:
   """Truncate nodepool name for AKS."""
--- a/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
@@ -393,8 +393,13 @@ class AksCluster(container_service.KubernetesCluster):
     return [nodepool['name'] for nodepool in nodepools]
   
 class AksAutomaticCluster(AksCluster):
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -392,6 +392,89 @@ def GetNodePoolNames(self) -> list[str]:
     nodepools = json.loads(stdout)
     return [nodepool['name'] for nodepool in nodepools]
 
+class AksAutomaticCluster(AksCluster):
+  """Class representing an AKS Automatic cluster, which has managed node pools."""
```
hubatish: I tried running this & got the semi-expected errors about the az extension not being installed. Can you link to here with the details:
https://learn.microsoft.com/en-us/azure/aks/automatic/quick-automatic-managed-network?pivots=azure-cli

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -392,6 +392,89 @@ def GetNodePoolNames(self) -> list[str]:
...
+
+  CLOUD = provider_info.AZURE
+  CLUSTER_TYPE = 'Auto'
+
+  def __init__(self, spec):
+    """Initializes the cluster."""
+    super().__init__(spec)
+
+  def _Create(self):
```
hubatish, line 411: Also on the extensions above.. we try not to install anything on the local machine, so we can just list those as requirements. But we do make changes to the account & cloud state. So the feature flags part https://learn.microsoft.com/en-us/azure/aks/automatic/quick-automatic-managed-network?pivots=azure-cli#register-the-feature-flags can be automatically done in the _CreateDependencies section to the account.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -392,6 +392,89 @@ def GetNodePoolNames(self) -> list[str]:
...
+        '--kubeconfig',
+        FLAGS.kubeconfig,
+        'get',
+        'serviceAccounts',
+    ]
+    stdout, _, _ = vm_util.IssueCommand(get_cmd)
+    return 'default' in stdout
+  
+  def _PostCreate(self):
+    pass
```
hubatish, line 525: Why are you overriding & skipping _PostCreate? We do want to tag the resource (which is what the parent _PostCreate does), as tagging is how we internally control what resources should be kept vs deleted. 

Ah, but perhaps parent is just for the node resource group? Does that still exist in AKS Automatic? I can't find any explicit documentation as to yes or no. Please investigate & justify with a comment if not, or fix if it is.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -392,6 +392,89 @@ def GetNodePoolNames(self) -> list[str]:
...
+        self.region,
+        '--ssh-key-value',
+        vm_util.GetPublicKeyPath(),
+        '--resource-group',
+        self.resource_group.name,
+        '--sku',
+        'automatic',
+        '--tags',
+    ] + tags_list
+    vm_util.Retry(timeout=300)(vm_util.IssueCommand)(
```
hubatish: Why retry here? what errors are being encountered & ignored but are fixed on a subsequent aks create? Failure rather than swallowing errors is sometimes preferred.

..ah I see this code matches AksCluster.. which also means I'm guessing you haven't run into the race condition mentioned above yet? I'd remove it atm & we can add it back if we run into the aforementioned race condition.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -392,6 +392,89 @@ def GetNodePoolNames(self) -> list[str]:
...
+            self.name,
+            '--file',
+            FLAGS.kubeconfig,
+        ]
+        + self.resource_group.args
+    )
+    version_cmd = [FLAGS.kubectl, '--kubeconfig', FLAGS.kubeconfig, 'version']
+    _, _, retcode = vm_util.IssueCommand(version_cmd, raise_on_failure=False)
+    if retcode:
+      return False
```
hubatish, line 471: Likewise we are swallowing a lot of errors in _IsReady.. but here it makes sense as we need to wait until these things are indeed ready & expect some errors before then. 

You still may encounter real vs expected errors here, but it mostly matters for debugging (hiding error as timeout rather than real error, which can still be found after further investigation) & faster timeouts. 

Please add some documentation of the error messages encountered. There should be an azure_kubernetes_service_test.py you could add documentation as code unit tests. Otherwise just in comments.

hubatish (changes requested):
This is pretty good overall. Glad it was a fairly simple change.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: A cluster is tagged within `_Create` method:
https://github.com/kiryl-filatau/PerfKitBenchmarker/blob/db028dca33c7d100a6ede895076bb4fbce8b62be/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py#L423-L424
That's why `_PostCreate` is overridden and skipped.

As it is an automatic cluster, there is no manually created nodepool:
`Node management is automatically handled without the need for manual node pool creation.`
https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic#node-management-scaling-and-cluster-operations

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Removed this retry.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Added

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Removed this

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Add extended `_CreateDependencies` method

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: Got it, sounds good. A comment about why there's no created/user managed node resource group would be good then.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: This is a really terse, no context link. Comment should be something like "This feature is in preview. To provision this resource you'll need to install an extension; see link for details" & should go in docstring rather than comment right below docstring.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: We definitely still need some handling.. this passes when you run it again? If only the get-credentials command was failing though that is great.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Docstring is updated and extended.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: This handling checks the cluster's `provisioningState` and waits until it becomes `Succeeded`.

However, in certain cases, the cluster may not reach the `"provisioningState": "Succeeded"` (`"provisioningState": "Updating"`), which could result in failed subsequent deployment attempts.

Please let me know how you would like to handle this.

hubatish: If it doesn't reach Succeeded or Updating, then that should be a failure right? It's good to just raise an error & fail the whole benchmark in that case.

vofish: Added cluster's provisioningState handling

PR #5923, opened 2025-07-29
Add image

vofish:
Allow specifying the container image to use


hubatish (approved):
This change & simple flag addition is in-line with the existing flags & code. Approving

PR #5940, opened 2025-07-31
allow customizing timeout values

ScottLinnn:
Add a flag to config timeout value for this benchmark because the time needed for index covering 100% varies a lot by provider. Hardcoded value fitting one provider will be too large/too small for other provider(s)


jnguertin (approved):
LGTM.

PR #5945, opened 2025-08-01
Divide resource Requests and Limits

haykking:
New variables responsible for Limits and Requests were created and changed.
A new Limits section was created in the Kubernetes manifest.


perfkitbenchmarker/linux_benchmarks/kubernetes_scale_benchmark.py
```diff
@@ -48,11 +48,17 @@
     False,
     'Whether to report individual event latencies',
 )
-CPUS_PER_POD = flags.DEFINE_string(
-    'kubernetes_scale_pod_cpus', '250m', 'CPU limit per pod'
+REQUEST_CPUS_PER_POD = flags.DEFINE_string(
+    'kubernetes_scale_min_pod_cpus', '250m', 'Min CPU limit per pod'
```
hubatish: No real need to change the flag name. It's somewhat inconvenient on my end as I then need to modify the scheduling code we have to change this as well. Updating the description is sufficient documentation.

perfkitbenchmarker/linux_benchmarks/kubernetes_scale_benchmark.py
```diff
@@ -48,11 +48,17 @@
...
 )
-MEMORY_PER_POD = flags.DEFINE_string(
-    'kubernetes_scale_pod_memory', '250M', 'Memory limit per pod'
+LIMIT_CPUS_PER_POD = flags.DEFINE_string(
+    'kubernetes_scale_max_pod_cpus', '250m', 'Max CPU limit per pod'
```
hubatish: But I think we also chatted offline about just needing one number for both limits & requests, so will wait on the larger modification to give next review.

haykking:
I returned the old names to the variables
Now I use only one variable for Requests
As a result, the changes were very small after all :)

hubatish:
Fair enough, very small indeed but all that's needed. I'll approve & we can check back in with Rich when he's back as well.

PR #5969, opened 2025-08-13
NodePool Provisioning AWS Karpenter setup

haykking:
To make Karpenter understand that it needs to create Nodes, I had to change its default settings. Here I set a new key "group" and a new label "pkb_nodepool" = default
image

### DIFF
--- a/tests/providers/aws/elastic_kubernetes_service_test.py
+++ b/tests/providers/aws/elastic_kubernetes_service_test.py
@@ -217,9 +217,16 @@ class ElasticKubernetesServiceTest(BaseEksTest):
   def testGetNodePoolNamesKarpenter(self):
     cluster = elastic_kubernetes_service.EksKarpenterCluster(EKS_SPEC)
     self.MockIssueCommand({
-        'kubectl get nodepool': [
-            (json.dumps([{'name': 'karpenter-ng'}, {'name': 'default'}]), '', 0)
-        ]
+        'kubectl --kubeconfig  get nodepool -o json': [(
+            json.dumps({
+                'items': [
+                    {'metadata': {'name': 'karpenter-ng'}},
+                    {'metadata': {'name': 'default'}},
+                ]
+            }),
+            '',
+            0,
+        )]
     })
     self.assertEqual(cluster.GetNodePoolNames(), ['karpenter-ng', 'default'])
 


perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -286,7 +286,20 @@ def DeployIngress(self, name: str, namespace: str, port: int) -> str:
...
+    stdout, _, _ = container_service.RunKubectlCommand([
+        'get',
+        'nodes',
+        '-o',
+        'json',
+    ])
+
+    data = json.loads(stdout)
+
+    return [item["metadata"]["name"] for item in data["items"]]
```
rsgowman: I don't have an aws cluster to test with, but this seems like it would return the node names rather than the nodepool names. Could you double check this? (If so, also check that the right number of node pools are being created by this test as a whole.)

Also, fwiw, I see that googke_kubernetes_engine_test.py and azure_kubernetes_service_test.py both add a unit tests for their GetNodePoolNames implementation. It might be a good idea to do the same here.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -248,6 +248,14 @@ def Run(benchmark_spec: bm_spec.BenchmarkSpec) -> List[sample.Sample]:
   """Runs the node pools provisioning benchmark."""
   cluster = benchmark_spec.container_cluster
   samples = []
+  cluster_type = getattr(cluster, "CLUSTER_TYPE", None)
+  if cluster_type == "Karpenter":
+    cluster.ApplyManifest(
+        "provision_node_pools/karpenter/nodepool.yaml.j2",
```
rsgowman: As I understand this, the current flow is:

* line 254: create all the nodepools
* line 259: start the timer
* line 261,262: ??? I think this will just create workloads within the already-existing nodepools
* line 263: stop the timer

I suspect instead, it should look more like:

* start the timer
* create all nodepools
* stop the timer

The existing _CreateNodePools implementation seems to just create a workload with some constraints and then assumes a nodepool will be created to handle it. With karpenter, that might not be possible. So replacing that with your yaml might be a good way forward.

perfkitbenchmarker/data/provision_node_pools/karpenter/nodepool.yaml.j2
```diff
@@ -0,0 +1,95 @@
...
+      nodeClassRef:
+        group: karpenter.k8s.aws
+        kind: EC2NodeClass
+        name: default
+      expireAfter: 720h # 30 * 24h = 720h
+  limits:
+    cpu: 1000
+  disruption:
+    consolidationPolicy: WhenEmptyOrUnderutilized
+    consolidateAfter: 1m
```
rsgowman: nit: I suspect we should increase this to some arbitrarily large number. For the purposes of this benchmark, consolidating node pools is best avoided.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -323,7 +323,41 @@ def DeployIngress(self, name: str, namespace: str, port: int) -> str:
...
+      # Get all Karpenter NodePool CRDs
+      cmd = [
+          FLAGS.kubectl,
+          "--kubeconfig", FLAGS.kubeconfig,
+          "get", "nodepool",
+          "-o", "json"
+      ]
+      stdout, _, retcode = vm_util.IssueCommand(cmd)
+      if retcode:
+        logging.warning("Failed to get Karpenter NodePools: %s", stdout)
```
rsgowman: kubectl typically (always?) outputs error messages on stderr; it would be more useful to capture that and output it here. (Alternatively, you can capture both stdout/stderr and output them both here.)

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -323,7 +323,41 @@ def DeployIngress(self, name: str, namespace: str, port: int) -> str:
...
+      # Default: return managed node group names
+      cmd = [
+          FLAGS.eksctl, "get", "nodegroup",
+          "--cluster", self.name,
+          "--region", self.region,
+          "-o", "json"
+      ]
+      stdout, _, retcode = vm_util.IssueCommand(cmd)
+      if retcode:
+        logging.warning("Failed to get nodegroups: %s", stdout)
```
rsgowman: Similar to above, you might want to capture stderr instead of (or in addition to) stdout. Caveat: I'm not familiar with eksctl, so don't know if it has differing conventions.

tests/providers/aws/elastic_kubernetes_service_test.py
```diff
@@ -194,10 +195,30 @@ def testEksClusterNodepools(self):
         node_groups[1],
     )
 
+  def testGetNodePoolNames(self):
+    # Mock the output of the aws cli command
+    self.MockIssueCommand({
+      'get nodegroup': [(
```
rsgowman: Could we also add a test for the karpenter case?

tests/providers/aws/elastic_kubernetes_service_test.py
```diff
@@ -194,10 +195,30 @@ def testEksClusterNodepools(self):
...
+    cluster = elastic_kubernetes_service.EksCluster(EKS_SPEC)
+    self.assertEqual(
+      cluster.GetNodePoolNames(),
+      ['default', 'nodegroup1', 'nodegroup2']
+    )
   @parameterized.named_parameters(
       ('default nodepool', 'default', 'default'),
       ('standard nodepool', 'nginx', 'nginx'),
   )
+
```
rsgowman: nit: blank line should be after line 216 rather than here

tests/providers/aws/elastic_kubernetes_service_test.py
```diff
@@ -194,6 +195,45 @@ def testEksClusterNodepools(self):
         node_groups[1],
     )
 
+  def testGetNodePoolNames(self):
```
rsgowman, line 198: This is fine.

You could optionally simplify this with one of the following approaches:

1. Do nothing.
2. Split into two test cases, one for karpenter, one for regular. This would allow you to eliminate the need for the `isinstance` check.
3. Parameterize the test. The test body itself wouldn't be much more than the assertEqual test - everything else would be in the parameters. This would look something like the following (untested):

```
@parameterized.named_parameters(
    ('regular cluster', lambda: self.MockIssueCommand({'eksctl ...), ["default', 'nodegroup1', 'nodegroup2'),
    ('karpenter cluster', ..., ...),
)
def testGetNodePoolNames(self, cluster, mockNodePoolsFn, expected):
  mockNodePoolsFn()
  self.assertEqual(cluster.GetNodePoolNames(), expected)

```

That last approach makes the test really simple, but puts a lot of text into the parameterization arguments due to the MockIssueCommand. Probably too much. You could simplify it with a helper function though.

Personally, I'd go with 2 - it's simplest, but all approaches are fine by me.

hubatish: Agreed, splitting into 2 tests would be good.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -323,7 +323,41 @@ def DeployIngress(self, name: str, namespace: str, port: int) -> str:
...
     ])
     return self._GetAddressFromIngress(stdout)
+  
+  def GetNodePoolNames(self) -> list[str]:
+    """Get node pool names for the cluster.
 
+    For Karpenter clusters, returns all CRD NodePool names created by Karpenter.
+    For other clusters, returns managed node group names.
+    """
+    if getattr(self, "CLUSTER_TYPE", None) == "Karpenter":
```
hubatish: We have 3 whole classes for EksCluster vs EksKarpenterCluster vs EksAutoCluster right? Using "getattr" is bad object oriented design. Just override this function for EksKarpenterCluster to run the below get nodepool command, & leave the get nodegroup command functionality either in BaseEksCluster like it is currently or in specifically EksCluster if it doesn't work/make sense for EksAutoCluster.

perfkitbenchmarker/data/provision_node_pools/karpenter/nodepool.yaml.j2
```diff
@@ -0,0 +1,95 @@
...
+        name: default
+      expireAfter: 720h # 30 * 24h = 720h
+  limits:
+    cpu: 1000
+  disruption:
+    consolidationPolicy: WhenEmptyOrUnderutilized
+    consolidateAfter: 720h
+---
+{%- for pool_num in range(1, batch_size|int + 1) %}
+{%- set pool_id = "%03d"|format(pool_num) %}
```
hubatish: The benchmark runs 2 nodepool creation operations right? One in the init batch as set by provision_node_pools_init_batch & one in the provision_node_pools_test_batch. The intent of this is to eg create 3 nodepools in the init batch & then create another 3 new nodepools in the test batch. But with this naming convention, it looks like you'd create eg nodepools with names:
nodepool1
nodepool2
nodepool3
in the first init batch. & then in the second test batch, you'd create nodepools with.. exactly the same names. Which uhh probably doesn't do anything? 
In other words, do you need an offset / prefix for these nodepool names? ..also looking below maybe this is handled by batch_name..

..ok nvm looks like that does handle it, all good. I'll still leave the comment but feel free to just resolve confirming behavior.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -222,6 +222,14 @@ def _CreateNodePools(
   nodes_before = len(cluster.GetNodeNames())
   nodes_pools_before = len(cluster.GetNodePoolNames())
   start = time.monotonic()
+  cluster_type = getattr(cluster, "CLUSTER_TYPE", None)
```
hubatish: Much like with the other "use object oriented design" comment for the GetNodepoolNames function, don't use getattr here. Instead move this code into the EksKarpenter class with an "AddNodepools" function. Also stub out that function in root BaseContainerCluster, but it can just pass / not do anything in the majority of implementations. It will do something in the Karpenter case, and do nothing in the GKE case. Probably will also do something unique in the AKS case that Volodymr can fill in.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -222,6 +222,14 @@ def _CreateNodePools(
...
+  if cluster_type == "Karpenter":
+    cluster.ApplyManifest(
```
hubatish: As a follow up to above, quite possibly also the current code run in the _CreateJobsAndWait function (eg the provision_node_pools/job_manifest.yaml.j2 applied at line 107) is incorrect for competitors / should be specific to a GKE implementation of "AddNodepools". Does is it in fact make sense to create the nodepools with the yaml above, & then also to create all the jobs as two separate steps? Or should the nodepool creation step just specify that each nodepool should have 1 node, & then we time how long those take to create? We don't need to change this just to change it, but if the behavior is in fact wrong (does it create double the number of nodes needed? do the created jobs each get created in their own nodepools?) then we should definitely change.

tests/providers/aws/elastic_kubernetes_service_test.py
hubatish: Thanks for addressing this.

Issue #5883, opened 2025-07-11
PerfKitBenchmarker: Certain Benchmarks Fail to Run on macOS


vofish:
While running a benchmark on MacOS using the following command:
```
./pkb.py --cloud=Azure --benchmarks=kubernetes_hpa --kubectl=$(which kubectl) --kubeconfig=$HOME/config --accept_licenses --zone centralus
```
an error occurs:
```
2025-07-11 12:00:24,909 b89c560b MainThread kubernetes_hpa(1/1) ERROR    Exception running benchmark
Traceback (most recent call last):
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 1368, in RunBenchmarkTask
    RunBenchmark(spec_for_run, collector)
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 1080, in RunBenchmark
    DoProvisionPhase(spec, detailed_timer)
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/pkb.py", line 670, in DoProvisionPhase
    spec.Provision()
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/benchmark_spec.py", line 898, in Provision
    self.container_cluster.Create()
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/container_service.py", line 1545, in Create
    super().Create(restore)
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/resource.py", line 401, in Create
    self._PostCreate()
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py", line 306, in _PostCreate
    super()._PostCreate()
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/container_service.py", line 1552, in _PostCreate
    self.event_poller.StartPolling()
  File "$HOME/repos/PerfKitBenchmarker/perfkitbenchmarker/container_service.py", line 1490, in StartPolling
    self.event_poller.start()
  File "$HOME/.pyenv/versions/3.11.13/lib/python3.11/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
                  ^^^^^^^^^^^^^^^^^
  File "$HOME/.pyenv/versions/3.11.13/lib/python3.11/multiprocessing/context.py", line 224, in _Popen
[...]

PR #6013, opened 2025-08-26
Karpenter hpa aws fix

kiryl-filatau:
[Add ALB controller installation and networking fixups]

- Install AWS Load Balancer Controller with pinned IAM policy (v2.13.4)

- Override DeployIngress to apply ALB Service+Ingress manifest

- Add post-ingress networking fixups (open 80/tcp on ALB SG, allow ALB→nodes)

- Fix multiprocessing issue on macOS

- Add .DS_Store to .gitignore

### DIFF
--- a/perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
@@ -837,7 +837,9 @@ class EksKarpenterCluster(BaseEksCluster):
         '--timeout=180s',
     ])
 
-  def DeployIngress(self, name: str, namespace: str, port: int, health_path: str) -> str:
+  def DeployIngress(
+      self, name: str, namespace: str, port: int, health_path: str
+  ) -> str:
     """Deploys only Service + Ingress (without IngressClass) for AWS Load Balancer Controller."""
     # Apply the custom manifest template (service + ingress with annotations).
     self.ApplyManifest(


perfkitbenchmarker/container_service.py
```diff
@@ -31,6 +31,7 @@
 import logging
 import multiprocessing
 from multiprocessing import synchronize
+multiprocessing.set_start_method("fork", force=True)
```
hubatish: The rest of your cl doesn't use multiprocessing. a) why change this? b) doesn't seem relevant to the rest of the change c) looks kinda sketchy in general.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -653,6 +867,8 @@ def _PostCreate(self):
         'logLevel=debug',
         '--wait',
     ])
+    # Ensure ALB ingress support: install AWS Load Balancer Controller.
+    self._InstallAwsLoadBalancerController()
```
hubatish: This looks fairly complex and quite possibly error prone. Let's gate behind a flag, config setting, or the benchmark calling DeployIngress so that this doesn't occur in eg scaling tests which currently work for Karpenter. The last of these (benchmark calling DeployIngress) would be the best code solution, but would also be the hardest change. An in-file flag would be an acceptable quick fix too.

perfkitbenchmarker/data/container/ingress_alb.yaml.j2
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
rsgowman, line 1: This file might be better placed in `./data/container/kubernetes_hpa` if it's specific to that benchmark. (And the presence of `/calculate` on line 23 suggests that it is.)

If you're intending to use it more generally, then this seems like an ok spot.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+    host = urlparse(raw).hostname if raw.startswith('http') else raw
+    normalized = (host or '').replace('dualstack.', '')
+    alb_sg = ''
+    lb_query = util.AWS_PREFIX + [
+        'elbv2', 'describe-load-balancers',
+        '--region', self.region,
+        '--query', f"LoadBalancers[?contains(DNSName, '{normalized}')].SecurityGroups[0]",
+        '--output', 'text',
+    ]
+    for attempt in range(24):  # ~120s
```
hubatish: O, so this just retries 24 times, essentially waiting 120s with the waits? Just use IssueRetryableCommand & set a timeout.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+        out, _, _ = vm_util.IssueCommand(lb_query, raise_on_failure=False)
+        alb_sg = (out or '').strip()
+        if alb_sg and alb_sg != 'None':
+            break
+        logging.info('[PKB][EKS] Waiting for ALB SG (%d/24) for %s', attempt + 1, normalized)
+        time.sleep(5)
+    if not alb_sg or alb_sg == 'None':
+        logging.warning('[PKB][EKS] Could not resolve ALB SG for %s; skipping fixups for now', normalized)
+        return
+    alb_sg = (out or '').strip()
```
hubatish, line 909: Seems to be a lot of string parsing. IMO that means a unit test is worthwhile.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+          'ec2', 'authorize-security-group-ingress',
+          '--region', self.region,
+          '--group-id', alb_sg,
+          '--protocol', 'tcp',
+          '--port', '80',
+          '--cidr', '0.0.0.0/0',
+      ], raise_on_failure=False)  # ok if rule already exists
+      logging.info('[PKB][EKS] Opened 80/tcp on ALB SG %s', alb_sg)
+    else:
+      logging.warning('[PKB][EKS] Could not resolve ALB SG for %s', address)
```
hubatish: We should likely throw if this doesn't get setup. In combination with a flag / some control to skip this code for scale, that'll make the HPA benchmark test easier to use (loud failure rather than silent failure that looks like a success)

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+            policy_path = tf.name
+        vm_util.IssueCommand([
+            'curl', '-sSL', '-o', policy_path,
+            'https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/'
+            'v2.13.4/docs/install/iam_policy.json',
+        ], raise_on_failure=True)
+        create_cmd = util.AWS_PREFIX + [
+            'iam', 'create-policy',
+            '--policy-name', 'AWSLoadBalancerControllerIAMPolicy',
+            '--policy-document', f'file://{policy_path}',
```
rsgowman: Is `file://` a mac-ism? I don't think it works elsewhere. You can probably just drop it to have it work on both mac/linux.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+    list_cmd = util.AWS_PREFIX + [
+        'iam', 'list-policies', '--scope', 'Local',
+        '--query', "Policies[?PolicyName=='AWSLoadBalancerControllerIAMPolicy'].Arn | [0]",
+        '--output', 'text',
+    ]
+    stdout, _, _ = vm_util.IssueCommand(list_cmd, raise_on_failure=False)
+    policy_arn = (stdout or '').strip()
+    if not policy_arn or policy_arn == 'None':
+        # Download the official policy JSON (pinned to a specific controller version).
+        with vm_util.NamedTemporaryFile(dir=vm_util.GetTempDir(), delete=False, mode='w') as tf:
```
rsgowman: > `delete=False`

Is there any reason not to delete this? i.e. you could refactor to:

```
with vm_util.NamedTemporaryFile(...) as tf:
  vm_util.IssueCommand([..., '-o', tf.name, ...])
  ...
```

Not only would it clean up the file, but it'd eliminate the now unnecessary policy_path variable too.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+      - Applies CRDs and installs/updates the Helm chart with a correct IngressClass.
+      - Waits for the controller rollout.
+    """
+    # 1) Ensure OIDC provider (IRSA)
+    vm_util.IssueCommand([
+        FLAGS.eksctl, 'utils', 'associate-iam-oidc-provider',
+        f'--region={self.region}',
+        f'--cluster={self.name}',
+        '--approve',
+    ], raise_on_failure=False)
```
rsgowman: I assume you're ignoring the potential failure b/c this command fails if it's previously been run?

It might be worth validating that somehow, if it's easy enough to do. Maybe check the return code and see if it's distinguishable from, e.g., a generic networking error. Alternatively, I've seen other places in the pkb code where the error message itself is matched with a regex, (which is a little gross).

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+    # 3) Ensure ServiceAccount with the attached IAM policy (IRSA)
+    vm_util.IssueCommand([
+        FLAGS.eksctl, 'create', 'iamserviceaccount',
+        '--cluster', self.name,
+        '--namespace', 'kube-system',
+        '--name', 'aws-load-balancer-controller',
+        '--attach-policy-arn', policy_arn,
+        '--approve',
+        '--override-existing-serviceaccounts',
+    ], raise_on_failure=False)
```
rsgowman: Same question re not checking command failure, with the additional remark that `--override-existing-serviceaccounts` seems like it might make this command succeed even if it was previously run. Maybe you should just raise on failure here.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+
+    # 4) Ensure CRDs (safe to re-apply)
+    container_service.RunKubectlCommand([
+        'apply', '-k',
+        'github.com/aws/eks-charts/stable/aws-load-balancer-controller/crds?ref=v2.13.4',
+    ], raise_on_failure=False)
```
rsgowman: Ah, now I'm on more familiar ground - kubectl should "just work" if you run it again (and indeed, your comment on line 681 states as much). It seems like a failure here should be fatal. (But if not, add a comment explaining why errors should be ignored.)

(I'm not going to comment on any further `raise_on_failure=False` issues - just take a look at all of them and make sure they make sense. Try to eliminate them where possible/reasonable.)

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+    policy_arn = (stdout or '').strip()
+    if not policy_arn or policy_arn == 'None':
+        # Download the official policy JSON (pinned to a specific controller version).
+        with vm_util.NamedTemporaryFile(dir=vm_util.GetTempDir(), delete=False, mode='w') as tf:
+            policy_path = tf.name
+        vm_util.IssueCommand([
+            'curl', '-sSL', '-o', policy_path,
+            'https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/'
+            'v2.13.4/docs/install/iam_policy.json',
+        ], raise_on_failure=True)
```
rsgowman: `raise_on_failure=True` is the default - you can just omit it.

OTOH, you may be including it here for stylistic purposes, since there's a lot of `raise_on_failure=False` throughout the rest of this function, in which case, that's fine by me. (But if you manage to remove most/all of the raise_on_failure=False elsewhere, then please do revisit.)

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +623,218 @@ def _Create(self):
...
+    container_service.RunKubectlCommand([
+        'rollout', 'status', 'deployment/aws-load-balancer-controller',
+        '-n', 'kube-system', '--timeout=180s',
+    ], raise_on_failure=False)
+
+  def DeployIngress(self, name: str, namespace: str, port: int) -> str:
+    """Deploys only Service + Ingress (without IngressClass) for AWS Load Balancer Controller."""
+    # Apply the custom manifest template (service + ingress with annotations).
+    self.ApplyManifest(
+        'container/ingress_alb.yaml.j2',
```
rsgowman: (Similar to my question from above.) You're unconditionally applying this ingress to all eks clusters, but really, I think we only need it for the kubernetes_hpa test. (At least for now). Maybe applying this ingress should be part of the test itself?

perfkitbenchmarker/container_service.py
rsgowman: What happens if you *don't* do this?

Also, I see that https://docs.python.org/3/library/multiprocessing.html claims:

> The default start method will change away from fork in Python 3.14. Code that requires fork should explicitly specify that via get_context() or set_start_method().

I wonder if we should get this to work with `spawn` instead...

kiryl-filatau: I added set_start_method("fork") because on macOS (Python 3.12) the benchmark crashed under spawn.
I agree this PR is not the right place for that fix, so I will remove it.
A separate issue has already been created: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/5883.
Thanks for pointing it out.

kiryl-filatau: I added set_start_method("fork") because on macOS (Python 3.12) the benchmark crashed under spawn.
I agree this PR is not the right place for that fix, so I will remove it.
A separate issue has already been created: https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/issues/5883.
Thanks for pointing it out.

hubatish: Ah, ok, glad y'all found a workaround for the Mac problems; I haven't looked at it in a while. May be worth sending that out as a PR as well (if we can figure out the correct spot and/or guards for it). I'll leave a comment on the other PR.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Good point, actually do not need delete=False here. I’ll refactor into a single with block so the tempfile is cleaned up automatically and the extra variable goes away. Thanks for catching that.

perfkitbenchmarker/data/container/ingress_alb.yaml.j2
kiryl-filatau: agree, will move to ./data/container/kubernetes_hpa

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: The file:// prefix is required by AWS CLI when passing a path to --policy-document. Without it, the CLI treats the argument as a literal JSON string. It’s not mac-specific, it works the same on Linux and other platforms. I’ll leave the file:// in place.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Just to clarify: DeployIngress is already implemented in the base class where it applies a Service and Ingress with an IngressClass. In this override in EksKarpenterCluster the manifest is the same except that we omit the IngressClass because it is created and managed by the AWS Load Balancer Controller Helm chart installed with createIngressClassResource=true and ingressClass=alb. This override is therefore only adapting the parent behavior to the AWS plus Karpenter case and does not apply to all clusters. For extra clarity we could also add a guard that falls back to the base implementation if the LBC is not installed, for example:
```
out, _, _ = container_service.RunKubectlCommand(
    ['get', 'deploy', 'aws-load-balancer-controller', '-n', 'kube-system', '-o', 'name'],
    raise_on_failure=False
)
if not (out or '').strip():
    return super().DeployIngress(name, namespace, port)
```
I did not include that here because the current flow always installs the controller but I am open to adding it if you think it improves clarity

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: I have added the check, if string is empty or none it will fail

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +622,158 @@ def _Create(self):
...
+            ])
+            stdout, _, _ = vm_util.IssueCommand(util.AWS_PREFIX + [
+                'iam', 'create-policy',
+                '--policy-name', 'AWSLoadBalancerControllerIAMPolicy',
+                '--policy-document', f'file://{tf.name}',
+                '--query', 'Policy.Arn', '--output', 'text',
+            ])
+            policy_arn = (stdout or '').strip()
+    # 3) Ensure ServiceAccount
+    vm_util.IssueCommand([
```
hubatish: Oof this whole section looks hella tricky. How many of these are theoretically one-off steps in the project? Some possible problems:
- I'm guessing I'm gonna get some errors here when running this in my project.
- Our cluster may not have permissions to run all these commands
- Permanent (rather than temporary) changes to a user's project or local environment cause semi-unintended side effects from running PKB - which we try to avoid. I've avoided this in the past (including for some contract runs) by moving these to an internal startup script / a series of bash commands to run + some comments in the file.

Should we instead just schedule a meeting between us where I try running the benchmark / these commands & see what errors I run into?
This is especially if indeed these are one-off steps.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +622,158 @@ def _Create(self):
     }
     self._EksCtlCreate(create_json)
 
+  def _InstallAwsLoadBalancerController(self) -> None:
+    """Install AWS Load Balancer Controller for the cluster."""
```
hubatish: nit: Use Installs with an s rather than Install without to match the style of the rest of the functions in this file. Also for _PostIngressNetworkingFixups below.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -621,6 +622,158 @@ def _Create(self):
...
+    container_service.RunKubectlCommand([
+        'rollout', 'status', 'deployment/aws-load-balancer-controller',
+        '-n', 'kube-system', '--timeout=180s',
+    ])
+
+  def DeployIngress(self, name: str, namespace: str, port: int) -> str:
+    """Deploys only Service + Ingress (without IngressClass) for AWS Load Balancer Controller."""
+    # Apply the custom manifest template (service + ingress with annotations).
+    self.ApplyManifest(
+        'container/kubernetes_hpa/ingress_alb.yaml.j2',
```
hubatish: Let's put this in the data/container/karpenter folder rather than data/container/kubernetes_hpa, since this is specific to Karpenter but could be used by other benchmarks.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
hubatish: Lots of good error checking going on here. I meant, can you also add some new unit tests to elastic_kubernetes_service_test.py confirming that this behavior works as intended.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -653,6 +806,9 @@ def _PostCreate(self):
         'logLevel=debug',
         '--wait',
     ])
+    # Ensure ALB ingress support: install AWS Load Balancer Controller.
+    if FLAGS.eks_install_alb_controller:
+        self._InstallAwsLoadBalancerController()
```
hubatish: Maybe we discussed offline, but why is this flag gated? / ok actually it makes sense that we don't want to run this in scale but would want to run this in kubernetes_hpa - is that the main reason for a flag? Should I set said flag when running the kubernetes_hpa_benchmark with EKS Karpenter every time?

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -690,7 +846,20 @@ def _PostCreate(self):
...
+        FLAGS.kubectl,
+        '--kubeconfig',
+        FLAGS.kubeconfig,
+        'delete',
+        'ingress',
+        '--all',
+        '--all-namespaces',
+        '--timeout=30s',
+    ], suppress_failure=lambda stdout, stderr, retcode: (
+        'deleted' in stdout and 'timed out waiting for the condition' in stderr
```
hubatish: Does this need to be retried if it failed? (this might actually be handled by _Delete automatically) But my main comment here is that timeout=30s seems kinda short, especially if we are getting a timed out message sometimes. Should we increase that timeout value?

kiryl-filatau: Agree, looks like a typo) thanks for catch, will update to 300, should be fine.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Agree

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Changed

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Done

kiryl-filatau: extended  kubectl delete ingress timeout to 600s as ALB finalizers need more time

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: I have removed the flag, should work only for kubernetes_hpa test

kiryl-filatau: I returned the flag, as it looks better. This was discussed in the chat

perfkitbenchmarker/data/container/karpenter/ingress_alb.yaml.j2
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  namespace: {{ namespace }}
+  name: {{ name }}
+  annotations:
+    alb.ingress.kubernetes.io/scheme: internet-facing 
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/healthcheck-path: /calculate
```
rsgowman: `/calculate` is specific to the (fibonacci) workload; it should probably be a variable too.

perfkitbenchmarker/providers/aws/flags.py
```diff
@@ -212,9 +212,17 @@
     1,
     249,
 )
+flags.DEFINE_boolean(
+    'eks_install_alb_controller',
+    False,
+    'Whether to install AWS Load Balancer Controller in EKS Karpenter clusters'
+    'Default value - don not install unless explicitly requested',
```
rsgowman: typo: s/don/do/

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
kiryl-filatau: Unit tests were added

PR #6029, opened 2025-09-03
AKS nodepool provisioning

vofish:
This PR adds support for `nodepool provisioning` in AKS with the auto-provisioning feature.

**Key Changes:**
1. Manifests for `AKSNodeClass` and `NodePool`:
    - Added manifests to handle tolerations and nodeSelector for job scheduling.
2. Modifications to the _Create Method in AksCluster:
    - Updated the `_Create` method to create an AKS cluster with the `--node-provisioning-mode=auto` feature.
3. Updates to the `Provisioning Node Pool Benchmark`:
    - Applied additional Azure-specific manifests.
    - Skipped the node pool count check, as AKS auto-provisioning does not manage node pools in the traditional sense.

**Additional Notes**
- This implementation uses the nodepools.karpenter.sh Kubernetes CRD to manage node provisioning policies and workload-specific configurations, as AKS auto-provisioning operates at the Kubernetes level, not the cloud-based node pool level.
- For more details, refer to the Microsoft documentation:
   [Node auto-provisioning overview](https://learn.microsoft.com/en-us/azure/aks/node-autoprovision?tabs=azure-cli)
   [Node auto-provisioning with node pools](https://learn.microsoft.com/en-us/azure/aks/node-autoprovision-node-pools)

### DIFF
--- a/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
+++ b/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
@@ -8,7 +8,7 @@ spec:
 apiVersion: karpenter.sh/v1
 kind: NodePool
 metadata:
-  name: batch-processing
+  name: {{ batch }}-nodepool
 spec:
   disruption:
     budgets:
--- a/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
+++ b/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
@@ -1,0 +1,21 @@
+# nodepool-azure.yaml.j2
+# Configures the Azure Node Auto Provisioner (Karpenter) to provision nodes
+# for pods that request a 'group' label matching the 'pool-{{ batch }}-{{ id }}' pattern.
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: {{ name | default('default') }}
+spec:
+  template:
+    metadata:
+      labels:
+        pkb_nodepool: {{ name | default('default') }}
+    spec:
+      # Define the rules for provisioning.
+      requirements:
+        - key: "group"
+          operator: In
+          values:
+          {% for i in range(1, job_count + 1) %}
+          - pool-{{ batch }}-{{ "%03d" | format(i) }}
+          {% endfor %}
--- a/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
+++ b/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
@@ -24,20 +24,16 @@ spec:
       requirements:
         - key: kubernetes.io/arch
           operator: In
-          values:
-          - amd64
+          values: ["amd64"]
         - key: kubernetes.io/os
           operator: In
-          values:
-          - linux
+          values: ["linux"]
         - key: karpenter.sh/capacity-type
           operator: In
-          values:
-          - on-demand
+          values: ["spot", "on-demand"]
         - key: karpenter.azure.com/sku-family
           operator: In
-          values:
-          - D
+          values: ["D", "E", "F"]
         - key: "group"
           operator: In
           values:
--- a/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
+++ b/perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
@@ -1,18 +1,43 @@
-# nodepool-azure.yaml.j2
-# Configures the Azure Node Auto Provisioner (Karpenter) to provision nodes
-# for pods that request a 'group' label matching the 'pool-{{ batch }}-{{ id }}' pattern.
-apiVersion: karpenter.sh/v1beta1
+apiVersion: karpenter.azure.com/v1beta1
+kind: AKSNodeClass
+metadata:
+  name: aks-managed-class
+spec:
+  imageFamily: "Ubuntu2204"
+---
+apiVersion: karpenter.sh/v1
 kind: NodePool
 metadata:
-  name: {{ name | default('default') }}
+  name: batch-processing
 spec:
+  disruption:
+    budgets:
+    - nodes: 30%
+    consolidateAfter: 0s
+    consolidationPolicy: WhenEmptyOrUnderutilized
   template:
-    metadata:
-      labels:
-        pkb_nodepool: {{ name | default('default') }}
     spec:
-      # Define the rules for provisioning.
+      nodeClassRef:
+        group: karpenter.azure.com
+        kind: AKSNodeClass
+        name: aks-managed-class
       requirements:
+        - key: kubernetes.io/arch
+          operator: In
+          values:
+          - amd64
+        - key: kubernetes.io/os
+          operator: In
+          values:
+          - linux
+        - key: karpenter.sh/capacity-type
+          operator: In
+          values:
+          - on-demand
+        - key: karpenter.azure.com/sku-family
+          operator: In
+          values:
+          - D
         - key: "group"
           operator: In
           values:
--- a/perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
@@ -84,7 +84,6 @@ TEST_BATCH_NAME = "test-batch"
 JOB_MANIFEST_TEMPLATE = "provision_node_pools/job_manifest.yaml.j2"
 AKS_NODEPOOL_MANIFEST_TEMPLATE = "provision_node_pools/nodepool_azure.yaml.j2"
 
-
 def GetConfig(user_config):
   return configs.LoadConfig(BENCHMARK_CONFIG, user_config, BENCHMARK_NAME)
 
@@ -107,9 +106,9 @@ def _CreateJobsAndWait(
   if cluster.CLOUD == 'Azure':
     cluster.ApplyManifest(
         AKS_NODEPOOL_MANIFEST_TEMPLATE,
-        name='default',
         batch=batch_name,
         job_count=jobs,
+        cloud=cluster.CLOUD,
     )
 
   apply_start = time.monotonic()
--- a/perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
@@ -82,6 +82,7 @@ provision_node_pools:
 INIT_BATCH_NAME = "init-batch"
 TEST_BATCH_NAME = "test-batch"
 JOB_MANIFEST_TEMPLATE = "provision_node_pools/job_manifest.yaml.j2"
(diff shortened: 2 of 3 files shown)


perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -169,13 +169,19 @@ def _Create(self):
...
         f'pkb_nodepool={container_service.DEFAULT_NODEPOOL}',
-    ] + self._GetNodeFlags(self.default_nodepool)
-    if self._IsAutoscalerEnabled():
-      cmd += [
-          '--enable-cluster-autoscaler',
-          f'--min-count={self.min_nodes}',
-          f'--max-count={self.max_nodes}',
-      ]
+    ]
+    if FLAGS.benchmarks and 'provision_node_pools' in FLAGS.benchmarks:
```
hubatish: Add a new flag in azure/flags.py which controls this behavior. Can set it manually or in the `flags:` section of the provision_node_pools_benchmark's config.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -169,13 +169,19 @@ def _Create(self):
...
+      # For provision_node_pools benchmark, add auto provisioning mode
+      cmd.append('--node-provisioning-mode=auto')
+      cmd += self._GetNodeFlags(self.default_nodepool)
```
hubatish: small: _GetNodeFlags looks 100% the same so should just be moved back to the `] + self._GetNodeFlags(...)` code it was prior to this change.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -169,13 +169,19 @@ def _Create(self):
...
+    else:
+      cmd += self._GetNodeFlags(self.default_nodepool)
+      if self._IsAutoscalerEnabled():
+        cmd += [
+            '--enable-cluster-autoscaler',
+            f'--min-count={self.min_nodes}',
+            f'--max-count={self.max_nodes}',
+        ]
```
hubatish, line 182: Does the below behavior for _CreateNodePool need to change? Seems like we shouldn't be mixing the concepts.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -102,6 +103,15 @@ def _CreateJobsAndWait(
       batch_name,
       jobs,
   )
+
+  if cluster.CLOUD == "Azure":
```
hubatish: This should go in an AddNodepools function in the resource. I also gave Hayk a similar comment in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/5969

perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
```diff
@@ -0,0 +1,42 @@
...
+        - key: kubernetes.io/os
+          operator: In
+          values: ["linux"]
+        - key: karpenter.sh/capacity-type
+          operator: In
+          values: ["spot", "on-demand"]
+        - key: karpenter.azure.com/sku-family
+          operator: In
+          values: ["D", "E", "F"]
+        - key: "group"
```
hubatish, line 40: Can you explain this "group" requirement? This seems like the part which could create a lot of nodepools.. but right now it looks like just 1-2 nodepools with the name eg init-nodepool or test-nodepool are created via the above root NodePool definition. Can also explain/ask about this offline in our chat.

vofish: This group is to handle:
```
nodeSelector:
   group: pool-{{ batch }}-{{ id }}
```

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: No additional flag is necessary.

The implementation automatically detects if the benchmark is `provision_node_pools` and includes the `--node-provisioning-mode=auto` option in the cluster creation command accordingly.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: We are not mixing the concepts.

This section is needed to create a cluster with the `--node-provisioning-mode=auto` feature, addressing the limitation:
```
You can't enable node autoprovisioning in a cluster 
where node pools have cluster autoscaler enabled
```

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
vofish: Move this logic to the jinja.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: Yes, it does, but I'm saying that checking benchmark_name in the resource is bad practice. Instead use a flag & can set the value of that flag from the benchmark / manually when running it instead.

perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
hubatish: Implementation swapped from this nodepool_azure.yaml.j2 to modifying perfkitbenchmarker/data/provision_node_pools/job_manifest.yaml.j2 right? I believe the more recent one looks like it's correctly adding more nodepools.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Flag added.

perfkitbenchmarker/data/provision_node_pools/nodepool_azure.yaml.j2
vofish: Yes, correct.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: Moved back

perfkitbenchmarker/data/provision_node_pools/job_manifest.yaml.j2
```diff
@@ -35,3 +35,48 @@ spec:
...
+{%- endif %}
+---
+apiVersion: karpenter.sh/v1
+kind: NodePool
+metadata:
+  name: {{ batch }}-nodepool-{{ id }}
+spec:
+  disruption:
+    budgets:
+    - nodes: 30%
```
hubatish: What does this value do? I generally lean towards using default behavior / underspecifying attributes when reasonable as it simplifies comparisons.

perfkitbenchmarker/data/provision_node_pools/job_manifest.yaml.j2
```diff
@@ -35,3 +35,48 @@ spec:
...
+          values: ["amd64"]
+        - key: kubernetes.io/os
+          operator: In
+          values: ["linux"]
+        - key: karpenter.sh/capacity-type
+          operator: In
+          values: ["on-demand"]
+        - key: karpenter.azure.com/sku-family
+          operator: In
+          values: ["D"]
```
hubatish: This should be a configurable value. Ideally we can use the config_override.container_cluster.vm_spec.machine_type value directly, eg Standard_D2s_v6. If not (ie we need to translate from Standard_D2s_v6 to these various sku family + size values), then this can be punted onto a later PR.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -196,6 +199,10 @@ def _AssertNodePools(
     added_node_pools: int,
 ) -> None:
   """Asserts expected number of node pools in the cluster."""
+  # Skip node pool count check for Azure AKS when using auto-provisioning, as nodes are added without associated node pools
+  if cluster.CLOUD == "Azure":
+    logging.info("Skipping node pool check for Azure AKS")
+    return
   node_pools = len(cluster.GetNodePoolNames())
```
hubatish, line 203: Can we not get the list of Karpenter Node pools? Maybe ask a stackoverflow (or possibly Azure or AWS site) question about this?

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
```diff
@@ -102,6 +104,7 @@ def _CreateJobsAndWait(
       batch_name,
       jobs,
   )
+
   apply_start = time.monotonic()
   for i in range(jobs):
     cluster.ApplyManifest(
```
hubatish, line 111: This is the part I'm waiting on Hayk's PR for. IMO we can call AddNodepool here.

perfkitbenchmarker/data/provision_node_pools/job_manifest.yaml.j2
vofish: The `disruption.budgets` value limits the percentage of node disruptions (e.g., during scaling or updates), ensuring availability of workloads. Can be removed.

perfkitbenchmarker/data/provision_node_pools/job_manifest.yaml.j2
vofish: This section is meant to define only the `sku-family` (e.g., "D", "E", "F") to provide flexibility within a family of VMs, allowing Karpenter to handle the selection of a specific SKU through `karpenter.azure.com/sku-name` when needed. Since it's Karpenter's responsibility to choose the most appropriate node, leaving `sku-name` unspecified aligns well with balancing configurability and leveraging Karpenter's capabilities.

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
vofish: Here we skip getting Karpenter Node pools

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
vofish: Will review once his PR is merged

vofish: Added and aligned with AWS logic

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
vofish: The AddNodepools function added

perfkitbenchmarker/linux_benchmarks/provisioning_benchmarks/provision_node_pools_benchmark.py
vofish: This part was refactored and aligned with AWS logic.

PR #6055, opened 2025-09-11
Fix issue5883

vofish:
Fix multiprocessing on macOS to use 'fork' instead of 'spawn'

### DIFF
--- a/perfkitbenchmarker/pkb.py
+++ b/perfkitbenchmarker/pkb.py
@@ -1509,13 +1509,9 @@ class ZoneRetryManager:
     """Changes zone to be a new zone in the different region."""
     region = self._utils.GetRegionFromZone(self._GetCurrentZoneFlag())
     self._regions_tried.add(region)
-    regions_to_try = (
-        {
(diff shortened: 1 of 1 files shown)


perfkitbenchmarker/pkb.py
```diff
@@ -1974,6 +1975,14 @@ def Main(argv):
   log_util.ConfigureBasicLogging()
   _InjectBenchmarkInfoIntoDocumentation()
 
+  # Fix multiprocessing on macOS to use 'fork' instead of 'spawn'
+  # to avoid pickle errors with local functions issue #5883
+  if platform.system() == 'Darwin':
+    try:
+      multiprocessing.set_start_method('fork', force=True)
+    except RuntimeError:
+      pass
```
hubatish: Why pass here? Is an error expected here or just playing it overly safe? Please remove or add a comment or log statement explaining why this is safe to skip.

vofish: Remove this and apply linting

PR #6089, opened 2025-09-22
Fix issue with AKS deployment failing due to admission webhook validation

vofish:
- Fixed issue with AKS deployment failing due to admission webhook validation
```
STDERR: Error from server (Forbidden): error when creating "/var/folders/_7/qlc76gs57gjg9dw3cjqy4zp00000gn/T/tmp88_902f_.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [azurepolicy-k8sazurev1antiaffinityrules-3dbdca1776c2f1f6a18a] Deployment with 5 replicas should have either podAntiAffinity or topologySpreadConstraints set to avoid disruptions due to nodes crashing
[azurepolicy-k8sazurev2containerenforceprob-9c81f9baeae421a77e19] [Implied by gatekeeper-expand-workloads] Container <kubernetes-scaleup> in your Pod <kubernetes-scaleup-pod> has no <livenessProbe>. Required probes: ["readinessProbe", "livenessProbe"]
[azurepolicy-k8sazurev2containerenforceprob-9c81f9baeae421a77e19] [Implied by gatekeeper-expand-workloads] Container <kubernetes-scaleup> in your Pod <kubernetes-scaleup-pod> has no <readinessProbe>. Required probes: ["readinessProbe", "livenessProbe"]
```

- Updated the `_GetCredentials` method to include the `--overwrite-existing` flag, resolving an issue when the kubeconfig file already exists.

- Updated the `_CreateRoleAssignment` method to fix an issue with the `create_role_assignment_cmd` command, which requires `user.objectId` instead of `user.name`
```
STDERR: WARNING: Failed to query `user.name` by invoking Graph API. If you don't have permission to query Graph API, please specify --assignee-object-id and --assignee-principal-type.
```

### DIFF
--- a/tests/providers/azure/azure_kubernetes_service_test.py
+++ b/tests/providers/azure/azure_kubernetes_service_test.py
@@ -73,7 +73,17 @@ class AzureKubernetesServiceTest(pkb_common_test_case.PkbCommonTestCase):
         {
             'az aks create': [('', '', 0)],
             'az aks nodepool': [('', '', 0)],
-            'az aks show --query provisioningState --output tsv': [("Succeeded", '', 0)],
+            'az aks show': [
+                (
+                    (
+                        '{"provisioningState": "Succeeded",'
+                        ' "nodeResourceGroup": "node-resource-group"}'
+                    ),
+                    '',
+                    0,
+                ),
+                ('Succeeded', '', 0),
+            ],
             'get serviceAccounts': [('default, foo', '', 0)],
         },
     )
@@ -179,9 +189,22 @@ class AzureKubernetesServiceTest(pkb_common_test_case.PkbCommonTestCase):
             'az aks create': [('', '', 0)],
             'az aks nodepool': [('', '', 0)],
             '--query id': [('cluster-id', '', 0)],
-            'az aks show --query provisioningState': [("Succeeded", '', 0)],
+            'az aks show': [
+                (
+                    (
+                        '{"provisioningState": "Succeeded",'
+                        ' "nodeResourceGroup": "node-resource-group"}'
+                    ),
+                    '',
+                    0,
+                ),
+                ('Succeeded', '', 0),
+            ],
             'get serviceAccounts': [('default, foo', '', 0)],
-            'az account show': [('user-name', '', 0)],
+            'az account show': [
+                ('servicePrincipal', '', 0),
+                ('user-name', '', 0),
+            ],
             'az role assignment': [('', '', 0)],
         },
     )
@@ -200,7 +223,8 @@ class AzureKubernetesServiceTest(pkb_common_test_case.PkbCommonTestCase):
                 '--scope',
                 'cluster-id',
             ],
-        ),])
+        ),
+    ])
 
   def testGetNodePoolNames(self):
     self.MockIssueCommand(


perfkitbenchmarker/data/container/kubernetes_scale/kubernetes_scale.yaml.j2
```diff
@@ -27,6 +27,20 @@ spec:
             {%- if NvidiaGpuRequest %}
             nvidia.com/gpu: {{ NvidiaGpuRequest }}
             {%- endif %}
+        # Added livenessProbe
+        livenessProbe:
```
hubatish, line 31: Can we pull either of these changes from the AKS Automatic implementation rather than baked into the yaml? 
/ ok we've fixed AKS Automatic on scale, but I'm wondering:
- Will AKS Automatic now fail for HPA & nodepool provisioning?
- Will other clouds now fail for scaling?

See eg the GetNodeSelectors code for an example.. There's probably also a somewhat clever refactor here using the yaml library to modify this better in code (ie replacing `GetNodeSelectors` with `ModifyPodSpecification(yaml)`) .. but that might be out of scope.

For now there are 2 possible solutions:
- Address the above questions about other clouds / benchmarks with manual runs + report back (does AKS Auto work for those other benchmarks right now)?
- Do a clever refactor / somehow move this code into the azure_kubernetes_service.py AKS Automatic resource.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -325,6 +325,7 @@ def _GetCredentials(self, use_admin: bool) -> None:
         self.name,
         '--file',
         FLAGS.kubeconfig,
+        '--overwrite-existing',
```
hubatish: azure_kubernetes_service_test is failing, might just need to update the mocks

perfkitbenchmarker/linux_benchmarks/kubernetes_scale_benchmark.py
```diff
@@ -62,7 +62,7 @@
 )
 
 MANIFEST_TEMPLATE = 'container/kubernetes_scale/kubernetes_scale.yaml.j2'
-DEFAULT_IMAGE = 'k8s.gcr.io/pause:3.1'
+DEFAULT_IMAGE = 'busybox:1.37'
```
vofish, line 65: Replace the default image with `busybox`, as the `pause` container does not support liveness or readiness probes.

perfkitbenchmarker/linux_benchmarks/kubernetes_scale_benchmark.py
```diff
@@ -155,11 +155,10 @@ def ScaleUpPods(
   initial_pods = set(cluster.GetPodNames())
   logging.info('Initial pods: %s', initial_pods)
 
-  command = None
   if virtual_machine.GPU_COUNT.value:
     # Use nvidia-smi to validate NVIDIA_GPU is available.
     command = ['sh', '-c', 'nvidia-smi && sleep 3600']
-  if CONTAINER_IMAGE.value:
+  else:
```
vofish, line 161: Update this logic because the command is also needed for default image.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -154,6 +154,8 @@ def _IsAutoscalerEnabled(self):
 
   def _Create(self):
     """Creates the AKS cluster."""
+    if self._Exists():
+      return
```
vofish: Add this check to resolve the issue where running with `--run_stage=provision,prepare,run` and `--run_uri=XXXX` fails due to the cluster already existing.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -339,11 +342,15 @@ def _IsReady(self) -> bool:
         'show',
         '--name',
         self.name,
+        '--query',
+        'provisioningState',
+        '--output',
+        'tsv',
```
vofish, line 345: Update this to suppress unnecessary output.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -458,6 +465,8 @@ class AksAutomaticCluster(AksCluster):
 
   def _Create(self):
     """Creates the Automatic AKS cluster with tags."""
+    if self._Exists():
+      return
```
vofish: Add this check to resolve the issue where running with --run_stage=provision,prepare,run and --run_uri=XXXX fails due to the cluster already existing.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -529,7 +538,18 @@ def _PostCreate(self):
...
+        'account',
+        'show',
+        '--query',
+        'user.type',
+        '--output',
+        'tsv',
+    ])
+    user_type = user_type.strip()
+    if user_type == 'servicePrincipal':
+      self._CreateRoleAssignment()
```
vofish, line 547: Add this logic to ensure the method is executed only for Service Principals.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: If the cluster already exists you should skip the provision phase & go straight to the run phase with just `run_stage=run`. I can see this code being useful for development (perhaps so you could try out the post create steps more quickly) but is not needed for production code.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: Same comment as above.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: thanks; I'll try running this again.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: Please fix this

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: This was removed; great.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
vofish: This flag has already been removed

hubatish (approved):
tests updated; pr still lgtm

hubatish (approved):

PR #6103, opened 2025-09-25
feat: Add llm_eval benchmark

anksena:
This pull request introduces a new benchmark, llm_eval, designed to evaluate the performance of various Large Language Model (LLM) providers. The benchmark measures key metrics such as time-to-first-token, full
  response time, and token counts for both streaming and non-streaming responses.

  Key Features

   * New Benchmark: llm_eval
   * Supported Providers: Google, OpenAI, Anthropic
   * Metrics:
       * Time to first token (streaming)
       * Full response time (streaming and non-streaming)
       * Input and output token counts

  Prerequisites

  This benchmark requires API keys for the respective LLM providers. These keys should be placed in a .env file and uploaded to a GCS bucket. The benchmark will download and use this file for authentication.

  Example .env file structure:

   1 GOOGLE_API_KEY=your_google_api_key
   2 OPENAI_API_KEY=your_openai_api_key
   3 ANTHROPIC_API_KEY=your_anthropic_api_key

  How to Run the Benchmark

  To run the benchmark, use the following pkb.py commands, replacing <your_gcs_bucket> with the name of the GCS bucket containing your .env file.

  Google:

   1 .venv/bin/python pkb.py --benchmarks=llm_eval --llm_eval_provider=google --gcp_preprovisioned_data_bucket=<your_gcs_bucket>

  OpenAI:

   1 .venv/bin/python pkb.py --benchmarks=llm_eval --llm_eval_provider=openai --gcp_preprovisioned_data_bucket=<your_gcs_bucket>

  Anthropic:

   1 .venv/bin/python pkb.py --benchmarks=llm_eval --llm_eval_provider=anthropic --gcp_preprovisioned_data_bucket=<your_gcs_bucket>

### DIFF
--- a/perfkitbenchmarker/data/llm_eval/logger.py
+++ b/perfkitbenchmarker/data/llm_eval/logger.py
@@ -1,0 +1,38 @@
+"""Configures and provides a logger for the application."""
+
+import logging
+import sys
+
+
+def get_logger(name: str) -> logging.Logger:
+  """Configures and returns a logger instance.
+
+  Args:
+    name: The name of the logger.
+
+  Returns:
+    A configured logger instance.
+  """
+  logger = logging.getLogger(name)
+
+  handler = logging.StreamHandler(sys.stdout)
+  formatter = logging.Formatter(
+      '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+      ' (%(filename)s:%(lineno)d)'
+  )
+  handler.setFormatter(formatter)
+
+  logger.addHandler(handler)
+  return logger
+
+
+def set_logging_level(debug: bool) -> None:
+  """Sets the logging level for the root logger.
+
+  Args:
+    debug: If True, sets the logging level to DEBUG, otherwise INFO.
+  """
+  if debug:
+    logging.getLogger().setLevel(logging.DEBUG)
+  else:
+    logging.getLogger().setLevel(logging.INFO)
--- a/perfkitbenchmarker/data/llm_eval/main.py
+++ b/perfkitbenchmarker/data/llm_eval/main.py
@@ -1,0 +1,217 @@
+"""Main script for running LLM benchmarks.
+
+This script provides a command-line interface for benchmarking large language
+models from various providers. It supports model discovery, benchmarking
+specific models, and various output formats.
+"""
+
+import enum
+import json
+from typing import Any, Dict, List, Tuple
+
+from absl import app
+from absl import flags
+import dotenv
+from logger import get_logger
+from logger import set_logging_level
+from prompts import OPEN_ENDED_PROMPTS
+from prompts import SUMMARIZATION_PROMPTS
+from providers.anthropic_provider import AnthropicProvider
+from providers.base_provider import BaseProvider
+from providers.google_provider import GoogleProvider
+from providers.openai_provider import OpenAIProvider
+
+dotenv.load_dotenv()
+
+FLAGS = flags.FLAGS
+logger = get_logger(__name__)
+
+
+class Provider(enum.Enum):
+  GOOGLE = 'google'
+  OPENAI = 'openai'
+  ANTHROPIC = 'anthropic'
+
+
+flags.DEFINE_enum(
+    'provider',
+    None,
+    [p.value for p in Provider],
+    'The provider to benchmark.',
+)
+flags.mark_flag_as_required('provider')
+flags.DEFINE_integer(
+    'max_models',
+    -1,
+    (
+        'The maximum number of models to test in a single invocation. This '
+        'flag has no effect if the --models flag is specified.'
+    ),
+)
+flags.DEFINE_list(
+    'specific_models',
+    None,
+    'An optional list of models to benchmark, overriding the discovery.',
+)
+flags.DEFINE_boolean(
+    'discover_models_only',
+    False,
+    'Discover models and exit without benchmarking. Can be used with'
+    ' --include_preview.',
+)
+flags.DEFINE_boolean(
+    'include_preview',
+    False,
+    'Include preview models from the benchmark. This flag has no effect when'
+    ' --models is used.',
+)
+flags.DEFINE_boolean('debug', False, 'Enable debug logging.')
+
+
+PROVIDER_MAPPING = {
+    Provider.GOOGLE.value: GoogleProvider,
+    Provider.OPENAI.value: OpenAIProvider,
+    Provider.ANTHROPIC.value: AnthropicProvider,
+}
+
+
+def _get_models_to_test(provider: BaseProvider) -> Tuple[List[str], List[str]]:
+  """Gets the list of models to test based on the command-line flags.
+
+  Args:
+    provider: The provider to get models from.
+
+  Returns:
+    A tuple containing the list of models to test and the list of skipped
+    models.
+  """
+  if FLAGS.specific_models:
+    if FLAGS.include_preview:
+      raise ValueError('Cannot use --specific_models with --include_preview.')
+    if FLAGS.max_models != 20:
+      raise ValueError('Cannot use --specific_models with --max_models.')
+    models_to_test = FLAGS.specific_models
+    skipped_models = []
+  else:
+    all_models = provider.get_models(include_preview=FLAGS.include_preview)
+    models_to_test = all_models[: FLAGS.max_models]
+    skipped_models = all_models[FLAGS.max_models :]
+
+  logger.debug(f'Models to be benchmarked: {models_to_test}')
+
+  return models_to_test, skipped_models
+
+
+def _get_provider() -> BaseProvider:
+  """Gets the provider instance based on the command-line flags.
+
+  Returns:
+    An instance of the selected provider.
+  """
+  logger.debug(f'Starting benchmark for provider: {FLAGS.provider}')
+
+  provider_class = PROVIDER_MAPPING.get(FLAGS.provider)
+  if not provider_class:
+    raise ValueError(f'Invalid provider: {FLAGS.provider}')
+  return provider_class()
+
+
+def _handle_discover_only(models_to_test: List[str]) -> bool:
+  """Handles the discover_models_only flag.
+
+  Args:
+    models_to_test: The list of models to test.
+
+  Returns:
+    True if the script should exit after discovering models, False otherwise.
+  """
+  if not FLAGS.discover_models_only:
+    return False
+
+  if FLAGS.specific_models:
+    raise ValueError(
+        'Cannot use --specific_models with --discover_models_only.'
+    )
+
+  logger.debug('Discovering models...')
+  print(json.dumps({'models': models_to_test}, indent=2))
+  return True
+
+
+def _run_benchmarks(
+    provider: BaseProvider, models_to_test: List[str]
+) -> Dict[str, Any]:
+  """Runs the benchmarks for the given models.
+
+  Args:
+    provider: The provider to use for benchmarking.
+    models_to_test: The list of models to test.
+
+  Returns:
+    A dictionary of benchmark results.
+  """
+  results = {}
+  for model_name in models_to_test:
+    logger.debug(f'Benchmarking model: {model_name}')
+    results[model_name] = {}
+    for prompt_id, prompt_data in SUMMARIZATION_PROMPTS.items():
+      try:
+        results[model_name][prompt_id] = provider.benchmark_model(
+            model_name,
+            prompt_data['text'],
+            prompt_id,
+            prompt_data['max_tokens'],
+        )
+      except Exception as e:  # pylint: disable=broad-exception-caught
+        # Catching broad exception to ensure one failed prompt doesn't stop
+        # the entire benchmark run.
+        logger.debug(f"  Error benchmarking prompt '{prompt_id}': {e}")
+        results[model_name][prompt_id] = {'error': str(e)}
+    for prompt_id, prompt_data in OPEN_ENDED_PROMPTS.items():
+      try:
+        results[model_name][prompt_id] = provider.benchmark_model(
+            model_name,
+            prompt_data['text'],
+            prompt_id,
+            prompt_data['max_tokens'],
+        )
+      except Exception as e:  # pylint: disable=broad-exception-caught
+        # Catching broad exception to ensure one failed prompt doesn't stop
+        # the entire benchmark run.
+        logger.debug(f"  Error benchmarking prompt '{prompt_id}': {e}")
+        results[model_name][prompt_id] = {'error': str(e)}
+    logger.debug(f'Finished benchmarking model: {model_name}')
+  return results
+
+
+def main(argv: List[str]) -> None:
+  """Main function for the benchmark script.
+
+  Args:
+    argv: The command-line arguments.
+  """
+  del argv  # Unused
+  set_logging_level(FLAGS.debug)
+  provider = _get_provider()
+  models_to_test, skipped_models = _get_models_to_test(provider)
+
+  if _handle_discover_only(models_to_test):
+    return
+
+  results = _run_benchmarks(provider, models_to_test)
+
+  logger.debug('Benchmark complete. Finalizing output.')
+
+  num_skipped_models = len(skipped_models)
+  output = {
+      'results': results,
+      'models_skipped': num_skipped_models,
+      'skipped_models': skipped_models,
+      'prompts_tested': {**SUMMARIZATION_PROMPTS, **OPEN_ENDED_PROMPTS},
+  }
+
+  print(json.dumps(output, indent=2))
+
+
+if __name__ == '__main__':
+  app.run(main)
--- a/perfkitbenchmarker/data/llm_eval/providers/base_provider.py
+++ b/perfkitbenchmarker/data/llm_eval/providers/base_provider.py
@@ -1,0 +1,44 @@
+"""Abstract base class for LLM providers."""
+
+import abc
+from typing import Any, Dict, List
+
+from logger import get_logger
+
+
+logger = get_logger(__name__)
+
+TEMPERATURE = 0.0
+
+
+class BaseProvider(abc.ABC):
+  """Abstract base class for LLM providers."""
+
+  @abc.abstractmethod
+  def get_models(self, include_preview: bool = False) -> List[str]:
+    """Fetches the latest generation of text-based models from the provider.
+
+    Args:
+      include_preview: Whether to include preview models in the list.
+
+    Returns:
+      A list of model names.
+    """
+    pass
+
+  @abc.abstractmethod
+  def benchmark_model(
+      self, model_name: str, prompt: str, prompt_id: str, max_output_tokens: int
+  ) -> Dict[str, Any]:
+    """Benchmarks a single model.
+
+    Args:
+      model_name: The name of the model to benchmark.
+      prompt: The prompt to send to the model.
+      prompt_id: The ID of the prompt.
+      max_output_tokens: The maximum number of tokens to generate.
+
+    Returns:
+      A dictionary of benchmark results.
+    """
+    logger.debug(f"Benchmarking model: {model_name} with prompt: '{prompt_id}'")
--- a/perfkitbenchmarker/data/llm_eval/providers/google_provider.py
+++ b/perfkitbenchmarker/data/llm_eval/providers/google_provider.py
@@ -1,0 +1,268 @@
+"""Provider for benchmarking Google models."""
+
+import datetime
+import itertools
+import os
+import re
+import time
+from typing import Any, Dict, List
+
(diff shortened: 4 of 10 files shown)


perfkitbenchmarker/data/llm_eval/logger.py
```diff
@@ -0,0 +1,38 @@
+"""Configures and provides a logger for the application."""
+
+import logging
+import sys
+
+
+def get_logger(name: str) -> logging.Logger:
```
hubatish, line 7: Why do we need these overrides / what's wrong with the defaults?

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+)
+_DISCOVER_MODELS_ONLY = flags.DEFINE_boolean(
+    'llm_eval_discover_models_only',
+    False,
+    'Discover models and exit without benchmarking.',
+)
+_INCLUDE_PREVIEW = flags.DEFINE_boolean(
+    'llm_eval_include_preview',
+    False,
+    'Include preview models from the benchmark.',
```
hubatish: nit: include preview models in the benchmark (exclude from)

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+    if _INCLUDE_PREVIEW.value:
+      raise errors.Config.InvalidValue(
+          'Cannot use --llm_eval_specific_models with'
+          ' --llm_eval_include_preview.'
+      )
+    if _MAX_MODELS.present:
+      raise errors.Config.InvalidValue(
+          'Cannot use --llm_eval_specific_models with --llm_eval_max_models.'
+      )
+  if _DISCOVER_MODELS_ONLY.value:
```
hubatish: small: indent this check to be under the existing if _SPECIFIC_MODELS.value if statement rather than a separate nesting.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+        cmd_parts.append('--include_preview')
+
+  run_command = ' '.join(cmd_parts)
+  full_cmd = f'cd {vm_util.VM_TMP_DIR}/llm_eval && {run_command}'
+  stdout, _ = vm.RemoteCommand(full_cmd)
+
+  if _DISCOVER_MODELS_ONLY.value:
+    metadata = {'discovered_models': json.loads(stdout)}
+    return [sample.Sample('discovered_models', 0, '', metadata)]
+  else:
```
hubatish: the joy of an early return is that you can unindent all this rather than have it in the else. go/tott/733

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+      if _INCLUDE_PREVIEW.value:
+        cmd_parts.append('--include_preview')
+
+  run_command = ' '.join(cmd_parts)
+  full_cmd = f'cd {vm_util.VM_TMP_DIR}/llm_eval && {run_command}'
+  stdout, _ = vm.RemoteCommand(full_cmd)
+
+  if _DISCOVER_MODELS_ONLY.value:
+    metadata = {'discovered_models': json.loads(stdout)}
+    return [sample.Sample('discovered_models', 0, '', metadata)]
```
hubatish: Should this data just be logged? I guess it is in the stdout, but:
- when would we use this? It seems mostly useful for debugging rather than production use. If so, yeah maybe just prettify the output a little & return with no samples.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+      for prompt, prompt_results in model_results.items():
+        metadata = {
+            'model': model,
+            'prompt': prompt,
+            'provider': _PROVIDER.value,
+            'prompt_text': prompts_tested.get(prompt, {}).get('text'),
+        }
+        if 'error' in prompt_results:
+          metadata['error_message'] = prompt_results['error']
+          samples.append(sample.Sample('error_count', 1, 'count', metadata))
```
hubatish: should we fail the benchmark? I guess if you're testing a bunch of models 1 or 2 failures could be acceptable, but if the same model fails every time that sounds like a problem.

perfkitbenchmarker/data/llm_eval/main.py
```diff
@@ -0,0 +1,217 @@
...
+  Args:
+    provider: The provider to get models from.
+
+  Returns:
+    A tuple containing the list of models to test and the list of skipped
+    models.
+  """
+  if FLAGS.specific_models:
+    if FLAGS.include_preview:
+      raise ValueError('Cannot use --specific_models with --include_preview.')
```
hubatish, line 90: we already validated this in CheckPrerequisites right? I guess this happens in the benchmark rather than in the script.. probably both are not necessary.

perfkitbenchmarker/data/llm_eval/main.py
```diff
@@ -0,0 +1,217 @@
...
+    if FLAGS.max_models != 20:
```
hubatish, line 91: Default values in this & pkb caller to None rather than 20.

perfkitbenchmarker/data/llm_eval/main.py
```diff
@@ -0,0 +1,217 @@
...
+
+  results = _run_benchmarks(provider, models_to_test)
+
+  logger.debug('Benchmark complete. Finalizing output.')
+
+  num_skipped_models = len(skipped_models)
+  output = {
+      'results': results,
+      'models_skipped': num_skipped_models,
+      'skipped_models': skipped_models,
```
hubatish, line 199: more on the discovery & max models stuff.. how do you plan to call this? Discover all models, then run the test on X=max models of them. and then you have some skipped models.. should you have some offset_models flag to jump past the first X you already tested & instead get the second X?

..also how many models are we talking about?

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Returns:
+      A list of model names.
+    """
+    client = genai.Client(api_key=os.environ['GOOGLE_API_KEY'])
+    models = []
+    for m in client.models.list():
+      if (
+          'generateContent' in m.supported_actions
+          and '2.5' in m.name
```
hubatish, line 58: nice filtering through all the duplicates here & in other providers.

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
```diff
@@ -0,0 +1,268 @@
...
+        """Extracts the date from a model name.
+
+        Args:
+          model_name: The name of the model.
+
+        Returns:
+          The date from the model name, or the minimum datetime if no date is
+          found.
+        """
+        match = re.search(r'(\d{2}-\d{2})$', model_name)
```
hubatish: There are a ton of whacky names.. wish this was better supported in general. For your code, a test would be nice.. It is hard to BUILD & test data code but not impossible.

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
```diff
@@ -0,0 +1,268 @@
...
+    Returns:
+      A dictionary of benchmark results.
+    """
+    logger.debug('Starting streaming benchmark...')
+    start_time = time.time()
+    first_token_time = None
+    config = types.GenerateContentConfig(temperature=base_provider.TEMPERATURE)
+    if max_output_tokens:
+      config.max_output_tokens = max_output_tokens
+    config.safety_settings = safety_settings
```
hubatish: IMO we don't need to pass safety settings around - just hardcode & use at the level you need it is fine.

perfkitbenchmarker/data/llm_eval/providers/base_provider.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns:
+      A list of model names.
+    """
+    pass
+
+  @abc.abstractmethod
+  def benchmark_model(
+      self, model_name: str, prompt: str, prompt_id: str, max_output_tokens: int
+  ) -> Dict[str, Any]:
```
hubatish: consider a more structured return type / object so we can be sure each provider returns most of the same metrics

perfkitbenchmarker/data/llm_eval/logger.py
anksena: We need a custom logger configuration here because the default Python logger doesn't print anything unless explicitly configured. This implementation serves three key purposes:

1.  Enables Logging: It adds a StreamHandler so that log messages can actually be displayed.
2.  Provides Richer Context: It sets a specific format ('%(asctime)s - %(name)s - ...') that includes timestamps and file/line numbers, which is invaluable for debugging on a remote VM.
3.  Controlled Debug Output: The logger is used exclusively for DEBUG level messages. This means that during a standard run, stdout is kept clean for the final JSON output. When a developer needs to troubleshoot, they can pass a --debug flag to see the detailed  logs, consciously prioritizing debugging information over the JSON output for that specific run.

perfkitbenchmarker/data/llm_eval/main.py
anksena: Yes we need in both places.
Typical SDLC pattern is to use benchmark to launch the utility on client vm, but subsequently iterate over the utility ON the client vm, so we need these fail fast checks in both places, for consistent execution.

perfkitbenchmarker/data/llm_eval/main.py
anksena: That's a very thoughtful point about the long-term usability when a provider has a variable number of models. I've considered the suggestion of an offset flag, but I'd prefer to stick with the current implementation for a couple of key reasons:

1.  API Instability: My primary concern with an offset is that we can't guarantee a stable, consistently ordered list from the provider APIs. The models returned could be sorted by creation date, alphabetically, or some other metric that might change. If the list order changes, an offset could lead to flaky, non-repeatable benchmarks.

2.  Precision with `specific_models`: The intended workflow for testing models that fall outside the default discovery is to use the --llm_eval_specific_models flag. A user can first run with --discover_models_only to get the complete, current list and then pass the exact models they want to benchmark. This approach is more explicit and guarantees the user is testing exactly what they intend to, regardless of any changes in the provider's model listing.

This is also related to your earlier question about the number of models, the current default of 20 is generally sufficient to cover the most relevant and recent GA (General Availability) models from each provider. The --specific_models flag serves as the robust tool for any deeper, more exhaustive testing scenarios that might arise. For extreme coverage --include_preview can be used to Include preview models from the benchmark.

perfkitbenchmarker/data/llm_eval/main.py
anksena: See rationale for setting this to 20 below.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: >>>>>> I guess if you're testing a bunch of models 1 or 2 failures could be acceptable
Exactly my thoughts.

>>>>>> if the same model fails every time that sounds like a problem.
If this happens then we can consider adding a metric for tracking failures and use results data to decide the course of action.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: My thinking was to prioritize the stability and consistency of the benchmark results by default.

 I'm hesitant to include preview models in the default run because their performance, and sometimes even their names, can be unpredictable. Including them by default could introduce noise into our core metrics, making it harder to track the performance of the stable, production-ready models over time.

I agree that we absolutely should be testing preview models. The --llm_eval_include_preview flag is designed for this purpose. A great way to handle this would be to configure a separate, recurring run in our automation that specifically enables this flag.

 This approach gives us the best of both worlds:
 1.  A clean, stable default benchmark for GA models.
 2.  Separate, consistent tracking of upcoming preview models.

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
anksena: Makes sense. The safety_settings parameter has been removed from the helper functions, which now use the module-level SAFETY_SETTINGS constant directly.

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
anksena: Thanks

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: Done.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: Done.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: That's a fair point. My thinking for generating a Sample is that this data is more valuable than just a debug log. The rationale is twofold:

Historical Tracking: By creating a sample, we can push this data into our standard analysis pipelines. This enables us to track the list of available models over time, creating a historical record of when models are added or removed by the provider.

Control Plane Monitoring: It serves as a simple but effective control plane test. We can monitor the success/failure of this benchmark to track the reliability of the provider's model-listing API.

To make this more concrete, I've just updated the code to change the sample from ('discovered_models', 0) to ('discovered_models_count', <model_count>). This provides a clear, quantitative metric that's much better suited for tracking and analysis.

perfkitbenchmarker/data/llm_eval/providers/base_provider.py
anksena: Thanks for the feedback.
Inspired from this I made several improvements

See changes to perfkitbenchmarker/data/llm_eval/providers/base_provider.py in 
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6103/commits/b4e3eaf0152e395e94ed49ddad0c1e2975ae30f6

Hope this satisfies.

perfkitbenchmarker/data/llm_eval/providers/google_provider.py
anksena: I'm deferring only this specific change to develop a more robust testing strategy. Since the impact is on durability, not just accuracy, it requires serious consideration. I'll follow up with the plan shortly.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
hubatish: Sorry, I agree that it's good to separate preview models out by default. My comment was simply about the wording of the description. I recommend changing "Include preview models from the benchmark" -> "Include preview models in the benchmark"

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
```diff
@@ -0,0 +1,237 @@
...
+    return [
+        sample.Sample(
+            'discovered_models_count', model_count, 'count', metadata
+        )
+    ]
+
+  results = json.loads(stdout)
+  samples = []
+  prompts_tested = results.get('prompts_tested', {})
+  for model, model_results in results['results'].items():
```
hubatish, line 208: Add a unit test for this manual string parsing. It would also be nice to add basic output string check unit tests and/or compilation tests for the data/ code but that is a bit more difficult. Both serve as examples of the interface between these 2 halves of the code.

perfkitbenchmarker/data/llm_eval/logger.py
hubatish: I thought logging.debug is suppressed by default & everything else shows up. 100% warn & error will print.

When we're maintaining automated runs of this we will likely want somewhat verbose logging. Your answer is great for a tool that we would manually run - but this will ultimately be run automatically. Both are flag controlled so probably no actual change needed here; feel free to resolve.

perfkitbenchmarker/data/llm_eval/main.py
hubatish: Are they/could we sort by release date?  Could be interesting to apply such a filter (X recent GAs) in python code rather than our flag gcl. If you plan to mostly run "discover" + "specific" this is moot, but then I would say remove the num_models entirely.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
hubatish: I started asking this question again but refound this comment - whew. 

IMO we should fail this sometimes - as is it will always pass which is not desirable. To be flexible / if we suspect some # of errors, do like a "if more than 20% fail, fail the benchmark" check. Otherwise we will silently pass even if something goes quite badly wrong & we get no results.

Also similar to above add some logging - probably which & how many models failed.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
hubatish: I like the model_count sample number & indeed could see that being interesting to track over time. I haven't seen the output but I do hope it's easy to also tell and/or copy paste the big list of discovered models. Probably it is easy to do that from the output of the data/ python script.

perfkitbenchmarker/data/llm_eval/main.py
```diff
@@ -0,0 +1,205 @@
...
+      raise ValueError('Cannot use --specific_models with --max_models.')
+    models_to_test = FLAGS.specific_models
+    skipped_models = []
+  else:
+    all_models = provider.get_models(include_preview=FLAGS.include_preview)
+    models_to_test = all_models[: FLAGS.max_models]
+    skipped_models = all_models[FLAGS.max_models :]
+
+  logger.debug(f'Models to be benchmarked: {models_to_test}')
+
```
hubatish, line 103: log skipped models too

perfkitbenchmarker/data/llm_eval/main.py
hubatish: I mean that this validation is odd. The default value in the caller is 20, and you check here to see if it is still that magic 20 value or not. But the default value in this file is -1, not 20.. So at minimum change the default in this file to be None rather than -1 & change this check to say if FLAGS.max_models rather than if != 20.

perfkitbenchmarker/data/llm_eval/providers/base_provider.py
hubatish: I have confidence in this interface to apply consistency without the providers (which is the code I've looked at the least). Thanks.

hubatish (approved):
I think most of my comments are pretty simple / easy to resolve now. Happy to chat offline / wait for seeing how this will be used and/or my inclination towards more logging rather than less.

perfkitbenchmarker/data/llm_eval/logger.py
anksena: Confirmed. The goal is to support dual-mode logging: quiet by default, with an option for verbose debug output. Let's onboard the utility into the automated benchmark now and set the appropriate logging level post-integration.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: Done

anksena: Done.

perfkitbenchmarker/linux_benchmarks/llm_eval_benchmark.py
anksena: >copy paste the big list of discovered models.
It is pretty straightforward. We can extract it in plx, if needed, easily

perfkitbenchmarker/data/llm_eval/main.py
anksena: >Could be interesting to apply such a filter (X recent GAs) in python code rather than our flag gcl.

I don't want to over optimize. 20 should be sufficient given the current models supported by providers.

perfkitbenchmarker/data/llm_eval/main.py
anksena: Done.

PR #6394, opened 2026-01-28
Fix call to ConvertManifestToYamlDicts method

vofish:
Fix call to the `ConvertManifestToYamlDicts` method.
The `ConvertManifestToYamlDicts` method is no longer on the cluster instance after switching to kubernetes_commands.


hubatish (approved):

vofish:
Already fixed

PR #6536, opened 2026-03-12
Enable kubernetes_node_scale benchmark (up to 5k nodes) on GCP GKE with CCC

vofish:
### Summary

Enables running the kubernetes_node_scale benchmark (scale up → scale down → scale up again, e.g., toward large node counts) on GCP GKE with Custom ComputeClass and node pool auto-creation (NAP) where configured. The benchmark applies a Deployment with pod anti-affinity (one pod per node), records scale-up / scale-down / second scale-up timing, then tears down.

### Main changes
- **kubernetes_node_scale.yaml.j2** - Deployment only (pause pods, anti-affinity). No ComputeClass in the manifest; the cluster path creates the ComputeClass (aligned with upstream PKB / custom compute class support).
- **GKE** - GetNodeSelectors sets cloud.google.com/compute-class to the default node pool name when `_UsesCustomComputeClass`(default_nodepool) is true, so `ModifyPodSpecPlacementYaml` adds the selector without cloud-specific YAML in the benchmark template.
- **GKE flags** - `gke_autoscaling_profile` `(optimize-utilization/balanced)` and `gke_cluster_ipv4_cidr_size` for large scale-outs; `gke_autoscaling_profile` is also stored in `GetResourceMetadata` for run-to-run comparison.
- **Machine families** - Use existing `--k8s_machine_families` (and `ContainerClusterSpec.machine_families`) instead of a dedicated NAP machine-type flag; `container_spec._ApplyFlags` reads `flag_values.k8s_machine_families` so the flag applies correctly.

### Usage notes
- For the Custom ComputeClass + NAP path on GCP, pass something like `--k8s_machine_families=e2`.
- Raise `container_cluster.max_vm_count` (or equivalent) so the cluster autoscaler can reach your target `--kubernetes_scale_num_nodes`.


perfkitbenchmarker/linux_benchmarks/kubernetes_node_scale_benchmark.py
```diff
@@ -45,6 +45,12 @@
     'kubernetes_scale_num_nodes', 5, 'Number of new nodes to create'
 )
 
+NAP_MACHINE_TYPE = flags.DEFINE_string(
+    'kubernetes_node_scale_nap_machine_type',
```
hubatish: Rather than this, use the flag K8S_MACHINE_FAMILIES created in:
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6559

(or set via config_overrides)

perfkitbenchmarker/data/container/kubernetes_scale/kubernetes_node_scale.yaml.j2
```diff
@@ -1,3 +1,20 @@
+{% if cloud == 'GCP' %}
+apiVersion: cloud.google.com/v1
+kind: ComputeClass
+metadata:
+  name: app-ccc
+spec:
```
hubatish: Most of this should be removed/duplicate with https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6559/changes

If eg storage & spot false are quite important we can make some revisions within that framework.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -394,7 +394,13 @@ def _Create(self):
       cmd.args.append('--enable-autoscaling')
       cmd.flags['max-nodes'] = self.max_nodes
       cmd.flags['min-nodes'] = self.min_nodes
-    cmd.flags['cluster-ipv4-cidr'] = f'/{_CalculateCidrSize(self.max_nodes)}'
+    if gcp_flags.GKE_AUTOSCALING_PROFILE.value:
+      cmd.flags['autoscaling-profile'] = gcp_flags.GKE_AUTOSCALING_PROFILE.value
```
hubatish, line 409: Let's also get this value into ResourceMetadata, since it sounds like something we might change over multiple runs (eg the only difference between 2 runs could be the autoscaling-profile & we'd like to compare them. ResourceMetadata lets us distinguish between the two from just the results).

perfkitbenchmarker/resources/container_service/kubernetes_cluster.py
```diff
@@ -20,6 +22,15 @@
 INGRESS_JSONPATH = '{.status.loadBalancer.ingress[0]}'
 RESOURCE_DELETE_SLEEP_SECONDS = 5
 
+# Timeout for "kubectl delete all --all" during teardown. Increase for
+# large-scale runs (e.g. 5000+ pods) to avoid benchmark failure.
+flags.DEFINE_integer(
+    'kubernetes_teardown_delete_timeout',
```
hubatish: Just looking at this PR, is this flag actually used? Seems reasonable enough.

perfkitbenchmarker/data/container/kubernetes_scale/kubernetes_node_scale.yaml.j2
```diff
@@ -24,6 +41,10 @@ spec:
                 values:
                 - app
             topologyKey: "kubernetes.io/hostname"
+      {% if cloud == 'GCP' %}
+      nodeSelector:
```
hubatish: Is this always the case? Prefer to pass in whether we're using custom compute classes or not.

perfkitbenchmarker/linux_benchmarks/kubernetes_node_scale_benchmark.py
```diff
@@ -71,9 +75,12 @@ def Prepare(bm_spec: benchmark_spec.BenchmarkSpec):
   bm_spec.always_call_cleanup = True
   assert bm_spec.container_cluster
   cluster = bm_spec.container_cluster
+  manifest_kwargs: dict[str, Any] = {'cloud': FLAGS.cloud}
+  if cluster.default_nodepool.machine_families:
+    manifest_kwargs['gcp_compute_class_name'] = cluster.default_nodepool.name
```
hubatish: Can this go in ModifyPodSpecPlacementYaml instead? I belive that code already deals with nodeSelector & it would obscure any reference to the cloud (eg cloud.google.com/compute-class) from the yaml file.

hubatish (approved):
Cool, I think that's all my comments addressed! Can you resolve them as well? Approving.

hubatish (approved):

hubatish (approved):

PR #6730, opened 2026-06-03
agent_sandbox: skeleton resource, spec, and cluster wiring

geojaz:
## What

First step of reshaping the agent sandbox from a `linux_package` into a proper PKB resource, modeled on the kubernetes inference server pattern. Per review feedback on #6718.

This PR is structure only: no install logic and no benchmark, so the new files can be reviewed (and mirrored internally for build files) before the implementation lands.

## Changes

- `BaseAgentSandbox` resource and `GetAgentSandbox` factory, keyed on `SANDBOX_TYPE` so multiple sandbox implementations can coexist.
- `BaseAgentSandboxConfigSpec` and `AgentSandboxConfigDecoder`, embeddable under `container_cluster` in a benchmark config.
- `K8sAgentSandbox` / `K8sAgentSandboxConfigSpec` (in `k8s_agent_sandbox.py` / `k8s_agent_sandbox_spec.py`): the Kubernetes (kubernetes-sigs/agent-sandbox) implementation stubs, registered under `type: Kubernetes`. They raise `NotImplementedError` until the follow-up.
- `KubernetesCluster` constructs and lifecycles `cluster.agent_sandbox` alongside `cluster.inference_server`. Cloud-agnostic; no GKE/EKS changes.

## Notes for reviewers

- The base classes are deliberately generic (`BaseAgentSandbox`, `BaseAgentSandboxConfigSpec`) so non-Kubernetes sandbox implementations can be added later under different `SANDBOX_TYPE`s. The concrete Kubernetes implementation is `K8sAgentSandbox`.
- Unrelated to the existing `SandboxSpec` (GKE Sandbox / sandboxed node pools). "Agent sandbox" here refers to the kubernetes-sigs/agent-sandbox SandboxClaim workload.

## Follow-ups

Subsequent PRs add the install logic (`K8sAgentSandbox._Create` / `_Delete`), the benchmark, and the GKE/EKS/AKS provider changes.

PR #6732, opened 2026-06-04
agent_sandbox: Kubernetes agent sandbox resource implementation

geojaz:
## What

Second step of reshaping the agent sandbox into a PKB resource (follows the skeleton in #6730). This fills in the Kubernetes implementation: a config-driven spec, the install orchestration, the data manifests, and a stub benchmark that provisions the resource.

> **Stacked on #6730** review and merge that first. Until #6730 merges, the diff below also includes the skeleton commit; GitHub narrows it to this PR's own commits once #6730 lands.

## Changes

- `K8sAgentSandboxConfigSpec` (`k8s_agent_sandbox_spec.py`): config-driven, with nested `controller`, `sandbox_template`, and `sandbox_warmpool` sub-specs. The `sandbox_template` block models the upstream `SandboxTemplateSpec` (pod shape rendered; template-level toggles like `network_policy_management` / `env_vars_injection_policy` / `service` are accepted and validated as stubs). The existing `agent_sandbox_*` flags are bridged via `_ApplyFlags`; the old `controller_ref` flag is renamed to `agent_sandbox_manifest_ref`.
- `K8sAgentSandbox` (`k8s_agent_sandbox.py`): `_Create` orchestrates gVisor install, controller install, SandboxTemplate apply, and SandboxWarmPool install (private methods reading `self.spec`). `_Delete` is a no-op (the ephemeral cluster teardown reclaims the stack). The controller-manifest configuration and install helpers are ported from the prior `linux_packages` implementation.
- Data manifests (`data/agent_sandbox/`): gVisor installer assets, and `sandbox-template.yaml.j2` parameterized for runtime class, image, resources, and labels. The SandboxTemplate and SandboxWarmPool share a single fixed name.
- Stub benchmark (`agent_sandbox_benchmark.py`): a `container_cluster.agent_sandbox` config that constructs a `K8sAgentSandbox`. `Run` returns no samples yet.
- Tests (`k8s_agent_sandbox_test.py`): spec decode + flag overrides, controller-manifest injection, `_Create` orchestration, `_Delete` no-op, and benchmark-config construction.

## Scope

Reviewable but not yet runnable end to end. The benchmark's `Run` (load generator + metrics) lands in the next PR, and the GKE/EKS/AKS nodepool changes that make it actually provision come after that. Cloud-agnostic: no provider changes here.

## Follow-ups

`Run` load generation + metrics; then GKE, EKS, AKS provider support.


perfkitbenchmarker/configs/container_spec.py
```diff
@@ -418,6 +419,7 @@ class ContainerClusterSpec(spec.BaseSpec):
   inference_server: (
       kubernetes_inference_server_spec.BaseInferenceServerConfigSpec | None
   )
+  agent_sandbox: agent_sandbox_spec.BaseAgentSandboxConfigSpec | None
```
hubatish: Inference server set the example here, but I'm not sure it's actually the correct location as opposed to having this in root benchmark_spec.py. Namely this approach creates some circular dependency issues, where agent_sandbox wants to reference a cluster (to call methods on it) & the cluster references it to create it. 
I believe wg_serving_inference_server.py & kubernetes_inference_server.py get around this by having a parent / abstract service & a child - and/or by not using pytype at that top level. But yeah putting this in benchmark_spec.py is probably the right place. 
See https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/commit/4daab75b21848ce46e5bf60760e363d98a147f6f for how example_resource.py was added to benchmark_spec.py. The ConstructAgentSandbox call can also take a container_cluster in its __init__ there.

perfkitbenchmarker/resources/kubernetes/k8s_agent_sandbox_spec.py
```diff
@@ -0,0 +1,309 @@
...
+    'agent_sandbox_controller_kube_api_qps', None,
+    'Controller --kube-api-qps value.')
+flags.DEFINE_boolean(
+    'agent_sandbox_controller_enable_tracing', False,
+    'Enable controller OpenTelemetry tracing.')
+flags.DEFINE_string(
+    'agent_sandbox_controller_otel_endpoint', None,
+    'OTLP exporter endpoint when tracing is enabled.')
+flags.DEFINE_boolean(
+    'agent_sandbox_controller_leader_elect', False,
```
hubatish: we should be able to set all these variables just through config_overrides like --config_override=kubernetes_redis_memtier.container_cluster.agent_sandbox.image=yada & the flags are just convenience. That's not bad to have the convenience flags, but it means we probably only need flags for the ones which we're most likely to actually manually change.

perfkitbenchmarker/resources/kubernetes/k8s_agent_sandbox_spec.py
```diff
@@ -0,0 +1,309 @@
...
+    return result
+
+  @classmethod
+  def _ApplyFlags(cls, config_values, flag_values):
+    super()._ApplyFlags(config_values, flag_values)
+    if flag_values['agent_sandbox_runtime_class'].present:
+      config_values['runtime_class'] = flag_values.agent_sandbox_runtime_class
+
+
+class SandboxWarmPoolSpec(spec.BaseSpec):
```
hubatish, line 129: Can you split some of these to their child PRs? IDK, this seems like it is implementing the base sandbox + a bunch of features. Each feature as an individual PR would be nice.

In general it seems like a ton of customization.. we can likely hardcode many of these values.

perfkitbenchmarker/data/agent_sandbox/gvisor-installer/daemonset.yaml
```diff
@@ -0,0 +1,71 @@
...
+# containerd-runsc shim onto the host, then sleeps as a pause container.
+#
+# The actual install logic comes from a ConfigMap named gvisor-installer-script
+# (created by install_gvisor() in resources/kubernetes/k8s_agent_sandbox.py from
+# data/agent_sandbox/gvisor-installer/install.sh before this DaemonSet is
+# applied). The ConfigMap key is "install.sh", mounted at /scripts.
+#
+# Targets nodes labelled sandbox.gke.io/runtime=runsc (the label the
+# benchmark applies to the sandbox node pool).
+apiVersion: apps/v1
```
hubatish: this turned into mostly a "setup spec" PR so I don't think the yamls are used right? Push to the PR where they are used.

perfkitbenchmarker/resources/kubernetes/k8s_agent_sandbox_spec.py
hubatish: Maybe I'm misinterpreting this (I note below they are all referenced by the parent spec)

PR #6747, opened 2026-06-05
google_kubernetes_engine: add management plane operations

ashishsuneja:
**Summary**
Adds GKE-specific implementations of management plane async methods.

**Main changes**
- _IssueAsync, CreateNodePoolAsync, DeleteNodePoolAsync,
  ResolveNodePoolVersions, _GetLatestOperationName, GetNodePoolNames

**How tested**
- Existing GKE test suite passing
- pyink + lint-diffs clean

### DIFF
--- a/perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
+++ b/perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
@@ -16,6 +16,7 @@
 import json
 import logging
 import math
+import time
 import os
 import re
 import typing
@@ -23,6 +24,7 @@ from typing import Any
 
 from absl import flags
 from perfkitbenchmarker import errors
+from perfkitbenchmarker import vm_util
 from perfkitbenchmarker import provider_info
 from perfkitbenchmarker import virtual_machine_spec
 from perfkitbenchmarker.configs import container_spec as container_spec_lib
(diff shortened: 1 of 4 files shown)


perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
...
+        op_start_time: Unix timestamp recorded just before the async gcloud
+            command was issued.  When provided, the status filter is broadened
+            to include DONE (so fast-completing operations are found) and a
+            startTime >= guard is added to avoid matching old operations.
+
+    Returns:
+        Operation name string, or empty string if none found.
+    """
+    link_target = target_name or self.name
+    if op_start_time:
```
hubatish: This is good logic; can we not just do it every time / require op_start_time? Additional conditions when not needed are just additional complexity.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
       cmd.flags['node-pool'] = node_pool
     cmd.Issue()
 
+  def _IssueAsync(self, cmd: util.GcloudCommand) -> str:
+    """Issues a gcloud command with --async, returns the operation name."""
+    cmd.args.append('--async')
```
hubatish: Ah, I was wondering what made these operations async. got it.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
...
+      status_filter = '(status=RUNNING OR status=PENDING)'
+      time_filter = ''
+
+    filter_str = (
+        f'operationType={operation_type} AND '
+        f'{status_filter} AND '
+        f'targetLink ~ {link_target}'
+        f'{time_filter}'
+    )
+    for attempt in range(1, max_attempts + 1):
```
hubatish: what are some examples where you try to call the operation but it doesn't get seen immediately? does it just take a minute for results to show up? 

A unit test with semi-real output examples would be helpful.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
...
+    cmd.flags['format'] = 'value(name)'
+    stdout, stderr, retcode = cmd.Issue(timeout=600, raise_on_failure=False)
+    if retcode:
+      raise errors.Resource.CreationError(stderr)
+    op_name = stdout.strip().splitlines()[-1].strip() if stdout else ''
+    if not op_name:
+      raise errors.Resource.CreationError(
+          f'GKE async command returned no operation name; stderr={stderr}'
+      )
+    return op_name
```
hubatish: below you have _GetLatestOperationName command.. does this command here not output the operation name? Shouldn't it just be:
- start operation, name is returned
- wait for operation to finish. now done.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
...
+    )
+    try:
+      return self._IssueAsync(cmd)
+    except errors.Resource.CreationError as e:
+      if 'returned no operation name' not in str(e):
+        raise
+      # Fallback: gcloud succeeded but printed nothing. Query the operations
+      # list scoped to this specific nodepool to find the operation name.
+      logging.info(
+          'UpgradeNodePoolAsync: ops list fallback for %s: %s',
```
hubatish: why/when/how frequently does this happen?

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,280 @@ def ResizeNodePool(
...
+    cmd.flags.pop('quiet', None)
+    # Record start time BEFORE issuing.  The label-update operation completes
+    # in seconds, so it may already be DONE by the time the fallback queries
+    # the operations list.  The timestamp lets us safely include DONE ops
+    # without matching older completed operations from previous runs.
+    op_start_time = time.time()
+    try:
+      return self._IssueAsync(cmd)
+    except errors.Resource.CreationError as e:
+      if 'returned no operation name' not in str(e):
```
hubatish: this logic is shared with UpgradeNodePoolAsync above. Refactor to only use in one location. Perhaps an optional parameter in _IssueAsync which can handle this "no operation name" case. If ofc this is necessary at all.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Done — collapsed it. _GetLatestOperationName now always uses the broadened filter (RUNNING/PENDING/DONE with a startTime>= guard) and always takes op_start_time. The guard already prevents matching stale completed ops, so the two-branch conditional was redundant.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: For create and delete, that's exactly the flow — _IssueAsync issues --async, gcloud prints the operation name, we wait on it. The exception is clusters upgrade --node-pool and clusters update: those reliably return success with empty stdout (gcloud just doesn't print the name for those two subcommands). Verified on a 100-pool GKE run — 99/99 upgrades and the cluster-update all came back with no operation name, so _GetLatestOperationName recovers it from the operations list. I didn't find a gcloud flag that makes upgrade/update print it; if one exists I'd happily drop the fallback.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Acknowledged

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Added unit tests (GoogleKubernetesEngineAsyncOpsTestCase) with captured-style gcloud output: create returns the op name directly (no fallback), upgrade and update fall back and recover from the operations list, and the no-fallback-configured case raises. On the timing: the operation can take a moment to leave PENDING, and fast ops (the label-update) may already be DONE by the time we query — both handled by the broadened filter. In practice on the 100-pool run it resolved on the first query every time; the 5×3s retry is a safety margin for the PENDING-transition window.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Every time, not intermittently — clusters upgrade --node-pool --async returned no operation name on 99/99 upgrades in the 100-pool run. So the fallback runs on every upgrade/update.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Done, exactly as suggested — moved the fallback into _IssueAsync behind optional fallback_op_type / fallback_target params, so the "no op name → query operations list" path lives in one place. UpgradeNodePoolAsync and UpdateClusterAsync are now one-liners

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,283 @@ def ResizeNodePool(
...
+    from_time = time.strftime(
+        '%Y-%m-%dT%H:%M:%SZ', time.gmtime((op_start_time or time.time()) - 30)
+    )
+    filter_str = (
+        f'operationType={operation_type} AND '
+        '(status=RUNNING OR status=PENDING OR status=DONE) AND '
+        f'targetLink ~ {link_target} AND '
+        f'startTime>="{from_time}"'
+    )
+    for attempt in range(1, max_attempts + 1):
```
hubatish: have you looked at vm_util.Retry? Used with @vm_util.Retry(poll_interval,max_retries,timeout). I suspect you're rewriting code rather than utilizing PKB standards/logic.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,283 @@ def ResizeNodePool(
...
+    if fallback_op_type is None:
+      raise errors.Resource.CreationError(
+          f'GKE async command returned no operation name; stderr={stderr}'
+      )
+    logging.info(
+        '_IssueAsync: no op name printed; ops-list fallback type=%s target=%s',
+        fallback_op_type,
+        fallback_target,
+    )
+    return self._GetLatestOperationName(
```
hubatish: so this is only for operations that don't return an op name? How common is that? Is it in the output but just harder to parse, or completely absent?

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,283 @@ def ResizeNodePool(
...
+      fallback_op_type: GKE operationType (e.g. 'UPGRADE_NODES',
+        'UPDATE_CLUSTER') to look up if stdout is empty. None disables the
+        fallback (create/delete, which always print the name).
+      fallback_target: targetLink substring for the fallback lookup (node-pool
+        or cluster name).
+
+    Returns:
+      The operation name.
+    """
+    cmd.args.append('--async')
```
hubatish: This method overall looks useful enough & broad enough we should consider putting it in providers/gcp/utils.py's GcloudCommand. The --async feature is supported across many gcloud commands & should be broadly applicable right? Even if the fallback "get lastest" behavior would vary - it could take like an optional "get_latest" lambda as an argument. (and given that said fallback is necessary).

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,283 @@ def ResizeNodePool(
...
+      cmd.flags['node-version'] = node_version
+    # --async is incompatible with the long --timeout flag in some gcloud
+    # builds; remove it so the CLI just hands back the op name immediately.
+    cmd.flags.pop('timeout', None)
+    return self._IssueAsync(cmd)
+
+  def UpgradeNodePoolAsync(self, name: str, target_version: str) -> str:
+    """Initiates node pool upgrade; returns op handle. Does NOT wait.
+
+    `clusters upgrade --node-pool --async` returns success with empty stdout
```
hubatish, line 773: got it ok I see where we need this now. Are we running multiple of these in parallel? That's where "get latest" would break.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
```diff
@@ -630,6 +632,283 @@ def ResizeNodePool(
...
+    `clusters upgrade`, `clusters update --async` returns success with empty
+    stdout, so the op name is recovered via _IssueAsync's fallback. The
+    label-update completes in seconds, so the fallback may find it already
+    DONE — handled by _GetLatestOperationName's startTime-guarded filter.
+    """
+    cmd = self._GcloudCommand('container', 'clusters', 'update', self.name)
+    cmd.flags['update-labels'] = f'k8s-mgmt-ts={int(time.time())}'
+    # GcloudCommand sets --quiet by default; the label update is
+    # non-interactive so it's safe to drop (matches how this was validated).
+    cmd.flags.pop('quiet', None)
```
hubatish, line 816: yeah this pop logic is very GcloudCommand specific. Seconding the "add async as option in utils.py" suggestion as a refactor. Might move some unit tests over as well.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Done — refactored to use vm_util.Retry instead of the manual for-loop/sleep. Since max_attempts/retry_delay are call-time parameters here (not fixed), I apply the decorator to a locally-defined inner function inside the method rather than statically on it — poll_interval/max_retries still fully respect the caller's values via closure. Scoped retryable_exceptions to errors.Resource.GetError so only the "not found yet" case retries, not unrelated failures.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Only UPGRADE_NODES and UPDATE_CLUSTER hit this path — create/delete always print the operation name directly (the fallback is disabled for those via fallback_op_type=None). For upgrade/update it's not intermittent — verified 99/99 upgrades plus the cluster update returned empty stdout in the 100-pool run. It's completely absent from gcloud's output for those two subcommands, not a parsing issue

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Yes — concurrent_node_pool_ops runs many of these in parallel (concurrent upgrades across node pools). It's safe: the filter includes targetLink ~ {link_target}, and each concurrent call passes its own specific fallback_target (the individual node-pool name) — so concurrent calls each resolve only operations matching their own target, not just "whatever's most recent." Combined with the op_start_time guard, this disambiguates correctly even when many ops start within the same few seconds. I'll add an explicit concurrent-calls test case to make this guarantee visible in the test suite.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Done — added GcloudCommand.IssueAsync(get_latest_op_fn=...) to util.py. GKE's _GetLatestOperationName is now passed in as the callable, so the GKE-specific filter logic stays in google_kubernetes_engine.py while the async-issue/fallback shape is reusable by any GCP resource. Purely additive — GcloudCommand.Issue() itself is unchanged, so none of the existing 194 call sites across GCP providers are affected. Verified with the full tests/providers/gcp/ suite.

perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
ashishsuneja: Done — added GcloudCommandIssueAsyncTestCase to util_test.py, covering direct op-name return, fallback invocation, raise-when-no-fallback, raise-on-command-failure, and the --async/format flag setup.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -399,6 +399,50 @@ def Issue(self, **kwargs: Any):
...
+    Returns:
+      The operation name.
+
+    Raises:
+      errors.Resource.CreationError: if stdout is empty and no
+        get_latest_op_fn is supplied, or if the command itself fails.
+    """
+    self.args.append('--async')
+    self.flags['format'] = 'value(name)'
+    stdout, stderr, retcode = self.Issue(raise_on_failure=False, **kwargs)
```
hubatish: Apologies if I asked this already, but why raise_on_failure=False ? That's a big problem for general use - maybe it should be an argument passed from google_kubernetes_engine. Especially given the "if retcode" below it just looks like you're changing the exception type returned.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -399,6 +399,50 @@ def Issue(self, **kwargs: Any):
...
+    if retcode:
+      raise errors.Resource.CreationError(stderr)
```
hubatish: Said exception type also doesn't really make sense in a more general location - this IssueAsync command doesn't have to be called in a _Create context.

perfkitbenchmarker/providers/gcp/util.py
```diff
@@ -399,6 +399,50 @@ def Issue(self, **kwargs: Any):
...
+    op_name = stdout.strip().splitlines()[-1].strip() if stdout else ''
+    if op_name:
+      return op_name
+    if get_latest_op_fn is None:
+      raise errors.Resource.CreationError(
```
hubatish: raise probably errors.VmUtil.IssueCommandError instead.

hubatish (approved):
looks pretty reasonable! Left a few more comments on the move to util.py but generally lgtm. Will start trying to approve/submit after last round of comments & the base is submitted.

perfkitbenchmarker/providers/gcp/util.py
ashishsuneja: Done - removed the hardcoded override. IssueAsync now just calls self.Issue(**kwargs), so it inherits Issue()'s own default (raise_on_failure=True) unless a caller explicitly opts out via **kwargs. The manual if retcode: raise is gone too — it was just reimplementing what the default already does.

perfkitbenchmarker/providers/gcp/util.py
ashishsuneja: Agreed — removed errors.Resource.CreationError from this path entirely. Since raise_on_failure now defaults to True, a failing command raises via Issue()/IssueCommand's own exception (errors.VmUtil.IssueCommandError) before we'd ever reach a retcode check here.

ashishsuneja: Agreed — removed errors.Resource.CreationError from this path entirely. Since raise_on_failure now defaults to True, a failing command raises via Issue()/IssueCommand's own exception (errors.VmUtil.IssueCommandError) before we'd ever reach a retcode check here.

perfkitbenchmarker/providers/gcp/util.py
ashishsuneja: Done — switched the "no op name and no fallback" raise to errors.VmUtil.IssueCommandError, matching what the command-failure path now raises by default too.

hubatish (approved):

hubatish (approved):

PR #6748, opened 2026-06-05
elastic_kubernetes_service: add management plane operations

ashishsuneja:
**Summary**
Adds EKS-specific implementations of management plane async methods.

**Main changes**
- _DiscoverSubnets, _DiscoverSubnetsPerAZ, _DiscoverNodeRoleArn,
  _ResolveReleaseVersion, CreateNodePoolAsync, DeleteNodePoolAsync,
  ResolveNodePoolVersions
- Capacity reservation support flag-gated behind
  --eks_reserve_capacity_per_az (default: false)
- logging.warning -> logging.info/raise per hubatish feedback

**How tested**
- 43/44 tests passing (1 pre-existing failure unrelated to this PR)
- EKS end-to-end: 99 pools, 100% success all 7 scenarios
- pyink + lint-diffs clean

### DIFF
--- a/perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
@@ -22,8 +22,8 @@ instructions.
 
 from collections import abc
 import copy
+import functools
 import json
-import threading
 import time
 import logging
 import math
@@ -552,12 +552,9 @@ class EksCluster(BaseEksCluster):
     ]
     vm_util.IssueCommand(cmd)
 
+  @functools.lru_cache(maxsize=None)
   def _DiscoverSubnets(self) -> list[str]:
     """Returns the EKS cluster's subnet IDs (cached after first call)."""
-    # pylint: disable-next=access-member-before-definition
-    if getattr(self, '_cached_subnets', None):
-      # pylint: disable-next=access-member-before-definition
-      return self._cached_subnets
     out, _, _ = vm_util.IssueCommand(
         util.AWS_PREFIX
         + [
@@ -570,25 +567,19 @@ class EksCluster(BaseEksCluster):
         ]
     )
     info = json.loads(out)
-    self._cached_subnets = info['cluster']['resourcesVpcConfig']['subnetIds']
-    return self._cached_subnets
+    return info['cluster']['resourcesVpcConfig']['subnetIds']
 
+  @functools.lru_cache(maxsize=None)
   def _DiscoverSubnetsPerAZ(self) -> dict[str, str]:
     """Returns a mapping of {AZ: subnet_id} for the cluster's subnets.
 
     Used by CreateNodePoolAsync to distribute nodegroups round-robin across
     AZs, avoiding per-AZ EC2 capacity limits when creating many pools.
     Only returns AZs that are in control_plane_zones (if specified).
-    Cached after first call.
+    Cached after first call (via functools.lru_cache).
     """
-    # pylint: disable-next=access-member-before-definition
-    if getattr(self, '_cached_subnets_per_az', None) is not None:
-      # pylint: disable-next=access-member-before-definition
-      return self._cached_subnets_per_az
-
     subnet_ids = self._DiscoverSubnets()
     if not subnet_ids:
-      self._cached_subnets_per_az = {}
       return {}
 
     # Describe subnets to get their AZ mapping
@@ -616,7 +607,6 @@ class EksCluster(BaseEksCluster):
           '[EKS] Could not describe subnets for AZ mapping —'
           ' falling back to all subnets'
       )
-      self._cached_subnets_per_az = {}
       return {}
 
     subnets = json.loads(out)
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -285,7 +287,7 @@ def GetNodePoolFromNodeName(
         nodepool = self.default_nodepool
       else:
         if nodepool_name not in self.nodepools:
-          logging.warning(
+          logging.info(
```
hubatish, line 290: revert changes to these pre-existing GetNodePool name functions; they're unrelated to your PR

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -549,6 +552,733 @@ def ResizeNodePool(
     ]
     vm_util.IssueCommand(cmd)
 
+  def _DiscoverSubnets(self) -> list[str]:
+    """Returns the EKS cluster's subnet IDs (cached after first call)."""
+    # pylint: disable-next=access-member-before-definition
```
hubatish: add this as a parameter in __init__ put set it to None. Should not need to pylint out of this

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -549,6 +552,733 @@ def ResizeNodePool(
...
+    subnets = json.loads(out)
+
+    # Do NOT filter by control_plane_zones — PKB truncates it to 2 AZs.
+    # Accept all subnets the VPC has across all AZs.
+    # Build AZ map — always prefer public subnets (MapPublicIpOnLaunch=True)
+    # which have an internet gateway route. Private subnets lack IGW routes
+    # and nodes launched there cannot reach the EKS API server to join.
+    az_map: dict[str, str] = {}
+    az_map_private: dict[str, str] = {}
+    for s in subnets:
```
hubatish, line 621: definitely need a unit test for this for loops & mapping logic

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -549,6 +552,733 @@ def ResizeNodePool(
...
+        ]
+    )
+    info = json.loads(out)
+    self._cached_subnets = info['cluster']['resourcesVpcConfig']['subnetIds']
+    return self._cached_subnets
+
+  def _DiscoverSubnetsPerAZ(self) -> dict[str, str]:
+    """Returns a mapping of {AZ: subnet_id} for the cluster's subnets.
+
+    Used by CreateNodePoolAsync to distribute nodegroups round-robin across
```
hubatish, line 576: This seems annoyingly complicated/overly specific. Why do you need to do any of this? Does EKS Auto or Karpenter do this for you?

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
```diff
@@ -549,6 +552,733 @@ def ResizeNodePool(
...
+            f'Failed to resolve EKS release version for minor {minor!r}: {err}'
+        )
+      cache[minor] = out.strip()
+      self._cached_release_versions = cache
+      return cache[minor]
+
+  def _DiscoverNodeRoleArn(self) -> str:
+    """Returns a node IAM role ARN by inspecting an existing nodegroup."""
+    # pylint: disable-next=access-member-before-definition
+    if getattr(self, '_cached_node_role_arn', None):
```
hubatish: I've put "add as variable, initialize to None in __init__" as a comment above. But perhaps since these are all functions, just use functools.lru_cache decorator instead. Then you don't even need an if statement; the second call will just return the results of the first.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
ashishsuneja: Checked the current diff against master — no GetNodePool-named functions are touched in this PR. Looks like this was already cleaned up in an earlier amend.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
ashishsuneja: Went with functools.lru_cache instead (per your alternative suggestion below) — cleaner than either the pylint-disabled getattr pattern or adding __init__ placeholders, since it removes the manual caching entirely. Applied the same fix to _DiscoverSubnets, _DiscoverSubnetsPerAZ, _DiscoverNodeRoleArn, and _ResolveReleaseVersion (which also dropped its threading.Lock — lru_cache is thread-safe by construction)

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
ashishsuneja: Added testDiscoverSubnetsPerAZPrefersPublicSubnets, covering the public-preferred-over-private-in-same-AZ logic and the private-fallback-when-no-public-exists case.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
ashishsuneja: Correct — EKS Auto Mode (EksAutoCluster) and Karpenter (EksKarpenterCluster) are already supported in PKB as separate cluster classes, and neither needs this AZ round-robin logic since they handle node placement automatically. This benchmark specifically uses the standard EksCluster (managed node groups) instead.

perfkitbenchmarker/providers/aws/elastic_kubernetes_service.py
ashishsuneja: Done — switched all four cached discovery methods (_DiscoverSubnets, _DiscoverSubnetsPerAZ, _DiscoverNodeRoleArn, _ResolveReleaseVersion) to @functools.lru_cache(maxsize=None), dropping the manual getattr/pylint-disable pattern and the threading.Lock entirely. This also resolves Comment 2 above — no __init__ cache attributes needed at all now.

PR #6749, opened 2026-06-05
azure_kubernetes_service: add management plane operations

ashishsuneja:
**Summary**
Adds AKS-specific implementations of management plane async methods.

**Main changes**
- GetNodePoolNames, CreateNodePoolAsync, DeleteNodePoolAsync,
  ResolveNodePoolVersions
- Retry logic via vm_util.Retry (max 5, 30s interval)

**How tested**
- Existing AKS test suite passing
- pyink + lint-diffs clean

### DIFF
--- a/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
@@ -614,70 +614,20 @@ class AksCluster(kubernetes_cluster.KubernetesCluster):
       nodepool_config: container.BaseNodePoolConfig,
       node_version: str | None = None,
   ) -> None:
-    """Creates a single named node pool on the cluster."""
-    node_flags = self._GetNodeFlags(nodepool_config)
-    if node_version:
-      # Remove any --kubernetes-version _GetNodeFlags may have added from
-      # self.cluster_version, then always append the caller-supplied version.
-      while '--kubernetes-version' in node_flags:
-        idx = node_flags.index('--kubernetes-version')
-        node_flags.pop(idx)
-        node_flags.pop(idx)
-      node_flags += ['--kubernetes-version', node_version]
-    cmd = self._BuildNodePoolAddCmd(nodepool_config, node_flags)
-    _, stderr, retcode = vm_util.IssueCommand(
-        cmd, timeout=1800, raise_on_failure=False
-    )
-    if retcode:
-      raise errors.Resource.CreationError(stderr)
+    """Creates a node pool and blocks until ready."""
+    self.WaitForOperation(self.CreateNodePoolAsync(nodepool_config, node_version))
 
   def DeleteNodePool(self, name: str) -> None:
-    """Deletes the named node pool."""
-    cmd = [
-        azure.AZURE_PATH,
-        'aks',
-        'nodepool',
-        'delete',
-        '--cluster-name',
-        self.name,
-        '--name',
-        _AzureNodePoolName(name),
-    ] + self.resource_group.args
-    self._RunCreateClusterCmd(cmd)
+    """Deletes the named node pool and blocks until removed."""
+    self.WaitForOperation(self.DeleteNodePoolAsync(name))
 
   def UpgradeNodePool(self, name: str, target_version: str) -> None:
-    """Upgrades the named node pool to target_version."""
-    cmd = [
-        azure.AZURE_PATH,
-        'aks',
-        'nodepool',
-        'upgrade',
-        '--cluster-name',
-        self.name,
-        '--name',
-        _AzureNodePoolName(name),
-        '--kubernetes-version',
-        target_version,
-    ] + self.resource_group.args
-    vm_util.IssueCommand(cmd, timeout=1800)
+    """Upgrades the named node pool and blocks until done."""
+    self.WaitForOperation(self.UpgradeNodePoolAsync(name, target_version))
 
   def UpdateCluster(self) -> None:
-    """Real cluster-level update via a unique-timestamp tag change.
-
-    Triggers a control-plane operation (cluster-scoped, not pool-scoped) by
-    updating the cluster tags. Always succeeds because the tag value changes
-    every call.
-    """
-    cmd = [
-        azure.AZURE_PATH,
-        'aks',
-        'update',
-        '--name',
-        self.name,
-        '--tags',
-        f'k8s-mgmt-ts={int(time.time())}',
-    ] + self.resource_group.args
-    vm_util.IssueCommand(cmd, timeout=1800)
+    """Performs a cluster-level update and blocks until done."""
+    self.WaitForOperation(self.UpdateClusterAsync())
 
   # ---- Async variants (return opaque handles) -------------------------------
 
--- a/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
+++ b/perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
@@ -16,6 +16,7 @@
 
 import base64
 import json
+import time
 import logging
 from typing import Any, List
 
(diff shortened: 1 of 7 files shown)


perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+    initial = f'{major}.{minor - 1}'
+    logging.info(
+        '[AKS] ResolveNodePoolVersions: cluster=%s initial=%s target=%s',
+        cluster_ver,
+        initial,
+        target,
+    )
+    return initial, target
+
+  def WaitForOperation(self, op_handle: str) -> None:
```
hubatish: so AKS doesn't have proper ids & therefore you have to split/format the op handle in a weird way huh? Understood this is needed as it's a top level method signature used from the benchmark. Add some examples in function comments & to unit tests.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+      status = out.strip()
+      if status == 'Succeeded':
+        return
+      if status == 'Failed':
+        raise errors.Resource.CreationError(f'nodepool {name} ended in Failed')
+      raise errors.Resource.RetryableCreationError(
+          f'nodepool {name} state={status}'
+      )
+
+    @vm_util.Retry(
```
hubatish: lots of duplicate code here. Can you create a shared function these call & call it from the others?
Some ideas:
- function signature would take a list of args & return like some object with either "not found", "succeeded", "failed" or "other error".
- Then the caller can decide whether to eg raise or silently return on not found.
- shared function can have the vm_util.Retry wrapper around it

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+            logging.info(
+                '[AKS] UpdateClusterAsync: scaling system pool %s %d->%d',
+                pool_name,
+                current_count,
+                new_count,
+            )
+            return 'cluster_succeeded'
+      except (ValueError, KeyError, json.JSONDecodeError) as e:
+        logging.info('[AKS] UpdateClusterAsync: pool parse error: %s', e)
+    # Fallback: tag update
```
hubatish: why do we need a fallback? When does a fallback happen instead of above code?

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+          current_count = int(pools[0]['count'])
+          # Toggle: scale to current+1 or current-1 (minimum 1)
+          new_count = (
+              current_count + 1 if current_count <= 1 else current_count - 1
+          )
+          scale_cmd = [
+              azure.AZURE_PATH,
+              'aks',
+              'nodepool',
+              'scale',
```
hubatish: how is a nodepool scale operation an "update" ?

IMO just stick with the tag & never attempt a scale. Looks weird & potentially produces non-deterministic results.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
     )
     logging.info('Successfully applied BlobFuse PVC.')
 
+  def CreateNodePool(
```
hubatish: remind me again why we need async vs sync versions of these?
Also looking at https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6751/ which is titled "sync operations wrap async operations" - it doesn't actually seem to do that & only modifies kubernetes_management_benchmark.

After you've convinced me we need sync & async versions, just wrap them in this PR. it should be fewer lines of code & therefore simpler & easier to review.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+      self,
+      nodepool_config: container.BaseNodePoolConfig,
+      node_version: str | None = None,
+  ) -> str:
+    """Initiates AKS nodepool create; returns np_succeeded handle."""
+    node_flags = self._GetNodeFlags(nodepool_config)
+    if node_version:
+      # _GetNodeFlags may have added self.cluster_version; replace or append.
+      if '--kubernetes-version' in node_flags:
+        node_flags[node_flags.index('--kubernetes-version') + 1] = node_version
```
hubatish: IMO if it is in node_flags, remove it first. Then either way second step add it back in.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +574,416 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+      else:
+        node_flags += ['--kubernetes-version', node_version]
+    cmd = [
```
hubatish: this cmd list of args looks the same as the list of args in _CreateNodepool above. It should share code with it. This might be the same as the wrapping comment, but generally too much duplication here.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Added WaitForOperation docstring examples showing the 3 handle formats (np_succeeded:<name>, np_gone:<name>, cluster_succeeded) and 7 unit tests covering happy paths and error paths for all 3 kinds (testWaitForOperationNpSucceeded, testWaitForOperationNpSucceededFailed, testWaitForOperationNpSucceededNotFound, testWaitForOperationNpGone, testWaitForOperationClusterSucceeded, testWaitForOperationClusterFailed, testWaitForOperationInvalidHandle

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Done — extracted _WaitForProvisioningState(cmd, resource_desc, not_found_is_success, retryable_exception_type) which wraps vm_util.Retry and handles Succeeded/Failed/NotFound/in-progress uniformly. WaitForOperation now just selects the right cmd and args for each handle kind and delegates to it — the three inner functions (WaitNpSucceeded, WaitNpGone, WaitClusterSucceeded) are gone.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Removed along with the scale branch (Comment 4) — there's no fallback anymore since UpdateClusterAsync now only does the tag update.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Done — removed the scale branch entirely. UpdateClusterAsync now only does the tag update, which is always cluster-scoped and always a real change (unix timestamp value).

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Folded #6751 into this PR as suggested. kubernetes_cluster.py sync wrappers (CreateNodePool, DeleteNodePool, UpgradeNodePool, UpdateCluster) now call *Async + WaitForOperation instead of raising NotImplementedError — so providers only need to implement the async variants. PR #6751 is closed.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Done — both CreateNodePool and CreateNodePoolAsync now use remove-then-append: strip any existing --kubernetes-version that _GetNodeFlags may have added from self.cluster_version, then always append the caller-supplied version.

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Done — extracted _BuildNodePoolAddCmd(nodepool_config, node_flags, no_wait=False) shared by both CreateNodePool and CreateNodePoolAsync. The only difference between them is no_wait=True for the async variant.

hubatish (approved):

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
hubatish: Is this updated? I still see distinct CreateNodePool vs AsyncCreateNodePool functions which duplciate code

hubatish (commented):
woops shouldn't have approved (thought I was looking at the base PR)

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Done — CreateNodePool, DeleteNodePool, UpgradeNodePool, and UpdateCluster now each delegate to their *Async counterpart + WaitForOperation in a single line. The redundant implementations are removed

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +565,338 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+      cmd.append('--no-wait')
+    return cmd + node_flags
+
+  def CreateNodePool(
+      self,
+      nodepool_config: container.BaseNodePoolConfig,
+      node_version: str | None = None,
+  ) -> None:
+    """Creates a node pool and blocks until ready."""
+    self.WaitForOperation(self.CreateNodePoolAsync(nodepool_config, node_version))
```
hubatish, line 609: these look nice & clean now

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
```diff
@@ -573,6 +565,338 @@ def ApplyFusePVC(self, pvc_name: str) -> None:
...
+    initial = f'{major}.{minor - 1}'
+    logging.info(
+        '[AKS] ResolveNodePoolVersions: cluster=%s initial=%s target=%s',
+        cluster_ver,
+        initial,
+        target,
+    )
+    return initial, target
+
+  def _WaitForProvisioningState(
```
hubatish, line 767: cool refactor worked well

hubatish (approved):

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Acknowledged

perfkitbenchmarker/providers/azure/azure_kubernetes_service.py
ashishsuneja: Acknowledged

PR #6758, opened 2026-06-09
Add swap_encryption benchmark (AWS EKS)

ashishsuneja:
Benchmarks Linux swap behavior on encrypted vs unencrypted backends on AWS EKS. Covers instance-store NVMe (i4i, always Nitro-encrypted) and io2 EBS (m6id, encrypted and unencrypted). Measures swap throughput, CPU overhead, and application-level impact (redis, kernel build, OpenSearch) under memory pressure.

Includes unit tests for the fio/memtier JSON parsers, cloud detection, and cost sampling. GKE support to follow in a separate change.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/swap_encryption_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/swap_encryption_benchmark.py
@@ -1,0 +1,2320 @@
+# Copyright 2026 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GKE vs. AWS EKS Swap Encryption and LSSD Performance Benchmark.
+
+Methodology: go/swap-encryption-and-lssd-performance-comparison:gke-vs-aws
+
+== Architecture ==
+
+Provisions a real GKE (GCP) or EKS (AWS) Kubernetes cluster via PKB's
+container_cluster abstraction, then deploys a privileged DaemonSet whose
+pod has full host-device access (/dev, /sys, hostPID).  All benchmark
+phases execute inside this pod via kubectl exec, so measurements reflect
+actual cluster-node behaviour including Kubernetes overhead (kubelet,
+containerd cgroup hierarchy, etc.).
+
+  GKE nodes  ── dm-crypt with ephemeral key (go/node:swap-encryption)
+                 swap device: /dev/mapper/swap_encrypted (over hyperdisk or
+                 LSSD RAID-0 /dev/md0)
+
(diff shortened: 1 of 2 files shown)


ajaysundark:
cc @ajaysundark

ajaysundark:
I'll leave more comments in detail, but my high level feedback is that GKE setup should use the managed swap configuration (https://docs.cloud.google.com/kubernetes-engine/docs/how-to/node-memory-swap) for provisioning swap instead of hacking per pod swap limits.. PKB already seem to support gke_node_system_config so it could be simpler to leverage this for GKE setup
 https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/blob/c0f8a7b40235490f7fa689c30c32ed56f0b7f189/perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py#L573

perfkitbenchmarker/linux_benchmarks/swap_encryption_benchmark.py
```diff
@@ -0,0 +1,2320 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GKE vs. AWS EKS Swap Encryption and LSSD Performance Benchmark.
+
+Methodology: go/swap-encryption-and-lssd-performance-comparison:gke-vs-aws
```
ajaysundark, line 17: I'm good to keep the swap+encryption micro-benchmarks separately in this file, but for the real workload comparisons it is better to keep them in the existing, respective workload benchmarks to keep the application level comparisons valuable

perfkitbenchmarker/linux_benchmarks/swap_encryption_benchmark.py
```diff
@@ -0,0 +1,2320 @@
...
+    )
+    time.sleep(10)
+  if not ignore_failure:
+    raise errors.VmUtil.IssueCommandError(
+        f"[swap_encryption] kubectl exec failed on {pod} (rc={rc}): {err}"
+    )
+  return out, err
+
+
+def _EnableSwapPressure(pod: str, high_pct: int = 75) -> None:
```
ajaysundark, line 933: I'm not clear on the need for these cgroup overrides. could you clarify the intent?

Kubernetes sets the swap cgroup limits based on the user memory request. overriding them for the benchmark run defeats the purpose as it may test the environment that's not representative of a real workload run.

perfkitbenchmarker/linux_benchmarks/swap_encryption_benchmark.py
```diff
@@ -0,0 +1,2320 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GKE vs. AWS EKS Swap Encryption and LSSD Performance Benchmark.
+
+Methodology: go/swap-encryption-and-lssd-performance-comparison:gke-vs-aws
+
+== Architecture ==
```
ajaysundark, line 19: For EKS, is it better to set kubelet swapBehavior in the launch-template or user-data and wire instance-store or io2 configurations for the swap device at boot instead of DS?

This could enable you to set swap as a capability on BaseNodePoolConfig so any k8s benchmark like redis or esrally, can request a swap-enabled pool declaratively. 

it'll help us land swap as a feature config to validate "present/absent" comparisons for different workloads.

PR #6772, opened 2026-06-16
add agentic benchmarking on gke

george-kalisse-sada:
# Agentic Workload Benchmarking for GKE (PKB Extension)

## Summary

Adds a complete benchmarking framework for **Agentic Workloads** on Google Kubernetes Engine (GKE) — specifically measuring per-operation performance of untrusted Python code execution and headless Chromium browser tasks running under **gVisor (GKE Agent Sandbox)** isolation.

---

## Motivation

AI agent systems require ephemeral, isolated execution environments (sandboxes) for running untrusted code. Understanding the performance characteristics of these sandboxes under gVisor — including cold-start latency, execution overhead, memory density limits, and scheduling throughput — is critical for production capacity planning.

This framework enables systematic, repeatable measurement of these characteristics across multiple GCP machine families.

---

### Benchmark Definitions (7 Use Cases)

| Benchmark | Use Case | Measures |
|:---|:---|:---|
| `gke_snapshot` | UC-A: Cold Start & Snapshot | Pod snapshot create/restore latency under CRIU |
| `gke_python_density` | UC-B: Python Density | CEL, TTFE, RSS growth at varying concurrency |
| `gke_chromium_density` | UC-C: Chromium Density | Interaction latency, screenshot time at scale |
| `gke_payload` | UC-D: Payload Transfer | Sandbox→orchestrator data transfer saturation |
| `gke_warmpool` | UC-E: Warmpool Scale-Up | Bulk provisioning speed (0→N pods) |
| `gke_qps` | UC-F: QPS Saturation | Scheduling throughput until pool drain |
| `gke_deletion` | UC-G: Deletion & Cleanup | Bulk deletion latency and IP reclamation |

### Shared Utilities

| Module | Purpose |
|:---|:---|
| `gke_benchmark_utils.py` | Agent API interaction, kubectl helpers, warm pool management, port-forward manager, sample construction |
| `gke_deploy_utils.py` | Idempotent workload deployment (CRDs, templates, warm pools, router, ADK agent, PSI reader) |
| `gke_provision_utils.py` | Full GKE infrastructure lifecycle (VPC, NAT, cluster, node pools, AR, IAM) |
| `gke_image_build_utils.py` | Container image builds via Cloud Build (ADK agent, Chrome sandbox, Sandbox Router) |
| `gke_prerequisite_setup.py` | Standalone script for pre-PKB infrastructure (VPC, NAT, AR, SA, images) |

### Dual Provisioning Modes

- **`custom` mode**: Direct `gcloud` calls for full infrastructure control
- **`native` mode**: Uses PKB's built-in `container_cluster` provisioner with prerequisite script for resources PKB cannot manage

### PKB Provider Extensions

Small additions to support GKE preview features:
- `--gke_use_beta` flag (forces `gcloud beta container clusters create`)
- `--gke_additional_flags` list (appended to cluster create)
- `--gke_additional_nodepool_flags` list (appended to node pool create)

---

### ADK Agent (`workloads/adk_agent/`)

A FastAPI service deployed inside GKE that:
- Exposes REST endpoints for each benchmark type (`/benchmark/python/density`, `/benchmark/python/payload`, `/benchmark/python/qps`, `/benchmark/chromium/density`)
- Uses a **Mock LLM** (no real model calls) to drive the ADK Runner through sandbox claim→execute→release cycles
- Connects to sandboxes via `DirectConnection` (in-cluster) or `kubectl port-forward` (dev mode)
- Measures both orchestrator-side and sandbox-side metrics

### Sandbox Scripts (`sandboxed_apps/`)

- `benchmark_density.py` — CPU-bound, syscall-heavy, and import-heavy tasks with RSS tracking
- `benchmark_payload.py` — Payload generation, serialization, and stdout transfer measurement
- `benchmark_qps.py` — Minimal script proving sandbox liveness
- `benchmark_density.js` — Playwright-driven Chromium interaction benchmark

### Vibe Coding Workloads (`workloads/vibe_coding/`)

Startup scripts simulating real-world agentic cold-starts:
- `startup_pip_fastapi.sh` — pip install + FastAPI server boot
- `startup_npm_vite.sh` — npm install + Vite dev server boot

---

### Prerequisites (once per environment)
```bash
python -m perfkitbenchmarker.linux_benchmarks.kubernetes.agentic.gke_prerequisite_setup \
    --project_id=sada-gke-benchmarking2 \
    --region=us-central1 \
    --zone=us-central1-a \
    --machine_type=c4-standard-8
```

### Provision Cluster
```bash
python pkb.py --benchmarks=gke_python_density \
    --run_stage=provision \
    --gke_provision_mode=native \
    --project=sada-gke-benchmarking2 \
    --owner=george-kalisse \
    --benchmark_config_file=k8s_agents/config/native_provision_config.yaml \
    --gce_network_name=george-agentic-vpc \
    --gce_subnet_region=us-central1 \
    --zone=us-central1-a \
    --container_cluster_version=1.35.3-gke.1389000 \
    --gke_use_beta=true \
    --gke_additional_flags="--enable-pod-snapshots,--enable-dataplane-v2,--enable-private-nodes,--enable-ip-alias,--master-ipv4-cidr=172.16.0.0/28,--workload-pool=sada-gke-benchmarking2.svc.id.goog,--subnetwork=george-agentic-subnet,--enable-master-authorized-networks,--master-authorized-networks=$(curl -s ifconfig.me)/32" \
    --gke_additional_nodepool_flags="--max-pods-per-node=250" \
    --gke_enable_shielded_nodes=false \
    --run_uri=test \
    --temp_dir=./testing/pkb/c4-standard-8/ucb```

### Run Benchmark
```bash
python pkb.py --benchmarks=gke_python_density \
    --run_stage=prepare,run,cleanup \
    --gke_provision_mode=native \
    --gke_project_id=sada-gke-benchmarking2 \
    --gke_region=us-central1 \
    --gke_zone=us-central1-a \
    --gke_sandbox_machine_type=c4-standard-8 \
    --gke_namespace=agentic \
    --gke_sandbox_version=v0.4.6 \
    --gke_python_density=4 \
    --gke_python_density_sample_count=20 \
    --gke_python_density_sample_warmup=0 \
    --gke_python_density_patch_warmpool=true \
    --gke_python_density_exec_timeout=600 \
    --gke_machine_type=c4-standard-8 \
    --gke_gvisor=true \
    --gke_api_url=http://localhost:8080 \
    --run_uri=test \
    --temp_dir=./testing/pkb/c4-standard-8/ucb
```


perfkitbenchmarker/data/k8s_agents/config/native_provision_config.yaml
```diff
@@ -0,0 +1,70 @@
+# Native PKB Provision Config for Agentic Benchmarks
+# Used with --gke_provision_mode=native
+#
+# Prerequisites (run once before PKB):
+#   python tools/agentic-benchmark/scripts/prerequisite_setup.py \
```
hubatish: This tool isn't being included & therefore this comment doesn't need to be here.

perfkitbenchmarker/data/k8s_agents/config/native_provision_config.yaml
```diff
@@ -0,0 +1,70 @@
...
+#       --zone=us-central1-a \
+#       --project=<project> \
+#       --owner=<owner> \
+#       --container_cluster_version=1.35.3-gke.1389000 \
+#       --gke_additional_flags="--subnetwork=<user>-agentic-subnet,--workload-pool=<project>.svc.id.goog"
+#
+# For sweeps (cluster pre-exists, PKB skips provision/teardown):
+#   The sweep bridge injects --run_stage=run,cleanup automatically.
+
+gke_python_density:
```
hubatish: Internally we put a lot of this info but externally it is useful.. it's probably a good addition.

perfkitbenchmarker/data/docker/agentic/adk-agent/performance_agent/agent.py
```diff
@@ -0,0 +1,240 @@
+from google.adk.agents import LlmAgent
+from google.adk.code_executors import GkeCodeExecutor
```
hubatish, line 29: Where is this file run? From the same machine running PKB or a different one?

requirements.txt
```diff
@@ -33,3 +33,4 @@ setuptools>=40.3.0,<81
 six>=1.13.0
 timeout-decorator
 scipy
+matplotlib
```
hubatish: I don't see a reference to this elsewhere with a ctrl-f; is it leftover from an earlier version?
In general we prefer not making many changes to requirements.txt.

perfkitbenchmarker/providers/gcp/flags.py
```diff
@@ -580,6 +580,27 @@
     ' the size derived from max_vm_count. Use when the cluster will scale'
     ' beyond the default node pool (e.g. kubernetes_node_scale with 5k nodes).',
 )
+
+GKE_USE_BETA = flags.DEFINE_boolean(
```
hubatish: If we add this flag, IMO just make it "gcloud_use_beta" (or actually an enum use alpha, beta, None "gcloud_beta_version") & being referenced from gcp/util.py directly seems best.

Alternatively we often will say in the provider "if preview feature used, cmd.use_beta_gcloud = True". In general what feature are you using that needs beta?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_provision_utils.py
```diff
@@ -0,0 +1,698 @@
+"""Shared Provision/Teardown utilities for GKE Agent Sandbox benchmarks.
+
+Provides the full GKE infrastructure lifecycle (create and destroy) used
+by all seven UC benchmark scripts.  Each benchmark's Provision() and
+Teardown() functions delegate to the public functions in this module.
+
+Infrastructure created (in order):
```
hubatish: The very premise of this file is incorrect. PKB (and esp eg google_kubernetes_engine.py _Create) should be handling all of the provisioning logic. 

I'm not sure how much of this is a) completely unnecessary because it's handled elsewhere in PKB (like we do setup subnets & networks automatically if you don't specify a network" or b) is indeed necessary but should be located in some other Resource.py class.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
```diff
@@ -0,0 +1,891 @@
...
+
+def _DeploySandboxTemplates(ns):
+    """Deploy SandboxTemplate + WarmPool for Python and Chromium."""
+    python_image = FLAGS.gke_python_image
+    chromium_image = FLAGS.gke_chromium_image or "chromium-placeholder:latest"
+    warmpool_replicas = FLAGS.gke_warmpool_replicas
+    chromium_replicas = FLAGS.gke_chromium_replicas
+
+    manifest = """---
+apiVersion: extensions.agents.x-k8s.io/v1alpha1
```
hubatish: should go in some .yaml.j2 file

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
```diff
@@ -0,0 +1,891 @@
...
+def _RunKubectl(args, check=True, timeout=120):
+    """Run kubectl with optional kubeconfig."""
+    cmd = ["kubectl"]
+    if FLAGS.gke_kubeconfig:
+        cmd += ["--kubeconfig", FLAGS.gke_kubeconfig]
+    cmd += list(args)
+    return _RunCmd(cmd, check=check, timeout=timeout)
+
+
+def _KubectlApply(manifest_str):
```
hubatish: why have you rewritten kubectl apply & _RunKubectl when implementations exist [container_service/kubectl.py](https://source.corp.google.com/piper///depot/google3/third_party/py/perfkitbenchmarker/resources/container_service/kubectl.py?sq=(file:%2F%2Fdepot%2Fgoogle3%2Fthird_party%2Fpy%2Fperfkitbenchmarker%20OR%20file:%2F%2Fdepot%2Fgoogle3%2Fcloud%2Fperformance)&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental&q=kubectl.py) ?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
```diff
@@ -0,0 +1,362 @@
+"""PKB Benchmark: GKE Agent Python Sandbox Density (Use Case B).
```
roycaihw: For easier review and faster iteration, I'd recommend keeping one benchmark in this PR and leave the other benchmarks for followup PRs. My recommendation is to keep the Python density benchmark.

roycaihw: Let's drop "(Use Case B)" from the description. For the published PKB benchmarks, the documentation should clearly state what the benchmarks are about. The ordering of A,B,C... will become stale and confusing to readers.

roycaihw: Can we drop "GKE" from the file name and the description? Based on the path this is a Kubernetes benchmark, and presumably this benchmark can be reused for other cloud provider without significant change, right?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
```diff
@@ -0,0 +1,362 @@
...
+
+_WARMPOOL_NAME = "python-sandbox-warmpool"
+_WARMPOOL_LABEL = "sandbox=python-sandbox-example"
+
+# ---------------------------------------------------------------------------
+# Benchmark-specific flags
+# ---------------------------------------------------------------------------
+
+flags.DEFINE_integer(
+    "gke_python_density",
```
roycaihw: > gke_python_density
nit: Shall we name the flag something like "concurrent_sandbox_count"? `gke` and `python` can already be implied based on the file name and description of the benchmark.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
```diff
@@ -0,0 +1,362 @@
...
+flags.DEFINE_integer(
+    "gke_python_density_sample_count",
+    20,
+    "Number of sample iterations per sandbox session.",
+)
+
+flags.DEFINE_integer(
+    "gke_python_density_sample_warmup",
+    0,
+    "Number of warmup iterations per session (excluded from stats).",
```
roycaihw: It's unclear what "warmup iterations" means as it's not mentioned before. Shall we document the workflow in the benchmark description?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_provision_utils.py
roycaihw: +1. Let's set up the cloud infra using PKB-native way.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_python_density_benchmark.py
```diff
@@ -0,0 +1,362 @@
...
+    """Teardown GKE cluster and all dependencies."""
+    gke_provision_utils.Teardown()
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _emit(samples, agg, agg_key, metric_suffix, unit, namespace, extra):
```
roycaihw: Can you document how the metrics emit works and what the parameters are?

roycaihw:
cc @yuanwang04 @oceanxie1

perfkitbenchmarker/data/k8s_agents/config/native_provision_config.yaml
george-kalisse-sada: Done, removed stale references in new commit.

perfkitbenchmarker/data/docker/agentic/adk-agent/performance_agent/agent.py
george-kalisse-sada: This is the ADK Agent we're benchmarking against (i.e calling its FASTAPI APIs). It Get's Docker-built and deployed to GKE. PKB Benchmarks target it via kubectl port-forward.

requirements.txt
george-kalisse-sada: Stale, Removed.

perfkitbenchmarker/providers/gcp/flags.py
george-kalisse-sada: I removed this Flag now. It was for --enable-pod-snapshots on GKE, which was BETA at the time of development.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_provision_utils.py
george-kalisse-sada: There were 2 approaches for Provisioning when this comment was made; 'Custom', and PKB 'Native'. I removed the 'Custom' option and any unnecessary code related to Custom; PKB-Native is the only way now.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
george-kalisse-sada: Done, moved all inline-templates to Jinja2 templates.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
george-kalisse-sada: Done, moved refactored all to use kubectl.py

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
george-kalisse-sada: The benchmarks have GKE-specific dependencies at the moment, such as Pod Snapshots (podsnapshot.gke.io/v1 CRD), image building using cloud build, ... etc. Abstracting this coupling would require some research and refactoring, and possibly another PR.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
george-kalisse-sada: renamed to 'gke_python_density_concurrent_sandbox_count'.  
Benchmark-/usecase- specific flags should maintain the benchmark name (for example: gke_python_density_) as a prefix in order not to have any potential 'namespace collisions' when multiple benchmarks are imported or executed.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_python_density_benchmark.py
george-kalisse-sada: added a description to the docstring at the top of the file.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_image_build_utils.py
```diff
@@ -0,0 +1,451 @@
...
+        machine_type=machine_type,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _BuildADKAgentImage(
```
hubatish: Probably this whole gke_image_build_utils file is not needed. See GoogleArtifactRegistry in google_kubernetes_engine.py & kubernetes_hpa   
with 
container_specs:
    kubernetes_fib:
      image: fibonacci
in the config + data/docker/fibonacci for the dockerimage

george-kalisse-sada: We're building 3 images.
The ADK Image, whose codebase is in the repo, and two other images (Sandbox Router and Chrom Sandbox) whose codebase is not in the repo; they're in https://github.com/kubernetes-sigs/agent-sandbox, and they are not built/published publically, they need to be built per use.

I can try to move ADK Image to PKB NAtive, but the other 2 will still require to be in a "repreq" python script, and that is because I do not want to import their static code into PKB.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
```diff
@@ -0,0 +1,497 @@
...
+        project = getattr(FLAGS, 'project', '') or ''
+    if not region:
+        zone = getattr(FLAGS, 'zone', '') or ''
+        region = zone[:-2] if zone else ''
+
+    # Derive image paths for template rendering.
+    # Chrome and Router images are built during prerequisites
+    # (gke_prerequisites.py), not during Prepare.
+    # ADK agent image is built by PKB container_specs during Provision.
+    from perfkitbenchmarker.linux_benchmarks.kubernetes.agentic import (
```
hubatish: across the board avoid importing within functions & import at the top instead.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/gke_deploy_utils.py
```diff
@@ -0,0 +1,497 @@
...
+    """Deploy Pod Snapshot infrastructure.
+
+    Idempotent: safe to call repeatedly. Sequence:
+      1. Create GCS bucket (hierarchical namespace)
+      2. Create managed folder
+      3. Create KSA for snapshots
+      4. Bind IAM roles
+      5. Deploy PodSnapshotStorageConfig + PodSnapshotPolicy
+    """
+    if FLAGS.skip_deploy_snapshots:
```
hubatish, line 257: Frequently prefer using https://absl.readthedocs.io/en/latest/absl.flags.html#absl.flags.FlagHolder which shows where the flag is defined.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,481 @@
...
+        utils.MakeSample(
+            f"{BENCHMARK_NAME}_total_drain_time",
+            round(total_drain_time, 2),
+            "seconds",
+            ns,
+            extra,
+        )
+    )
+
+    if n > 0:
```
hubatish, line 324: Consider throwing if eg n == 0 would be consider an error. In general throw somewhat liberally rather than handling errors that resulted in unexpected state.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+# Set by each benchmark's Run() via set_benchmark_spec().
+_current_benchmark_spec = None
+
+
+# ---------------------------------------------------------------------------
+# Shared flags (registered once; importable by benchmark modules)
+# ---------------------------------------------------------------------------
+
+flags.DEFINE_string(
+    "k8s_namespace",
```
hubatish: should be named k8s_agentic_namespace & could possibly just be a hardcoded constant K8S_AGENTIC_NAMESPACE = "agentic" as well.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+    "agentic",
+    "Kubernetes namespace where the agentic workloads are deployed.",
+)
+
+flags.DEFINE_bool(
+    "k8s_gvisor",
```
hubatish: likewise prefix all the flags with k8s_agent since they're only used in these files.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+    and retries for transient connection errors automatically.
+    """
+    return kubectl.RunKubectlCommand(
+        list(args),
+        timeout=timeout,
+        raise_on_failure=raise_on_failure,
+    )
+
+
+def CountPods(namespace, label, phase=None):
```
hubatish: Use pytyping pretty much everywhere: https://google.github.io/pytype/user_guide.html

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Port-forward manager
+# ---------------------------------------------------------------------------
+
+import atexit
+import os as _os
+import signal
+import threading
```
hubatish: same comments about imports at the top. Probably just move this to its own file given its independence

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+def BuildMetadata(namespace, extra=None):
+    """Construct the common metadata dict for all samples."""
+    metadata = {
+        "namespace": namespace,
+        "gvisor": FLAGS.k8s_gvisor,
+    }
+    # Derive machine_type from benchmark_spec (set via set_benchmark_spec)
+    machine_type = None
+    if _current_benchmark_spec:
+        cluster = getattr(_current_benchmark_spec, 'container_cluster', None)
```
hubatish, line 240: is this 2 or 4 spaces for the indent? Should be 2. That might be handled by the linter from CONTRIBUTING.md or might not be.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_benchmark_utils.py
```diff
@@ -0,0 +1,506 @@
...
+        time.sleep(2)
+    logging.warning("Drain timed out, %d pods still present",
+                    CountPods(namespace, label))
+    return False
+
+
+def set_benchmark_spec(benchmark_spec):
+    """Store benchmark_spec for metadata derivation (called by Run())."""
+    global _current_benchmark_spec
+    _current_benchmark_spec = benchmark_spec
```
hubatish, line 221: This is kind of a smell. What it indicates to me is that you should be considering using more classes to manage state. In general IMO there ought to be some "Sandbox" class/resource. A resource is expected to handle creation,deletion, metadata, & flags/spec. I believe you are doing all of those things just with global functions.
See the new README in https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/tree/master/perfkitbenchmarker/resources

george-kalisse-sada: there seems to be work done on that in different PRs.
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6732/changes
https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6740/changes

We can just refactor and consume those resources in a different Effort.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+(see gke_deploy_utils.py for the K8s manifest). It is NOT run from the
+machine executing PKB. The ADK agent pod serves a FastAPI app (main.py)
+that PKB calls via HTTP through a kubectl port-forward tunnel.
+
+Execution flow:
+  PKB (your laptop/CI) -> kubectl port-forward -> adk-agent pod -> this file
+  -> GkeCodeExecutor -> SandboxClient -> gVisor sandbox pod
+"""
+
+"""GKE Performance Agent â ADK agent definition for sandbox benchmarking.
```
roycaihw: There is a weird symbol. Drop it? 
```
â
```

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+    density_benchmark_code = "import os; print(os.uname())"
+
+payload_script_path = os.path.join(
+    basedir, "../sandboxed_apps/python_test_app/benchmark_payload.py"
+)
+try:
+    with open(payload_script_path, "r") as f:
+        payload_benchmark_code = f.read()
+except Exception:
+    payload_benchmark_code = "import os; print(os.uname())"
```
roycaihw: This is a programming error. It should fail more loudly rather than continuing in an invalid state.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+
+# Load the benchmark scripts
+density_script_path = os.path.join(
+    basedir, "../sandboxed_apps/python_test_app/benchmark_density.py"
+)
+try:
+    with open(density_script_path, "r") as f:
+        density_benchmark_code = f.read()
+except Exception:
+    density_benchmark_code = "import os; print(os.uname())"
```
roycaihw: This is a programming error. It should fail more loudly rather than continuing in an invalid state.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+    payload_benchmark_code = "import os; print(os.uname())"
+
+qps_script_path = os.path.join(
+    basedir, "../sandboxed_apps/python_test_app/benchmark_qps.py"
+)
+try:
+    with open(qps_script_path, "r") as f:
+        qps_benchmark_code = f.read()
+except Exception:
+    qps_benchmark_code = "import json; print(json.dumps({'sandbox_status': 'ok'}))"
```
roycaihw: This is a programming error. It should fail more loudly rather than continuing in an invalid state.

roycaihw: Write a reusable function for the scripts loading:
```
density_benchmark_code = _load_script("benchmark_density.py")
payload_benchmark_code = _load_script("benchmark_payload.py")
qps_benchmark_code = _load_script("benchmark_qps.py")
```

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+
+# =========================================================================
+# 3. Agent Initialization
+# =========================================================================
+
+
+# Module-level thread pool for sandbox I/O operations.
+# Initialized once at import time to avoid thread-safety issues
+# with lazy initialization inside _execute_in_sandbox().
+_SANDBOX_POOL = ThreadPoolExecutor(max_workers=16)
```
roycaihw: The number of workers should be configurable benchmark config. Otherwise we are limited with 16 concurrent agent sandbox executions.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class V3GkeCodeExecutor(GkeCodeExecutor):
+    def _execute_in_sandbox(self, code: str) -> CodeExecutionResult:
+        """Executes code using the v0.4.6 compatible SandboxClient."""
+        from k8s_agent_sandbox.sandbox_client import SandboxClient
+        from k8s_agent_sandbox.models import SandboxDirectConnectionConfig
+        import logging
+        import time
```
roycaihw: As we discussed in the meeting, please move imports to the top of the file.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+# =========================================================================
+
+
+# Module-level thread pool for sandbox I/O operations.
+# Initialized once at import time to avoid thread-safety issues
+# with lazy initialization inside _execute_in_sandbox().
+_SANDBOX_POOL = ThreadPoolExecutor(max_workers=16)
+
+
+class V3GkeCodeExecutor(GkeCodeExecutor):
```
roycaihw: Where is `V3` coming from?

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
...
+    def _execute_in_sandbox(self, code: str) -> CodeExecutionResult:
+        """Executes code using the v0.4.6 compatible SandboxClient."""
```
roycaihw: Let's drop the version "v0.4.6" from the code, since it will easily get stale. You already have the version tracked in requirements.txt

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
```diff
@@ -0,0 +1,276 @@
+"""GKE Performance Agent -- ADK agent definition.
```
roycaihw: There is no GKE-specific content in this file. Let's replace "GKE" with "k8s"

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
```diff
@@ -0,0 +1,196 @@
...
+    return sorted_vals[min(idx, len(sorted_vals) - 1)]
+
+
+def run_benchmark():
+    results = {"ttfe_ms": None, "cel_ms": [], "rss_mb_start": None, "rss_mb_end": None}
+
+    # Measure TTFE
+    ttfe_start = time.perf_counter()
+    exec("x = 1 + 1", globals())
+    results["ttfe_ms"] = round((time.perf_counter() - ttfe_start) * 1000, 6)
```
roycaihw: This is wrong. Time-To-First-Execution is the time between the agent triggering execution and the sandbox conducting the actual execution. Here you're measuring a period of time within the sandbox

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
```diff
@@ -0,0 +1,196 @@
...
+
+    results["rss_mb_start"] = get_rss_mb()
+
+    tasks = get_static_tasks()
+    sampled_tasks = [t for t in tasks if t["type"] != "import"]
+    import_task = next((t for t in tasks if t["type"] == "import"), None)
```
roycaihw: It's weird that we are using `next()` while the hardcoded list has exactly one "import" task. Why is that?

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # Warmup — sampled tasks only (import uses C-extension modules that
+    # error on repeated reimport, so it runs once outside the loop)
+    for _ in range(SAMPLE_WARMUP):
+        for task in sampled_tasks:
+            exec(task["code"], globals())
```
roycaihw, line 127: Can you document why we need a warmup phase here? What things are we warming up?

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
```diff
@@ -0,0 +1,196 @@
...
+                "        except Exception:\n"
+                "            pass\n"
+                "# Build a large dict to add memory pressure\n"
+                "_ = {str(i): list(range(100)) for i in range(10_000)}\n"
+            ),
+        },
+    ]
+
+
+def _percentile(sorted_vals, pct):
```
roycaihw: _percentile is not used in this file. Drop it?

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_qps.py
```diff
@@ -0,0 +1,24 @@
...
+t0 = time.perf_counter()
+
+# Trivial computation to prove the sandbox is functional
+result = sum(range(10_000))
+
+elapsed_ms = (time.perf_counter() - t0) * 1000
+
+print(json.dumps({
+    "sandbox_status": "ok",
+    "sandbox_qps_exec_ms": round(elapsed_ms, 3),
```
roycaihw: Let's drop "qps" from the metrics name since this metrics measures how long a simple python sandbox execution takes-- nothing related to qps here.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_qps.py
```diff
@@ -0,0 +1,24 @@
...
+    "sandbox_compute_result": result,
```
roycaihw: Let's drop the "sandbox_compute_result" since it's just confusing to people unless they read the actual source of what the sandbox does.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_qps.py
```diff
@@ -0,0 +1,24 @@
...
+
+t0 = time.perf_counter()
+
+# Trivial computation to prove the sandbox is functional
+result = sum(range(10_000))
+
+elapsed_ms = (time.perf_counter() - t0) * 1000
+
+print(json.dumps({
+    "sandbox_status": "ok",
```
roycaihw: Let's make this a "success" v.s. "fail" based on whether `result` matches the expectation (49,995,000-- the sum of all integers from 0 up to 9,999)

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
george-kalisse-sada: because `import_task = tasks[2]` is fragile.
If next is unapealing, I can replace with `import_task = [t for t in tasks if t["type"] == "import"][0]`, same effect.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
george-kalisse-sada: Will remove, this is within sandbox anyways. the real TTFE is orchestrator-side measured elsewhere in code.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
george-kalisse-sada: will remove.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_density.py
george-kalisse-sada: will add documentation, something like:

    # Warmup iterations stabilize measurements by pre-warming:
    #   - Python bytecode cache (first exec compiles .pyc)
    #   - gVisor sentry page cache (first memory access triggers page faults)
    #   - OS dentry/inode cache (first stat/listdir populates kernel caches)
    #   - CPU branch predictor (first iterations train prediction tables)

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
george-kalisse-sada: Agree. Will put in a fix to crash the agent pod (CrashLoopBackOff) when reading the file fails.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
george-kalisse-sada: Agree. Will put in a fix to crash the agent pod (CrashLoopBackOff) when reading the file fails.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
george-kalisse-sada: Will Fix.Agree. Will put in a fix to crash the agent pod (CrashLoopBackOff) when reading the file fails.

perfkitbenchmarker/data/docker/agentic/adk-agent/gke_performance_agent/agent.py
george-kalisse-sada: just a lingering development iteration name from iterating against older k8s-agent-sandbox API versions. Will rename to `BenchmarkGkeCodeExecutor`, more meaningful.

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_payload.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env python3
+"""Agentic Payload Transfer Benchmark (Use Case D).
```
roycaihw: Let's drop "Use Case D"

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_payload.py
```diff
@@ -0,0 +1,199 @@
...
+    print(msg, file=sys.stderr, flush=True)
+
+
+_log(f"PAYLOAD_SIZE_MB: {PAYLOAD_SIZE_MB}")
+_log(f"PAYLOAD_ITERATIONS: {PAYLOAD_ITERATIONS}")
+
+
+def get_rss_mb() -> float:
+    """Get current RSS memory in MB."""
+    return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024
```
roycaihw, line 52: This function is also defined in benchmark_density.py. Can we have it in a util package and reuse it?

perfkitbenchmarker/data/docker/agentic/adk-agent/sandboxed_apps/python_test_app/benchmark_payload.py
```diff
@@ -0,0 +1,199 @@
...
+    generation_times = []
+    serialization_times = []
+    stdout_times = []  # stdout write syscall time
+    transfer_times = []  # serialize + stdout write (the threshold metric)
+    throughputs = []  # MB/s based on stdout write time
+
+    # --- Warmup (2 iterations, not recorded) ---
+    # Stabilizes os.urandom() entropy pool and base64 codec
+    # initialization before measured iterations begin.
+    for _ in range(2):
```
roycaihw, line 93: Similar to the density benchmark, can you document what the warmup phase is for, and make the number of iterations configurable?

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+
+Endpoints:
+  GET  /healthz                      → liveness probe
+  POST /benchmark/python/density     → run the Python density benchmark (UC-B)
+  POST /benchmark/python/payload     → run the payload transfer benchmark (UC-D)
+  POST /benchmark/python/qps         → run the QPS saturation benchmark (UC-F)
+  POST /benchmark/chromium/density   → run the Chromium density benchmark (UC-C)
+  POST /run                          → raw ADK agent interaction
+
+POST /benchmark/python/density — Request:
```
roycaihw: The documentation seems incomplete as it only has the request and response for `POST /benchmark/python/density`. I'd recommend putting the API endpoint documentation in a separate doc file.

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+        yield
+    finally:
+        try:
+            executor.shutdown(wait=False)
+            logging.info("ThreadPoolExecutor shut down")
+        except Exception:
+            logging.exception("Error shutting down ThreadPoolExecutor")
+
+
+app = FastAPI(title="GKE Benchmark Agent", version="0.2.0", lifespan=lifespan)
```
roycaihw: Let's replace GKE with Kubernetes

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+
+# Serialise benchmark requests so concurrent POSTs cannot clobber the
+# shared env vars (BENCHMARK_MODE, SAMPLE_COUNT, …) that agent.py reads.
+_benchmark_lock = asyncio.Lock()
+
+
+def _percentile_stats(sorted_values: list, prefix: str) -> dict:
```
roycaihw: Can you turn this function into a util for reuse?

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+    )
+    duration_s: float = Field(
+        default=60.0, ge=5.0, description="Duration of the QPS burst in seconds"
+    )
+    sandbox_exec_timeout_s: int = Field(
+        default=30, ge=10, description="Sandbox command execution timeout in seconds"
+    )
+
+
+class ChromiumBenchmarkRequest(BaseModel):
```
roycaihw: Right now the main.py combines all the endpoints and their Request class. For better readability, can you split each endpoint and its Request class into a separate file?

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+        os.environ["SANDBOX_EXEC_TIMEOUT_S"] = str(req.sandbox_exec_timeout_s)
+
+        logger.info(
+            "Starting Python benchmark: sample_count=%d sample_warmup=%d concurrent_sessions=%d",
+            req.sample_count,
+            req.sample_warmup,
+            req.concurrent_sessions,
+        )
+
+        prompt = "Please start the GKE performance benchmark workflow."
```
roycaihw: The prompt is not used, correct? Can we drop it?

perfkitbenchmarker/data/docker/agentic/adk-agent/main.py
```diff
@@ -0,0 +1,1109 @@
...
+async def benchmark_python_density(req: BenchmarkRequest):
+    """Trigger the Python density benchmark (Use Case B).
+
+    Fires `concurrent_sessions` parallel agent sessions.  Each session
+    claims its own sandbox, runs the benchmark script with the given
+    iteration/warmup counts, and returns both orchestrator-side and
+    sandbox-side metrics.
+    """
+    async with _benchmark_lock:
+        os.environ["BENCHMARK_MODE"] = "density"
```
roycaihw: Not urgent, but maybe for a followup PR:

Using os env to pass configs is not very intuitive. Maybe consider using the `prompt` to pass the configs
```
POST /benchmark/python/density
 | 
 v
FastAPI service 
 |
 | (prompt: {mode: "density", sample_count: 100...})
 v
ADK agent 
 |
 | (prompt: {mode: "density", sample_count: 100...})
 v
MockLLM
 |
 | (script instruction based on the prompt)
 v
ADK agent (execution)
```

PR #6795, opened 2026-06-25
Fix GPU for AKS Automatic

vofish:
### Fix for the GPU capacity issue in AKS Automatic

**The Problem**

Applying the DaemonSet to install the GPU device plugin manually fails with the following error:
```
STDERR: Error from server (Forbidden): error when creating "/var/folders/_7/qlc76gs57gjg9dw3cjqy4zp00000gn/T/tmp5egszs8u.yaml":
daemonsets.apps "nvidia-device-plugin-daemonset" is forbidden: ValidatingAdmissionPolicy
'aks-managed-protect-system-namespaces' with binding 
'aks-managed-protect-system-namespaces-binding' denied request: 
Modification of resources in managed system namespaces is not allowed
```
As a result, although the GPU node is successfully added to the cluster, the GPU capacity is not available because we are blocked from deploying the custom device plugin.

**How This Solution was Found**

Through collaboration with Azure Support, I received a workaround to unblock GPU usage on AKS Automatic / Node Auto-Provisioning (NAP) clusters (specifically targeting the Automatic cluster) while Microsoft continues to develop official managed GPU API field support natively with NAP.

**The Solution & What It Does**

Instead of manually deploying custom DaemonSets (which conflicts with the security policies on AKS Automatic), this PR leverages the EnableManagedGPUExperience VMSS tag (preview) directly within the NAP configuration.

When applied, this tag delegates the responsibility of setting up the GPU stack entirely to Azure. When an NVIDIA GPU VM size is selected by NAP (via NodePool and AKSNodeClass configuration), the tag automatically controls and triggers:

- The installation of the official NVIDIA GPU driver.
- The deployment of the GPU device plugin.

This approach is much simpler, aligns with AKS Automatic security standards (preventing system namespace violation errors), and is officially supported and managed by the cloud provider.

### DIFF
--- a/perfkitbenchmarker/data/container/kubernetes_ai_inference/azure-gpu-nodepool.yaml.j2
+++ b/perfkitbenchmarker/data/container/kubernetes_ai_inference/azure-gpu-nodepool.yaml.j2
@@ -4,6 +4,8 @@ metadata:
   name: gpu-nodeclass
 spec:
   imageFamily: Ubuntu2204
+  tags:
+    EnableManagedGPUExperience: 'true'
 ---
 apiVersion: karpenter.sh/v1
 kind: NodePool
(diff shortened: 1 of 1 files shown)


hubatish:
Your PR description describes the problem well - can you add more about why this is the solution, how you found it, & what it does? In general this looks very nice, much simpler & hopefully better supported by AKS.

vofish:
> Your PR description describes the problem well - can you add more about why this is the solution, how you found it, & what it does? In general this looks very nice, much simpler & hopefully better supported by AKS.

The description has been updated.

hubatish (approved):

PR #6803, opened 2026-06-29
providers/gcp: generalize async op handling into GcloudCommand.IssueAsync

ashishsuneja:
Follow-up to PR #6747 per hubatish's review comments (https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6747#discussion_r... and the seconding comment below it).

Moves the `--async` issue + 'no op name printed, fall back to listing operations' pattern out of GKE-specific `_IssueAsync` into a reusable `GcloudCommand.IssueAsync` method on the shared GCP util class, with the resource-specific lookup (`_GetLatestOperationName` for GKE) passed in as an optional `get_latest_op_fn` callable.

**Scope:** Purely additive — `GcloudCommand.Issue()` itself is unchanged, so none of the 194 existing call sites across GCP providers are affected.

**Tested:**
- All 4 `GoogleKubernetesEngineAsyncOpsTestCase` tests pass
- Full `tests/providers/gcp/` suite: 278/281 passed (3 failures are pre-existing test-isolation flakiness in `gce_managed_instance_group_test.py`/`gcp_cluster_test.py`, unrelated to this change — confirmed they fail identically without this patch applied)
- pyink + lint-diffs clean

### DIFF
--- a/perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
+++ b/perfkitbenchmarker/providers/gcp/google_kubernetes_engine.py
@@ -16,6 +16,7 @@
 import json
 import logging
 import math
+import time
 import os
 import re
 import typing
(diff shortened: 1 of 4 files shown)


ashishsuneja:
Closing — folded these commits directly into #6747 per reviewer feedback, since the suggestion originated there. See #6747 for the IssueAsync refactor + tests.

PR #6862, opened 2026-07-10
kubernetes_cluster: revert AddNodepool to no-op to fix provision_node…

ashishsuneja:
…_pools on GCP

PR #6746 changed AddNodepool to call CreateNodePool, breaking the provision_node_pools benchmark on GCP (reported by Zach Howell). GkeCluster has no AddNodepool override and CreateNodePool is not implemented for that use case.

Revert to pass (no-op) — AKS, EKS, and Karpenter variants already override AddNodepool with the correct provider-specific implementation.


hubatish (approved):
sorry about the waffling back / forth here & that we landed on "maintain 2 similar implementations". Thanks for the fix!

PR #6899, opened 2026-07-17
Warmpool & Deletion Benchmarks

george-kalisse-sada:
**Files:** 2 new files
- `perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_warmpool_benchmark.py`
- `perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py`

**Description:** Adds two infrastructure lifecycle benchmarks:

**Warmpool Scale-Up:** Measures how quickly N sandbox pods can be provisioned from zero via the SandboxWarmPool controller. Scrapes pod lifecycle timestamps (creationTimestamp, PodScheduled condition, Ready condition) for time-to-created/scheduled/running percentiles. No agent API needed — interacts directly with the Kubernetes API.

**Deletion & Cleanup:** Measures bulk deletion efficiency and IP address reclamation. Provisions N pods via SandboxWarmPool, bulk-deletes them by scaling to zero, and tracks per-pod deletion latency and IPAM release timing. Tests whether "Terminating" pods hold IP addresses longer than expected.

### DIFF
--- a/perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
+++ b/perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
@@ -1,0 +1,511 @@
+# Copyright 2026 PerfKitBenchmarker Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""PKB Benchmark: GKE Agent Deletion & Cleanup .
+
+Atomic single-point measurement of bulk deletion efficiency and IP
+reclamation on a pre-provisioned GKE cluster with gVisor isolation.
+Provisions N sandbox pods via SandboxWarmPool, then bulk-deletes them
+and measures per-pod deletion latency, aggregate deletion stats, and
+IP address reclamation timing.
+
+This benchmark is designed to be invoked repeatedly by an external sweep
+controller that varies the batch_size parameter across iterations to find
+the deletion saturation point.
+
+Usage:
+  python pkb.py --benchmarks=gke_deletion \\
+                --k8s_deletion_batch_size=100 \\
+                --k8s_deletion_warmpool_name=python-sandbox-warmpool \\
+                --k8s_deletion_pod_label=sandbox=python-sandbox-example \\
+                --k8s_deletion_poll_interval_s=1.0 \\
+                --k8s_deletion_provision_timeout_s=120.0 \\
+                --k8s_deletion_drain_timeout_s=300.0 \\
+                --k8s_agentic_namespace=agentic \\
+                --gke_machine_type=c4-standard-8
+
+Samples emitted (per run):
+  - gke_deletion_provision_time              (seconds)
+  - gke_deletion_total_drain_time            (seconds)
+  - gke_deletion_latency_p50                 (seconds)
+  - gke_deletion_latency_p95                 (seconds)
+  - gke_deletion_latency_p99                 (seconds)
+  - gke_deletion_latency_max                 (seconds)
+  - gke_deletion_rate                        (pods/sec)
+  - gke_deletion_ip_before                   (count)
+  - gke_deletion_ip_after                    (count)
+  - gke_deletion_ip_reclaim_time             (seconds)
+  - gke_deletion_final_running_count         (count)
+  - gke_deletion_wall_time                   (seconds)
+"""
+
+from __future__ import annotations
+
+
+import json
+import logging
+import time
+import uuid
+
+from absl import flags
+from perfkitbenchmarker import sample
+from perfkitbenchmarker import configs
+from perfkitbenchmarker.linux_benchmarks.kubernetes.agentic import (
+    k8s_benchmark_utils as utils,
+)
+from perfkitbenchmarker.linux_benchmarks.kubernetes.agentic import (
+    gke_deploy_utils as deploy_utils,
+)
+
+FLAGS = flags.FLAGS
+
+BENCHMARK_NAME = "k8s_deletion"
+BENCHMARK_CONFIG = """
+k8s_deletion:
+  description: >
+    Atomic single-point bulk deletion and IP reclamation measurement on a
+    pre-provisioned GKE cluster with gVisor isolation.
+  flags: {}
+  container_registry: {}
+  container_specs: {}
+  container_cluster: {}
+"""
+
+# ---------------------------------------------------------------------------
+# Benchmark-specific flags
+# ---------------------------------------------------------------------------
+
+flags.DEFINE_integer(
+    "k8s_deletion_batch_size",
+    100,
+    "Number of sandbox pods to provision then bulk-delete.",
+)
+
+flags.DEFINE_string(
+    "k8s_deletion_warmpool_name",
+    "python-sandbox-warmpool",
+    "SandboxWarmPool resource name.",
+)
+
+flags.DEFINE_string(
+    "k8s_deletion_pod_label",
+    "sandbox=python-sandbox-example",
+    "Label selector for warm pool pods.",
+)
+
+flags.DEFINE_float(
+    "k8s_deletion_poll_interval_s",
+    1.0,
+    "Seconds between kubectl polls during deletion.",
+)
+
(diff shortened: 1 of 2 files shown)


perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,511 @@
...
+    label = FLAGS.k8s_deletion_pod_label
+    poll_interval = FLAGS.k8s_deletion_poll_interval_s
+    provision_timeout = FLAGS.k8s_deletion_provision_timeout_s
+    drain_timeout = FLAGS.k8s_deletion_drain_timeout_s
+
+    logging.info("=== Run: batch_size=%d ===", batch_size)
+
+    # Drain to 0 for clean measurement (moved from Prepare for sweep compatibility)
+    utils.DrainWarmPool(ns, warmpool_name, label, timeout=int(drain_timeout))
+    time.sleep(2)
```
pmandewalkar: Saw these a couple times in the test, could we maybe use a more deterministic method to avoid waiting too long/too little if the system resolves earlier/later? Maybe a polling loop or kubectl command checking the state of what we are waiting for.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,511 @@
...
+    _PatchReplicas(ns, warmpool_name, batch_size)
+
+    deadline = time.time() + provision_timeout
+    while time.time() < deadline:
+        running = utils.CountPods(ns, label, phase="Running")
+        pct = (running / batch_size * 100) if batch_size > 0 else 0
+        logging.info("Provisioning... %d/%d (%.0f%%)", running, batch_size, pct)
+        if running >= batch_size:
+            break
+        time.sleep(3)
```
pmandewalkar: Since we have a configurable poll_interval that is used during the deletion loop on line 258, should we replace this hardcoded time.sleep(3) with time.sleep(poll_interval) for consistency?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,511 @@
...
+            break
+
+        time.sleep(poll_interval)
+
+    total_drain_time = time.monotonic() - t_delete
+
+    # Pods we never saw disappear (stuck) get the full drain time
+    for pn in pod_names_before:
+        if pn not in pod_gone_times:
+            pod_gone_times[pn] = total_drain_time
```
pmandewalkar, line 266: Is there a way we are differentiating pods in the edge case that take the whole drain time to successfully delete vs pods that are stuck and are assigned the same drain time? Maybe we could track a count of stuck pods and push it through as an additional metric (like k8s_deletion_stuck_pods_count) so we can separate that?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,511 @@
...
+
+    # 5. Compute per-pod deletion latencies
+    deletion_latencies = sorted(pod_gone_times.values())
+    n = len(deletion_latencies)
+
+    ip_after = _CountAllocatedIPs(ns, label)
+    deletion_rate = (
+        (len(pod_names_before) / total_drain_time) if total_drain_time > 0 else 0
```
pmandewalkar: If pods are left stuck, when we use len(pod_names_before) will it assume all pods were successfully deleted and inflate the deletion_rate? Since we are already tracking the exact number of deleted pods inside the loop on line 244, could we use the deleted variable as the numerator here?

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
```diff
@@ -0,0 +1,511 @@
...
+    time.sleep(2)
+
+    t_wall_start = time.monotonic()
+
+    # 1. Provision N pods
+    logging.info("Provisioning %d pods...", batch_size)
+    provision_start = time.monotonic()
+    _PatchReplicas(ns, warmpool_name, batch_size)
+
+    deadline = time.time() + provision_timeout
```
pmandewalkar: Would it be better to stick to time.monotonic() rather than using and comparing different measures? Since time.time() could be affected by system clock adjustments, it could theoretically make the timeout weird.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_warmpool_benchmark.py
```diff
@@ -0,0 +1,456 @@
...
+        ]
+    )
+
+    # 3. Poll until ready or timeout
+    t_scale = time.monotonic()
+    scale_start_epoch = time.time()
+    deadline = t_scale + threshold_s
+    first_pod_time = None
+
+    while time.time() < deadline:
```
pmandewalkar: Same question as before regarding monotonic() vs time()

pmandewalkar (commented):
Left a few comments regarding clock and hardcoded sleeps.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_warmpool_benchmark.py
```diff
@@ -0,0 +1,456 @@
...
+        }
+
+    return {
+        "time_to_created_s": _pcts(created_deltas),
+        "time_to_scheduled_s": _pcts(scheduled_deltas),
+        "time_to_running_s": _pcts(running_deltas),
+    }
+
+
+def _EmitLifecycleSamples(samples: list, lifecycle: dict, namespace: str, extra: dict) -> None:
```
roycaihw, line 427: Please document what's "lifecycle", and what's the difference between `_emit` and `_EmitLifecycleSamples`

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
george-kalisse-sada: The `utils.DrainWarmPool` actually is a deterministic polling loop (it polls kubectl get pods until the count is exactly 0). The time.sleep(2) that follows it is just a brief buffer to ensure the Kubernetes API cache has fully settled before we start the timer for the next phase. I've added an inline comment to clarify this.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
george-kalisse-sada: yeah That was a hardcoded oversight, thanks for catching this. I've updated it to use time.sleep(poll_interval) for consistency.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
george-kalisse-sada: Modified this so we now track stuck_pods_count during the loop. The deletion rate is now calculated strictly against `actual_deleted = len(pod_names_before) - stuck_pods_count`.
I've added a new metric gke_deletion_stuck_pods_count to the emitted samples so we can track this edge case if needed somewhere downstream.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
george-kalisse-sada: updates so that both to use time.monotonic().

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_warmpool_benchmark.py
george-kalisse-sada: updates so that both to use time.monotonic().

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_warmpool_benchmark.py
george-kalisse-sada: Done.

perfkitbenchmarker/linux_benchmarks/kubernetes/agentic/k8s_deletion_benchmark.py
roycaihw: L226 still uses `while time.time() < deadline_drain:`

roycaihw (approved):
One comment https://github.com/GoogleCloudPlatform/PerfKitBenchmarker/pull/6899#discussion_r3867379149. LGTM otherwise. Please squash the commits

